PR #4, opened 2019-03-17
Gail2

wichersn:
Implements the GAIL algorithm in the codebase.


sirspinach:
Tried getting the GAIL "generator increases discriminator loss" tests by increasing the number of PPO steps by 20x. Still failed about 50% of the time. Similar results for AIRL if we don't cache rollouts.

@AdamGleave thinks that in general end-to-end ML tests are more useful. Starting to agree with this, especially since static typing would catch most of the "no_crash" cases which are usually due to NameErrors.

yairl/util/util.py
```diff
@@ -298,3 +298,15 @@ def build_placeholders(env, include_new_obs):
...
+    ndim = len(space_shape)
+    if ndim== 0:
+        return tf.reshape(tensor, [-1, 1])
+    elif ndim == 1:
+        return tf.reshape(tensor, [-1, space_shape[0]])
+    else:
+        # TODO: Take the product(space_shape) and use that as the final
+        # dimension. In fact, product could encompass all the previous
+        # cases.
+        raise NotImplementedError
```
sirspinach: Could you add a newline?

wichersn:
> Tried getting the GAIL "generator increases discriminator loss" tests by increasing the number of PPO steps by 20x. Still failed about 50% of the time. Similar results for AIRL if we don't cache rollouts.
> 
> @AdamGleave thinks that in general end-to-end ML tests are more useful. Starting to agree with this, especially since static typing would catch most of the "no_crash" cases which are usually due to NameErrors.

I had the same result when trying the test. I think that if we can come up with an end to end ML test that passes reliably we should use it. But of course a test that fails half of the time is useless.

yairl/discrim_net.py
```diff
@@ -90,3 +97,57 @@ def build_policy_train_reward(self):
...
+        super().__init__()
+
+        tf.logging.info("using GAIL")
+
+    def build_discrm_network(self, obs_input, act_input):
+        inputs = tf.concat([
+            util.flat(obs_input, self.env.observation_space.shape),
+            util.flat(act_input, self.env.action_space.shape)], axis=1)
+
+        discrim_logits = util.apply_ff(inputs, hid_sizes=[])
```
sirspinach: The discriminator network is just linear in the inputs! I was surprised that this is enough, but Cartpole is a simple environment

yairl/util/util.py
wichersn: Done

PR #5, opened 2019-03-20
Cleanup

wichersn:
Uses gin, removes unused code and autoformates code.
Modifies so there is only one function to load the expert data, and only one function to train the model.


sirspinach:
I'll take a look at this tomorrow, right now I'm trying to get something done fast for another project

yairl/util/util.py
```diff
@@ -88,69 +94,12 @@ def make_blank_policy(env, policy_class=stable_baselines.PPO2,
     policy (stable_baselines.BaseRLModel)
     """
     env = maybe_load_env(env)
+    print("kwargs", kwargs)
```
sirspinach: Probably want log instead of print here

scripts/run_training.py
```diff
@@ -1,13 +1,12 @@
...
 from yairl.util.trainer import init_trainer
 import gin.tf
 import argparse
+import tensorflow as tf
 
 
 @gin.configurable
 def main(env_name):
-    logging.getLogger().setLevel(logging.INFO)
+    tf.logging.set_verbosity(tf.logging.INFO)
```
sirspinach: What are the advantages of using tf.logging relative to logging?

Guessing that by default it has a prettier print format than the root logger we've been using here

conftest.py
```diff
@@ -7,9 +7,11 @@
 seconds.
 """
 
+
 def pytest_addoption(parser):
     parser.addoption("--expensive", action="store_true",
-            help="run expensive tests (which are otherwise skipped).")
```
sirspinach: I'm curious -- did you use PyCharm's autoformat tool?

yairl/scripts.py
```diff
@@ -148,77 +39,112 @@ def plot_fight_loss(env='CartPole-v1',
...
         _savefig_timestamp("plot_fight_epreward_gen", interactive)
 
     add_plot_disc(False)
     add_plot_gen()
+
+    if n_plots_each_per_epoch <= 0:
+        n_gen_steps_per_plot = float('Inf')
+        n_disc_steps_per_plot = float('Inf')
+    else:
+        n_gen_steps_per_plot = int(n_gen_steps_per_epoch / n_plots_each_per_epoch + .5)
```
sirspinach: Consider `int(math.ceil(...))`? Not sure if that's actually any cleaner

sirspinach:
Thanks for the cleanup! I like having config files + gin here, seems like this will be handy for saving experiment configs outside of Jupyter notebooks.

The configuration framework that seems to be the most popular at CHAI is Sacred.

yairl/util/util.py
```diff
@@ -179,40 +128,42 @@ def make_save_policy_callback(savedir, file_prefix, save_interval):
...
-def load_expert_policy(env, **kwargs):
-    """
-    Load an expert policy from expert_models/.
-    """
-    return load_policy(env, basedir="expert_models", **kwargs)
+    paths.sort(key=lambda f: int(''.join(filter(str.isdigit, f))))
+
+    if len(paths) < n_experts:
+        raise ValueError(
+            """
```
AdamGleave: I'd favour using implicit line joining as per http://google.github.io/styleguide/pyguide.html#310-strings rather than `"""`.
Alternatively, could use a variable for the string, e.g. `msg = "Wanted to load ..."`, `raise ValueError(msg.format(...))`.

yairl/util/util.py
```diff
@@ -228,32 +179,34 @@ def load_policy(env, policy_class=stable_baselines.PPO2, basedir="",
...
-    else:
-        logging.info("couldn't find policy at '{}'".format(path))
-        return None
 
+    paths = get_policy_paths(env, policy_model_class, basedir, n_experts)
+
+    env = maybe_load_env(env)
+
+    if policy_network_class is not None:
+        kwargs["policy"] = policy_network_class
```
AdamGleave: What if the user specifies `policy` in `kwargs`? Are we OK overwriting this? Perhaps rename `policy_network_class` to just `policy?`

AdamGleave:
Thanks for this @wichersn, definitely an improvement.

I agree with @sirspinach that using Sacred would be nice, it has some helpful features not present in `gin` (e.g. storing git commit hashes, support for some other logging). However, I don't think we need to hold up these cleanup on that; best to add Sacred in a separate PR at a later date.

scripts/run_training.py
wichersn: I just thought that since we're using tf we should use its logging too.

conftest.py
wichersn: Yep

yairl/scripts.py
wichersn: I actually want to round though, not always round up.

wichersn:
> Thanks for this @wichersn, definitely an improvement.
> 
> I agree with @sirspinach that using Sacred would be nice, it has some helpful features not present in `gin` (e.g. storing git commit hashes, support for some other logging). However, I don't think we need to hold up these cleanup on that; best to add Sacred in a separate PR at a later date.

I just used gin since that's the only one I'm familiar with, but sure let's switch in another CL.

yairl/util/util.py
wichersn: Done

yairl/util/util.py
wichersn: Done

yairl/util/util.py
wichersn: Fixed

setup.py
```diff
@@ -3,5 +3,5 @@
 setup(
     name='yairl',
     version='0.1',
-    packages=['yairl',],
+    packages=['yairl', ],
 )
```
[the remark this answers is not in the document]
wichersn: Done

PR #6, opened 2019-03-21
Continuous integration & PEP8 fixes

AdamGleave:
Adds Travis continuous integration performing linting, unit tests and code coverage (via codecov.io).

Unit tests all passed but there were a lot of linting errors. For semantically meaningful changes, I did them manually and *hopefully* did not break anything. I let `autopep8` and `isort` do a lot of the reformatting, which may have occasionally made bad style decisions but should not break anything.


wichersn:
Looks like we'll have a lot of merge conflicts with mine. How would you like to handle them?

AdamGleave:
I think the conflicts largely come from the autopep8 and isort commits? If
we merge your PR first, we can probably revert those two commits and just
rerun the code formatting.

.travis.yml
```diff
@@ -0,0 +1,33 @@
...
+    install:
+      - pip install flake8 isort
+    script:
+      - ci/code_checks.sh
+    after_success: []
+
+  - name: "3.7 Coverage Tests"
+    python: "3.7"
+    script:
+      - py.test --cov-report=xml --cov=yairl yairl/
```
wichersn: Add --expensive to also run the expensive tests. These are also important to make sure it works.

.travis.yml
```diff
@@ -0,0 +1,33 @@
...
+    after_success:
+      - codecov
+
+  - name: "3.5 Unit Tests"
+    python: "3.5"
+  - name: "3.6 Unit Tests"
+    python: "3.6"
+  - name: "3.7 Unit Tests"
+    python: "3.7"
```
wichersn: Is there a way to also run the data_collection, and training python programs to make sure those don't crash? Sometimes the tests pass, but one of those still crashes.

README.md
```diff
@@ -1,27 +1,21 @@
...
 This repository reimplements AIRL with modern dependencies.
 Paper: https://arxiv.org/abs/1710.11248
 Original Implementation: https://github.com/justinjfu/inverse_rl
 
 To install:
-
 ```
-conda create -n yairl python=3.6.8  # Tensorflow doesn't work with python 3.7 at the time of writing.
+sudo apt install libopenmpi-dev
+conda create -n yairl python=3.7.2
```
wichersn: If it works with the latest version of python, just leave off the 'python=3.7.2' so conda can install the most up to date one.

requirements.txt
```diff
@@ -1,5 +1,8 @@
 gym
 numpy>=1.15
-pytest
-tensorflow  # Replace with tensorflow-gpu if you have a GPU
+tensorflow>=1.13.1,<2  # Replace with tensorflow-gpu if you have a GPU
 tqdm
+git+git://github.com/hill-a/stable-baselines.git@e59f32#egg=stable-baselines
```
wichersn: I'm curious how you got this to work.

README.md
```diff
@@ -1,27 +1,21 @@
...
 
 This repository reimplements AIRL with modern dependencies.
 Paper: https://arxiv.org/abs/1710.11248
 Original Implementation: https://github.com/justinjfu/inverse_rl
 
 To install:
-
 ```
-conda create -n yairl python=3.6.8  # Tensorflow doesn't work with python 3.7 at the time of writing.
+sudo apt install libopenmpi-dev
```
wichersn: Consider making a setup.sh that does all of this.

setup.cfg
```diff
@@ -0,0 +1,16 @@
+[flake8]
+max-line-length=99
```
wichersn: Why not 100? Much easier to remember.

requirements.txt
AdamGleave: I had to make this patch to Stable Baselines: https://github.com/hill-a/stable-baselines/pull/240

setup.cfg
AdamGleave: I based this off PEP8 having a 79 character limit for an 80 character target. It seems that some editors start line wrapping one character earlier. See https://www.reddit.com/r/learnpython/comments/1h2eug/pep_8_why_is_the_character_limit_79_and_not_80/

However, I think it'd be fine to set this to 100 as well.

.travis.yml
AdamGleave: Yes we should do this, I don't have bandwidth to add this but have created issue #8 to track this.

README.md
AdamGleave: I feel this is sufficiently platform dependent (e.g. people might need to use `yum` rather than `apt`, `virtualenv` rather than `conda`) I'd rather not make it a script. However, we should definitely consider making a script if it makes our life easier e.g. for testing or deployment (although I'd probably favor Docker at that point).

PR #11, opened 2019-03-25
Adds pendulum configs. Improves logging. 

wichersn:


configs/pendulum_data_collect.gin
```diff
@@ -1,2 +1,2 @@
 main.env_name = "Pendulum-v0"
-main.total_timesteps = 600000
\ No newline at end of file
+main.total_timesteps = 1000000
```
sirspinach: Missing a newline. Are you working on a Windows computer? :)

requirements-dev.txt
```diff
@@ -3,3 +3,4 @@ ipdb
 jupyter
 flake8
 isort
+autopep8
```
sirspinach: For this change, I would have preferred it be bundled in a patch other than "cleanup" (maybe even in its own "Add missing linting requirement" commit). That being said, building a meticulous commit history is probably less important than quickly iterating on research code like you are doing right now.

imitation/scripts.py
```diff
@@ -80,51 +81,62 @@ def show_plot_disc():
...
+            plt.plot(rand_ep_reward[name],
+                     label="avg random ep reward", c="black")
+            plt.plot(exp_ep_reward[name], label="avg exp ep reward", c="blue")
+            plt.legend()
+            _savefig_timestamp("plot_fight_epreward_gen", interactive)
 
     add_plot_disc(False)
-    add_plot_gen()
+    add_plot_gen(env, "True Reward")
+    add_plot_gen(trainer.wrap_env_test_reward(env), "Learned Reward")
```
sirspinach: Nice, I wonder what the learned reward plots look like :)

In the case of GAIL, the learned and train rewards are the same, right?

imitation/scripts.py
```diff
@@ -80,51 +81,62 @@ def show_plot_disc():
...
     if n_plots_each_per_epoch <= 0:
         n_gen_steps_per_plot = float('Inf')
         n_disc_steps_per_plot = float('Inf')
     else:
         n_gen_steps_per_plot = int(
-                n_gen_steps_per_epoch / n_plots_each_per_epoch + .5)
+            n_gen_steps_per_epoch / n_plots_each_per_epoch + .5)
         n_disc_steps_per_plot = int(
-                n_disc_steps_per_epoch / n_plots_each_per_epoch + .5)
+            n_disc_steps_per_epoch / n_plots_each_per_epoch + .5)
```
sirspinach: Could we use `int(round( . ))` here?

imitation/util/util.py
```diff
@@ -40,7 +41,10 @@ def make_vec_env(env_id, n_envs=8):
     env_id (str): The Env's string id in Gym.
     n_envs (int): The number of duplicate environments.
     """
-    return DummyVecEnv([lambda: gym.make(env_id) for _ in range(n_envs)])
+    # Use Monitor to support logging the episode reward and length.
+    return DummyVecEnv(
+        [lambda: Monitor(gym.make(env_id),
```
AdamGleave: Maybe make this an inline nested function `def` rather than `lambda`?

requirements-dev.txt
shwang: Ah, this isn't actually a linting requirement

AdamGleave:
@shwang do we know why the test cases are failing on this? We should either fix it and merge, or close the PR without merging -- don't want to leave this hanging indefinitely.

shwang:
1. Linting stuff, which I'll fix now.
2. AIRL's PPO step doesn't reliably decrease increase discriminator loss, so the tests don't always pass here. I think I will disable this for now, and look more closely into whether this test is worthwhile / should be seeded when I revisit AIRL.

shwang:
@AdamGleave 

I did some repeat runs of "generator increases discriminator loss" tests.
`test_train_disc_then_gen` (AIRL) passed 19/30 times.
`test_train_gen_degrade_D` (AIRL) passed 36/40 times.

For now I've left the flaky generator tests as `xfail`. This way the continue to run both on Travis CI and locally, but don't cause the overall test suite to fail.

Maybe we can try seeding in the future, but that seems problematic to me in a different way. (Should the contributor have to search for good seeds for stochastic tests every time they make changes to the reward network code?)

PR #15, opened 2019-06-20
Fix code coverage

AdamGleave:
codecov broke since we were installing imitation in a virtual env and then testing the installed package. This is the Right Thing To Do, but codecov by default ignores files in virtual env (since otherwise it'd measure coverage for all installed libraries). This PR explicitly points codecov at the appropriate package in the virtual env, and then rewrites the path to make it match the source.

I've also moved some things that should *not* be  in the code coverage out of `imitation/`, notably:
  - `imitation/tests` to `tests` -- we don't want to measure coverage of the tests themselves!
  - `imitation/scripts.py` into `scripts/train.py`. I also merged this with `scripts/run_training.py`, since this seemed to be the only file using it. However, there may be some reason to keep them separate, in which case feel free to demerge them.


shwang:
Back in December, the rationale for keeping the two scripts separate was to directly mirror the scripts in the inverse_rl repository. Seems okay to merge them.

shwang:
This doesn't have to be done now, but I'm thinking that it might be a good idea to check coverage of `scripts/*.py`, because these scripts are major entry points into the repository.

`tests/*.py` won't be able to import from `scripts/` unless scripts is a module though. (Or we use some pythonpath hack)

scripts/train.py
```diff
@@ -1,10 +1,9 @@
-"""
-Random experiments. As this file expands, I will probably move individual
-experiments into an scripts/ directory.
+"""Train an IRL algorithm and plot its output.
 
-These scripts are meant to be run in a Jupyter notebook (displays figures)
-but also automatically save timestamped figures to the output/ directory.
-"""
+Can be used as a CLI script, or the `train_and_plot` function can be called
```
shwang: > or the `train_and_plot` function can be called directly

I don't think anyone outside of `scripts/train.py` can import `train_and_plot`, since `scripts/` is not installed via `pip install -e .`.

shwang: Merging `imitation/scripts.py` into this file breaks the Jupyter notebook right now, which has an `import imitation.scripts as scripts`.

scripts/run_training.py
```diff
@@ -1,23 +0,0 @@
-import argparse
```
shwang: README.md needs an update since run_training.py was renamed

shwang:
What are the requirements for the `codecov/patch` and `codecov/project` checks to pass?

shwang:
The codecov check is failing right now because we moved scripts/*.py into the main module. (Not a problem)

AdamGleave:
> The codecov check is failing right now because we moved scripts/*.py into the main module. (Not a problem)

It's more subtle than that. I'm trying to configure it to apply different thresholds to scripts/ using flags, but Codecov does not seem to support this feature well.

The coverage checks currently don't make much sense: apparently main is covered 100% but only 9% of the diff is hit...

AdamGleave:
Things are ready for another review now. To summarize the changes:
  - I moved `scripts` into `imitation`, so it's now a package. This will let things like the Jupyter notebook work.
  - I think it's a good idea to test the scripts, but it's not as high priority. It's easy enough in the codecov UI to see coverage for different parts of the project. I've also now configured a `,codecov.yml` which will report on PRs coverage for the auxiliary (examples and scripts) and main (imitation learning algorithms) part of the codebase. Specifically, it reports the *relative change in the absolute coverage* (this is weird, but how CodeCov works).

Note:
  - The CodeCov patch check is failing for this PR. This is because I introduced a few lines of untested code (in the scripts). I think we can just ignore this. (I could also reconfigure what paths it looks at for the patch; but I think it's better for us just to scrutinize the diff and make a decisions on a case-by-case basis.)
  - The CodeCov project check is passing with 100% code coverage for main! This is because previously there was no test coverage here (due to a misconfig), and so any non-zero coverage is a 100% relative coverage increase. This should go back to sane values for future PRs.

shwang:
To continue an tradition from NetSys lab, posting a cat pic with every PR

image

decodyng:
Sorry I missed giving a review on this; was out on Friday for EAG volunteering

PR #18, opened 2019-06-25
Misc clean up

shwang:
* Convert to two space indent.
  * Add a python3.6 tag to setup.py
  * Add env/ to .gitignore for virtualenv users


.gitignore
```diff
@@ -1,16 +1,15 @@
 __pycache__
```
AdamGleave: I suggest we base our .gitignore on https://github.com/github/gitignore/blob/master/Python.gitignore This is nicely structured, has ignores for common IDEs (except PyCharm; we should also ignore .idea) and various possible directories for virtual environments.

shwang:
There might be some indent problems in docstrings, like those fixed [here](https://github.com/HumanCompatibleAI/imitation/pull/20/commits/983df06d86d6452aebcc95ec14b4625f5c364587).

I'll catch them while doing #19

Issue #19, opened 2019-06-25
Docstrings


AdamGleave:
Consistent use of Google style docstrings. This is the best [summary](https://gist.github.com/redlotus/3bc387c2591e3e908c9b63b97b11d24e) I've seen; also see [Napolean](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/sphinxcontrib.napoleon.html#sphinxcontrib.napoleon.docstring.GoogleDocstring) and [official style guide](http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings). Main violations I see in the codebase right now are:
  - Starts with a paragraph. Should be first line summary, new line, then additional info.
  - Using `Params` rather than `Args`, not indenting things.
  - Being imperative: it should be descriptive. So, `Computes fizzbuzz`, not `Compute fizzbuzz. 
  - Having first line just be `"""`. Should be followed with text.

We're also including types in the documentation, which is not usual Google style, but I think is good practice. I suggest we do it in the form:
  `n: (int) a natural number`
not:
  `n (int): a natural number`
so that automatic documentation generation tools like `Sphinx` will be happy.

We should also make sure each module has at least a one-sentence docstring. I'd also suggest doing this for all methods, there are a few without. I don't think it's necessarily worth writing copious documentation for each function at this stage, though: many are self-explanatory.

shwang:
What do you think about make 2-space indentation everywhere including docstring for consistency?

For what it's worth, TF does 2-space [indentation everywhere](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/composite_tensor.py#L147).

AdamGleave:
That's the style I believe I followed in reward_net (except for line
continuations which I made 4 spaces for legibility). Think it's a good idea.

AdamGleave:
Also it seems I'm wrong about where the types should go: both the example I linked and the Napolean definition have types like `n (int): a natural number`. I don't see this actually specified anywhere in the Google Style Guide. Think I got confused as Stable Baselines seems to use the other convention. I'm not sure it's worth changing back though...

PR #20, opened 2019-06-25
Reward Network: make hidden units configurable and fix docstrings

AdamGleave:
Main material change is making `BasicRewardNet` and `BasicShapedRewardNet` support configurable numbers and sizes of hidden layers. I believe I've kept the defaults the same so should be backward compatible. I also factored out `build_theta_network` as code was duplicated between the classes.

I've updated docstrings to be compliant with the Google style mentioned in contribution guidelines. Also added a few docstrings.

*NOTE*: I've made this 2-space indent and put `2space` as the base so you don't have to change my formatting. Make sure to change the base to `master` after `2space` gets merged, and then merge into master.


imitation/reward_net.py
```diff
@@ -181,94 +183,113 @@ def build_summaries(self):
...
+
+        Args:
+            env: (gym.Env or str) The environment we are predicitng reward for.
+            units: (int) The number of hidden units in each of the two layers
+                of the feed forward phi network.
+            state_only: (bool) If True, then ignore the action when predicting
+                and training the reward network phi.
+            theta_units: (List[int]) number of hidden units at each layer of
+                the feedforward theta network.
+            phi_units: (List[int] number of hidden units at each layer of
```
shwang: missing ')', but I can fix this up when I go through all the comments a la #19

shwang:
Didn't know the base branch trick, that's pretty cool!

PR #21, opened 2019-06-25
Improve documentation: consistent style & linting

### LINKED ISSUE #19: Docstrings

shwang:
closes #19


imitation/util/rollout.py
```diff
@@ -131,56 +131,55 @@ def rollout_done():
...
-  """
-  Get the undiscounted reward after rolling out `n_timestep` steps in
-  of the policy. With large n_timesteps, this can be a decent metric
-  for policy performance.
+  """Get the undiscounted reward after rolling out `n_timestep` steps in
+  of the policy.
+
+  With large n_timesteps, this can be a decent metric for policy performance.
 
   Params:
```
shwang: Missed Params here

AdamGleave:
I've added HTML documentation building via Sphinx, and this is now part of `ci/code_checks.sh`. This caught a few remaining errors in the docstrings that I fixed. I've also put the types in the correct places (my bad -- I got the convention wrong previously) -- we should perhaps consider using type annotations in the future instead?

I'd be in favor of changing docstrings to 2-spaces, but only if this is relatively simple (i.e. we can do a regex or there's a tool for it); I don't think this is the best use of engineering time :)

Since I've now touched the code substantially as well can @decodyng review?

imitation/airl.py
```diff
@@ -105,85 +103,81 @@ def train_gen(self, n_steps=10000):
...
+        gen_act (np.ndarray): A numpy array with shape
+            `[n_timesteps] + env.action_space.shape`.
+        gen_new_obs (np.ndarray): A numpy array with shape
+            `[n_timesteps] + env.observation_space.shape`. The ith observation
+            in this array is from the transition state after the generator
+            chooses action `gen_act[i]`.
+
+    Returns:
+        discriminator_loss (float): The total cross-entropy error in the
+            discriminator's clasifications.
```
decodyng: Nit: "classifications" needs an extra "s"

imitation/airl.py
```diff
@@ -105,85 +103,81 @@ def train_gen(self, n_steps=10000):
...
   def wrap_env_test_reward(self, env):
-    """
-    Returns the given Env wrapped with a reward function that returns
+    """Returns the given Env wrapped with a reward function that returns
     the reward learned by this AIRLTrainer.
 
-    The reward network referenced (not copied) into the Env
-    wrapper, and therefore the rewards are changed by calls to
-    AIRLTrainer.train().
+    References the reward network (not copied) into the `Env`
```
decodyng: Coming from having less context on this project, this isn't super clear: it seems like we're using the reward function attached to self? What's meaningful about it being copied vs not copied?

imitation/airl.py
```diff
@@ -105,85 +103,81 @@ def train_gen(self, n_steps=10000):
...
-      string environment name is given or a Env is given, then we first
-      make a VecEnv before continuing.
+    Args:
+        env (str, Env, or VecEnv): The Env that should be wrapped. If a
+            string environment name is given or a Env is given, then we first
+            make a VecEnv before continuing.
 
     Returns:
-    env
+        wrapped_env (Env): The wrapped environment with a new reward.
```
decodyng: At this point would this be an Env or a VecEnv? (since you say you'll convert from the former to the latter)

imitation/reward_net.py
```diff
@@ -53,7 +53,7 @@ def reward_output_train(self):
     ```
 
     Returns:
-      _reward_output_train: (tf.Tensor) A (None,) shaped Tensor holding
+      _reward_output_train (tf.Tensor) A (None,): shaped Tensor holding
```
decodyng: Seems like the colon drifted a bit too far right here ;)

imitation/reward_net.py
```diff
@@ -75,7 +75,7 @@ def reward_output_test(self):
     ```
 
     Returns:
-      _reward_output_test: (tf.Tensor) A (None,) shaped Tensor holding
+      _reward_output_test (tf.Tensor) A (None,): shaped Tensor holding
```
decodyng: Same as above, colon should probably come right after tf.Tensor parens

imitation/reward_net.py
```diff
@@ -146,7 +146,7 @@ def reward_output_train(self):
     ```
 
     Returns:
-      _reward_output_train: (tf.Tensor) A (None,) shaped Tensor holding
+      _reward_output_train (tf.Tensor) A (None,): shaped Tensor holding
```
decodyng: Again re colon

imitation/reward_net.py
```diff
@@ -252,16 +252,16 @@ def __init__(self, env, *, state_only=False,
     """Builds a simple shaped reward network.
 
     Args:
-      env: (gym.Env or str) The environment we are predicitng reward for.
-      units: (int) The number of hidden units in each of the two layers
+      env (gym.Env or str): The environment we are predicitng reward for.
```
decodyng: "predicting" is misspelled

imitation/util/trainer.py
```diff
@@ -17,15 +17,15 @@
...
-  env_id (str): The string id of a gym environment.
-  use_random_expert (bool):
-    If True, then use a blank (random) policy to generate rollouts.
-    If False, then load an expert policy. Will crash if DNE.
-  **kwargs -- Pass additional arguments to the AIRLTrainer constructor.
+  Args:
+    env_id (str): The string id of a gym environment.
+    use_random_expert (bool):
+        If True, then use a blank (random) policy to generate rollouts.
+        If False, then load an expert policy. Will crash if DNE.
```
decodyng: Needs docstring for policy_dir and use_gail (specifically clarifying what policy_dir is meant to be if use_random_expert=True, since it seems like in that case we wouldn't be loading a policy?)

imitation/util/util.py
```diff
@@ -11,16 +11,16 @@
 
 
 def maybe_load_env(env_or_str, vectorize=True):
-  """
-  Params:
-  env_or_str (str or gym.Env): The Env or its string id in Gym.
-  vectorize (bool): If True, then vectorize the environment before returning,
-    if it isn't already vectorized.
+  """Maybe load and process an environment or environment id.
```
decodyng: Can we explain what the "maybe" is doing here? When would or wouldn't this happen?

imitation/airl.py
shwang: It seems fine to remove the first part and just keep "the wrapped reward is changed when .train() is called".

My documenting motive was to emphasize that the caller should note that the wrapped reward isn't fixed.

imitation/util/trainer.py
shwang: Yeah, kinda confusing that AIRLTrainer can `use_gail` here, now that I think about it. Nevan and I weren't very careful about cleanliness during the review for this change/

AdamGleave:
@decodyng if you could look at this sometime today it'd be appreciated. It's blocking merging `rewardnet-refactor` which I need for some of my code. I've got my own fork but it'd be easier to work on the main branch...

PR #23, opened 2019-06-26
Expose theta and phi builders publicly

AdamGleave:
For my use case it's helpful to be able to build the networks outside of the reward shaping classes when desired. I expect this to also be helpful to other users, who might be working outside of `reward_net.py`  but still want to use similar architectures.

I've also generalized `build_basic_theta_network` to be able to depend on the next observation.

*Note*: the `docstrings` base will need to be changed.

### DIFF
--- a/imitation/reward_net.py
+++ b/imitation/reward_net.py
@@ -194,11 +194,12 @@ def build_basic_theta_network(hid_sizes, obs_space, act_space,
 
   Arguments:
     hid_sizes (Optional[List[int]]): Number of units at each hidden layer.
+        Default is [], i.e. linear.
     obs_space (gym.Space): Observation space.
     act_space (gym.Space): Action space.
     old_obs_input (Optional[tf.Tensor]): Previous observation.
     new_obs_input (Optional[tf.Tensor]): Next observation.
-    act_input (tf.Tensor): Action.
+    act_input (Optional[tf.Tensor]): Action.
 
   Returns:
     tf.Tensor: Predicted reward.
@@ -266,12 +267,13 @@ def build_basic_phi_network(hid_sizes, obs_space, old_obs_input, new_obs_input):
 
   Arguments:
     hid_sizes (Optional[List[int]]): Number of units at each hidden layer.
+        Default is [32, 32].
     obs_space (gym.Space): Observation space.
     old_obs_input (Optional[tf.Tensor]): Previous observation.
     new_obs_input (Optional[tf.Tensor]): Next observation.
 
   Returns:
-    Tuple[tf.Tensor]: potential for the old and new observations.
+    Tuple[tf.Tensor, tf.Tensor]: potential for the old and new observations.
   """
   if hid_sizes is None:
     hid_sizes = [32, 32]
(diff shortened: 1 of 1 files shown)


imitation/reward_net.py
```diff
@@ -183,19 +183,42 @@ def build_summaries(self):
...
+  Conversely, if `new_obs_input` and `act_input` are both set to `None`, it
+  will depend just on the current observation: :math:`R(o)`.
+
+  Arguments:
+    hid_sizes (Optional[List[int]]): Number of units at each hidden layer.
+    obs_space (gym.Space): Observation space.
+    act_space (gym.Space): Action space.
+    old_obs_input (Optional[tf.Tensor]): Previous observation.
+    new_obs_input (Optional[tf.Tensor]): Next observation.
+    act_input (tf.Tensor): Action.
```
shwang: `act_input` should also be `Optional`.

imitation/reward_net.py
```diff
@@ -183,19 +183,42 @@ def build_summaries(self):
...
+                              old_obs_input, new_obs_input, act_input):
+  """Builds a reward network depending on specified observations and actions.
+
+  All specified inputs will be preprocessed and then concatenated. If all
+  inputs are specified, then it will be a :math:`R(o,a,o')` network.
+  Conversely, if `new_obs_input` and `act_input` are both set to `None`, it
+  will depend just on the current observation: :math:`R(o)`.
+
+  Arguments:
+    hid_sizes (Optional[List[int]]): Number of units at each hidden layer.
```
shwang: Note that default is linear

imitation/reward_net.py
```diff
@@ -224,16 +247,50 @@ def __init__(self, env, *, state_only=False, theta_units=None, **kwargs):
...
   def reward_output_train(self):
     """Training reward is the same as the test reward, since no shaping."""
     return self.reward_output_test
 
 
+def build_basic_phi_network(hid_sizes, obs_space, old_obs_input, new_obs_input):
+  """Builds a potential network depending on specified observation.
+
+  Arguments:
+    hid_sizes (Optional[List[int]]): Number of units at each hidden layer.
```
shwang: Note that default is [32, 32]

imitation/reward_net.py
```diff
@@ -224,16 +247,50 @@ def __init__(self, env, *, state_only=False, theta_units=None, **kwargs):
...
+    obs_space (gym.Space): Observation space.
+    old_obs_input (Optional[tf.Tensor]): Previous observation.
+    new_obs_input (Optional[tf.Tensor]): Next observation.
+
+  Returns:
+    Tuple[tf.Tensor]: potential for the old and new observations.
```
shwang: For documentation purposes this is fine, but fun-fact:
`Tuple[tf.Tensor, tf.Tensor]` is actually the return type would make a type-checker pass.

(Tuple is weird; `Sequence[tf.Tensor]` and `Iterable[tf.Tensor]` would have passed okay.)

AdamGleave:
Thanks for the comments Steven, addressed. Holding off on merging since `docstrings` is still open; feel free to merge once that's closed.

shwang:
image

PR #27, opened 2019-07-02
Take mean of loss for consistent loss magnitude

shwang:
The scale of the discrim loss shown the following plot was previously taken as a sum over losses. This led to discrim loss plots with magnitudes proportional to the discrim batch size. Now we take the average for consistency.

image

Also made the discriminator gradient step over a `tf.reduce_mean(loss)` rather than just `loss` in case we had the same magnitude problem.

Based on a few runs of Cartpole+GAIL, I don't think that using `opt.minimize(loss)` (where loss is a Tensor with shape `(4000,)`) is different than using `opt.minimize(tf.reduce_meanloss))`. 

Before:
image

After (trial 1):
image

After (trial 2):
image

EDIT: This is CartPole+GAIL, not CartPole+AIRL is less stable

### DIFF
--- a/imitation/airl.py
+++ b/imitation/airl.py
@@ -145,7 +145,7 @@ class AIRLTrainer():
             discriminator's classification.
     """
     fd = self._build_disc_feed_dict(**kwargs)
-    return np.sum(self._sess.run(self.discrim.disc_loss, feed_dict=fd))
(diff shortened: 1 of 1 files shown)


shwang:
note: the previous plots were for GAIL, not AIRL. AIRL is not that stable on Cartpole

shwang:
image

PR #28, opened 2019-07-02
Test cleanup: eliminate spurious warnings, use modern markers

AdamGleave:
Two changes:
  - pytest reports all warnings on tests, which is useful. But right now they're being drowned out by warnings triggered by third-party packages. Tidy them up so we can see any important warnings.
  - I also noticed pytest was complaining about `expensive` being an unregistered marker. Docs suggest declaring them in `setup.cfg`, and can then skip tests marked in that way using a CLI flag: `-m not expensive`. This lets us get rid of the logic in `tests/conftest.py`.


shwang:
image

AdamGleave:
I think the Stable Baselines warning is legit (I assume from the models in `expert_models` being from an older version of Stable Baselines). I guess it's OK to ignore, but it might be better to just re-save the policies.

PR #33, opened 2019-07-03
Refactor and document train_and_plot

shwang:
(Wait for replay_buffer to get merged and change base to master before merging this one)

Changes in this PR somehow have aggravated a `rollouts.total_reward(n_episodes)` problem where we accidentally added rewards from partial episodes to the total_episodes_rewards. I believe that this is because we are now correctly passing around a length-8-VecEnv where we were passing around a env-id-string that got loaded as a length-1-VecEnv.

The rollouts problem should be fixed by changes to `rollouts` made in #32.


shwang:
A review is low priority right now because I won't merge until after two other PRs are done, and I can check whether plots look sane with the rollout fix.

imitation/airl.py
```diff
@@ -62,6 +63,8 @@ def __init__(self, env, gen_policy, discrim, expert_policies, *,
 
     self.env = util.maybe_load_env(env, vectorize=True)
     self.gen_policy = gen_policy
+    if not isinstance(expert_policies, collections.abc.Sequence):
```
AdamGleave: OK, although this API will be changing soon anyway I believe (to take demos)?

imitation/scripts/train.py
```diff
@@ -38,130 +43,149 @@ def train_and_plot(policy_dir, env='CartPole-v1',
...
+        instead of initializing a new one using `airl_trainer_kwargs`. Also,
+        ignore the `env` argument.
   """
+  assert n_epochs_per_plot is None or n_epochs_per_plot >= 1
   if trainer is None:
+    assert env is not None
     trainer = init_trainer(env, policy_dir=policy_dir, **trainer_kwargs)
-  if trainer_hook_fn:
-    trainer_hook_fn(trainer)
+  env = trainer.env
```
AdamGleave: Check that `env` is None? If we must have two ways of specifying the same thing, at least make sure users are certain which one they're using.

imitation/trainer.py
```diff
@@ -59,6 +60,8 @@ def __init__(self, env, gen_policy, discrim, expert_policies, *,
 
     self.env = util.maybe_load_env(env, vectorize=True)
     self.gen_policy = gen_policy
+    if not isinstance(expert_policies, collections.abc.Sequence):
```
AdamGleave: Not a fan of multiple expert policies being an option in the first place. But we're soon switching to an API where this takes demos, so can probably ignore this for now.

imitation/scripts/train.py
```diff
@@ -38,130 +43,149 @@ def train_and_plot(policy_dir, env='CartPole-v1',
...
+    plot_num = math.floor(n_plots_per_epoch * epoch)
+    prev_plot_num = math.floor(n_plots_per_epoch * (epoch - 1))
+    assert abs(plot_num - prev_plot_num) <= 1
+    return plot_num != prev_plot_num
+
+  # Collect data for epoch 0.
+  if n_epochs_per_plot is not None:
+    disc_plot_add_data(False)
+    ep_reward_plot_add_data(env, "True Reward")
+    ep_reward_plot_add_data(trainer.env_wrapped_test, "Learned Reward")
```
AdamGleave: Maybe specify in the title this is test reward? It's a bit misleading in so far as the generator is being trained with something else. Possibly log both?

imitation/trainer.py
shwang: Agreed, that we shouldn't have multiple expert policies. This change was a crash-fix for the PR.

imitation/scripts/train.py
shwang: Yes, I intend on making adding better plot titles to this PR

shwang:
Force-push was because of a merge mistake. Merging again.....

shwang:
Waiting on #39

imitation/scripts/train.py
shwang: ended up just removing the optional `trainer` parameter

shwang:
Looks like Cartpole AIRL training has slowed down since #37. I could investigate why if this is important, not sure if it is because we are moving towards automatically hyperparam tuning later anyways @AdamGleave 

#37, with Adam's hyperparam tuning
image


master (h5d18e395a3)
image

This PR: (61daec0)
image

AdamGleave:
Thanks for noting the regression in performance. I'm not too concerned provided it does still get to stable expert performance eventually. Plausibly some of the changes I made when tuning for MuJoCo may have harmed performance in Cartpole. We should keep an eye on this though.

PR #34, opened 2019-07-04
Miscellaneous clean up

### LINKED ISSUE #22: Belated code review
Meta issue I will be updating with points I notice as I read through the code, that seem too small to be worth creating their own issue about.

- [x] Consider using stable_baselines `observation_input` to build placeholders with preprocessed output. This will let us work smoothly with a variety of observation and action spaces.
- [x] Use `tf.layers.flatten` instead of `util.flat` (which I think can be deleted). I believe this would also let us get rid of passing the shape around explicitly with the tensor (since `tf.layers.flatten` extracts it automatically).
- [ ] In `RewardModel`, pass in `observation_space` and `action_space` rather than a `gym.Env`, since we do not seem to ever need other properties of the `gym.Env`.

shwang:
Closes #22.

I realize that trainer.py and util/trainer.py have confusingly similar names. `util/trainer.py` only contains `init_trainer()`, an alternate initializer for Trainer that loads expert policies from pickles. I think `init_trainer()` should be deleted or subsumed into `class Trainer` once we start loading expert trajectories from pickles instead of expert policies.


shwang:
The base branch is `replay_buffer` because that branch made changes to AIRLTrainer docstrings and parameters.

AdamGleave:
Note we're getting some warnings because `CartPoleEnv` sets its `Box` high and low limits to `np.finfo(np.float32).max`, which then overflow if you try and subtract. Dunno why they don't just set it to `np.inf`. We may want to add these warnings to the ignore list.

PR #35, opened 2019-07-04
Hotfix: ReplayBuffer returns in-order entries

AdamGleave:
`ReplayBuffer` had a bug: we sampled *independently* from the different buffers, so did not sample state-action-next state transitions. This PR fixes this by changing `Buffer` to operate on a dict of NumPy arrays, with `ReplayBuffer` being a simple wrapper around this.

There are some things here that could be improved. I haven't really pushed through the fact that `ReplayBuffer` is now a simple wrapper. So for example there's some testing specific to `ReplayBuffer` to check for out-of-order sequences, that should really be in the test for `Buffer` instead. I don't have time to do this now, but someone else is welcome to clean up.

Reviewing this is moderately urgent (i.e. look at this before you do other work, but don't work on a holiday because of it) -- the codebase as-is is currently broken, so will be hard to make progress on anything.

### DIFF
--- a/imitation/util/buffer.py
+++ b/imitation/util/buffer.py
@@ -215,9 +215,17 @@ class ReplayBuffer:
         raise ValueError("Shape or dtype missing and no environment specified.")
 
     self.capacity = capacity
-    self._old_obs_buffer = Buffer(capacity, obs_shape, obs_dtype)
-    self._act_buffer = Buffer(capacity, act_shape, act_dtype)
-    self._new_obs_buffer = Buffer(capacity, obs_shape, obs_dtype)
+    sample_shapes = {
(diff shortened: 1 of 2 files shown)


shwang:
> So for example there's some testing specific to ReplayBuffer to check for out-of-order sequences, that should really be in the test for Buffer instead.

I'm guessing that you are talking about calls to `_check_bound()` with multiple data keys in `test_replay_buffer` and `test_replay_from_data`? 

I added a second data key to `test_buffer` so that it can check for out-of-order sequences. I could also remove/replace the `_check_bound` parts from ReplayBuffer for something else, but I don't think it's necessary for now.

PR #37, opened 2019-07-04
Hyperparameter tuning and continuous action spaces

AdamGleave:
I have added support for configuring in GIN a variety of previously hardcoded parameters, and have found configs that reliably achieve near-expert performance in AIRL and GAIL for `CartPole-v1` and `Pendulum-v0`.

When investigating `Pendulum-v0` I found training immediately blew up into NaNs. This is because Stable Baselines was not computing action probabilities for continuous action spaces. I've opened a PR to address that; for now, I'm pointing this repo towards my branch on Stable Baselines.

There's also a lot of minor changes, e.g. fixing up docstrings and adding comments as I went. Apologies for the noise; I can try and split these out if needed.

I think this would be good to get reviewed and merged soon so that we have a known-good config that we can use to test any future changes.


AdamGleave:
For posterity, attaching results from latest version (03e485). I observed some variance between runs. I re-ran the CartPole GAIL once as the first result looked broken, all others are the first run.

**CartPole, GAIL**

cartpole_gail_disc
cartpole_gail_true
cartpole_gail_learned

**CartPole, AIRL**

cartpole_airl_disc
cartpole_airl_true
cartpole_airl_learned

**Pendulum, GAIL**

pendulum_gail_disc
pendulum_gail_true
pendulum_gail_learned

**Pendulum, AIRL**

pendulum_airl_disc
pendulum_airl_true
pendulum_airl_learned

**Saved notebooks**
[notebooks.zip](https://github.com/HumanCompatibleAI/imitation/files/3363311/notebooks.zip)

imitation/trainer.py
```diff
@@ -1,69 +1,90 @@
...
-               n_disc_samples_per_buffer=200, n_expert_samples=4000,
+  """Trainer for GAIL and AIRL."""
+
+  def __init__(self,
+               env: Union[gym.Env, str],
+               gen_policy: BaseRLModel,
+               discrim: DiscrimNet,
+               expert_policies: Sequence[BaseRLModel], *,
+               disc_opt_cls: tf.train.Optimizer = tf.train.AdamOptimizer,
+               disc_opt_kwargs: Dict = {},
```
shwang: It's considered better form, I think, to use `dict` when using the generic properties of `Dict`

AdamGleave: :+1: did not know this, makes sense

PR #38, opened 2019-07-06
Beta MuJoCo Configs

AdamGleave:
Some experimentation trying to get MuJoCo environments to work. HalfCheetah kind of works: getting around 75% of expert, but very unstable across time and between random seeds.

Haven't actually tested `Ant` or `Swimmer` yet. I couldn't train a good expert policy using `data_collect`, it needs `VecNormalize` support.

Postponing further experimentation until we have better tooling (e.g. Sacred config, logging, ability to run hyperparam sweeps). But seems worth including so we can test on HalfCheetah at least and make sure there aren't any regressions.

Also made `hid_sizes` in `DiscrimNetGAIL` configurable.


AdamGleave:
Some training curves.

One theory I have for instability is that the learned reward seems to be on a large scale, effectively making the learning rate larger than intended. But when I decreased the generator learning rate by 10x it no longer learnt (stayed flat around random). So, that might not be the reason. Still, I am interested in thinking of trying ways of normalizing the reward during training.

**HalfCheetah, GAIL**
halfcheetah_gail_disc
halfcheetah_gail_true
halfcheetah_gail_learned

**HalfCheetah, AIRL**
halfcheetah_airl_disc
halfcheetah_airl_true
halfcheetah_airl_learned

shwang:
LGTM

shwang:
image

PR #39, opened 2019-07-09
Rollout hotfix: add truncate option

shwang:
Fixes a notebook-crashing problem where `rollouts.flatten_trajectory(rollouts.generate(..., n_timesteps=1000))` was generating more than 1000 samples, leading to an error from ReplayBuffer of capacity 1000.

* Re-do rollouts API: now `generate` and `generate_multiple` both return obs-act-obs-rew tuples, and `generate_traj` returns lists of trajectory dictionaries. (`generate` used to return list of traj dicts)
* `generate` now automatically truncates the length of trajectories to `n_timesteps` if `truncate=True`. (Matching the behavior of `generate_traj`.)
* Add a test to check if `scripts/train.py` is running properly, using a gin configuration similar to the notebook.

### DIFF
--- a/imitation/util/rollout.py
+++ b/imitation/util/rollout.py
@@ -63,6 +63,7 @@ class _TrajectoryAccumulator:
 
   def finish_trajectory(self, idx) -> Dict[str, np.ndarray]:
     """Complete the trajectory labelled with `idx`.
+
     Return list of completed trajectories popped from
     `self.partial_trajectories`.
     """
@@ -79,16 +80,16 @@ class _TrajectoryAccumulator:
     return out_dict_stacked
 
   def add_step(self, idx, step_dict: Dict[str, np.ndarray]):
-    """Add a single step to the partial trajectory identified by `idx` (this
-    could correspond to, e.g., one environment managed by a vecenv)."""
+    """Add a single step to the partial trajectory identified by `idx`.
+
+    This could correspond to, e.g., one environment managed by a VecEnv.
+    """
     self.partial_trajectories[idx].append(step_dict)
 
 
 def generate_trajectories(policy, env, *, n_timesteps=None, n_episodes=None
                           ) -> List[Dict[str, np.ndarray]]:
-  """
-  Generate a list of old_obs-action-new_obs-reward trajectories from a policy
-  and an environment.
+  """Generate trajectory dictionaries from a policy and an environment.
 
   Args:
     policy (BasePolicy or BaseRLModel): A stable_baselines policy or RLModel,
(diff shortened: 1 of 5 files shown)


imitation/util/rollout.py
```diff
@@ -297,8 +297,52 @@ def flatten_trajectories(trajectories: Sequence[Dict[str, np.ndarray]]
       cat_parts["rew"]
 
 
-def generate_multiple(policies, env, n_timesteps, *, truncate=True):
-  """Generate obs-act-obs triples from several policies.
+def generate(policy, env, *, n_timesteps=None, n_episodes=None, truncate=True,
```
qxcv: Maybe call this `generate_transitions` to make it clear that they're undistinguished transitions from a bunch of underlying environment copies?

imitation/util/rollout.py
```diff
@@ -84,11 +84,11 @@ def add_step(self, idx, step_dict: Dict[str, np.ndarray]):
...
 
 
-def generate(policy, env, *, n_timesteps=None, n_episodes=None
-             ) -> List[Dict[str, np.ndarray]]:
+def generate_trajectories(policy, env, *, n_timesteps=None, n_episodes=None
+                          ) -> List[Dict[str, np.ndarray]]:
   """
-  Generate old_obs-action-new_obs-reward tuples from a policy and an
-  environment.
+  Generate a list of old_obs-action-new_obs-reward trajectories from a policy
```
qxcv: Good catch on the incorrect docstring. The `old_obs-action-new_obs-reward` tuple part needs to be removed too, since now it's returning separate vectors of concatenated observations,  concatenated actions, and concatenated rewards.

imitation/util/rollout.py
```diff
@@ -312,25 +356,28 @@ def generate_multiple(policies, env, n_timesteps, *, truncate=True):
...
+          If the number of policies given doesn't divide this number evenly,
+          then the last policy generates more timesteps than the other policies.
+      truncate (bool): If True and n_timesteps is not None, then drop any
+          additional samples to ensure that exactly `n_timesteps` samples are
+          returned.
 
   Returns:
       rollout_obs_old (array): A numpy array with shape
-          `[n_timesteps] + env.observation_space.shape`. The ith observation in
+          `[n_samples] + env.observation_space.shape`. The ith observation in
```
qxcv: Are you sure you want this to be `n_samples` and not `n_timesteps` for consistency with the `n_timesteps` arg?

tests/test_trainer.py
```diff
@@ -27,8 +27,8 @@ def test_init_no_crash(use_gail, env='CartPole-v1'):
 def test_train_disc_no_crash(use_gail, env='CartPole-v1', n_timesteps=200):
   trainer = init_trainer(env, use_gail=use_gail)
   trainer.train_disc()
-  obs_old, act, obs_new, _ = rollout.flatten_trajectories(rollout.generate(
```
qxcv: Nice, much cleaner.

imitation/util/rollout.py
shwang: It isn't always `n_timesteps` because we can go over. I'll add this as a clarification

shwang: (note: also do this for `generate()`)

tests/test_script_train.py
```diff
@@ -0,0 +1,21 @@
...
+def test_train_and_plot_no_crash():
+  train_and_plot(n_epochs=2,
+                 n_plots_each_per_epoch=1,
+                 n_disc_steps_per_epoch=1,
+                 n_gen_steps_per_epoch=1,
+                 interactive=False)
+  tf.reset_default_graph()
+
+
+# TODO(shwang): edit notebooks to match new params
```
shwang: (note: rm this)

imitation/util/rollout.py
shwang: Sounds good to me. `generate_multiple` will become `generate_transitions_multiple`

PR #41, opened 2019-07-10
Migrate experiment configs from Gin to Sacred

shwang:
notes:
* Corrected a use_gail=False error in `halfcheetah_gail.gin`.
* Renamed test.gin => debug.
* API CHANGE: Removed policy_dir as required first argument from imitation.scripts.train.train_and_plot().
  * This was a quick way to prevent a sacred error with doubly filled parameter.
* Set default arg for policy_dir and stop passing policy_dir explicitly in a lot of places.
* Sacred and TF both recently released some broken builds, so explicitly skipping them in the `requirements.txt`. In particular, sacred 0.7.5 inadvertently fails to import without mongodb, and tensorflow 1.14.0 is missing a `tf.__spec__` variable that sacred requires.

Merge #39 first, since notebook and classical_control_experiements are broken until then.


shwang:
Oh, I need to update README.md

imitation/scripts/data_collect.py
```diff
@@ -1,9 +1,6 @@
-import argparse
```
AdamGleave: Add an observer to the experiment by default. That way it'll save the config, Git commit hashes and other nice things. Without it we're not getting much benefit from switching to Sacred.

imitation/scripts/data_collect.py
```diff
@@ -16,22 +13,11 @@ def make_PPO2(env_name, num_vec):
   return policy
 
 
-@gin.configurable
+@data_collection_exp.automain
```
AdamGleave: I ended up opting for just using `@exp.main` in my code and having a `if name == '__main__'` block that calls it, see e.g. https://github.com/HumanCompatibleAI/adversarial-policies/blob/master/src/modelfree/score_agent.py

My reason for this is that you might want to run the experiment from multiple endpoints: as a stand-alone script, or from another script that's running the experiment multiple times (perhaps in parallel) to perform a hyperparameter sweep.

If you use `automain`, then you have to have the same environment (and in particular observers) regardless. If you have a `__main__` block manually defined, you can add the observers and anything else you want in the stand-alone script, leaving the experiment that's exported to other files barebones.

Given that we're not currently using the experiment from multiple files, I think it's fine to keep using `automain`, but just beware you may need to change it later.

imitation/util/util.py
```diff
@@ -89,14 +88,14 @@ def __init__(self, *args, **kwargs):
...
 
-@gin.configurable
 class FeedForward64Policy(FeedForwardPolicy):
   def __init__(self, *args, **kwargs):
     super().__init__(*args, **kwargs,
                      net_arch=[64, 64], feature_extraction="mlp")
 
 
-@gin.configurable
+@train_exp.capture(prefix="make_blank_policy_kwargs")
```
AdamGleave: Consider making `make_blank_policy` an ingredient, which both experiments depend on. https://sacred.readthedocs.io/en/latest/ingredients.html

You could alternately just not make it depend on Sacred at all, and feed in the arguments manually from `train` and `data_collect`. The `capture` style can avoid huge parameter lists making things more readable, but it also makes it hard to understand where all your parameters have an effect. Because of this I like to try and limit the scope of Sacred `capture` and `config` to my script/experiment files only.

imitation/scripts/train.py
```diff
@@ -1,25 +1,24 @@
 """Train an IRL algorithm and plot its output.
```
AdamGleave: Add an observer to the experiment by default

imitation/scripts/train.py
```diff
@@ -1,25 +1,24 @@
...
 import tqdm
 
+from imitation.experiments import train_exp
 import imitation.util as util
 from imitation.util.trainer import init_trainer
 
 
-@gin.configurable
-def train_and_plot(policy_dir, env='CartPole-v1',
+@train_exp.automain
```
AdamGleave: Same comment as before about automain

AdamGleave:
I've updated the code coverage config to count `experiments.py` under auxiliary (configs, scripts) not main (core algorithms). If you rename/split up the file, you may want to update `.codecov.yml` accordingly

imitation/experiments.py
```diff
@@ -0,0 +1,217 @@
+import sacred
```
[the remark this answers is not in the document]
shwang: At least if we use two experiments (one for data collection and one for train), there's an annoying import cycle that occurs between the script file and the utilities module that does the `experiment.capture`.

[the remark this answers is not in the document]
shwang: Oh I see, you are probably talking about leaving the experiments in the current file, and then defining the configs in the scripts. That will probably work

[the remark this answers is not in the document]
shwang: I think I'm going to do config files in `imitation/scripts/config/*.py`.

imitation/experiments.py
```diff
@@ -0,0 +1,217 @@
...
+            # speed, but may come at a cost of stability.
+            gen_replay_buffer_capacity=1000,
+            n_expert_samples=1000,
+        ),
+
+        discrim_scale=False,
+    )
+
+
+@train_exp.named_config
```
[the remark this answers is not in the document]
shwang: Nice, I think there's only a single thing gail/airl pair that had a difference

imitation/experiments.py
```diff
@@ -0,0 +1,217 @@
...
+    nminibatches=32,
+    noptepochs=10,
+    # WARNING: this is actually 8*256=2048 due to 8 vector environments
+    n_steps=256,
+)
+
+
+@train_exp.config
+def train_defaults():
+    n_epochs = 50
```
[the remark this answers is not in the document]
shwang: I'll leave comments for later, since I'm expecting some of these parameters to change as we transition to training from demonstrations (as opposed to policies).

imitation/scripts/data_collect.py
shwang: done, I think

imitation/scripts/data_collect.py
shwang: I used a __main block to easily add an observer

imitation/util/util.py
shwang: I limited `capture` to my script file

shwang:
Annoyingly, `sacred.namedconfig` doesn't allow the decorated function to accept previously defined configuration keys as arguments. Consequently, for the big dictionaries that were being passed around and updated, I "passed them around" by repeatedly referencing a module-level dictionary variable.

This should be fine for single-threaded experiments, might become problematic for a multi-threaded process spawning lots of different experiments (I heard that Ray+Sacred can get kinda weird).

AdamGleave:
> Annoyingly, `sacred.namedconfig` doesn't allow the decorated function to accept previously defined configuration keys as arguments. Consequently, for the big dictionaries that were being passed around and updated, I "passed them around" by repeatedly referencing a module-level dictionary variable.
> 
> This should be fine for single-threaded experiments, might become problematic for a multi-threaded process spawning lots of different experiments (I heard that Ray+Sacred can get kinda weird).

I think you've misunderstood how Sacred config parsing works. It's confusing and IMO fundamentally the wrong design. I've pushed a change that I believe does what you want without resorting to module-level variables; please review and check that I've not broken anything.

My feeling is that modifying global state in Sacred config should be avoided at all costs. Global state is confusing at the best of time. Add to that Sacred does not actually call the functions: it reads the source code via reflection, and evaluates them internally. It captures the changes to state during the evaluation, which it then adds to the config. I would not rely on the order in which the functions are evaluated, or on modifications to global state in the function actually affecting global state elsewhere in the program. And, yes, adding parallelism to the mix will make things even more confusing.

The Sacred config process is described [here](https://sacred.readthedocs.io/en/latest/internals.html). The key point is that named configs execute before the normal configs. This is why you saw the error of being unable to capture parameters defined in the normal config: they haven't actually been defined yet!

You are, however, free to define parameters in multiple configs. In this case, the value set in the named config takes priority over the value defined in the normal config; if it's defined in multiple named/normal configs, then the later one takes priority.

So for example, if I run `python -m imitation.scripts.train print_config with swimmer pendulum airl` I get:
Screenshot from 2019-07-12 14-02-00

Note that `pendulum` has overidden the `env_name`, but the `n_epochs` and `policy_network_class` set by `swimmer` are still there. Anything not specified in a named config is from the default normal config.

Hope this helps. Feel free to ping me or @decodyng in the future if you run into Sacred issues. We've both had to deal with a lot of problems like this in the past.

imitation/scripts/config/common.py
```diff
@@ -0,0 +1,7 @@
+DEFAULT_BLANK_POLICY_KWARGS = dict(
+    learning_rate=3e-4,
+    nminibatches=32,
+    noptepochs=10,
+    # WARNING: this is actually 8*256=2048 due to 8 vector environments
+    n_steps=256,
```
AdamGleave: In `adversarial-policies`, I instead introduced my own `batch_size` variable, and then computed `n_steps = batch_size / num_envs`. It made the config easier, but did have the downside of needing to write an adapter for each Stable Baselines policy. Something worth considering if this becomes painful.

imitation/scripts/config/train.py
```diff
@@ -0,0 +1,97 @@
...
+
+@train_ex.config
+def train_defaults():
+    n_epochs = 50
+    n_disc_steps_per_epoch = 50
+    n_gen_steps_per_epoch = 2048
+
+    init_trainer_kwargs = dict(
+        use_random_expert=False,
+        num_vec=8,  # NOTE: changing this also changes the effective nsteps!
```
AdamGleave: `n_steps`?

PR #51, opened 2019-07-19
Add script to evaluate trained policies

AdamGleave:
We have a script to train policies (and produce rollouts from it as a side-effect), and a script to train an IRL reward model and policy, but nothing to evaluate policies.

It's not obvious this really belongs in `imitation`: we're just using the Stable Baselines policy format, so this could live there or in some more general RL utility library.

Right now this is very much an MVP. A much more complicated version of this script (entangled with adversarial policies codebase though): https://github.com/HumanCompatibleAI/adversarial-policies/blob/master/src/modelfree/score_agent.py


shwang:
I wonder if there's a way to split `patch` into `patch/{auxiliary,main}`

imitation/scripts/policy_eval.py
```diff
@@ -0,0 +1,78 @@
...
+from imitation.scripts.config.policy_eval import policy_eval_ex
+from imitation.util import rollout, util
+
+POLICY_CLASSES = {
+    'ppo1': stable_baselines.PPO1,
+    'ppo2': stable_baselines.PPO2,
+}
+
+
+class InteractiveRender(VecEnvWrapper):
```
shwang: Oh, this is neat!

tests/test_scripts.py
```diff
@@ -0,0 +1,17 @@
...
+  """Smoke test for imitation.scripts.policy_eval"""
+  run = policy_eval_ex.run(config_updates={'render': False},
+                           named_configs=['fast'])
+  assert run.status == 'COMPLETED'
+  assert isinstance(run.result, dict)
+
+
+def test_train():
+  train_ex.run(named_configs=['cartpole', 'gail', 'debug'])
+  tf.reset_default_graph()
```
shwang: It might be annoying to have test sacred logs clutter the log dir whenever I run `pytest tests/`. Do you know of a way to easily separate them from actual logs? Probably still want to save them for debugging purposes. Maybe we could save them into `output/sacred_test` instead of `output/sacred`.

Anyhow this shouldn't block this PR.

AdamGleave: The `FileStorageObserver` isn't attached (since it gets added in the `if name == '__main__'` guarded block), so no Sacred output will be produced. However, it will still produce output such as the checkpoints and other non-Sacred logging info. I've changed the `log_root` so this goes into `output/tests` now.

PR #53, opened 2019-07-20
Move source files to src/ subdirectory

AdamGleave:
Main motivation for a `src` directory is to have a clean developer mode installation, especially important now we're using that for coverage testing. See https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure for an argument for this structure

I'm aware this is unconventional in Python though, so feel free to reject if you disagree with it!


shwang:
Found another [blog post](https://hynek.me/articles/testing-packaging/). I don't understand based on these readings why this will affect coverage when we are not running tests in parallel, but it looks like `src/` can avoid some package testing confusions. (Probably the "clean dev mode installation" you were mentioning)

AdamGleave:
Sorry for the confusion: I meant that you can easily have different code installed in developer mode vs normal. Not a big deal when all our testing is using normal installs. But problematic when we start testing using developer mode.

Issue #40, opened 2019-07-09
Absorbing state fix


qxcv:
For environments that can terminate before some fixed horizon, it would be good to have the absorbing state fix from [discriminator-actor-critic](https://openreview.net/pdf?id=Hk4fpoA5Km).

Programatically, I expect we could do this by making a new env wrapper that always produces a final `(s_absorb,a_0, s_absorb)` transition at the end of each trajectory before propagating `done=True` up. `s_absorb` could be represented by a zero vector of size `d_obs`, or by a zero vector of size `d_obs` with an extra `1` flag appended to the end to signal that it's terminal. The only really tricky bit would be ensuring that `a_0` is always the same for the last trajectory, which might require some changes to the rollout code.

AdamGleave:
A fix for this issue would certainly be valuable. I'm not sure on the details of how best to implement it.

For reproducibility, we should definitely make this optional, although perhaps the default.

If I understand the fix correctly, we're effectively learning just one additional parameter: the reward in an absorbing state. And then to compute returns, we're summing the discounted absorbing reward up over the remaining (possibly infinite) steps in the horizon.

This suggests an implementation where we instead have a wrapper around the `RewardNet` that adds this extra parameter and, for terminal state transitions, computes the terminal return. This might require adding some extra placeholders, e.g. for the `dones` flags and the number of remaining timesteps.

This feels cleaner to me than changing the observation space, where I'd be concerned the reward model might have smoothness constraints that make it unable to assign arbitrary values to the terminal reward without also changing the reward assigned to nearby states. Near-zero states especially often occur naturally.

qxcv:
A bit late, but…

> I'd be concerned the reward model might have smoothness constraints that make it unable to assign arbitrary values to the terminal reward without also changing the reward assigned to nearby states

Ah, that's why I suggested adding the extra 1 indicator at the end for absorbing states (which would be 0 in non-absorbing states). Unless the model has extremely strong smoothness constraints, it should be able to get a lot out of a single dimension that changes value completely in absorbing states. Doing it this way also means that we don't need a wrapper that messes with TF graphs to introduce a new parameter & conditional, which is possibly (?) harder than just changing observations. Then again, changing observations would get harder for image spaces (where in the image do you tack on the extra dim?), so perhaps the extra parameter way is more general :man_shrugging:

AdamGleave:
[...]

Issue #31, opened 2019-07-03
API Changes


AdamGleave:
Issue to collect possible changes to the AIRL API. Feel free to treat as a Wiki and edit the main post to add issues.

- [x] Adam: Have `AIRLTrainer` take a set of expert *demonstrations*, rather than expert policies. Requiring expert policies effectively precludes using real human data.
- [x] Sam: Replace uses of `BaseRLModel` with `BasePolicy` wherever possible. e.g rollout functions should only need a _policy_, not a complete RL algorithm. Some imitation algorithms also only need policies.
  - (Adam: `BaseRLModel` does some preprocessing in `predict`, so the distinction between policy and RL algorithm is not totally clear in Stable Baselines.)
  - (Steven: In Stable Baselines, there isn't an easy `save`/`load` API for policies. In this respect it's much easier to work with `BaseRLModel`.)
  - (Sam: Those are both good points. However, the SB abstractions don't make sense for this project, where some of the algorithms we use don't do RL, and thus do not have a `BaseRLModel` to begin with. Possibly I should create an issue upstream in SB calling for policies, algorithms, preprocessors, and rollout utilities to be decoupled so that we can re-use their policy abstractions.)
- [x] Sam: Make rollout functions trajectory-centric instead of transition-centric. That should make it easier to compute trajectory-wise return, to work with trajectory-centric algorithms, to add absorbing states to the end of trajectories, etc. Can just have `flatten()` fn that goes from trajectories back to transitions. (Adam: :+1:)
- [x] Adam: not a fan of letting `env` be either an `Env`, `VecEnv`, `str`. It shouldn't be `imitation`'s responsibility to create an environment: how would it know what preprocessing the user wants, how many copies of the environment, how to parallelize (`DummyVecEnv` or `SubprocVecEnv`), etc? It is convenient but I think we can get most of that benefit just by providing some helper methods to easily create e.g. vector environments.
- [x] Adam: who's responsibility is it to create the session? There are two sensible options IMO. 1) imitation is responsible for creating a fresh session and graph. Everything it does is isolated to this. The user can access these if they want (through an instance variable `sess` and `graph`) but may also have their own sessions. This is how Stable Baselines does it. 2) imitation uses the default session and graph. The user is responsible for setting them up. If they want to isolate it, they must create a different session for imitation. This is how I've been writing my own project.

[...]

PR #54, opened 2019-07-20
Density-based imitation learning baselines

### LINKED ISSUE #1: Wrapping VecEnv -- Final (old_obs, act, obs) triple is incorrect.
Suppose I am at timestep 5 and I took actions that a triggered a                                                                                                       
transition such that my 0th environment ends its episode. Vec Env implementations of step() will internally reset() the 0th environment and return the observation associated with the episode start instead of the episode end.

The episode end is never returned by step(), so there is no way for me to see it.

### LINKED ISSUE #16: Meta-issue: dumb imitation baselines
In addition to our fancy adversarial & maxent algorithms, we should also consider having some simple baselines for imitation. In descending order of importance, these could/should include:

- [x] Behavioural cloning (in progress by @qxcv)
- [ ] Soft RL with soft behavioural cloning. Fit a rough density estimate `\hat\pi_E(a|s)` to the expert demonstrations with BC, then use `\log \hat \pi_E(a|s) \approx A^\star(s,a)` as a reward function to train an imitation agent.
- [ ] Try state-based density estimates instead of action-based density estimates. Specifically, you could put a Gaussian around each of the demonstration trajectories and reward the agent with the log likelihood of the states it visits under those Gaussians. If we only have one trajectory `\tau = (s_1, s_2, \ldots, s_T)` (ignoring actions), then the reward function would be `r_t(\hat s_t, \hat a_t) = \|\hat s_t - s_t\|^2`, where `s_t` is the demo state at time `t` from trajectory `\tau`. This is a silly version of what [DeepMimic](https://xbpeng.github.io/projects/DeepMimic/2018_TOG_DeepMimic.pdf) does. I don't expect this to work very well, but it would be good at keeping us honest on how well the parametric methods are doing.

qxcv:
This PR adds some simple baseline imitation learning algorithms to the repo. Specifically, the baselines fit a rough density model to the demonstrator trajectory, then perform RL on a reward function based on the log likelihood of that trajectory. The algorithms in this PR are very silly: they only do density estimation with some KDEs from scikit-learn, & do not yet handle termination conditions correctly. However, I expect they'll be useful as baselines, and as a foundation for imitation learning algorithms built on more powerful forms of density estimation (e.g Real NVP or whatever deep learning people use for density estimation nowadays).

There are also a few other changes in this PR:

- Fixes #1 by recording terminal transition during rollouts.
- Adds an IPython notebook demonstrating one of the density-based baselines on Pendulum, where it goes from -1,200 reward down to around -400 reward in a few hundred thousand timesteps. That's slightly worse than demonstrator, which IIRC was closer to -300 after 1,000,000 time steps.
- Increases the default `n_steps` collected by PPO to be consistent with the Pendulum-v0 example in the SB agent zoo (I found this made it much more stable).
- Adds some Pendulum-v0 demo trajectories to the test data directory.
- Moves `RewardVecEnvWrapper` into its own class, and gives it the ability to condition rewards on the time step `t`.
- Adds some script entry points to `setup.py` (`imitation-data-collect`, `imitation-train`).
- Adds the remaining two algorithms mentioned in #16 (so fixes #16).


shwang:
I'll review this by the end of the week

src/imitation/scripts/config/common.py
```diff
@@ -1,7 +1,10 @@
...
+    # For recommended PPO hyperparams in each environment, see:
+    # https://github.com/araffin/rl-baselines-zoo/blob/master/hyperparams/ppo2.yml
     learning_rate=3e-4,
     nminibatches=32,
     noptepochs=10,
-    # WARNING: this is actually 8*256=2048 due to 8 vector environments
-    n_steps=256,
+    # WARNING: this is actually 8*2048=16384 due to 8 vector environments
+    # (n_steps=2048 across 8 vector environments is also what the Zoo uses)
+    n_steps=2048,
```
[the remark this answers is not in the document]
AdamGleave: Bigger `n_steps` means lower variance gradients but fewer overall gradient steps. So it should always work but might require `total_timesteps` be increased.

I ended up using a batch size of 16,384 / n_steps=2048 for Ant and Humanoid and it worked much better in those environments than n_steps=256. For things like CartPole and even Pendulum I expect a smaller n_steps will work fine and be faster overall. For things like HalfCheetah it could go either way, would need to be benchmarked.

src/imitation/util/__init__.py
```diff
@@ -1,4 +1,10 @@
 # flake8: noqa: F401
 
-import imitation.util.rollout
+from imitation.util.reward_wrapper import RewardVecEnvWrapper
+from imitation.util import rollout
+from imitation.util.rollout import (RandomPolicy, flatten_trajectories,
```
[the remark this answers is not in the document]
AdamGleave: I've been updating against importing stuff in `__init__.py`'s recently: it does save some key strokes but makes it much less clear where things live. Probably appropriate for libraries where you may want to hide some implementation details, but in internal code I think it's generally better to be explicit and use fully-qualified imports.

src/imitation/scripts/config/common.py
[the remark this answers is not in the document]
qxcv: Perhaps it's worth setting up the default config so that it's environment-dependent? e.g. if someone runs on Pendulum then it should use this value, if someone runs on CartPole it should use 32, etc. I could add such a change to this PR or include it in a new one.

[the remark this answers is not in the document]
AdamGleave: I like having sensible defaults, but making the hyperparams change based on `env_name` feels too magical. How about just a named config for each environment, setting the `env_name` and appropriate hyperparams?

shwang:
(This closes #1)

tests/test_density_baselines.py
```diff
@@ -0,0 +1,86 @@
...
+    ret = 0.0
+    for step, (obs, act, next_obs) in enumerate(traj_zip):
+      ret += reward_fn([obs], [act], [next_obs], steps=[step])
+    returns.append(ret)
+  return np.mean(returns)
+
+
+@parametrize_density_stationary
+def test_density_reward(density_type, is_stationary):
+  # use Pendulum b/c I don't handle termination correctly yet
```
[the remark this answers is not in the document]
qxcv: This refers to #40: like AIRL/GAIL, this density baselines don't correctly handle episodes that terminate before reaching the horizon. I'm using Pendulum instead of Cartpole because Pendulum doesn't terminate early. I'll update the comment.

tests/test_reward_vec_env_wrapper.py
```diff
@@ -0,0 +1,31 @@
...
+from imitation import util
+
+
+class FunkyReward:
+  def __init__(self):
+    self.counter = 0
+
+  def __call__(self, obs, act, next_obs, *, steps=None):
+    # give each environment number from 1 to num_envs
+    return (np.arange(len(obs)) + 1).astype('float32')
```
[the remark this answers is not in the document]
qxcv: I'm a little confused. How does that help test `wrapped_env_rew`?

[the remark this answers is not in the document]
shwang: Not sure (I am confused too). But could you make a modification to this test that checks whether `wrapped_env_rew` is working properly?

Oh, I think my idea was that adding a decimal reward to the wrapper reward only would be an easy way to make the wrapped reward distinguishable from the inner reward. (-XXX.1 vs -XXX).

src/imitation/density_baselines.py
```diff
@@ -0,0 +1,267 @@
...
+  def _fit_single_density(self, flat_transitions):
+    # This bandwidth was chosen to make sense with standardised inputs that
+    # have unit variance in each component. There might be a better way to
+    # choose it automatically.
+    density_model = KernelDensity(kernel=self.kernel,
+                                  bandwidth=self.kernel_bandwidth)
+    density_model.fit(flat_transitions)
+    return density_model
+
+  def _flatten_trajectories(self, trajectories):
```
shwang: Doing more than just flattening here. (Totally different behavior from `rollout.flatten_trajectories`). Consider renaming to `process_trajectories`? A brief docstring describing the returned value (params not necessary) would be great too.

src/imitation/density_baselines.py
```diff
@@ -0,0 +1,267 @@
...
+      flat_traj = []
+      for step_num in range(len(traj['act'])):
+        flat_trans = self._flatten_transition(obs_vec[step_num],
+                                              act_vec[step_num],
+                                              obs_vec[step_num + 1])
+        flat_traj.append(flat_trans)
+      flat_trajectories.append(flat_traj)
+    return flat_trajectories
+
+  def _flatten_transition(self, obs, act, next_obs):
```
shwang: consider `_process_transition`

src/imitation/density_baselines.py
```diff
@@ -0,0 +1,267 @@
...
+    Returns:
+      rew_array (np.ndarray): array of scalar rewards of the form `r_t(s,a,s')
+        = \log \hat p_t(s,a,s')` (one for each environment), where `\log \hat
+        p` is the underlying density model (and may be independent of s', a, or
+        t, depending on options passed to constructor)."""
+    rew_list = []
+    assert len(obs_b) == len(act_b) and len(obs_b) == len(next_obs_b)
+    for idx, (obs, act, next_obs) in enumerate(zip(obs_b, act_b, next_obs_b)):
+      flat_trans = self._flatten_transition(obs, act, next_obs)
+      scaled_padded_trans = self._scaler.transform(flat_trans[None])
```
shwang: nit: `np.newaxis` instead of None here

src/imitation/density_baselines.py
```diff
@@ -0,0 +1,267 @@
...
+               is_stationary=False):
+    r"""Family of simple imitation learning baseline algorithms that apply RL to
+    maximise a rough density estimate of the demonstration trajectories.
+    Specifically, it constructs a non-parametric estimate of `p(s)`, `p(s,s')`,
+    `p_t(s,a)`, etc. (depending on options), then rewards the imitation learner
+    with `r_t(s,a,s')=\log p_t(s,a,s')` (or `\log p(s,s')`, or whatever the
+    user wants the model to condition on).
+
+    Args:
+      env (gym.Env or str): environment to train on.
```
shwang: Missing `rollouts` arg

src/imitation/density_baselines.py
```diff
@@ -0,0 +1,267 @@
...
+      imitation_trainer (BaseRLModel): RL algorithm & initial policy that will
+        be used to train the imitation learner.
+      expert_trainer (BaseRLModel): snapshot of RL algorithm that has been
+        trained to solve `env`.
+      n_expert_trajectories (int): number of expert trajectories to sample and
```
shwang: expert_trainer, n_expert_trajectories not included in parameters

src/imitation/density_baselines.py
shwang: Could use `TrajectoryList` [named type](https://github.com/HumanCompatibleAI/imitation/blob/a0f2f7b88212851058ba6224b60750fae01d8c4d/src/imitation/util/rollout.py#L19) for doc.

src/imitation/density_baselines.py
```diff
@@ -0,0 +1,267 @@
...
+                                   kernel=kernel,
+                                   kernel_bandwidth=kernel_bandwidth,
+                                   standardise_inputs=standardise_inputs)
+    self.wrapped_env = reward_wrapper.RewardVecEnvWrapper(self.env,
+                                                          self.reward_fn,
+                                                          include_steps=True)
+    self.graph = tf.Graph()
+    self.sess = tf.Session(graph=self.graph)
+    with self.graph.as_default():
+      self.sess.run(tf.global_variables_initializer())
```
shwang: [Else](https://github.com/HumanCompatibleAI/imitation/blob/a0f2f7b88212851058ba6224b60750fae01d8c4d/src/imitation/trainer.py#L77)[where](https://github.com/HumanCompatibleAI/imitation/blob/5d775e7bac886a067bc4c08faa9b4701a7a26164/src/imitation/discrim_net.py#L18) in the repo we expect a default session to exist.

We haven't decided which of the two tf.Session strategies that @AdamGleave described #31 we are going for, so it's okay to leave this for now (fixing it later should be easy).

src/imitation/density_baselines.py
```diff
@@ -0,0 +1,267 @@
...
+        method of the imitation RL model. Refer to Stable Baselines docs for
+        details."""
+    self.imitation_trainer.set_env(self.wrapped_env)
+    # FIXME: learn() is not meant to be called frequently; there are
+    # significant per-call overheads (see Adam's comment in trainer.py)
+    # FIXME: the ep_reward_mean reported by SB is wrong; it comes from a
+    # Monitor() that is being (incorrectly) used to wrap the underlying
+    # environment.
+    self.imitation_trainer.learn(n_timesteps,
+                                 reset_num_timesteps=False,
```
shwang: Why is this set to False rather than letting `kwargs` decide? I'm not sure why this is important. (maybe leave a comment)

src/imitation/density_baselines.py
```diff
@@ -0,0 +1,267 @@
...
+                                 **kwargs)
+
+  def test_policy(self, *, n_trajectories=10, true_reward=True):
+    """Test current imitation policy on environment & give some rollout
+    stats.
+
+    Args:
+      n_trajectories (int): number of rolled-out trajectories.
```
shwang: Add true_reward param

src/imitation/util/reward_wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+from stable_baselines.common.vec_env import VecEnvWrapper
+
+
+class RewardVecEnvWrapper(VecEnvWrapper):
+  def __init__(self, venv, reward_fn, *, include_steps=False):
+    """A RewardVecEnvWrapper uses a provided reward_fn to replace
+    the reward function returned by `step()`.
+
+    Automatically resets the inner VecEnv upon initialization. A tricky part
+    about this class keeping track of the most recent observation from each
```
shwang: "is keeping"

src/imitation/bc.py
```diff
@@ -36,7 +36,7 @@ def __init__(self,
     self.env = env
     self.policy_class = policy_class
     self.batch_size = batch_size
-    expert_obs, expert_acts, _ = expert_rollouts
+    expert_obs, expert_acts, _, _ = expert_rollouts
```
shwang: If `Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]` is too long for your taste, could use `Tuple[np.ndarray, ...]` to fix pytype error

src/imitation/density_baselines.py
qxcv: So that we can see the total number of timesteps of training. Without an appropriate value, the timesteps shown in the log don't reset to 0 each time `.learn` is called. Will comment.

src/imitation/density_baselines.py
qxcv: Does it actually matter?

qxcv: (changed it anyway, I'm just curious why you prefer `np.newaxis`)

shwang: It doesn't matter, seemed clearer to me. Probably because I keep on forgetting what None does.

shwang:
image

PR #55, opened 2019-07-22
AIRL envs

shwang:
Import and smoke test environments from `inverse_rl`


src/imitation/envs/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+              'disabled': False,
+          })
+_register('imitation/DisabledAnt-v0',
+          entry_point='imitation.envs.ant_env:CustomAntEnv',
+          kwargs={
+              'gear': 30,
+              'disabled': True,
+          })
+
+# FIXME: There are some import issues with VisualPointMaze, not sure if
```
AdamGleave: I think this can be deleted -- never used in the AIRL paper AFAIK

src/imitation/envs/ant_env.py
```diff
@@ -0,0 +1,604 @@
...
+        self.np_random.uniform(size=self.model.nq, low=-.1, high=.1)
+    qvel = self.init_qvel + self.np_random.randn(self.model.nv) * .1
+    self.set_state(qpos, qvel)
+    return self._get_obs()
+
+  def viewer_setup(self):
+    self.viewer.cam.distance = self.model.stat.extent * 0.5
+
+
+if __name__ == "__main__":
```
AdamGleave: I don't think we need to be able to run this as a CLI script?

src/imitation/envs/visual_pointmass.py
```diff
@@ -0,0 +1,165 @@
+# pytype: skip-file
```
AdamGleave: This file can probably be deleted as discussed above

tests/test_envs.py
```diff
@@ -0,0 +1,33 @@
+import gym
+import pytest
+
+from imitation.envs import ENV_NAMES
+from imitation.util import make_vec_env
+
+PARALLEL = [False, True]
+
+try:
```
AdamGleave: codecov will complain as one side of this branch isn't taken so we don't achieve 100% test coverage. Add a "# pragma: nocover" comment.

tests/test_envs.py
```diff
@@ -0,0 +1,33 @@
...
+  MUJOCO_OK = False
+
+
+@pytest.mark.skipif(not MUJOCO_OK,
+                    reason="Requires `mujoco_py`, which isn't installed.")
+@pytest.mark.parametrize("parallel", PARALLEL)
+def test_envs(parallel):
+  """Check that our custom environments don't crash on `step`, and `reset`."""
+  for env_name in ENV_NAMES:
+    if parallel:
```
AdamGleave: `parallel` is misleading here as it suggests true parallelism, i.e. `SubprocVecEnv`, but `make_vec_env` can also make `dummy_vec_env`. I'd suggest renaming as `vector`.

I'm also not sure why we need to test both. The *environment* doesn't know if it's in a `VecEnv` or not. So can probably just test one or the other.

tests/test_envs.py
```diff
@@ -0,0 +1,33 @@
...
+      env = make_vec_env(env_name)
+    else:
+      env = gym.make(env_name)
+
+      env.reset()
+      obs_space = env.observation_space
+      for _ in range(4):
+        act = env.action_space.sample()
+        obs, rew, done, info = env.step(act)
```
AdamGleave: I don't think this works in vector environments, since the `action_space` is not vectorized (i.e. will return an action for just one environment)

tests/test_envs.py
AdamGleave: `pytype` is also unhappy about this and will need a comment.

In general I'm uneasy about disabling some tests (e.g. I'm pretty sure this test will *not* pass right now, but it's still green on Travis because it skips the test). However, getting MuJoCo working on Travis is a bit of a pain, and some people may not have license keys so probably best to avoid it being a forced dependency.

AdamGleave:
We should also be sure to add attribution to Justin's codebase for these environments, e.g. a copyright declaration at the top of the files or reproducing his LICENSE file. It was MIT licensed which makes it easy though.

tests/test_envs.py
shwang: That sounds right, I'll remove parallel if I keep this test

tests/test_envs.py
shwang: Lol it works, but only because the for loop is part of the `else` branch! (oops)

shwang:
Moving to examples sounds good.

tests/test_envs.py
shwang: stable-baselines zoo [uses `pybullet`](https://github.com/araffin/rl-baselines-zoo#pybullet-environments) for its "mujoco" environments. It would be neat if this project usurped Mujoco

shwang: I think I'll leave the `skipif` test in for now. Arguably we could also omit the test because we shouldn't ever make changes to the AIRL environments

shwang:
> More generally, what's the game plan here? I'd recommend against trying to match the results in the AIRL paper.

Thanks for bring this point up! Just finished a brief discussion about this with Sam. We agree with you that it's going to be hard to match the exact environments/code from the paper. 

My current plan is to compare results from `adamgleave/inverse_rl` on environments compatible with that codebase's dependencies and and results from this repo on the analogous environments compatible with this codebase's dependencies. (EG: disabled-ant from `airl.envs` vs disabled-ant from `imitation.examples.airl_envs`, or cartpole from `adamgleave/inverse_rl`'s Gym version and CartPole from the latest Gym version)

AdamGleave:
> > More generally, what's the game plan here? I'd recommend against trying to match the results in the AIRL paper.
> 
> My current plan is to compare results from `adamgleave/inverse_rl` on environments compatible with that codebase's dependencies and and results from this repo on the analogous environments compatible with this codebase's dependencies. (EG: disabled-ant from `airl.envs` vs disabled-ant from `imitation.examples.airl_envs`, or cartpole from `adamgleave/inverse_rl`'s Gym version and CartPole from the latest Gym version)

Seems like a good plan. Environments you've copied from the `adamgleave/inverse_rl` should match pretty closely. You can also run `adamgleave/inverse_rl` on Gym environments, which will be more different from the paper but perhaps more convenient/more diverse.

I'd check the experts you train are getting similar reward (ideally, test them in both environments). If there is a discrepancy maybe still OK if you understand why, just look at what % of performance the IRL algorithm gets.

Will review code again tomorrow (is late in the UK).

AdamGleave:
Also added tests for `CliffWorld`.

There was a bug where we tried each environment inside `test_envs`, even though it was parameterized by `ENV_NAMES`.

Please review my changes and merge if happy.

@qxcv I registered `CliffWorld` but I wasn't sure which values were sensible, so you may want to change those to be better defaults.

qxcv:
Thanks! I made the cliffworlds a bit wider & shorter. The agent only gets rewarded for getting from the far left to the far right, so usually it only moves up a few squares (say 2-3 squares in an environment that's 7 squares wide) to avoid the cliff before moving across.

shwang:
image

PR #57, opened 2019-07-24
Train high-performance expert policies in MuJoCo environments

AdamGleave:
Todos:
 - [X] Consistent format for serializing/deserializing policies
 - [X] Support for saving/restoring VecNormalize statistics
 - [x] Hyperparameter tuning
 - [x] Train experts on all MuJoCo tasks

PR #58, opened 2019-07-25
Gail benchmark

shwang:
`imitation.scripts.gail_benchmark` is a Sacred-based script meant to generate the results in the following table:
image

Started coding this up before I read Adam's PR #57 and knew about `parallel`. It seemed pretty cumbersome trying to shoehorn `train_ex` into `gail_benchmark`. Maybe it would have been easier had I used `Ingredient`s?

Commit 4f3fb24 "generate_trajectories_multiple" doesn't actually seem useful (was confused about whether loading experts or trajectory pickles was easier), I think I will drop this.


shwang:
@AdamGleave: Wondering if you could give `scripts/{config/,}gail_benchmark.py` a quick look? This seems unnecessarily messy.

I'm starting to think that instead of having a single `Experiment.main` `imitation.scripts.gail_benchmark.sh` that calls every possible `imitation.scripts.train_ex`, it might be better to have a shell script that uses `parallel` to call `imitation.scripts.train_ex`.

For easy scraping by another shell script afterwards, `train_ex` could write the mean and std_err of episode reward into `{log_dir}/results.txt`.

Maybe `sacred.Ingredients` are enough to fix this. A benefit to keeping everything in Python would be Ray-compatibility.

src/imitation/scripts/gail_benchmark.py
```diff
@@ -0,0 +1,50 @@
...
+      of `imitation.scripts.train.`).
+  """
+  all_results = {}
+  for config_updates in config_updates_list:
+    # TODO(shwang): This seems embarassingly parallellizable. (Ray later?)
+    config_updates["seed"] = _seed
+    config_updates["rollout_glob"] = rollout_glob.format(
+        env_name=config_updates["env_name"])
+    run = train_ex.run(config_updates=config_updates,
+                       named_configs=train_ex_named_configs)
```
shwang: Maybe "treating `train_ex` as an `Ingredient`" will automatically allow me to access namedconfigs of `train_ex` via the command line when calling `python -m imitation.scripts.train_ex`.

I can look closer at Ingredients API if this sounds about right.

experiments/download_experts.py
```diff
@@ -0,0 +1,14 @@
+import urllib
```
AdamGleave: This is OK for now, but I feel like making an S3 bucket or GCP Cloud Storage will be cleaner in the long run. If we want to add or update a model, we have to make a new ZIP file, change the URL, etc.

It's pretty easy to sync with S3, see e.g. https://github.com/HumanCompatibleAI/adversarial-policies/blob/master/experiments/pull_public_s3.sh

src/imitation/scripts/gail_benchmark.py
AdamGleave: Yes, I've added an ingredient and you can now use `python -m imitation.scripts.gail_benchmark with train.fast` (for example).

src/imitation/scripts/train.py
```diff
@@ -209,6 +219,16 @@ def should_plot_now(epoch) -> bool:
 
     save(trainer, os.path.join(log_dir, "final"))
 
+    # Final evaluation of imitation policy.
```
AdamGleave: This seems useful. If we end up wanting more complex evaluation we might want to consider putting it into `imitation.scripts.policy_eval` to stick to UNIX philosophy of one-tool-one-job.

src/imitation/util/rollout.py
```diff
@@ -321,85 +322,58 @@ def generate_transitions(policy, env, *, n_timesteps=None, n_episodes=None,
   return rollout_arrays
 
 
-def generate_transitions_multiple(policies, env, n_timesteps, *, truncate=True,
-                                  ) -> TransitionsTuple:
-  """Generate obs-act-obs-rew arrays from several policies.
+def generate_trajectories_multiple(
```
AdamGleave: I agree with you it seems like we can drop this function.

src/imitation/util/trainer.py
```diff
@@ -29,7 +32,14 @@ def init_trainer(env_id: str,
 
   Args:
     env_id: The string id of a gym environment.
-    rollout_glob: Argument for `imitation.util.rollout.load_transitions`.
+    rollout_glob: Argument for `imitation.util.rollout.load_trajectories`.
+    n_expert_demos: The number of expert trajectories to actually use
```
AdamGleave: :+1: good to be able to limit number of trajectories

shwang:
> In either case you should think about logging. By default I think environment name is included in output directory name, but not the number of demonstrations. You may want to make this hierarchical, or otherwise include the # of demonstrations and other config parameters.

I structured my script / results output in a way that I'm getting hierarchical directories for stdout.txt via `parallel --header : --results`.

shwang:
Example output (used `fast` named_config here):
```
/env_config/acrobot/n_demonstrations/1/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/0/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/1/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/1/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/1/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/2/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/10/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/0/stdout <==
[result] Mean Episode Return: -398 ± 0 (n=1)
/env_config/acrobot/n_demonstrations/10/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/1/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/10/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/2/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/4/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/0/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/4/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/1/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/4/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/2/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/7/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/0/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/7/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/1/stdout <==
[result] Mean Episode Return: -499 ± 0 (n=8)
/env_config/acrobot/n_demonstrations/7/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/Acrobot-v1.pkl/seed/2/stdout <==
[result] Mean Episode Return: -467 ± 0 (n=1)
/env_config/cartpole/n_demonstrations/1/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/CartPole-v1.pkl/seed/0/stdout <==
[result] Mean Episode Return: 17 ± 0 (n=2)
/env_config/cartpole/n_demonstrations/1/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/CartPole-v1.pkl/seed/1/stdout <==
[result] Mean Episode Return: 10 ± 0 (n=1)
...
/env_config/mountain_car/n_demonstrations/7/n_epochs/300/n_gen_steps_per_epoch/5000/rollout_glob/MountainCar-v0.pkl/seed/2/stdout <==
[result] Mean Episode Return: -199 ± 0 (n=8)
```

shwang:
I used #59 to generate demonstrations for every environment. (Ended up using about 233MB in demonstrations).

Until everything is merged together I don't expect others to be able to run this script your computer. I plan on adding something to the README that explains how to run `gail_benchmark` when I do #62).

shwang:
Interestingly, I've never gotten any Humanoid training to finish, even with "fast" and at least an hour on my MacBook Air. I'll have to take another look at this when doing #62. My first hypothesis is that even with `fast`, it's taking a really long time because Humanoid uses at least 80 trajectories at once. (Although the mechanism that causes so much slowness is not obvious to me)

experiments/gail_benchmark.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+# This script finds the the mean and standard error of episode return after
+# training GAIL on benchmark tasks.
+#
+# The CSV can be modified and regenerated from following spreadsheet:
```
AdamGleave: Explain this is a CSV config. At first I assumed it was output and was confused.

experiments/gail_benchmark.csv
```diff
@@ -0,0 +1,32 @@
+env_config,rollout_glob,n_gen_steps_per_epoch,n_epochs,n_demonstrations,
```
AdamGleave: Consider renaming to `gail_benchmark_config.csv`?

AdamGleave: Why is there an empty column at the end?

experiments/gail_benchmark.sh
```diff
@@ -0,0 +1,45 @@
...
+# https://docs.google.com/spreadsheets/d/1MZYTr23ddwhO2PrNI2EalKleoULXlsH5y9oqawt22pY/edit?usp=sharing
+#
+# When regenerating csv, remember that each line needs to end with ',' to
+# work properly.
+#
+# NOTE: Since I don't have walker experts available right now, I also
```
AdamGleave: This should ideally be addressed before merging.

experiments/gail_benchmark.sh
```diff
@@ -0,0 +1,45 @@
...
+TIMESTAMP=$(${DATE_CMD} --iso-8601=seconds)
+OUTPUT_DIR="output/gail_benchmark/${TIMESTAMP}"
+mkdir -p "${OUTPUT_DIR}"
+
+SEEDS="0 1 2"
+
+parallel -j 25% --header : --results ${OUTPUT_DIR}/parallel/ --colsep , --progress \
+  python -m imitation.scripts.train \
+  with \
+  fast \
```
AdamGleave: Do we want `fast`? We won't expect to reproduce GAIL results with this, right? Perhaps change to `with "$@"` so the bash script passes through command line arguments, letting you run it with `fast` if you want.

experiments/gail_benchmark.sh
```diff
@@ -0,0 +1,45 @@
...
+# by regenerating the csv after these experts are available.
+
+if $(command -v gdate > /dev/null); then
+  DATE_CMD=gdate  # macOS compatibility
+else
+  DATE_CMD=date
+fi
+
+TIMESTAMP=$(${DATE_CMD} --iso-8601=seconds)
+OUTPUT_DIR="output/gail_benchmark/${TIMESTAMP}"
```
AdamGleave: Consider adding an echo statement saying where it's logging to.

experiments/download_experts.py
AdamGleave: Why do we have this and `src/imitation/scripts/download_experts.py`? Surely only one is needed?

src/imitation/util/trainer.py
```diff
@@ -40,6 +52,9 @@ def init_trainer(env_id: str,
         files, as sorted by modification times.
     policy_dir: The directory containing the pickled experts for
         generating rollouts.
+    scale: If True, then scale input Tensors to the interval [0, 1].
+    entropy_weight: Only applicable for AIRL. The entropy_weight argument of
```
AdamGleave: Can't this be part of `discrim_kwargs`?

src/imitation/scripts/download_experts.py
```diff
@@ -0,0 +1,32 @@
+import urllib.request
+import zipfile
+
+experts_url = ("https://www.dl.dropboxusercontent.com/"
```
AdamGleave: FYI this zip contains a bunch of https://en.wikipedia.org/wiki/.DS_Store files, you may want to add that to the exclude list when generating the ZIP

shwang: Thanks for catching this, I'll keep the exclude in mind next time I generate

src/imitation/util/trainer.py
shwang: The GAIL discrim net doesn't have entropy_weight right now, so this breaks GAIL

shwang: Ah, we could just add an analogous entropy weight here:
image . I'll do this as part of the PR


There's a second entropy weight (\lambda in eq 18) that I'll want to add in later because some tuning this was important for Reacher results.
image

experiments/download_experts.py
shwang: My bad, I thought I had deleted this one

src/imitation/scripts/download_experts.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def main(verbose=True):
+  try:
+    zip_path, _ = urllib.request.urlretrieve(experts_url)
+  except urllib.error.URLError as e:
+    if "SSL" in str(e):
+      print(
+        "Suggestion: If on macOS, run "
+        r"`/Applications/Python\ 3.6/Install\ Certificates.command`"
```
shwang: newline or a space would be nice here

experiments/gail_benchmark.sh
shwang: My bad, I thought I had pushed a change that removed this to this. `"$@"` sounds like a great idea

experiments/gail_benchmark.csv
shwang: The empty column is somehow necessary for `parallel` to parse the `n_demonstrations` column correctly. Without it, the `n_demonstrations` is simply ignored. (I should clarify this in the script's comments)

experiments/download_experts.py
shwang: The reason I moved to `imitation.scripts` was because I believed it would be important for integration script or test later. Not sure why now. This seems more appropriate for `experiments/` so I'll move it back.

shwang: Ah, I think the reason I put it there was so that it would get linted and type checked for free. I'll quickly try adding lint and type check to `experiments/`.

shwang:
I'll have to make sure to read through my code next time I request a review! Habits have slipped since I've started research :)

experiments/gail_benchmark.csv
shwang: https://stackoverflow.com/questions/57280651/gnu-parallel-doesnt-work-without-trailing-commas-in-csv

experiments/gail_benchmark.sh
shwang: Looks like this isn't sufficient to have fast runs, because the `n_epochs=50000` overrides the named config! I'll probably add a `--fast` flag later when doing #62.

src/imitation/util/trainer.py
shwang: Actually, I'm not even sure if this is analogous/useful, because the DiscrimNetGAIL entropy weight would go in `DiscrimNetGAIL.discriminator_loss`, whereas the DiscrimNetAIRL entropy weight is in `DiscrimNetAIRL.build_policy_train`.

Let me know what you think

shwang:
> It'd be nice to have a single script or clear documentation. [...] I understand they're in separate PRs so integration may need to happen later.
Plan on working on this after the merges (related to #62, which is basically a integration test that does a fast version of the single-pipeline script.)

shwang:
Oops, n_gen_steps_per_epochs from CSV is ignored right now!

experiments/gail_benchmark.csv
AdamGleave: I see you've removed the empty columns now -- did you fix this? It seems like an odd error. I notice you have trailing spaces at the end of each line, which could potentially cause problems, but is probably OK (`int("4 ") == 4 == int("4")`).

experiments/gail_benchmark.sh
AdamGleave: Ah that is annoying! You could relocate the `"$@"` to the end, and then the user could override the previous `n_gen_steps_per_epoch` manually: `fast n_gen_steps_per_epoch=1`. But that's quite hacky. `--fast` flag is probably better.

src/imitation/util/trainer.py
AdamGleave: If there was a direct analogue of AIRL's `entropy_weight` in GAIL then it would make more sense to put it at the top-level, although I'm sitll hesitant about making too many things top-level configuration parameters (what if we want `train` to be used for something other than AIRL/GAIL in the future?) I'm not that familiar with GAIL but I think you're right there are similar but different concepts, which should probably have different names.

Configuration parameters that are silently ignored under some settings can get confusing quickly. Originally I was imagining that a default `entropy_weight` would get set by e.g. the `airl` named config. I now understand that you might want to set it on a per-environment basis. The options I see are:
  - 1. Use current approach of a top level config parameter that is ignored under GAIL, but add documentation explaining this.
  - 2. Set it in `discrim_kwargs`, but in the environment named config check that `use_gail` is False before setting it. This is self-documenting, but would break if you executed the environment named config before the AIRL config.
  - 3. Make `discrim_kwargs` and `reward_kwargs` be a dict of algorithm names to parameters, with a `'default'` key that applies to all algorithms. The environment named config can then just set `discrim_kwargs['airl']['entropy_weight'] = 1.0`.
I think right now I favour solution (3), but none of them seem great.

experiments/gail_benchmark.csv
shwang: Hmm, I don't see trailing spaces in the CSV. 
```
shwangster@airbears2-10-142-35-202:/tmp/parallel$ cat -vet gail_benchmark_config.csv 
env_config,rollout_glob,n_gen_steps_per_epoch,n_epochs,n_demonstrations$
cartpole,CartPole-v1.pkl,5000,300,1$
cartpole,CartPole-v1.pkl,5000,300,4$
cartpole,CartPole-v1.pkl,5000,300,7$
cartpole,CartPole-v1.pkl,5000,300,10$
mountain_car,MountainCar-v0.pkl,5000,300,1$
```

I suppose you mean the actual parallel command.

shwang: Yeah, it's fixed now! This is the title of one of the commits 
e0d9ebb

src/imitation/util/trainer.py
shwang: thanks for the ideas. (iii) seems like the best long term solution but also crufty

shwang: I think I will do something like (iii) if we get another algorithm-specific variable like `entropy_weight` -- (i) seems not-too-bad to me right now.  Adding in a sentence or two of documentation and renaming variable to `airl_entropy_weight` for now.

AdamGleave:
I swear I saw trailing spaces in the CSV but only have my phone now so
can't check. Maybe I was looking at an old revision.

shwang:
image

PR #59, opened 2019-07-26
Add "rollouts_from_policy" command to `data_collect`

shwang:
The new command takes a `policy_path` and loads the pickled policy to generate `rollouts/{env_name}.pkl`.

I used this to [generate demonstrations](https://gist.github.com/shwang/010da07b8d0a1232f3f5f00ac18bfa5b) from Adam's trained expert models in #57.

### DIFF
--- a/src/imitation/scripts/data_collect.py
+++ b/src/imitation/scripts/data_collect.py
@@ -25,7 +25,7 @@ def data_collect(_seed: int,
 
                  rollout_save_interval: int = 0,
                  rollout_save_final: bool = False,
-                 rollout_save_n_samples: Optional[int] = None,
+                 rollout_save_n_timesteps: Optional[int] = None,
                  rollout_save_n_episodes: Optional[int] = None,
 
                  policy_save_interval: int = -1,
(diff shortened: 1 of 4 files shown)


src/imitation/scripts/data_collect.py
```diff
@@ -46,16 +48,19 @@ def data_collect(_seed: int,
...
-      rollout_save_n_samples: The minimum number of timesteps saved in every
-          file. Could be more than `rollout_save_n_samples` because trajectories
-          are saved by episode rather than by transition.
-
+      rollout_save_n_timesteps: The minimum number of timesteps saved in every
+          file. Could be more than `rollout_save_n_timesteps` because
+          trajectories are saved by episode rather than by transition.
+          Must set exactly one of `rollout_save_n_timesteps`
+          and `rollout_save_n_episodes`.
+      rollout_save_n_episodes: The number of episodes saved in every
```
AdamGleave: Can we add some input validation for this so it fails fast? Right now it will only fail when we actually try to save the rollouts, which in the worst case could be at the very end of policy training.

shwang: Good call

PR #67, opened 2019-07-31
Use Stable Baselines 2.7.0 PyPi release

AdamGleave:
Stable Baselines just made a release, we can stop tracking their master branch


shwang:
image

PR #68, opened 2019-07-31
Train experts for AIRL environments

shwang:
For AIRL custom envs that were reported in the paper:
  * Add named configs
  * Include in `experiments/mujoco_experts.sh`


shwang:
Training right now, will report results when they are available.

shwang:
Expert results:

```
(imit) steven@astar:~/imitation/output/mujoco_experts/2019-07-31T16:41:33-07:00$ find . -name stdout | xargs tail -n 15 | grep -E '(==|ep_reward_mean)'
==> ./parallel/env/acrobot/seed/0/stdout <==
| ep_reward_mean     | -75.1         |
==> ./parallel/env/acrobot/seed/2/stdout <==
| ep_reward_mean     | -71.2         |
==> ./parallel/env/acrobot/seed/1/stdout <==
| ep_reward_mean     | -72.8         |

==> ./parallel/env/reacher/seed/0/stdout <==
| ep_reward_mean     | -6.04        |
==> ./parallel/env/reacher/seed/2/stdout <==
| ep_reward_mean     | -6.27          |
==> ./parallel/env/reacher/seed/1/stdout <==
| ep_reward_mean     | -6.02        |

==> ./parallel/env/swimmer/seed/0/stdout <==
| ep_reward_mean     | 43.2          |
==> ./parallel/env/swimmer/seed/2/stdout <==
| ep_reward_mean     | 48.5          |
==> ./parallel/env/swimmer/seed/1/stdout <==
| ep_reward_mean     | 52.6         |

==> ./parallel/env/cartpole/seed/0/stdout <==
| ep_reward_mean     | 480           |
==> ./parallel/env/cartpole/seed/2/stdout <==
| ep_reward_mean     | 386           |
==> ./parallel/env/cartpole/seed/1/stdout <==
| ep_reward_mean     | 443           |

==> ./parallel/env/hopper/seed/0/stdout <==
| ep_reward_mean     | 1.71e+03       |
==> ./parallel/env/hopper/seed/2/stdout <==
| ep_reward_mean     | 1.87e+03     |
==> ./parallel/env/hopper/seed/1/stdout <==
| ep_reward_mean     | 1.87e+03     |

==> ./parallel/env/disabled_ant/seed/0/stdout <==
| ep_reward_mean     | 884          |
==> ./parallel/env/disabled_ant/seed/2/stdout <==
| ep_reward_mean     | 885           |
==> ./parallel/env/disabled_ant/seed/1/stdout <==
| ep_reward_mean     | 805          |

==> ./parallel/env/humanoid/seed/0/stdout <==
| ep_reward_mean     | 4.22e+03     |
==> ./parallel/env/humanoid/seed/2/stdout <==
| ep_reward_mean     | 3.49e+03     |
==> ./parallel/env/humanoid/seed/1/stdout <==
| ep_reward_mean     | 2.53e+03    |

==> ./parallel/env/mountain_car/seed/0/stdout <==
| ep_reward_mean     | -101          |
==> ./parallel/env/mountain_car/seed/2/stdout <==
| ep_reward_mean     | -100          |
==> ./parallel/env/mountain_car/seed/1/stdout <==
| ep_reward_mean     | -97.6         |

==> ./parallel/env/walker/seed/0/stdout <==
| ep_reward_mean     | 3.7e+03      |
==> ./parallel/env/walker/seed/2/stdout <==
| ep_reward_mean     | 2.89e+03     |
==> ./parallel/env/walker/seed/1/stdout <==
| ep_reward_mean     | 3.94e+03    |

==> ./parallel/env/two_d_maze/seed/0/stdout <==
| ep_reward_mean     | -5.02         |
==> ./parallel/env/two_d_maze/seed/2/stdout <==
| ep_reward_mean     | -40.8       |
==> ./parallel/env/two_d_maze/seed/1/stdout <==
| ep_reward_mean     | -41.7       |

==> ./parallel/env/ant/seed/0/stdout <==
| ep_reward_mean     | 4.09e+03     |
==> ./parallel/env/ant/seed/2/stdout <==
| ep_reward_mean     | 3.22e+03     |
==> ./parallel/env/ant/seed/1/stdout <==
| ep_reward_mean     | 2.53e+03     |

==> ./parallel/env/half_cheetah/seed/0/stdout <==
| ep_reward_mean     | 4.86e+03     |
==> ./parallel/env/half_cheetah/seed/2/stdout <==
| ep_reward_mean     | 2.41e+03     |
==> ./parallel/env/half_cheetah/seed/1/stdout <==
| ep_reward_mean     | 3.82e+03     |

==> ./parallel/env/custom_ant/seed/0/stdout <==
| ep_reward_mean     | 1.83e+03     |
==> ./parallel/env/custom_ant/seed/2/stdout <==
| ep_reward_mean     | 1.94e+03     |
==> ./parallel/env/custom_ant/seed/1/stdout <==
| ep_reward_mean     | 1.96e+03     |
```

shwang:
We didn't reach expert on CartPole (500)! The other tasks look okay AFAIK without knowing the expert reward of any of them except of MountainCar (-110 is considered solved).

AdamGleave:
Most of these tasks don't have an established threshold for optimal performance, but look in RL papers to see what rewards people typically get. PPO, SAC and Rainbow paper are good sources.

Eyeballing it, Reacher, Hopper, HalfCheetah, Humanoid, Ant, MountainCar all seem to be performing well. Cartpole should be able to get to 500 but 450 isn't all that bad. I don't know about the other environments.

I'm not familiar with the custom environments, but Justin's paper gives some numbers for a TRPO expert (table 1 & 2). We seem to do a lot better on `disabled_ant` (880 vs 315). `custom_ant` seems closer, we get ~1.8k and they report ~1.5k. I'd expect PPO to do a little better than TRPO, but the difference in `disabled_ant` seems extreme. Possible they had different length episodes, or that the MuJoCo version change is really significant.

shwang:
eyeballed some numbers and noted version changes.
image

For these mujoco environments, the difference between -v1 and -v2 should be insignificant. (See https://github.com/openai/gym/pull/834)

shwang:
image

Our best results out of 3 do better than some of their reported PPO means but worse on Humanoid and Hopper. I'm not going to worry about our experts not matching previously reported experts for now -- just noting in case we want to think about this later on.

shwang:
> but the difference in disabled_ant seems extreme. Possible they had different length episodes, or that the MuJoCo version change is really significant.

Maybe (don't have a good prior on how much Mujoco change matters). I don't see why the episode length would change if I ported the code correctly.

If disabled_ant reward discrepancies seems important to look over later, I can carefully read through https://github.com/HumanCompatibleAI/imitation/pull/55 to make sure that I didn't accidentally make a change to the environment when porting it over.

shwang:
I preemptively commited `registry.load` into this PR because it was helpful for handling a PPO1 failure when `mpipy4` isn't installed.

AdamGleave:
> Maybe (don't have a good prior on how much Mujoco change matters). I don't see why the episode length would change if I ported the code correctly.
> 
> If disabled_ant reward discrepancies seems important to look over later, I can carefully read through #55 to make sure that I didn't accidentally make a change to the environment when porting it over.

Have we replicated the disabled ant experiment using Justin's codebase? If not, we shouldn't assume that the environment (and in particular episode length) in that code and the paper is the same.

A 2x difference feels significant to me. Worth quickly running the code in Justin's codebase. Seems worth running https://github.com/AdamGleave/inverse_rl/blob/master/scripts/ant_data_collect.py but with the environment changed to 'airl/DisabledAnt-v0' to see what expert performance we get with TRPO in original codebase.

> Our best results out of 3 do better than some of their reported PPO means but worse on Humanoid and Hopper. I'm not going to worry about our experts not matching previously reported experts for now -- just noting in case we want to think about this later on.

This seems reasoable. The differences aren't huge nor are they always in one direction, so seems plausibly down to differently tuned hyperparameters and random seeds. If we were going to do a publication-quality benchmark of different IRL methods we might want to push more on getting the best possible experts, but for testing correctness of our implementation of IRL methods, I think the current experts are fine.

> For these mujoco environments, the difference between -v1 and -v2 should be insignificant. (See openai/gym#834)

Thanks, I'd seen this a while ago but forgot about it, does seem like the MuJoCo version different shouldn't matter that much.

src/imitation/policies/serialize.py
```diff
@@ -85,17 +85,21 @@ def _load_zero(path: str, env: gym.Env) -> ZeroPolicy:
...
     'random': _load_random,
     'zero': _load_zero,
 }
-for k, (cls, attr) in STABLE_BASELINES_CLASSES.items():
-  AGENT_LOADERS[k] = _load_stable_baselines(cls, attr)
+for k, (cls_name, attr) in STABLE_BASELINES_CLASSES.items():
+  try:
+    cls = registry.load_attr(cls_name)
+    AGENT_LOADERS[k] = _load_stable_baselines(cls, attr)
+  except (AttributeError, ImportError):
```
AdamGleave: Consider adding comment to explain we have optional dependencies so this is an expected/non-critical error

src/imitation/scripts/config/data_collect.py
```diff
@@ -100,6 +103,33 @@ def walker():
...
+
+@data_collect_ex.named_config
+def two_d_maze():
+  env_name = "imitation/TwoDMaze-v0"
+
+
+@data_collect_ex.named_config
+def custom_ant():
+  env_name = "imitation/CustomAnt-v0"
+  make_blank_policy_kwargs = dict(
```
AdamGleave: Do we envisage a situation where `ant`, `custom_ant` and `disabled_ant` would have different hyperparameters? If not, can we factor out into a helper function/dictionary? Sadly might need to be a bit hacky, along the lines of `locals().update(...)`

src/imitation/scripts/config/train.py
```diff
@@ -140,6 +141,30 @@ def swimmer():
...
+
+# Custom env configs
+
+@train_ex.named_config
+def two_d_maze():
+  env_name = "imitation/TwoDMaze-v0"
+
+
+@train_ex.named_config
+def custom_ant():
```
AdamGleave: Why does `ant()` get `n_epochs = 2000` but not `custom_ant` or `disabled_ant`?

I also suspect we may need to do more hyperparameter tuning here, but that's probably another PR.

tests/test_policies.py
```diff
@@ -42,7 +42,14 @@ def test_serialize_identity(env_name, model_cfg, normalize):
   vec_normalize = None
   if normalize:
     venv = vec_normalize = VecNormalize(venv)
-  model_name, model_cls = model_cfg
+
+  model_name, model_cls_name = model_cfg
```
AdamGleave: I think `cls` in `BASELINE_MODELS` should be renamed to `cls_name` then?

src/imitation/scripts/config/data_collect.py
shwang: Not sure if `locals().update(...)` is okay. (Maybe sacred does some magic that makes this work.)

https://docs.python.org/3/library/functions.html#locals

I'll give it a quick shot though

src/imitation/scripts/config/train.py
shwang: Haven't thought about hp tuning at all. Plan is to think about that after there's most of the pieces of the AIRL/GAIL pipelines available.

Adding `n_epochs = 2000` to the new configs.

src/imitation/scripts/config/data_collect.py
shwang: It works!

shwang:
> Have we replicated the disabled ant experiment using Justin's codebase? If not, we shouldn't assume that the environment (and in particular episode length) in that code and the paper is the same.
Looking into this now

shwang:
## DisabledAnt
For DisabledAnt experiment (legacy TRPO, 1 seed) I get around 390 Average Return.

    2019-08-17 13:25:20.114989 PDT | AverageReturn              380.602
    2019-08-17 13:25:33.891566 PDT | AverageReturn              391.266
    2019-08-17 13:26:17.628041 PDT | AverageReturn              396.033
    2019-08-17 13:26:20.762647 PDT | AverageReturn              390.332
    2019-08-17 13:26:29.459419 PDT | AverageReturn              379.702
    2019-08-17 13:26:42.913788 PDT | AverageReturn              396.057

The paper reports 315.5.
Our PPO (best of 3) reports 880, more than a 2x change.

## CustomAnt
For CustomAnt experiment (legacy TRPO, 1 seed) I get around 800 Average Return.

The paper reports 1537.9 Average Return.

My PPO (best of 3) reports 1960 Average Return, more than 2x change.

AdamGleave:
These are substantial differences. It could just be a PPO vs TRPO difference: the PPO paper did report quite substantial improvements on some MuJoCo differences (e.g. HalfCheetah, Reacher) while others were comparable (e.g. Hopper, Swimmer, InvertedPendulum). 

Do we definitely have a consistent episode length (it should be reported somewhere...)? In https://github.com/shwang/inverse_rl/blob/master/scripts/ant_data_collect.py the `max_path_length` is set to `500`, whereas `CustomAntEnv` in https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/envs/examples/airl_envs/ant_env.py#L546 has `max_timesteps` set to `1000`. So if you're not correcting for this, we will get around a 2x better episode reward.

shwang:
Ah, thanks for catching that! That would definitely make a ~2x change.

PR #71, opened 2019-08-07
Toy MNIST environment

mtrazzi:
I've built a straightforward gridworld environment where observations are MNIST digits and  rewards are the digit displayed. The end goal would be to add more environments in the folder `env_suite`, for the two use cases:
- _Engineering_: trying to implement an existing algorithm, and wanting to test if it meets the spec.
- _Research_: understanding the strengths and weaknesses of different algorithms.

For now:
- I've collected expert data for a 2x2 Mnist gridworld [here](https://www.dropbox.com/s/bikprbaaxlip27k/Mnist2x2-v0-PPO2.pkl?dl=0).
- I have tested the env with AIRL (using [this](https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/scripts/train.py) script), but `ep_reward_mean` didn't improve after 1e6 episodes. The policy training seems functional though (when I set `debug_use_ground_truth` to `True`, it correctly learns the optimal policy).

@AdamGleave Does the current mnist env./TODO correspond to what you had in mind? Any guesses for why the AIRL script doesn't work?


AdamGleave:
> @AdamGleave Does the current mnist env./TODO correspond to what you had in mind? Any guesses for why the AIRL script doesn't work?

I've only looked at the implementation briefly, but if I'm understanding it correctly it's a partially observable gridworld where:
  - You see the digit in the cell you're currently in.
  - You get a reward equal to the value of the digit.
  - The digits in the gridworld are preserved across episodes (unclear if they're preserved across environment reloading). Resetting an episode leads to starting in a random position.

I'm a bit puzzled as to how a policy without memory can solve this. It only sees the current digit, and doesn't know anything about other cells or even where they are. One possibility is to learn a heuristic that if the current cell is less than e.g. 5, take a random action; otherwise, stay put.

If the expert policy is doing better than this, then either I've misunderstood something (very possible) or it must have memorized the particular digits it saw. In this case I don't think there's enough data for IRL learn from, and I suspect the policy would not transfer to an environment with different randomly sampled MNIST digits.

I'd suggest modifying the environment to be closer to the original MNIST classification task. You observe a digit, you take an action from 0..9, and get reward if your action matches the true label. Then, regardless of the action you took, you see a new digit.

I like the gridworld idea, I think that's worth including too, but for it to work well I think you need to give the agent a view of the whole world and to randomize the digits across episodes. This'll be harder to train, so I'd check everything works with the classification task first.

I don't know what network architecture you were using, but note the default in this code (fully-connected, two-hidden layers, 64 hidden units) is meant for low-dimensional MuJoCo tasks and is probably too narrow and shallow for MNIST. I'd lean towards using CNNs, although you could probably do it with an appropriately sized FCN.

mtrazzi:
> I'm a bit puzzled as to how a policy without memory can solve this.

Agreed. The only reason it worked with behavourial cloning was that the map didn't change.

> I'd suggest modifying the environment to be closer to the original MNIST classification task.

I've updated the environment to a "1 step per episode" task where it gets a reward if it guessed the true label. Regardless of its action, it then moves to some absorbing state (i.e. observation is an array of zeroes and the episode is done). I previously implemented "sees a new digit, regardless of the action" with only one step, but removed it as it only complexified the task. I can add it again if necessary.

> the default in this code (fully-connected, two-hidden layers, 64 hidden units) is meant for low-dimensional MuJoCo tasks and is probably too narrow and shallow for MNIST. I'd lean towards using CNNs

I guess you're refering to the `MlpPolicy` in stable-baselines, which reaches a threshold of about 70% accuracy after 100k timesteps with AIRL. I've added a `MnistCnnPolicy` (modified version of the `CnnPolicy` from stable-baselines to match the dimensions of Mnist) to see if I could get better results, but training is taking much longer, so I'm still waiting for the results.

src/imitation/policies/base.py
```diff
@@ -101,3 +121,24 @@ def get_action_policy(policy, observation, deterministic=True):
...
+  """
+  Policy object that implements actor critic, using a CNN (the Mnist CNN)
+  :param sess: (TensorFlow session) The current TensorFlow session
+  :param ob_space: (Gym Space) The observation space of the environment
+  :param ac_space: (Gym Space) The action space of the environment
+  :param n_env: (int) The number of environments to run
+  :param n_steps: (int) The number of steps to run for each environment
+  :param n_batch: (int) The number of batch to run (n_envs * n_steps)
+  :param reuse: (bool) If the policy is reusable or not
+  :param _kwargs: (dict) Extra keyword arguments for feature extraction
```
AdamGleave: Why `_kwargs` not `kwargs`?

src/imitation/policies/base.py
```diff
@@ -4,8 +4,28 @@
...
 import numpy as np
+from stable_baselines.a2c.utils import conv, conv_to_fc, linear
 from stable_baselines.common import BaseRLModel
 from stable_baselines.common.policies import BasePolicy, FeedForwardPolicy
+import tensorflow as tf
+
+
+def mnist_cnn(scaled_images, **kwargs):
+  """
+  Tweakeable CNN.
```
AdamGleave: Should be "Tweakable"

src/imitation/scripts/config/data_collect.py
```diff
@@ -104,3 +104,10 @@ def walker():
 def fast():
   """Intended for testing purposes: small # of updates, ends quickly."""
   total_timesteps = int(1e4)
+
+
+@data_collect_ex.named_config
+def mnist():
+  env_name = "imitation/Mnist-v0"
+  total_timesteps = int(1e4)
```
AdamGleave: This seems far too small to get a good policy -- intentional?

src/imitation/util/util.py
```diff
@@ -64,6 +64,7 @@ def make_vec_env(env_id: str,
       log_dir: If specified, saves Monitor output to this directory.
   """
   def make_env(i):
+    import imitation.examples.env_suite  # noqa: F401
```
AdamGleave: Best to have imports be top-level where possible, see http://google.github.io/styleguide/pyguide.html#313-imports-formatting

tests/test_envs.py
```diff
@@ -3,6 +3,7 @@
 
 # Import for side-effect of registering environment
 import imitation.examples.airl_envs  # noqa: F401
+import imitation.examples.env_suite  # noqa: F401
```
AdamGleave: @shwang it seems like we're collecting a number of custom environments, how do you think we should structure them? Perhaps have a single import that registers all of them, but have a hierarchy underneath. Gym should only actually import the entry point when we try and make the environment.

AdamGleave:
Thanks for the changes, let me know when you have results with the new environment & CNN policy.

src/imitation/policies/base.py
mtrazzi: No strong feelings for `_kwargs` (copy-pasted from [stable-baselines](https://github.com/hill-a/stable-baselines/blob/63f208eb21c67fd1a50a691b98ee54cdabc1308a/stable_baselines/common/policies.py#L598)). Updated it to kwargs.

src/imitation/scripts/config/data_collect.py
mtrazzi: Used it when debugging, nice catch. Updating it to `1e6` to get a 99% optimal policy.

mtrazzi:
*For the CNN*: using AIRL, the current training config (i.e. `n_gen_steps_per_epoch = 10000`, `n_disc_steps_per_epoch = 10`, `n_epochs = 100`) and the ground_truth as the expert model, it reaches 82% of optimal performance. This is 10% better than just MLP, but still way inferior to the 99.25% accuracy we should get from [supervised learning with same architecture](https://github.com/keras-team/keras/blob/master/examples/mnist_cnn.py). This is even more concerning as the environment always sample from the same training set, so it should be able to overfit at some point.

*For the build*: there was a problem with previous builds in building html with Spinx (not finding `contents.rst`). I have no clue how to fix that. As for codecov, it seems to be only in the new configs & CNN, so will fix that with 2/3 additional unit tests.

AdamGleave:
> _For the CNN_: using AIRL, the current training config (i.e. `n_gen_steps_per_epoch = 10000`, `n_disc_steps_per_epoch = 10`, `n_epochs = 100`) and the ground_truth as the expert model, it reaches 82% of optimal performance. This is 10% better than just MLP, but still way inferior to the 99.25% accuracy we should get from [supervised learning with same architecture](https://github.com/keras-team/keras/blob/master/examples/mnist_cnn.py). This is even more concerning as the environment always sample from the same training set, so it should be able to overfit at some point.
This is worse than I'd have expected, although AIRL hyperparameters are not tuned for image-based problems, so it's possible.

I'd suggest trying an RL algorithm on this environment, to see if it gets to similar performance to a supervised learner. The hyperparameters people use for Atari should work reasonably well as similar dimensions.

Will try to take a closer look at the code & build errors this week, but may not have time.

mtrazzi:
> I'd suggest trying an RL algorithm on this environment, to see if it gets to similar performance to a supervised learner. The hyperparameters people use for Atari should work reasonably well as similar dimensions.

Using the `data_collect` script with the atari hyperparams, a RL agent using PPO2 & CNN achieves about 99.75%, and PPO2 & MLP reaches 99%, similar to what we would get using supervised learning.

> Will try to take a closer look at the code & build errors this week, but may not have time.

No worries, I'll try to fix the build errors today.

tests/test_envs.py
shwang: A single import like `import imitation.examples` registering all the environments sounds good. Also, we should maybe rename `examples` to `envs` (or `examples.envs`) sometime.

AdamGleave:
> Using the `data_collect` script with the atari hyperparams, a RL agent using PPO2 & CNN achieves about 99.75%, and PPO2 & MLP reaches 99%, similar to what we would get using supervised learning.

Thanks for running this experiment, that RL works confirms the environemnt is reasonable. It'd be interesting to work out why AIRL does  so badly here. Some experiments I'd suggest, roughly in order:
  - Set the `debug_use_ground_truth` flag (via `init_trainer_kwargs` and `trainer_kwargs`) to `True`. This effectively runs AIRL short-circuiting the discriminator, and giving the true environment reward. If this fails, then either the RL part of AIRL is broken/badly tuned or we are using a policy with insufficient capacity. Conversely, if it works (should get similar performance to using PPO2 directly), then it's a problem with the discriminator.
  - Hyperparameter tuning: just try a small random search. Also look at training curves, see if it looks like it's converged or not. `n_epochs=100` isn't that much (produces ~10M generator timesteps; about what's needed for good performance in Atari in RL, would expect MNIST to be easier though, but IRL is harder).
  - Train a `RewardNet` via regression on the ground-truth reward. Now substitute in that pre-trained reward model and keep training. Does it work now?
  - If not, load the pre-trained reward model *and* pre-trained policy (on the ground truth). Does it work now (i.e. is a correct reward and policy a stable fixed point at least)?

No obligation to run these, of course -- it's not your responsibility to debug our AIRL implementation :) But it'd be helpful, and might give you a better sense of how these test environments can be useful.

mtrazzi:
With `debug_use_ground_truth flag = True`, it gets about perfect accuracy, so closer to what we get with PPO2, which means there might be a problem with the discriminator.

What do you think would be the relevant hyperparameters for a random search? The most obvious one seems to be the `airl_entropy_weight`, but there's also `n_{disc,gen}_steps_per_epoch`, buffer capacity and disc samples in `trainer_kwargs`, the hyperparams for `make_blank_policy_kwargs` or even the shape of the `{theta,phi}_units`.

As for the low number of epochs and tensorboards, [here](https://www.dropbox.com/s/8fzmhqyn8an6nho/events.out.tfevents.1566223410.fhiserver?dl=0) are the tensorboards I get with 200 epochs, with  n_gen_steps_per_epoch=2048 and n_disc_steps_per_epoch=50. There seems to be convergence, at least in terms of `ep_reward_mean` around 0.85 (ie. 85% accuracy).

Will try to figure out the last two bullet points today or tomorrow.

AdamGleave:
> With `debug_use_ground_truth flag = True`, it gets about perfect accuracy, so closer to what we get with PPO2, which means there might be a problem with the discriminator.
OK, this is good to know.

> What do you think would be the relevant hyperparameters for a random search? The most obvious one seems to be the `airl_entropy_weight`, but there's also `n_{disc,gen}_steps_per_epoch`, buffer capacity and disc samples in `trainer_kwargs`, the hyperparams for `make_blank_policy_kwargs` or even the shape of the `{theta,phi}_units`.

I'd guess the most important ones for this problem are `n_{disc,gen}_steps_per_epoch`. GAN training is often unstable: if either the discriminator or generator gets too strong too quickly it can prevent the other network from getting a gradient it needs to train. Changing these parameters (and especially their ratio) lets us control this.

The buffer size and discriminator samples are also important. For now you can just set this to a large value and not bother searching over it: larger values will almost always work better, but will slow down training.

Given it works OK with `debug_use_ground_truth`, the policy parameters seem OK. I'd recommend using the same architecture for the reward and policy network. That's generally a safe bet, and here it seems like the only sensible option, since in a 1-step decision problem the reward and policy are doing basically the same amount of work.

> As for the low number of epochs and tensorboards, [here](https://www.dropbox.com/s/8fzmhqyn8an6nho/events.out.tfevents.1566223410.fhiserver?dl=0) are the tensorboards I get with 200 epochs, with n_gen_steps_per_epoch=2048 and n_disc_steps_per_epoch=50. There seems to be convergence, at least in terms of `ep_reward_mean` around 0.85 (ie. 85% accuracy).
>
I agree that shows clear convergence. Now more suspicious the discriminator capacity is too small. Are you using a smaller reward net than policy net?

> Will try to figure out the last two bullet points today or tomorrow.

mtrazzi:
> I'd guess the most important ones for this problem are n_{disc,gen}_steps_per_epoch.

Thanks, will tweak those hyperparams.

> Are you using a smaller reward net than policy net?

I'm using this 3 layer CNN with more capacity than the reward net, nice catch.

> I'd recommend using the same architecture for the reward and policy network.

Got it. Two possibilities: 1) go back to just using MLP everywhere, but with 300+100 hidden units instead of 32+32, both for the reward and policy network. 2) plug in CNNs in the theta and phi networks of the reward net.

Will start with 1), as it seems the most likely to work given the architectures in the AIRL paper & the repo.

mtrazzi:
*Update*:
- Last week I've tried to have the same architecture/capacity for both the reward and policy net. By using FeedForward with 300+100 hidden layers, I get some improvement compared to just 32+32, but it's still somehow inferior than using CNN policy & MLP reward net. I've also tried using CNN for both the reward net and the policy net, but that required changing most of the codebase for the reward nets, so I gave up. I got some marginal gains by tweaking the generator/discriminator steps, but nothing that improved substantially the results. A google sheets for some runs can be found [here](https://bit.ly/2L8XaGo).
- I'm closing this issue as my internship at FHI is ending, and I won't have time to work on this in the next days/weeks. Happy to re-open this whenever I get more time & we define more steps for debugging.

AdamGleave:
I hope you enjoyed your internship at FHI! Thanks for sharing those results. It's unfortunate we don't have a CNN reward net implementation, we should consider adding one at some point. It's good to see results improved using a larger MLP, although I'm still surprised they're not better, given fairly small MLPs can get good performance on MNIST in supervised learning setting. It's possible this is a real limitation of AIRL's performance.

PR #72, opened 2019-08-08
Model-based environment ABC for non-tabular environments

AdamGleave:
This PR introduces a new abstract class for model-based environments that does not make any assumption about the observation/action space. The current `ModelBasedEnv` is renamed to `TabularModelEnv` and subclasses this.

There are a very broad class of environments in common use that are model-based but not tabular. Any MuJoCo environment, for example -- although exposing the dynamics of MuJoCo environments is a little involved due to internal state in the optimizer (see e.g. https://github.com/HumanCompatibleAI/adversarial-policies/blob/master/src/aprl/agents/monte_carlo.py#L12)

Although we don't currently have algorithms that can benefit from model-based but non-tabular environments, this seems like a plausible category: there've been various discussions of trying model-based RL and IRL. Additionally, it's useful for post-hoc analysis of reward models. For example, it lets you compute a variety of state-action-next state distributions, e.g. you can randomly sample an observation and action and then compute the next state. Having a state distribution is essential for knowing what areas of the reward model to evaluate and which can be safely ignored (unreachable states, physically impossible transitions, etc).

### DIFF
--- a/src/imitation/model_env.py
+++ b/src/imitation/model_env.py
@@ -2,6 +2,7 @@
 are handy when you want to perform exact maxent policy optimisation."""
 
 import abc
+from typing import Optional
 
 import gym
 from gym import spaces
@@ -9,6 +10,64 @@ import numpy as np
 
 
 class ModelBasedEnv(gym.Env, abc.ABC):
+  """ABC for environments with known dynamics."""
+
+  n_actions_taken: Optional[int]
+  """Number of steps taken so far"""
+
+  def __init__(self):
+    # TODO(gleave): anything needed here?
+    self.n_actions_taken = None
+
+  def seed(self, seed=None):
+    if seed is None:
+      # Gym API wants list of seeds to be returned for some reason, so
+      # generate a seed explicitly in this case
+      seed = np.random.randint(0, 1 << 31)
+    self.rand_state = np.random.RandomState(seed)
+    return [seed]
+
+  @abc.abstractmethod
+  def initial_state(self):
+    """Samples from the initial state distribution."""
+
+  @abc.abstractmethod
+  def obs_from_state(self, state):
+    """Returns observation produced by a given state."""
+
+  @abc.abstractmethod
+  def reward(self, old_state, action, new_state):
+    """Computes reward for a given transition."""
+
+  @abc.abstractmethod
+  def transition(self, old_state, action):
+    """Samples from transition distribution."""
+
+  @abc.abstractmethod
+  def terminal(self, state, step: int) -> bool:
+    """Is the state terminal?"""
+
+  def reset(self):
+    self.cur_state = self.initial_state()
+    self.n_actions_taken = 0
+    return self.obs_from_state(self.cur_state)
+
+  def step(self, action):
+    if self.cur_state is None or self.n_actions_taken is None:
+      raise ValueError("Need to call reset() before first step()")
+
+    old_state = self.cur_state
+    self.cur_state = self.transition(self.cur_state, action)
+    obs = self.obs_from_state(self.cur_state)
+    rew = self.reward(old_state, action, self.cur_state)
+    done = self.terminal(self.cur_state, self.n_actions_taken)
+    self.n_actions_taken += 1
+
+    infos = {"old_state": old_state, "new_state": self.cur_state}
+    return obs, rew, done, infos
+
+
+class TabularModelEnv(ModelBasedEnv, abc.ABC):
   """ABC for tabular environments with known dynamics."""
 
   def __init__(self):
@@ -16,6 +75,7 @@ class ModelBasedEnv(gym.Env, abc.ABC):
     including current state & number of actions taken so far (initial None,
     so that error can be thrown if reset() is not called), attributes for
     cached observation/action space, and random seed for rollouts."""
+    super().__init__()
     self.cur_state = None
     self.n_actions_taken = None
     # Constructing action & observation spaces requires self.n_actions and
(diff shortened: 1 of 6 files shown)


tests/test_envs.py
```diff
@@ -1,27 +1,114 @@
...
-# Import for side-effect of registering environment
+# Unused imports are for side-effect of registering environments
 import imitation.examples.airl_envs  # noqa: F401
 import imitation.examples.model_envs  # noqa: F401
 
 ENV_NAMES = [env_spec.id for env_spec in gym.envs.registration.registry.all()
              if env_spec.id.startswith('imitation/')]
 
+DETERMINISTIC_ENVS = [
+    r"imitation/CliffWorld.*-v0",
```
qxcv: This name is confusing because CliffWorld is not actually deterministic (it just has a deterministic initial state distribution). Maybe call it `DETERMINSTIC_INITIAL_STATE_ENVS` or `DETERMINISTIC_S0_ENVS`?

src/imitation/model_env.py
```diff
@@ -9,40 +9,56 @@
 
 
 class ModelBasedEnv(gym.Env, abc.ABC):
```
qxcv: Nitpick: none of the abstract methods in this class actually require known dynamics. All you need is a re-settable simulator, which is a much weaker condition that all current Gym environments could satisfy anyway. We could just as easily use this class to wrap Gym, Atari, etc. environments for non-model-based algorithms like vine-mode TRPO, which needs the ability to "branch" from states. Maybe a name like `ResettableEnv` or `SimulatorEnv` better reflects the generality of this interface?

tests/test_envs.py
```diff
@@ -1,27 +1,114 @@
...
+    r"imitation/PointMaze.*-v0",
+]
 
-@pytest.mark.parametrize("env_name", ENV_NAMES)
-def test_envs(env_name):
-  """Check that our custom environments don't crash on `step`, and `reset`."""
+
+def matches_list(env_name, patterns):
+  for pattern in patterns:
+    if re.compile(pattern).match(env_name):
```
qxcv: Just use `re.match(p,s)`.

qxcv: (or make `DETERMINISTIC_ENVS` a list of regexes rather than strings)

tests/test_envs.py
```diff
@@ -1,27 +1,114 @@
...
     if e.args[0].find('mujoco_py') != -1:
       pytest.skip("Requires `mujoco_py`, which isn't installed.")
     else:
       raise
-  env.reset()
+  return env
+
+
+@pytest.mark.parametrize("env_name", ENV_NAMES)
+def test_seed(env, env_name):
```
qxcv: Should either be renamed to reflect the fact that it only tests initial observation, or changed to take a random action before testing for observation equality (which would require seeding action space, annoyingly).

tests/test_envs.py
```diff
@@ -1,27 +1,114 @@
...
+
+  act = env.action_space.sample()
+  with pytest.raises(Exception):  # need to call env.reset() first
+    env.step(act)
+
+
+@pytest.mark.parametrize("env_name", ENV_NAMES)
+def test_model_based(env):
+  """Smoke test for each of the ModelBasedEnv methods with basic type checks."""
+  if not hasattr(env, 'state_space'):  # pragma: no cover
```
qxcv: Why use duck typing instead of `isinstance`?

tests/test_tabular.py
```diff
@@ -74,6 +75,7 @@ def test_random_mdp():
 def test_policy_om_random_mdp():
   """Test that optimal policy occupancy measure ("om") for a random MDP makes
   sense."""
+  mdp = gym.make('imitation/Random-v0')
```
qxcv: Comment (e.g. `# test env is registered under some name`) to make overwrite of `mdp` less confusing.

qxcv:
Good stuff! I think the API makes sense, although I did have a (minor) objection to naming it a `ModelBasedEnv` given that it does not expose any models.

tests/test_tabular.py
shwang: I still don't understand why you need initialize RandomMDP through both `gym.make` and `RandomMDP.__init__`. Doesn't `gym.make` just call `RandomMDP.__init__` in the first place?

tests/test_envs.py
AdamGleave: Changed to take random sequence of actions.

tests/test_envs.py
AdamGleave: Duck typing is more Pythonic, but more practically `gym.make` often wraps the environments, so we'd actually need to use `isinstance(env.unwrapped, ModelBasedEnv)` which starts getting a bit hacky IMO.

tests/test_tabular.py
AdamGleave: You're right, I don't need the `RandomMDP` call any longer.

tests/test_envs.py
```diff
@@ -36,28 +33,50 @@ def env(env_name):
...
-    obs = env.reset()
-    if not np.all(obs == first_obs):
-      same_obs = False
-      break
+    new_rollout = rollout(env, actions)
+    for step_a, step_new in zip(rollout_a, new_rollout):
+      obs_a = step_a[0]
+      obs_new = step_new[0]
+      if np.any(obs_a != obs_new):
+        same_obs = False
```
shwang: suggestion: Keep the `break` statement here. (Removed during 36b7d9f)

tests/test_envs.py
```diff
@@ -11,15 +11,12 @@
 ENV_NAMES = [env_spec.id for env_spec in gym.envs.registration.registry.all()
              if env_spec.id.startswith('imitation/')]
 
-DETERMINISTIC_ENVS = [
-    r"imitation/CliffWorld.*-v0",
-    r"imitation/PointMaze.*-v0",
-]
+DETERMINISTIC_ENVS = []
```
shwang: Huh. Guess we just don't have any deterministic models (not even PointMaze).

PR #74, opened 2019-08-09
Have `make_session` close session be optional

AdamGleave:
Sometimes the Session needs to persist, e.g. when loading a serialized model, when we return the Session needs to still exist so the user can query the model.

### DIFF
--- a/src/imitation/util/util.py
+++ b/src/imitation/util/util.py
@@ -193,7 +193,9 @@ def make_session(close_on_exit: bool = True, **kwargs):
   graph = tf.Graph()
   with graph.as_default():
     session = tf.Session(graph=graph, **kwargs)
-    with session.as_default():
-      yield graph, session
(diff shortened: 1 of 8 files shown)


src/imitation/util/util.py
```diff
@@ -176,21 +176,24 @@ def build_inputs(observation_space: gym.Space,
...
     (graph, session) where graph is a `tf.Graph` and `session` a `tf.Session`.
   """
   graph = tf.Graph()
   with graph.as_default():
-    with tf.Session(graph=graph, **kwargs) as session:
-      with session.as_default():
-        yield graph, session
+    session = tf.Session(graph=graph, **kwargs)
+    with session.as_default():
+      yield graph, session
```
shwang: suggestion: `try: yield; finally: close()` here but might be overkill.

Issue #62, opened 2019-07-29
GAIL end-to-end integration test


shwang:
A test that
  * Downloads expert models [from S3/Dropbox](https://github.com/HumanCompatibleAI/imitation/pull/58#discussion_r307239717)
  * Runs generate_policy script to make rollouts
  * Executes a fast version of `experiments/gail_benchmark.sh` as smoke test and checks that exit code is 0 for each process (Nice, `parallel` exits with 0 iff every process succeeded.)
    * (Add `--fast` flag to experts/gail_benchmark.sh)

PR #77, opened 2019-08-14
train_experts: Save results.txt and build zip file

shwang:
Modifies the script to automatically build the zip file that I previously was building manually. Also archives a `results.txt` file into the zip file, containing:
```
...
==> ./parallel/env/half_cheetah/seed/2/stdout <==                                                                                                                                                           
| ep_reward_mean     | 2.41e+03     |                                                                                                                                                                       
==> ./parallel/env/half_cheetah/seed/1/stdout <==                                                                                                                                                           
| ep_reward_mean     | 3.82e+03     |                                       
==> ./parallel/env/custom_ant/seed/0/stdout <==                                              
| ep_reward_mean     | 1.83e+03     |                                    
==> ./parallel/env/custom_ant/seed/2/stdout <==    
...
```

Note that the zip file only saves one expert model for each environment. This is because `gail_benchmark.sh` expects to find a single model at `{env_name}/`.

### DIFF
--- a/experiments/train_experts.sh
+++ b/experiments/train_experts.sh
@@ -1,8 +1,14 @@
 #!/usr/bin/env bash
 
+# This script trains experts for AIRL and GAIL benchmark scripts.
+# When training is finished, it reports the mean episode reward of each
+# expert and builds a zip file of expert models that can be used for
+# `experiments/gail_benchmark.sh`.
+
 ENVS+="acrobot cartpole mountain_car "
 ENVS+="reacher half_cheetah hopper ant humanoid swimmer walker "
 ENVS+="two_d_maze custom_ant disabled_ant "
+
 SEEDS="0 1 2"
 
 if $(command -v gdate > /dev/null); then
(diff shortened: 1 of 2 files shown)


shwang:
It might be a good idea to just save all of the models into the zip like `${env_name}/{0,1,2}` and then modify `gail_benchmark.sh` to account for the change.

shwang:
> Why are we training with multiple seeds if we only save the first one? It'd make sense to use the best of three seeds as an expert (to reduce variance due to RL), or to use each seed as a different expert (in case some policies are easier to learn from than others). But keeping just the first policy seems odd.

Yeah, this is pretty wasteful. I don't have enough bash-fu to easily identify and archive the best policy. (And I think that script would be pretty convoluted)

I think I'll try to archive the models as ${env_name}/${seed}, and update the train script to load an expert from `experts/${env_name}/0`.

shwang:
For each environment, #62's rollout generate script could generate a equal number of trajectories from each expert.

shwang:
Didn't know that S3 could sync entire directories like that! Look forward to updating that later.

shwang:
Looks like macOS bash doesn't support `${X::N}` for negative N.
```diff
-  env_name=${env_name::-1}  # Remove trailing "/"
+  env_name=$(echo ${env_name} | sed -e 's|/$||')  # Remove trailing slash
```

shwang:
Only a minor change, so I'm not going to ask for another review

shwang:
rebased on master, now that airl_data_collect was squash merged

shwang:
image

PR #80, opened 2019-08-14
Restructure codebase

AdamGleave:
This PR reorganizes the codebase to reflect a hierarchical structure, and renames some modules for clarity. Notable changes:
  - Have a single package `imitation.envs.examples` which can be imported to register all the optional example environments.
  - Merge `imitation.util.trainer` with `imitation.algorithms.adversarial` (previously `imitation.trainer`). There were some warnings about circular imports, but these don't seem to be an issue any longer.
  - New package `algorithms` collecting the implementations of AIRL/GAIL, behavioral cloning, density modeling and tabular MCE IRL.
  - New package `rewards` collecting the reward and discriminator networks. Arguably the discriminator networks belong under `algorithms` with AIRL/GAIL, however since you can treat the GAIL discriminator as a reward (even if this isn't quite the right thing to do) I think it's more natural here.
  - Renamed `data_collect` to `expert_demos` to clarify the kind of data that is being collected, `train` to `train_adversarial` since it does not currently support e.g. BC or MCE IRL, and `policy_eval` to `eval_policy` so we keep the verb first for consistency. I'm not totally happy with these names, so would welcome sugestions on improvements.
  - Added some Sphinx autodoc stubs in docs/ using `sphinx-apidoc`, to make sure all the modules get built (we were missing a few). Also fixed some docstring syntax errors that got caught by this.


shwang:
rebased onto master now that `transfer_learn` was squash merged

PR #86, opened 2019-08-21
Time limit

shwang:
Useful for training experts to compare against `inverse_rl` repo.

In particular, the `inverse_rl` repo uses 500 max episode length for Ant environments, whereas the default episode length is 1000. The shorter time horizon might be important for AIRL training.

### DIFF
--- a/src/imitation/scripts/eval_policy.py
+++ b/src/imitation/scripts/eval_policy.py
@@ -72,8 +72,7 @@ def eval_policy(_seed: int,
         a reward of this.
     reward_path: If reward_type is specified, the path to a serialized reward
         of `reward_type` to override the environment reward with.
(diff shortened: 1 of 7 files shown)


src/imitation/scripts/eval_policy.py
```diff
@@ -67,15 +71,17 @@ def eval_policy(_seed: int,
         a reward of this.
     reward_path: If reward_type is specified, the path to a serialized reward
         of `reward_type` to override the environment reward with.
-
+    log_dir: The directory to log intermediate output to. (As of 2019-07-19
+        this is just episode-by-episode reward from bench.Monitor.)
```
shwang: (remove this)

shwang:
Agreed, force 500 step seems like a good idea

PR #87, opened 2019-08-22
pipeline scripts: Generate rollouts from expert policies

shwang:
This script generates the demonstrations necessary to run `experiments/gail_benchmark.sh`.

There's a tricky (messy) nested `parallel` in the script. I wanted at least one `parallel` to load the values of {env_name}, {config_name}, and {n_demonstrations} from the CSV. {uuid} needed to be dynamically generated.

Another solution would be to replace the inner parallel with a `for $(find ...)`.

* [x] `gail_benchmark.sh` uses the right globs to load from output of `rollout_from_policies.sh`.
* [x] Add custom gym envs to `rollouts_from_policies.csv`.
  * [x] ~~I probably want to simplify all the save structures to not include `env_name` as an intermediate directory or file basename. Instead, they should use the primary named config for the same purpose.~~
    * ~~This semi-blocks adding custom gym envs because I think it's too messy to use `imitation_Ant_v2` in some places and `imitation/Ant_v2` in other parts of the script.~~
    * It's messy that I have to list the {env_name} next to every {config_name} in the CSVs.
* [x] Rename `rollouts_from_policies{,_config}.csv` for consistency with `gail_benchmark*`


shwang:
echoed `python` calls can be found here:
https://gist.github.com/shwang/b2d24125676de5d68897d47a4991aaa2

AdamGleave:
I've run into similar awkwardness with file globs and needing either nested `parallel` or a `for` loop. I don't have a good solution. One hacky approach is to just make the file structure consistent, so you can construct it all from within parallel without needing to glob. You could do this just-in-time by constructing symbolic links. But I'm not sure this is overall better than nested `parallel`.

`parallel` is a nice tool but eventually it does show its limitations. I'd just keep hacking until it seems easier to use Python, then switch :)

experiments/gail_benchmark_config.csv
```diff
@@ -1,36 +1,36 @@
...
-ant,Ant-v2.pkl,50000,500,18
-ant,Ant-v2.pkl,50000,500,25
-humanoid,Humanoid-v2.pkl,50000,1500,80
-humanoid,Humanoid-v2.pkl,50000,1500,160
-humanoid,Humanoid-v2.pkl,50000,1500,240
-reacher,Reacher-v2.pkl,50000,200,4
-reacher,Reacher-v2.pkl,50000,200,11
-reacher,Reacher-v2.pkl,50000,200,18
-reacher,Reacher-v2.pkl,50000,200,25
+env_config,env_name,n_gen_steps_per_epoch,n_epochs,n_demonstrations
```
shwang: It's unfortunate that I have to keep track of the `env_name` right now to access the logs. It would be nice if on the python side of things, I could save artifacts into `$LOG_DIR/{env_config}` instead of `$LOG_DIR/{env_name}`.

Migrating to ray and taking advantage of `run.results` wouldn't automatically solve this issue because Phase 2 (this script) usually won't run straight after Phase 1. We will typically download a zip file to skip Phase 1.

shwang:
Haven't bothered to check whether the scripts still run with the new changes (which seemed minor). If seems important at this stage I can give that a shot.

I wouldn't be surprised if these bash scripts are replaced with Ray things later.

experiments/gail_benchmark_config.csv
AdamGleave: Why not just set `log_dir="${OUTPUT_DIR}/{env_config}` in the `experiments/gail_benchmark.sh`? All `log_dir` in the Sacred config is log root plus environment plus timestap. You already have a timestamp in OUTPUT_DIR so we don't need that again.

I'd also be fine with replacing `env_name` in `log_dir` with `exp_name`, which should default to `env_name` but could be overidden in particulra named configs. We just want a sensible default for `log_dir`.

experiments/rollouts_from_policies.sh
```diff
@@ -0,0 +1,71 @@
...
+# Then it executes `find` to to dynamically generate every {uuid}
+# associated with {env_name}.
+#
+# The inner parallel reads in values of {uuid} from STDIN and executes
+# `python -m imitation.scripts.expert_demos` with the appropriate Sacred options.
+
+parallel -j 25% --header : --results ${output_dir_abs}/parallel/ --colsep , \
+  "${FIND_CMD} policies -mindepth 2 -maxdepth 2 -type d -path '*{env_name}/*' \
+    -printf '%f\n' \
+  | parallel -I '{uuid}' -j 1 \
```
AdamGleave: Why no parallelism (`-j 1`)?

experiments/rollouts_from_policies.sh
```diff
@@ -0,0 +1,71 @@
...
+    python -m imitation.scripts.expert_demos rollouts_from_policy \
+    with \
+    {config_name} \
+    log_root='${output_dir_abs}' \
+    policy_path='policies/{env_name}/{uuid}' \
```
AdamGleave: I'm confused by your directory structure. I thought `expert_demos` generate directories of structure:
  - `output/expert_demos/{env_name}/{timestamp}`
    - `policies`
      - `%5d` -- checkpoints
     - `final` -- the final policy saved

So shouldn't this be `{env_name}/{uuid}/policies/final`?

experiments/rollouts_from_policies_config.csv
```diff
@@ -0,0 +1,13 @@
+config_name,env_name,n_demonstrations
```
AdamGleave: Curious how you chose these numbers, but they seem reasonable (10 for simple tasks, 25 for most MuJoco, 240 for Humanoid).

experiments/gail_benchmark.sh
```diff
@@ -31,7 +31,7 @@ parallel -j 25% --header : --results ${OUTPUT_DIR}/parallel/ --colsep , --progre
   {env_config} \
   log_root="${OUTPUT_DIR}" \
   n_gen_steps_per_epoch={n_gen_steps_per_epoch} \
-  init_trainer_kwargs.rollout_glob=expert_models/rollouts/{rollout_glob} \
+  init_trainer_kwargs.rollout_glob=expert_models/rollouts/{env_name}/*.pkl \
```
AdamGleave: When would we have multiple rollout pickles (and so need the `*`)? Perhaps make this `.../rollouts/{env_name}/{env_name}.pkl` (which I think is the naming convention)?

experiments/gail_benchmark_config.csv
AdamGleave: Why not just set `log_dir="${OUTPUT_DIR}"/{env_config}` from inside the bash script? The `log_dir` in Sacred is just meant to be a sensible default (log_root + env_name + timestamp), and since we already have a timestamp in OUTPUT_DIR seems fine to override.

You could also change this on the Python end, by e.g. changing `env_name` in the default `log_dir` to `exp_name`, and then setting `exp_name = env_name` by default but letting named configs override it. Seems unnecessary for now though.

experiments/rollouts_from_policies.sh
shwang: There's already `-j 25%` on the first level of parallelism. Don't think it's necessary to nested `-j`.

experiments/rollouts_from_policies_config.csv
shwang: With the exception of the custom envs, I used the maximum number of demonstrations used in the GAIL paper's benchmarks (almost every env had three datapoints with 3 different # of demos).

For the ant custom envs I just copied the original ant env's # of demos.

For the 2DMaze I thought 10 would be fine, but did 25 just in case. (The imitation learning phase can choose the number of demonstrations to use from disk)

experiments/rollouts_from_policies.sh
shwang: Yes, `expert_demos_ex` saves artifacts to that directory.

This scripts assumes that we are loading artifacts from the `expert_models.zip` file, extracted to `./expert_models/`. This zip file right now is rearranged in a particular way in `train_experts.sh`.
I took notes on the format in this [Notion doc](https://www.notion.so/chaiberkeley/Pipeline-design-doc-231509121fad45ae9377cfab6858afd7#1a319e1d27124583b089a09160a8dc2b).

shwang: Doesn't seem to me like a good format anymore. It's certainly simpler to not rearrange things before zipping, and simply access `{env_name}/{uuid}/policies/final` when generating the zip.

This will probably be more amenable to scripted EC2 upload later too.

shwang:
> Why not just set `log_dir="${OUTPUT_DIR}"/{env_config}` from inside the bash script? 

Neat, setting `log_dir` externally would be a clean solution. Not sure why I hadn't thought of this earlier.

experiments/gail_benchmark.sh
shwang: Right now we generate multiple demos because there are multiple experts generated in `train_experts.sh` (one for each seed) and I haven't bothered to figure out how to pick the best expert from within bash. The inner parallel in `rollouts_from_policies.sh` which iterates over {uuid} is in fact iterating over expert seeds.

Another thing I tried in the past was copying only one expert `train_experts.sh`. (In bash, picking a single expert surprisingly messy because I didn't know what {uuid}/ would be ahead of time.)

shwang: The current format is `expert_demos/rollouts/{env_name}/{uuid}.pkl`.

experiments/rollouts_from_policies.sh
AdamGleave: OK, I missed that.

experiments/rollouts_from_policies.sh
AdamGleave: Directory structures always get confusing so generally rearranging things as little as possible is good since then at least it's consistent. But I'm fine approving this so long as things work and we can clean up later.

experiments/gail_benchmark.sh
AdamGleave: The glob means "load all of them", though, right? So you're doing IRL on demos from a mixture of policies. This is an interesting thing to do, but is non-standard, and could easily make results better (especially generalization) or much worse (it might be hard to model a mixture of policies).

shwang:
Thanks for the comments, Adam. It's getting a bit late here so I'm going to summarize these changes tomorrow. Made quite a few directory-related changes in the past few commits so it would probably be easier to review with an explanation.

experiments/rollouts_from_policies.sh
shwang: The log directory is no longer rearranged before zipping.

experiments/gail_benchmark.sh
shwang: Now we only load a single expert (the one corresponding to seed=0, `expert_models/{env_config_name}_{seed}/rollouts`).

shwang:
- I decided on a directory structure for expert_models that looks like this:
    - `log_dir=expert_models/{env_config_name}_{seed}/`
    - After `train_experts.sh` (Phase 1) is finished, `$log_dir` contains exactly the log_dir contents from running `expert_demos_ex`.
    - After `rollouts_from_policies.sh` (Phase 2) is finished, we have a rollout `expert_models/{env_config_name}_0/rollouts/auto.pkl` .
    - `gail_benchmark.sh` (Phase 3) loads rollouts from the previous path.
- All of the repository's pkls, (policies in `expert_models` and rollouts in `test/data`) were moved to `toy_models/` , with the same directory structure as `expert_models`.
    - Removed PPO1 cartpole expert (seemed redundant)
- Test cases in `test_scripts.py` used to have `env_name` parameters, which were always set to `"CartPole-v1"` . I removed this parameter so that I could hardcode a path to the policy `toy_models/cartpole_0/...` without having to parse the config names like `cartpole` from arbitrary environment names.
- Added fast mode flag `-f` to all the aforementioned scripts.

shwang:
Added -r flag to `train_experts.py`. Also renamed `toy_models` to `tests/data`.

(The overall change to `tests/data` that results from this PR is that `tests/data` now mirrors the format of `expert_models` described [here](https://github.com/HumanCompatibleAI/imitation/pull/87#issuecomment-530113330).)

PR #88, opened 2019-08-26
Ray Tune parallelized meta-experiment and various experiment refactoring

shwang:
* Adds `parallel_ex`, a Sacred meta-experiment that wraps other experiments in `ray.tune`.
* Moves most of the plotting logic inside `train_adversarial.py:train()` into a helper class `_Visualizer`.
  * The plot logic no longer accepts an expert policy for the purpose of plotting expert performance, instead it accepts a single number `expert_ep_mean_reward`, which is averaged from the loaded expert demonstrations.
* Adds new dictionary entries to the return values of `train_ex` and `expert_demos_ex`. (Will probably add even more in future PRs.)
* Remove default params from scripts/*.py.


src/imitation/scripts/config/multi_expert_demos.py
```diff
@@ -0,0 +1,32 @@
...
+
+from imitation.util import util
+
+multi_expert_demos_ex = sacred.Experiment("multi_expert_demos")
+
+
+@multi_expert_demos_ex.config
+def config():
+  csv_config_path = "FILL ME"  # Config file
+  n_workers = math.ceil(
```
AdamGleave: Why is the integer ceiling more appropriate than the floor here? (Note for floor you can just write `multiprocessing.cpu_count() // 4`.)

src/imitation/scripts/config/multi_expert_demos.py
```diff
@@ -0,0 +1,32 @@
...
+  log_dir = osp.join(log_root, util.make_unique_timestamp())
+  seeds = list(range(3))
+
+
+@multi_expert_demos_ex.named_config
+def fast():
+  """Minimize computation for debugging purposes."""
+  seeds = [666]
+  extra_named_configs = ["fast"]
+  parallel = False  # False may be necessary for proper code coverage.
```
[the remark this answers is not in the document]
AdamGleave: Ouch. It seems like this should reduce performance (only run one experiment at a time)? https://stackoverflow.com/a/53180921 provides a viable workaround, although I'm not a fan of `multiprocessing` in general (see overall comments).

src/imitation/scripts/config/multi_train_adversarial.py
```diff
@@ -0,0 +1,42 @@
...
+import sacred
+
+from imitation.util import util
+
+multi_train_ex = sacred.Experiment("multi_train_adversarial")
+
+
+@multi_train_ex.config
+def config():
+  n_workers = math.ceil(
```
AdamGleave: As above, is `ceil` or `floor` rbetter here?

src/imitation/scripts/expert_demos.py
```diff
@@ -40,7 +43,7 @@ def rollouts_and_policy(
 
   policy_save_interval: int = -1,
   policy_save_final: bool = True,
-) -> None:
+) -> dict:
```
AdamGleave: :+1: this is useful even without `multi_expert_demos` (Sacred will save it so it's a useful summary)

src/imitation/scripts/multi_expert_demos.py
```diff
@@ -0,0 +1,150 @@
...
+
+  Returns:
+    Path to the output CSV file, `f"{log_dir}/results.csv"`.
+  """
+  if extra_named_configs is None:
+    extra_named_configs = []
+  if extra_config_updates is None:
+    extra_config_updates = {}
+  if parallel:
+    mp = multiprocessing
```
AdamGleave: You don't seem to import `multiprocessing` (just `multiprocessing.dummy`) so I'm confused how this can work

src/imitation/scripts/multi_expert_demos.py
```diff
@@ -0,0 +1,150 @@
...
+def _job(
+  row: OrderedDict,
+  log_dir: str,
+  extra_named_configs: Iterable[str],
+  extra_config_updates: dict,
+) -> Tuple[OrderedDict, dict]:
+  """Start a `expert_demos_ex` run.
+
+  Params:
+    row: A CSV row encoding config settings for this run. Its fields should
```
AdamGleave: Avoid passing arguments inside a dictionary. This makes it harder to reason about function parameters, and loses a lot of the static type analysis. Instead make them top-level arguments and pass them as **row if necessary.

src/imitation/scripts/multi_expert_demos.py
```diff
@@ -0,0 +1,150 @@
...
+  Params:
+    log_dir: Main logging directory.
+    n_workers: The number of jobs that are processed simultaneously.
+    csv_config_path: Path to CSV configuration. Columns should
+      include 'env_config' (str), 'reward_path' (str), 'reward_type' (str),
+      and 'phase4_seed' (int).
+    seeds: Every row in `csv_config_path` is run using each of these seeds.
+      The seed associated with each row in the output CSV,
+      `f"{log_dir}/results.csv"`, is saved in a column called "phase3_seed".
+      extra_named_configs: Extra named configs to apply to every call of
```
AdamGleave: Needs to be de-indented.

src/imitation/scripts/multi_expert_demos.py
```diff
@@ -0,0 +1,150 @@
...
+  stderr_path = osp.join(log_dir, "stderr")
+  run = multi_util.redirect_to_files(run, stdout_path, stderr_path)
+
+  named_configs = [row['env_config']] + list(extra_named_configs)
+
+  config_updates = {
+    "log_dir": log_dir,
+    "reward_path": row["transfer_reward_path"],
+    "reward_type": row["transfer_reward_type"],
+    "seed": int(row['phase4_seed']),
```
AdamGleave: Why is this called "phase4" (what're phases 1-3?)

src/imitation/scripts/multi_expert_demos.py
```diff
@@ -0,0 +1,150 @@
...
+    mp = multiprocessing.dummy
+
+  # Construct job args for `Pool.imap_unnordered`.
+  # There is one job for every combination of seed and CSV row.
+  job_args = []  # type: List[tuple]
+  with open(csv_config_path, newline='') as csv_file:
+    for row in csv.DictReader(csv_file):
+      for seed in seeds:
+        new_row = row.copy()  # type: OrderedDict
+        new_row["phase4_seed"] = str(seed)
```
AdamGleave: This isn't part of the row, so why are we including it here?

src/imitation/scripts/multi_expert_demos.py
```diff
@@ -0,0 +1,150 @@
...
+  row: OrderedDict,
+  log_dir: str,
+  extra_named_configs: Iterable[str],
+  extra_config_updates: dict,
+) -> Tuple[OrderedDict, dict]:
+  """Start a `expert_demos_ex` run.
+
+  Params:
+    row: A CSV row encoding config settings for this run. Its fields should
+      include 'env_config' (str), 'reward_path' (str), 'reward_type' (str),
```
AdamGleave: This function looks for keys `"transfer_reward_path"`and `"transfer_reward_type"` not `"reward_path"` and `"reward_type"`.

src/imitation/scripts/multi_expert_demos.py
```diff
@@ -0,0 +1,150 @@
...
+    return csv_output_path
+
+
+def main_console():
+  obs_path = osp.join('output', 'sacred', 'multi_expert_demos')
+  job_obs_path = osp.join(obs_path, 'jobs')
+
+  multi_expert_demos_ex.observers.append(
+    FileStorageObserver.create(obs_path))
+  expert_demos_ex.observers.append(
```
AdamGleave: Doesn't seem to be causing problems right now, but just to flag that pickling `Observer` objects (which can happen implicitly as functions are closing on it) will fail as it contains IO handles.

src/imitation/scripts/multi_train_adversarial.py
```diff
@@ -0,0 +1,149 @@
...
+import tensorflow as tf
+
+from imitation.scripts.config.multi_train_adversarial import multi_train_ex
+from imitation.scripts.train_adversarial import train_ex
+import imitation.util.multi as multi_util
+
+
+@multi_util.args_single_to_star
+def _job(
+  row: OrderedDict,
```
AdamGleave: As above, would prefer this to be fully-fledged keyword arguments.

src/imitation/scripts/multi_train_adversarial.py
```diff
@@ -0,0 +1,149 @@
...
+    mp = multiprocessing.dummy
+
+  # Construct job args for `Pool.imap_unnordered`.
+  # There is one job for every combination of seed and CSV row.
+  job_args = []  # type: List[tuple]
+  with open(csv_config_path, newline='') as csv_file:
+    for row in csv.DictReader(csv_file):
+      for seed in seeds:
+        new_row = row.copy()  # type: OrderedDict
+        new_row["phase3_seed"] = str(seed)
```
AdamGleave: Why is this phase3? Isn't this after `expert_demos` which was phase4?

src/imitation/scripts/multi_train_adversarial.py
```diff
@@ -0,0 +1,149 @@
...
+
+  # Run all jobs and write results to "results.csv".
+  csv_output_path = osp.join(log_dir, "results.csv")
+  with mp.Pool(n_workers) as pool:
+    with open(csv_output_path, 'w', newline='') as csv_file:
+      writer = None
+      for row, results in pool.imap_unordered(_job, job_args):
+        row = multi_util.ordered_dict_rename(
+          row,
+          {"log_dir": "phase4_log_dir",
```
AdamGleave: Inconsistent: is this phase 3 or 4?

src/imitation/scripts/train_adversarial.py
```diff
@@ -236,12 +237,27 @@ def should_plot_now(epoch) -> bool:
...
-    print(f"[result] Mean Episode Return: {mean:.4g} ± {std_err:.3g} "
+    ep_reward_mean = stats["return_mean"]
+    ep_reward_std_err = stats["return_std"] / math.sqrt(n_episodes_eval)
+    print("[result] Mean Episode Return: "
+          f"{ep_reward_mean:.4g} ± {ep_reward_std_err:.3g} "
           f"(n={stats['n_traj']})")
 
-    return dict(mean=mean, std_err=std_err)
+    reward_path = os.path.join(log_dir, "checkpoints", "final", "discrim")
+    # TODO(shwang): I think Serializable should store the save_type, and
```
AdamGleave: This seems like a reasonable modification.

src/imitation/util/multi.py
```diff
@@ -0,0 +1,91 @@
...
+  try:
+    os.makedirs(os.path.dirname(path), exist_ok=True)
+    with open(path, "w") as f:
+      sys.stderr = f
+      yield
+  finally:
+    sys.stderr = temp
+
+
+def redirect_to_files(
```
AdamGleave: This is nice but FYI Sacred should already be saving `stdout` / `stderr`.

tests/test_scripts.py
```diff
@@ -6,19 +6,20 @@
...
 import pytest
 
 from imitation.scripts.eval_policy import eval_policy_ex
 from imitation.scripts.expert_demos import expert_demos_ex
+from imitation.scripts.multi_expert_demos import multi_expert_demos_ex
+from imitation.scripts.multi_train_adversarial import multi_train_ex
 from imitation.scripts.train_adversarial import train_ex
 
 
 def test_expert_demos_main():
```
AdamGleave: Now that `expert_demos` returns a value, add a check on `run.result`?

src/imitation/scripts/config/multi_expert_demos.py
AdamGleave: :+1: makes sense

shwang:
Actually haven't read `modelfree...multi.*` before. Thanks for linking, gave it a skim and will reference it.

My bad for not properly documenting/explaining the design decisions here. I'll try to be more explicit next time. (Not bothering to elaborate too much right now because I think the design needs to be changed to something else)

I'm thinking that the right way to move forward is to get rid of `multiprocessing` (b/c tf.logging interactions and harder scaling to multiple machines) and move to Ray.

I believe that there are two types of Ray routines that would be nice here. Maybe they will live in the same file, or maybe they will be different enough to be separate.

1. A script "multi_train_adversarial" script or command that can take in a long CSV of desired runs, (maybe convert the CSV into a generator of configs,) and find the average imitation learning performance for each config. (perhaps continues to find the transfer learning performance). The output of this script is a big table of results in a similar format to the table in the GAIL paper.

2. A script "tune_train_adversarial" script or command that has a `ray.tune.track` hook and can hyperparameter tune imitation learning for a single configuration (e.g.: "with cartpole gail"). The output of this script is the performance of a single config over a lot of different hyperparameter settings.

src/imitation/scripts/multi_train_adversarial.py
AdamGleave: Thanks for clarifying, this diagram is really useful. Might we want to use `multi_expert_demos` for phase 1 as well as phase 4 though?

AdamGleave:
> 1. A script "multi_train_adversarial" script or command that can take in a long CSV of desired runs, (maybe convert the CSV into a generator of configs,) and find the average imitation learning performance for each config. (perhaps continues to find the transfer learning performance). The output of this script is a big table of results in a similar format to the table in the GAIL paper.
> 2. A script "tune_train_adversarial" script or command that has a `ray.tune.track` hook and can hyperparameter tune imitation learning for a single configuration (e.g.: "with cartpole gail"). The output of this script is the performance of a single config over a lot of different hyperparameter settings.

I agree these seem like two distinct use cases. Leaning towards trying to make a single script that is flexible enough to support both of them though. Mainly I think the lines between these two use cases can become fuzzy. For example, we might want to do a hyperparameter sweep on multiple environment, to find hyperparameters that are good on a range of environments, not just one.

src/imitation/scripts/multi_train_adversarial.py
shwang: yes, this would be nice. Phase 1 is currently performed by `experiments/train_experts.sh`. When I made this PR I intended to punt this issue till later (another thing that would have been good to mention in the PR summary)

shwang: For completion I'll write here that Phase 2 is roughly done by #87

src/imitation/scripts/expert_demos.py
```diff
@@ -38,13 +44,20 @@ def rollouts_and_policy(
   rollout_save_n_timesteps: Optional[int] = None,
   rollout_save_n_episodes: Optional[int] = None,
 
+  ray_tune_interval: int = 1,
```
AdamGleave: Is this meant to be `1` not `-1` (usually we've kept the default value here and in the Sacred config the same)?

src/imitation/scripts/expert_demos.py
```diff
@@ -142,6 +172,14 @@ def callback(locals_: dict, _) -> bool:
         if policy_save_interval > 0 and step % policy_save_interval == 0:
           output_dir = os.path.join(policy_dir, f'{step:05d}')
           serialize.save_stable_model(output_dir, policy, vec_normalize)
+        if ray_tune_interval > 0 and step % ray_tune_interval == 0:
+          # TODO(shwang): Not sure if `venv` has the ground truth reward
```
AdamGleave: `VecNormalize` clips rewards, it can also rescale them but doesn't do so by default IIRC. It's probably best to compute this before VecNormalize. 

I'm not sure on the best way to do this: we could look at the `ep_rew` added by `Monitor`? This might be too constrained, in which case we could add our own wrapper for instrumentation.

src/imitation/scripts/expert_demos.py
```diff
@@ -156,6 +194,21 @@ def callback(locals_: dict, _) -> bool:
         output_dir = os.path.join(policy_dir, "final")
         serialize.save_stable_model(output_dir, policy, vec_normalize)
 
+      # Final evaluation of expert policy.
+      stats = util.rollout.rollout_stats(policy,
+                                         venv,
```
AdamGleave: Same issue as above re: `VecNormalize` (and probably elsewhere in our code)

src/imitation/scripts/expert_demos.py
```diff
@@ -142,6 +172,14 @@ def callback(locals_: dict, _) -> bool:
         if policy_save_interval > 0 and step % policy_save_interval == 0:
           output_dir = os.path.join(policy_dir, f'{step:05d}')
           serialize.save_stable_model(output_dir, policy, vec_normalize)
+        if ray_tune_interval > 0 and step % ray_tune_interval == 0:
+          # TODO(shwang): Not sure if `venv` has the ground truth reward
+          # after it is wrapped in VecNormalize (does VecNormalize normalize
+          # both rewards and observations?) Check this with Adam.
+          mean_return = util.rollout.mean_return(policy, venv,
```
AdamGleave: Most (all?) RL algorithms report mean episode reward obtained during training, so doing evaluation periodically is unnecessary and will slow things down slightly. However, I don't have strong feelings about this. It's a little awkward (but possible) to intercept the logging entry from the RL algorithm, and you have to handle different keys used by different RL algorithms, so this approach is probably cleaner.

src/imitation/scripts/train_adversarial.py
```diff
@@ -36,73 +38,91 @@ def save(trainer, save_path):
...
-          then only save weights after training is complete.
-      interactive: Figures are always saved to `output/*.png`. If `interactive`
-        is True, then also show plots as they are created.
-      expert_policy (BasePolicy or BaseRLModel, optional): If provided, then
-          also plot the performance of this expert policy.
-      init_trainer_kwargs: Keyword arguments passed to `init_trainer`,
-        used to initialize the trainer.
+    _seed: Random seed.
+    env_name: The environment to train in.
+    log_dir: Directory to save models and other logging to.
```
AdamGleave: Missing: `rollout_glob`.

src/imitation/scripts/train_adversarial.py
```diff
@@ -36,73 +38,91 @@ def save(trainer, save_path):
...
+      reward over time. The script will crash unless `ray.tune` was
+      externally initialized.
+
+    n_episodes_eval: The number of episodes to average over when calculating
+      the average ground truth reward return of the imitation policy for return
+      and for `ray.tune.track`.
+    checkpoint_interval: Save the discriminator and generator models every
+      `checkpoint_interval` epochs and after training is complete. If <=0,
+      then only save weights after training is complete.
+    init_trainer_kwargs: Keyword arguments passed to `init_trainer`,
```
AdamGleave: Move earlier?

src/imitation/scripts/train_adversarial.py
```diff
@@ -36,73 +38,91 @@ def save(trainer, save_path):
...
+    show_plots: Figures are always saved to `output/*.png`. If `show_plots`
+      is True, then also show plots as they are created.
+
+    ray_tune_interval: The number of epochs between calls to `ray.tune.track`.
+      If nonpositive, disables ray tune. Otherwise, enables hooks
+      that call `ray.tune.track` to track the imitation policy's mean episode
+      reward over time. The script will crash unless `ray.tune` was
+      externally initialized.
+
+    n_episodes_eval: The number of episodes to average over when calculating
```
AdamGleave: Move earlier?

src/imitation/scripts/train_adversarial.py
```diff
@@ -36,73 +38,91 @@ def save(trainer, save_path):
...
+      `n_disc_steps_per_epoch` discriminator steps followed by
+      `n_gen_steps_per_epoch` generator steps.
+    n_gen_steps_per_epoch: The number of generator update steps during every
+      generator epoch.
+    n_disc_steps_per_epoch: The number of discriminator update steps during
+      every training epoch.
+
+    plot_interval: The number of epochs between each plot. (If nonpositive,
+      then plots are disabled).
+    n_episodes_plot: The number of episodes averaged over when
```
AdamGleave: Renamed to `n_plot_episodes`

src/imitation/scripts/train_adversarial.py
```diff
@@ -36,73 +38,91 @@ def save(trainer, save_path):
...
+    env_name: The environment to train in.
+    log_dir: Directory to save models and other logging to.
+
+    n_epochs: The number of epochs to train. Each epoch consists of
+      `n_disc_steps_per_epoch` discriminator steps followed by
+      `n_gen_steps_per_epoch` generator steps.
+    n_gen_steps_per_epoch: The number of generator update steps during every
+      generator epoch.
+    n_disc_steps_per_epoch: The number of discriminator update steps during
+      every training epoch.
```
AdamGleave: For consistency with above: should this be "every discriminator epoch"?

src/imitation/scripts/train_adversarial.py
```diff
@@ -36,73 +38,91 @@ def save(trainer, save_path):
...
+
+    plot_interval: The number of epochs between each plot. (If nonpositive,
+      then plots are disabled).
+    n_episodes_plot: The number of episodes averaged over when
+      calculating the average episode reward of a policy for the performance
+      plots.
+    expert_policy_plot (BasePolicy or BaseRLModel, optional): If provided,
+      then also plot the performance of this expert policy.
+    show_plots: Figures are always saved to `output/*.png`. If `show_plots`
```
AdamGleave: I think we now save plots to `log_dir` not `"output/"`

src/imitation/scripts/train_adversarial.py
```diff
@@ -36,73 +38,91 @@ def save(trainer, save_path):
...
+      `checkpoint_interval` epochs and after training is complete. If <=0,
+      then only save weights after training is complete.
+    init_trainer_kwargs: Keyword arguments passed to `init_trainer`,
+      used to initialize the trainer.
 
   Returns:
-      results: A dictionary with two keys, "mean" and "std_err". The
-          corresponding values are the mean and standard error of
-          ground truth episode return for the imitation learning algorithm.
+    A dictionary with the following keys: "ep_reward_mean" and
```
AdamGleave: Normally you have "and" as the penultimate entry in a list, unless you want to emphasize that "ep_reward_mean" and "ep_reward_std_err" are paired.

src/imitation/scripts/train_adversarial.py
```diff
@@ -114,116 +134,37 @@ def train_and_plot(_seed: int,
...
-        plt.plot(rand_ep_reward[name],
-                 label="avg random ep reward", c="black")
-        plt.plot(exp_ep_reward[name], label="avg exp ep reward", c="blue")
-        plt.legend()
-        _savefig_timestamp("plot_fight_epreward_gen", interactive)
-
-    if n_epochs_per_plot is not None:
-      n_plots_per_epoch = 1 / n_epochs_per_plot
+    if plot_interval > 0:
+      visualizer = _TrainVisualizer(
```
AdamGleave: :+1: nice to factor out the plotting code from the essential training logic.

src/imitation/scripts/train_adversarial.py
```diff
@@ -114,116 +134,37 @@ def train_and_plot(_seed: int,
...
-        ep_reward_plot_show()
+      if visualizer and epoch % plot_interval == 0:
+        visualizer.disc_plot_show()
+        visualizer.ep_reward_plot_add_data(trainer.env, "Ground Truth Reward")
+        visualizer.ep_reward_plot_add_data(trainer.env_train, "Train Reward")
+        visualizer.ep_reward_plot_add_data(trainer.env_test, "Test Reward")
+        visualizer.ep_reward_plot_show()
+
+      if ray_tune_interval > 0 and epoch % ray_tune_interval == 0:
+        gen_ret = util.rollout.mean_return(
```
AdamGleave: Potential issues with `VecNormalize` (along with potentially other uses in the code, e.g. `visualizer` logging.)

src/imitation/scripts/train_adversarial.py
```diff
@@ -236,20 +177,136 @@ def should_plot_now(epoch) -> bool:
...
+    return dict(ep_reward_mean=ep_reward_mean,
+                ep_reward_std_err=ep_reward_std_err,
+                log_dir=log_dir,
+                transfer_reward_path=reward_path,
+                transfer_reward_type=reward_type)
+
+
+class _TrainVisualizer:
+  def __init__(self,
+               trainer: "imitation.algorithms.adversarial.AdversarialTrainer",
```
AdamGleave: Why does this need to be in quotes? Can't we just import `AdversarialTrainer` (we're already importing the module it's contained in...)?

src/imitation/scripts/train_adversarial.py
```diff
@@ -236,20 +177,136 @@ def should_plot_now(epoch) -> bool:
...
+    X, Y = self.gen_data if gen_mode else self.disc_data
+    # Divide by two since we get two data points (gen and disc) per epoch.
+    X.append(self.plot_idx / 2)
+    Y.append(self.trainer.eval_disc_loss())
+    tf.logging.info(
+        "plot idx ({}): {} disc loss: {}"
+        .format(mode, self.plot_idx, Y[-1]))
+    self.plot_idx += 1
+
+  def disc_plot_show(self):
```
AdamGleave: Method name is a bit misleading when it may never call `plt.show()`. Maybe just `plot_disc`?

src/imitation/util/multi.py
```diff
@@ -0,0 +1,10 @@
+"""Utilities for distributed experiments."""
+import ray.tune.track
+
+
+def ray_tune_active() -> bool:
+  """Returns True if a Ray Tune track session has been initialized."""
+  try:
+    return ray.tune.track.get_session() is not None
+  except ValueError:
```
AdamGleave: Consider adding comment to explain when/why this occurs?

tests/test_scripts.py
```diff
@@ -65,34 +63,37 @@ def test_eval_policy(config):
...
-          'rollout_glob': "tests/data/rollouts/CartPole*.pkl",
-      }
-      run = train_ex.run(
-          named_configs=['cartpole', 'gail', 'fast'],
-          config_updates=config_updates,
-      )
-      assert run.status == 'COMPLETED'
+  with TemporaryDirectory(prefix='imitation-train') as tmpdir:
+    named_configs = ['cartpole', 'gail', 'fast']
+    if plots:
```
AdamGleave: Why not always test with plots? Are they any code paths not tested with plots that are tested without?

tests/data/multi_expert_demos_config.csv
```diff
@@ -0,0 +1,3 @@
+env_config,transfer_reward_path,transfer_reward_type,phase4_seed
```
AdamGleave: Is this used anywhere now?

tests/data/multi_train_adv_config.csv
```diff
@@ -0,0 +1,3 @@
+env_config,n_gen_steps_per_epoch,n_epochs,n_expert_demos
```
AdamGleave: Is this used anywhere now?

tests/data/multi_expert_demos_config.csv
shwang: no, removing.

tests/test_scripts.py
shwang: No doesn't look like it. I'm removing this now.

src/imitation/scripts/train_adversarial.py
shwang: I renamed to `plot_disc_loss`. Also renamed some other stuff in the class for consistency/clarity.

src/imitation/util/multi.py
shwang: This method doesn't make sense for the current release of Ray. I'm removing this file.

src/imitation/scripts/train_adversarial.py
shwang: That's right. I corrected this, and moved figures to `{log_dir}/plots/*.png`.

src/imitation/scripts/train_adversarial.py
shwang: Will try to address this in #90 along with `ray.tune` episode reward mean reports.

src/imitation/scripts/train_adversarial.py
shwang: Should be "every training epoch" for both parameters. Changed to this.

src/imitation/scripts/expert_demos.py
shwang: Extracting rewards from a vectorized monitors seems like a pain. I could imagine wrapping the VecEnv result from `make_vec_env` in a custom `VecEnvWrapper` that knows how to access all the inner monitor logs and concatenate the dataframes from `stable_baselines...monitor.load_result`.

I'm partial to making an ad-hoc-y context that temporarily sets `venv.training = False` if `venv` is a VecNormalize (this will have to assume that we never wrap `VecNormalize` in order to work). Within this context we could call `util.rollout.mean_return(venv)` to get unnormalized rewards and also don't have to worry about this evaluation affecting the statistics tracked by `VecNormalize`.

AdamGleave: I'm taking this as being in response to both this and the above comment I made about `VecNormalize` clipping reward.

I don't think 2) would work: setting `training` to false just stops it updating statistics, it'll still clip reward & rescale it. I guess you could have a context that changes `self.norm_reward` and `self.clip_reward` but in any case this feels hacky. If you wanted to do that I'd just an outer wrapper that wraps an inner wrapp and lets you temporarily bypass the inner wrapper.

What's the purpose of the rollouts here? As far as I can tell it's just for logging. Do we need to log this at all, given PPO2 (and other RL algorithms) already log mean rewards during training? We can perhaps eliminate `ray.tune.track.log` entirely, or if we want it just log everything that Stable Baselines logs (see https://github.com/HumanCompatibleAI/adversarial-policies/blob/master/src/modelfree/multi/train_worker.py#L13 for an example of how to do this). This would be my favoured solution if possible.

However, it may be we actually need to know the mean reward in code. For example, if you wanted to do early stopping in Ray based on this, or if you were dissatisfied with what Stable Baselines logs. In that case a solution like (1) seems cleanest. I would not want to use `load_result`: that reads from a CSV that is still being written, which invites problems with buffering, race conditions, etc. Instead why not just look for the `episode` key in the `info` dict, which is how `Monitor` is intended to be used (and how PPO2 etc reports things)? If that doesn't work, accessing the `ep_rew` property directly still feels better than loading a CSV.

Note scripts like `eval_policy` could also run into this issue (we restore the `VecNormalize`), but here since we are not training a solution like 2) seems more palatable (only a one-off change, not messing around with contexts). It'd be nice not to have to do this though.

src/imitation/scripts/train_adversarial.py
shwang: Not necessary any more since "DiscrimNet" is now a valid key.

src/imitation/scripts/expert_demos.py
shwang: Decided not to use `Monitor`. As discussed in [design docs](https://www.notion.so/chaiberkeley/PR-88-Transfer-learning-from-reward-models-Ray-a2dbae9c22014c8ebc14e6c1740dcc35#16acf51dd42145f9be0024973e6c1778) I decided to use Tensorboard+SB.

src/imitation/scripts/expert_demos.py
shwang: Decided not to use `Monitor`. As discussed in [design docs](https://www.notion.so/chaiberkeley/PR-88-Transfer-learning-from-reward-models-Ray-a2dbae9c22014c8ebc14e6c1740dcc35#16acf51dd42145f9be0024973e6c1778) I decided to use Tensorboard+SB.

shwang: I think I will use non-context-variant of 2) for evaluating mean episode reward in `eval_policy`. I'm leaning towards using 2) for calculating the results dictionary at the end of `expert_demos_ex` as well.

src/imitation/scripts/expert_demos.py
shwang: (This will be removed because we are no longer using Ray Tune track or reporter)

src/imitation/scripts/expert_demos.py
shwang: Will address Tensorboard+SB in a new PR.

src/imitation/scripts/train_adversarial.py
shwang: Nevermind, I will address `VecNormalize` in a separate PR.

src/imitation/scripts/expert_demos.py
shwang: Also will address VecNormalize in a new PR.

shwang:
The overlap between editting this PR and #90 was too much so I merged that PR into this one.

I've updated the description of this PR so that it reflects the current contents of the PR and not the original `multiprocessing` phase3/phase4 contents.

There's a couple of issues that you and I brought up in this PR and #90 that I'm punting to new PRs, as this combined PR already is a mishmash of a lot of things.
  * How to track the performance of the mean episode reward under different hyperparameter settings? I decided on visual inspection via SB+Tensorboard and will be implementing this later.
  * How do we deal with interactions between VecNormalize + ep reward mean evaluation? Using SB+Tensorboard over `ray.tune` avoided most of the problem. However, I still want `eval_policy` and `expert_demos` to return results dicts containing the final policy's (unnormalized) ep reward mean. I think that in both cases I will just directly disable reward normalization in the VecEnv. (It will probably be cleaner than one would think, because after training, there will never be any need for reward normalization.)

setup.py
```diff
@@ -22,6 +22,7 @@
     install_requires=[
         'gym',
         'numpy>=1.15',
+        'ray[debug]==0.7.3',
```
AdamGleave: Why not the latest release 0.7.4?

src/imitation/scripts/expert_demos.py
```diff
@@ -179,6 +190,16 @@ def callback(locals_: dict, _) -> bool:
         output_dir = os.path.join(policy_dir, "final")
         serialize.save_stable_model(output_dir, policy, vec_normalize)
 
+      # Final evaluation of expert policy.
```
AdamGleave: Consider adding a TODO here so we don't forget to fix `VecNormalize` related issues.

src/imitation/scripts/parallel.py
```diff
@@ -0,0 +1,66 @@
...
+    # are not picklable, and Ray requires this function to be picklable.
+    from imitation.scripts.expert_demos import expert_demos_ex
+    from imitation.scripts.train_adversarial import train_ex
+    experiments = {
+      "expert_demos": expert_demos_ex,
+      "train_adversarial": train_ex,
+    }
+    ex = experiments[inner_experiment_name]
+
+    with util.make_session():
```
AdamGleave: Why is this necessary? Don't experiments make a session inside their `@ex.main` functions?

src/imitation/scripts/parallel.py
```diff
@@ -0,0 +1,66 @@
...
+
+import ray
+import ray.tune
+
+from imitation.scripts.config.parallel import parallel_ex
+import imitation.util as util
+
+
+@parallel_ex.main
+def parallel(inner_experiment_name: str, search_space: dict) -> None:
```
AdamGleave: How do I set fixed parameter values (that differ from the defaults for the experiment)? Do I have to make it part of `search_space` (even though I'm not searching over it)?

shwang: Yes, I think that's right

setup.py
shwang: updated

shwang:
> How about changing util.rollout.rollout_stats to use infos['epinfo']['r'] as the episode reward -- that way it'll be fixed automatically everywhere?

That sounds good -- implemented this in #97.

> I would strongly recommend thinking about Ray's upload_dir.

Ok, I will probably do this in another PR.

src/imitation/scripts/parallel.py
AdamGleave: This is a little unfortunate as it's often useful to distinction between fixed and variable parameters, as you might e.g. want to report the variable parameters as the TB run name but not the fixed ones. I'd suggest considering restructuring this to distinguish them, but if it's not causing you problems feel free to ignore.

shwang: Sounds good, I'll do that if it makes TB too hard to read later.

src/imitation/scripts/parallel.py
```diff
@@ -0,0 +1,63 @@
...
+  `ex.run(...)` because we want to be able to hyperparameter tune over both the
+  `named_configs` and `config_updates` arguments.
+
+  The Ray Tune `reporter` is not passed to the inner experiment.
+
+  Args:
+    inner_experiment_name: The experiment to tune. Either "expert_demos" or
+      "train_adversarial".
+    command_name: If provided, then run this particular command. Otherwise, run
+      the Sacred Experiment's main command.
```
shwang:
```suggestion
```

src/imitation/scripts/parallel.py
```diff
@@ -0,0 +1,63 @@
...
+  argument `config`. The wrapped function uses `config` as keyword args for
+  `ex.run(...)` because we want to be able to hyperparameter tune over both the
+  `named_configs` and `config_updates` arguments.
+
+  The Ray Tune `reporter` is not passed to the inner experiment.
+
+  Args:
+    inner_experiment_name: The experiment to tune. Either "expert_demos" or
+      "train_adversarial".
+    command_name: If provided, then run this particular command. Otherwise, run
```
shwang:
```suggestion
```

src/imitation/scripts/parallel.py
shwang: Implemented in #101

PR #93, opened 2019-09-06
Explicit types in util.rollout

AdamGleave:
Introduces two new types: `Trajectory`, for a single trajectory, and `Transitions`, for flattened trajectories.

I think this makes the code easier to understand (as opposed to having to memorize which index elements are in, leading to code like `flatten_trajectories(...)[:3]`) and enables more static analysis.

I've added two fields to Transitions:
  - `rew` for reward. This was present in some places but not others. It's not always needed, and I think there's a case to be made that it should not be in e.g. `ReplayBuffer` (which is only used by AIRL code, that really should not be looking at rewards...) However it doesn't take much extra space to store, so I think it's easiest to just always have it.
  - `done` for episode termination. This isn't used anywhere, but should be (important for potential shaping.) Will leave that to another PR.

Minor question: I'm unsure if the field names should be singular or plural. Leaning towards plural, but what about `old_obs` and `new_obs`? Should they be `old_obss`, `old_obses`? Or could make it long form: `old_observations`?

@qxcv: added you as a reviewer since I know you wrote some of the rollout code originally, but feel free to ignore this.


src/imitation/algorithms/adversarial.py
```diff
@@ -251,28 +250,31 @@ def _build_disc_feed_dict(self, *,
       tf.logging.debug("_build_disc_feed_dict: No generator rollout "
                        "parameters were "
                        "provided, so we are generating them now.")
-      gen_old_obs, gen_act, gen_new_obs = self._gen_replay_buffer.sample(
+      gen_sample = self._gen_replay_buffer.sample(
           self._n_disc_samples_per_buffer)
+      gen_old_obs = gen_sample.old_obs
+      gen_act = gen_sample.act
+      gen_new_obs = gen_sample.new_obs
```
qxcv: May as well delete these and `s/gen_\(old_obs\|act\|new_obs\)/gen_sample.\1/g`.

src/imitation/algorithms/bc.py
```diff
@@ -30,17 +29,16 @@ class BCTrainer:
...
                batch_size: int = 32):
     self.env = env
     self.policy_class = policy_class
     self.batch_size = batch_size
-    expert_obs, expert_acts, _, _ = expert_rollouts
     self.expert_dataset = Dataset(
         {
-            "obs": expert_obs,
-            "act": expert_acts,
+            "obs": expert_demos.old_obs,
```
qxcv: This is confusing naming. If `old_obs`, why not `old_act`? Going to add comment on the `Transitions` class below.

src/imitation/util/rollout.py
```diff
@@ -16,18 +16,44 @@
...
+    rew: Reward. Shape: (batch_size, ).
+        The reward `rew[i]` at the i'th timestep is received after the agent has
+        taken action `act[i]`.
+    done: Boolean array indicating episode termination. Shape: (batch_size, ).
+        `done[i]` is true iff `new_obs[i]` the last observatoin of an episode.
+  """
 
-For details see the docstring for `generate_transitions`.
-"""
+  old_obs: np.ndarray
```
qxcv: Most RL code I've seen uses `act` and `obs` to refer to observation seen at time `t` and action taken at time `t`, and `next_obs`, `nobs`, `obsn`, `obs_next`, etc. to refer to the next observation at time `t+1` (in analogy to the `(s,a,r,s')` convention in textbooks). The current naming scheme leads to confusing stuff like the `old_obs` vs. `act` that I highlighted in the BC code. IMO a better scheme would be `s/old_obs/obs/` and `s/new_obs/next_obs/` or something.

tests/test_density_baselines.py
```diff
@@ -40,8 +41,8 @@ def test_density_reward(density_type, is_stationary):
   env = util.make_vec_env(env_id, 2)
 
   # construct density-based reward from expert rollouts
-  expert_trajectories_all = rollout.load_trajectories(
-    f"tests/data/rollouts/{env_id}_*.pkl")
+  glob = f"tests/data/rollouts/{env_id}_*.pkl"
```
qxcv: Hyper-pedantry: maybe `pattern` is more appropriate than `glob` (`glob` is a verb AFAIK, and is also the name of both a built-in module and the only useful function in that module).

src/imitation/util/rollout.py
```diff
@@ -16,18 +16,44 @@
...
+        by the agent when choosing action `act[i]`.
+    act: Actions. Shape: (batch_size, ) + action_shape.
+    new_obs: New observation. Shape: (batch_size, ) + observation_shape.
+        The i'th observation `new_obs[i]` in this array is the observation after
+        the agent has taken action `act[i]`.
+    rew: Reward. Shape: (batch_size, ).
+        The reward `rew[i]` at the i'th timestep is received after the agent has
+        taken action `act[i]`.
+    done: Boolean array indicating episode termination. Shape: (batch_size, ).
+        `done[i]` is true iff `new_obs[i]` the last observatoin of an episode.
```
shwang:
```suggestion
        `done[i]` is true iff `new_obs[i]` the last observation of an episode.
```

src/imitation/util/rollout.py
```diff
@@ -16,18 +16,44 @@
...
+    rew: Reward, shape (trajectory_len, ).
+  """
+
+  act: np.ndarray
+  obs: np.ndarray
+  rew: np.ndarray
+
+
+class Transitions(NamedTuple):
+  """A batch of transition obs-act-obs-rew transitions.
```
shwang:
```suggestion
  """A batch of obs-act-obs-rew-done transitions.
  
  Usually generated by combining and processing several Trajectories via `flatten_trajectory()`.
```

src/imitation/util/rollout.py
```diff
@@ -16,18 +16,44 @@
...
 from . import util  # Relative import needed to prevent cycle with __init__.py
 
-TrajectoryList = List[Dict[str, np.ndarray]]
-"""A list of trajectory dicts.
 
-Each dict contains the keys 'act', 'obs', and 'rew'. For details on these
-key-value pairs, see the docstring for `generate_trajectories`.
-"""
+class Trajectory(NamedTuple):
+  """A trajectory.
```
shwang:
```suggestion
  """A trajectory.
  
  e.g.: A single demonstration from an expert policy.
```

src/imitation/util/rollout.py
```diff
@@ -16,18 +16,44 @@
 
 from . import util  # Relative import needed to prevent cycle with __init__.py
 
-TrajectoryList = List[Dict[str, np.ndarray]]
-"""A list of trajectory dicts.
 
-Each dict contains the keys 'act', 'obs', and 'rew'. For details on these
-key-value pairs, see the docstring for `generate_trajectories`.
-"""
+class Trajectory(NamedTuple):
```
shwang: I think it's easy to mix up the two NamedTuples since they have similar fields and class names.

Renaming `Trajectory` => `Demonstration` might be nice.
This would necessitate a lot of downstream manual renaming of local variables and methods though, so probably best to make to punt this to an issue.

shwang:
> Minor question: I'm unsure if the field names should be singular or plural. Leaning towards plural, but what about old_obs and new_obs? Should they be old_obss, old_obses? Or could make it long form: old_observations?

Prefer short names over "long form" because these names are used so often and usually one-after-another. Don't feel strongly about changing to plural except that `old_obss` looks silly 😛

We could do what Gym does with VecEnv and use "{old,new}_obs", "acts", "rews", "dones".

Again, feel free to make this into an issue instead of refactoring yourself

src/imitation/algorithms/adversarial.py
AdamGleave: I don't think we can do this as `gen_*`  can also be supplied as keyword arguments. I could change the argument to optionally take a `Transition` object instead of `gen_*`, but that would then force the user to provide some irrelevant attributes (e.g. `rew`, `done`).

src/imitation/util/rollout.py
AdamGleave: I find `Demonstration` more confusing. Both `Trajectory` and `Transitions` can hold demonstrations. Some algorithms are trajectory-centric (e.g. GCL, deep RL from human preferences), others are fine with unordered transitions (e.g. most IRL algorithms, BC).

I'd be open to renaming these in general though.

AdamGleave:
Thanks for the comments, I believe I've addressed everything except the two unresolved comments (name of `Trajectory`, and the presence of `gen_*`).

src/imitation/algorithms/adversarial.py
qxcv: Oh, my suggestion was more minor than that: I don't mean passing `gen_sample` to downstream functions, I just mean referring to the attributes directly within this function instead of creating variables.

AdamGleave: This function, `_build_disc_feed_dict`, takes `gen_*` as arguments -- it only computes `gen_sample` if they're `None`. (This isn't obvious from the context provided in the diff.)

qxcv: Thank you for the clarification! I should have read more carefully.

src/imitation/util/rollout.py
shwang: Ok, given that explanation these names seem fine!

PR #94, opened 2019-09-06
Serialize: Add abstract methods `_load` and `_save`

shwang:
`Serializable.load` is no longer abstract. Instead, it depends on abstract `Serializable._load`.

Also added `ABC` mixin to abstract classes `DiscrimNet` and `RewardNet`. Somehow `pytype` started calling out their unimplemented `build_policy_train_reward` after some changes to serializable.py.

### DIFF
--- a/src/imitation/rewards/discrim_net.py
+++ b/src/imitation/rewards/discrim_net.py
@@ -81,6 +81,7 @@ class DiscrimNet(serialize.Serializable, ABC):
     Returns:
         The rewards. Its shape is `(batch_size,)`.
     """
+    del steps
     log_act_prob = np.squeeze(
       gen_log_prob_fn(observation=old_obs, actions=act, logp=True))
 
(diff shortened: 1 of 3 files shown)


shwang:
This PR is sparse on documentation (in particular it deleted some new docstrings from #92 that no longer make sense after these changes). If this PR seems like a good idea then I can add new docstrings

src/imitation/rewards/discrim_net.py
```diff
@@ -12,7 +12,7 @@
 from imitation.util import serialize
 
 
-class DiscrimNet(serialize.Serializable):
+class DiscrimNet(serialize.Serializable, ABC):
```
AdamGleave: This is unnecessary. Subclasses of an abstract class are abstract until they implement all abstract methods. See https://stackoverflow.com/questions/48814211/python3-how-does-one-define-an-abstract-subclass-from-an-existing-abstract-cla

src/imitation/util/serialize.py
```diff
@@ -9,22 +9,23 @@
 
 from imitation.util import util
 
-T = TypeVar('T')
+T = TypeVar('T', bound='Serializable')
```
AdamGleave: :+1:

src/imitation/rewards/discrim_net.py
AdamGleave: Oh I see, `pytype` is complaining about it, very odd.

PR #99, opened 2019-09-18
Test suite improvements: use tmpdir fixture and make test_envs a test class

AdamGleave:
Minor clean-up of the test suite. No (intentional) semantic changes. Two main differences:
  - Turns out pytest implements a `tmpdir` fixture. Bye-bye `with tempfile.TemporaryDirectory(...)` that was cluttering the code. Also has advantage that temporary directories follow a logical naming convention (based on the test name).
  - Created a new module `imitation.testing.envs` that factors out some generally useful methods from `imitation.tests.test_envs`, which could be used by end-users to test their own custom Gym environments (for example, in my codebase). `tests.test_envs` is now just a thin-wrapper around this.

  I've modeled this loosely on NumPy, which has `np.testing` containing generally helpful methods like `np.testing.assert_equal` and `np.tests` containing the unit tests themselves.


AdamGleave:
Supporting the second use case (`test_envs` from outside imitation) is more complicated than I thought. Have to make `tests` part of the `imitation` package. This is common (e.g. `import numpy.tests` works) but goes against https://blog.ionelmc.ro/2014/05/25/python-packaging/ which we've been otherwise following. I think his reason for excluding tests is relatively weak (about it not being the responsibility of end-users to run library tests) so I feel OK doing this, but it is a bit annoying.

An alternative is to just lift the relevant methods from `test_env` into e.g. `imitation.testing` (similar to how `numpy` has `numpy.testing` that exports generally useful utility methods, and `numpy.tests` that is the actual test suite). Or we can just drop the second use case entirely.

tests/test_envs.py
```diff
@@ -20,18 +21,6 @@ def matches_list(env_name, patterns):
...
-    env = gym.make(env_name)
-  except gym.error.DependencyNotInstalled as e:  # pragma: no cover
-    if e.args[0].find('mujoco_py') != -1:
-      pytest.skip("Requires `mujoco_py`, which isn't installed.")
-    else:
-      raise
-  return env
-
-
 def rollout(env, actions):
```
shwang: Looks like this is a duplicate of envs.rollout

tests/test_envs.py
```diff
@@ -49,90 +38,41 @@ def assert_equal_rollout(rollout_a, rollout_b):
...
-  state = env.initial_state()
-  assert env.state_space.contains(state)
-
-  action = env.action_space.sample()
-  new_state = env.transition(state, action)
-  assert env.state_space.contains(new_state)
-
-  reward = env.reward(state, action, new_state)
-  assert isinstance(reward, float)
+class TestEnvs:
```
shwang: I'm curious why you chose to add a class wrapper here. Does this serve any purpose besides adding another layer of hierarchy in the test logs?

tests/test_envs.py
shwang: `assert_equal_rollout` (next function down) doesn't have coverage either

AdamGleave: thanks, fixed

tests/test_envs.py
AdamGleave: All the methods are parameterized by `env_name`, so it seems a bit cleaner to parameterize the class rather than individual methods.

PR #100, opened 2019-09-18
Upload dir

shwang:
Upload `parallel`'s Ray Tune results to my S3 bucket at `s3://shwang-chai/{inner_experiment_name}/{uuid}/`.


src/imitation/scripts/config/parallel.py
```diff
@@ -21,6 +25,21 @@ def config():
...
     "config_updates": {},
   }  # `config` argument to `ray.tune.run(trainable, config)`
+  s3_bucket = None  # Used to create default `upload_dir` if not None.
+
+
+@parallel_ex.config
+def ray_upload_dir(inner_experiment_name, s3_bucket):
+  if s3_bucket is not None:
+    upload_dir = "s3://{}".format(
+      osp.join(s3_bucket, inner_experiment_name, util.make_unique_timestamp()))
```
AdamGleave: This is a very flat directory structure. Fine if your workflow is: find the Sacred config.json, look up the upload_dir, grab from S3. But you won't be able to find the experiment you want just from looking at the directory.

src/imitation/scripts/config/parallel.py
```diff
@@ -21,6 +25,21 @@ def config():
     "named_configs": [],
     "config_updates": {},
   }  # `config` argument to `ray.tune.run(trainable, config)`
+  s3_bucket = None  # Used to create default `upload_dir` if not None.
+
+
+@parallel_ex.config
+def ray_upload_dir(inner_experiment_name, s3_bucket):
```
AdamGleave: If you're only running stuff locally, not clear you want/need S3. Although if you expect to run stuff on AWS in the future, keeping everything on S3 is reasonable (so you always have a single source for experimental results).

shwang: `upload_dir` doesn't support local file syncing at least by default (maybe you guys were able to do this via the `rsync` driver thing in the aprl repo).

Right I'm locally running experiments which dump things into S3, and which I download via `aws s3 sync` into `astar` to read. It takes a minute or so, but I get this nice directory hierachy with hyperparams inside the directory name for free.  Not ideal, but not bad either.

I suppose that I could eliminate the S3 download step and retain directory hierachy by using an appropriate `local_dir`. But now that I think about it, adding organizing levels (timestamp, inner experiment name, ...) to the `local_dir` never carries over to the `upload_dir`. (So I'd end up with an S3 hierachy looking like `timestamp/inner_experiment` and a local directory hierachy looking like `timestamp/inner_experiment/timestamp/inner_experiment`.)

src/imitation/scripts/config/parallel.py
shwang: I've only been looking at the most recent S3 uploads during prototyping (copy and paste to cd into the the final output of `ls /tmp/shwang-chai/`). I can imagine it being hard to distinguish between all the uuids later. Probably at this point I will support adding a prefix to the `util.make_unique_timestamp`.

src/imitation/scripts/config/parallel.py
AdamGleave: Directory structure shouldn't change whether you're using S3 or local directory syncing?

Yeah, I think local file syncing isn't supported out of the box, but it's pretty easy to add. Ray supports any shell command, and you're also welcome to steal `_rsync_func` from https://github.com/HumanCompatibleAI/adversarial-policies/blob/master/src/modelfree/multi/common.py#L39

shwang: > Directory structure shouldn't change whether you're using S3 or local directory syncing?

Yeah, that's correct. `timestamp/inner_experiment/timestamp/inner_experiment` is roughly what would happen if I didn't have `_rsync_func` and I tried to make a directory hierachy in `local_dir` somehow in addition to making hierarchy in `upload_dir`. (Not important, was a weird solution).

PR #101, opened 2019-09-19
Parallel tensorboard for parallel.py

shwang:
* Renames `make_blank_policy_kwargs` to `init_rl_kwargs` (since the function was renamed previously)
* Adds base config options to `parallel_ex`. These config options don't show up in the Tensorboard run name.
* Add `init_tensorboard` options to the `train_ex` and `expert_demos_ex` experiments.

Example TB run names:
` python -m imitation.scripts.parallel with example_cartpole_rl s3` (note that `named_config=["cartpole", "fast"]` and `init_tensorboard=True`, which were set using `base_named_configs` and `base_config_updates` have been omitted).

image


shwang:
Unfortunately it looks like Tensorboard is showing the VecNormalize rewards and not the Monitor rewards

image

```
(pid=28292) --------------------------------------
(pid=28292) | approxkl           | 0.00092371664 |
(pid=28292) | clipfrac           | 0.01015625    |
(pid=28292) | ep_len_mean        | 289           |
(pid=28292) | ep_reward_mean     | 289           |
(pid=28292) | explained_variance | 0.754         |
(pid=28292) | fps                | 1151          |
(pid=28292) | n_updates          | 48            |
(pid=28292) | policy_entropy     | 0.34824577    |
(pid=28292) | policy_loss        | -0.0004412593 |
(pid=28292) | serial_timesteps   | 12288         |
(pid=28292) | time_elapsed       | 49.8          |
(pid=28292) | total_timesteps    | 98304         |
(pid=28292) | value_loss         | 0.040454395   |
(pid=28292) --------------------------------------   
```

Can't see the episode lengths either. Maybe need to train longer?

src/imitation/scripts/expert_demos.py
```diff
@@ -61,7 +63,7 @@ def rollouts_and_policy(
           TimeLimit so that they have at most `max_episode_steps` steps per
           episode.
       normalize: If True, then rescale observations and reward.
-      make_blank_policy_kwargs: Kwargs for `make_blank_policy`.
+      init_rl_kwargs: Kwargs for `make_blank_policy`.
```
AdamGleave: Should `kwargs` be lower case?

src/imitation/scripts/parallel.py
```diff
@@ -24,17 +27,30 @@ def parallel(inner_experiment_name: str,
...
+      Ray directory name.
+    base_config_updates: `search_space["config_updates"]` is applied to this
+      dict before it is passed to the inner experiment's `run()`.
     upload_dir: `upload_dir` argument to `ray.tune.run()`.
   """
-  trainable = _ray_tune_sacred_wrapper(inner_experiment_name)
+  trainable = _ray_tune_sacred_wrapper(inner_experiment_name,
+                                       base_named_configs,
+                                       base_config_updates)
+  ray_loggers = (tune_logger.JsonLogger,)  # Prevent unused Tensorboard write.
```
AdamGleave: Make comment more explicit: does Ray write TensorBoard logs by default? 

For that matter, why do we want any Ray Tune logs -- I don't think we report any metrics? Is this to future-proof it in case we start using that feature?

shwang: It was only annoying to have TB logs because they would clutter up the runs dashboard in Tensorboard.

The JsonLogger seems harmless to include right now.

shwang: Here's an example JSON log:
```
$ cat result.json
{"done": true, "time_this_iter_s": 44.65022039413452, "timesteps_total": null, "episodes_total": null, "training_iteration": 1, "experiment_id": "cead010016b74119b7d941f19f55a728", "date": "2019-09-18_17-54-05", "timestamp": 1568854445, "time_total_s": 44.65022039413452, "pid": 11013, "hostname": "astar", "node_ip": "169.229.222.160", "config": {"config_updates": {"init_rl_kwargs": {"learning_rate": 0.0003, "nminibatches": 16}, "seed": 1}, "named_configs": []}, "time_since_restore": 44.65022039413452, "timesteps_since_restore": 0, "iterations_since_restore": 1, "trial_id": "dea0dbde"}
```

shwang: Ended up removing JSON logs too because for the most part they are a subset of Sacred logs

PR #102, opened 2019-09-20
Add MonitorPlus and Trajectory unwrapping

shwang:
It turns out that `rollout.save` (used by expert_demos.py) saved VecNormalized observations in its Trajectory pickles!

This PR solves this problem by creating a subclass of `Monitor`, `MonitorPlus`, which also adds raw observations and raw rewards to the `info` dict.

`unwrap_traj(traj)` unpacks `traj.info` to generate a copy of `traj` with the original observations and rewards.

`test_bc.py` is back to matching the expert's 500.0. (Although it took me 5 or so tries to train an expert that consistently makes reaches 500.)


shwang:
Originally `unwrap_traj` was a method called `Trajectory.unwrap`. Moved it outside because I was wondering if it was taking up memory in the pkl. (I thought there was a nontrivial chance that the method was passed along with every single named tuple, rather than living with the class. Doesn't look like it).

PR #105, opened 2019-09-27
Imit table

shwang:
Changes necessary for parsing and analyzing Sacred data in imitation learning experiments.

- [x] Currently copying the entire `log_root` via `cp -RL` to copy Sacred file observer directory to `tests/data/imit_benchmark/`. Maybe we simplify by using a [command line option](https://sacred.readthedocs.io/en/latest/observers.html#adding-a-filestorageobserver) to add a second file observer.
 - [x] Add CSV for complete `gail_benchmark.sh` once that is finished running.
- [x] Review changes here and summarize

My apologies for the horrific commit history. I tried rebasing+squashing earlier commits away, but I ran into repeated merge conflicts. I believe that 83bfecb  is the first PR related commit.

PR notes
==

`experiments/`
* Rename gail_benchmark.sh to imit_benchmark.sh
* imit_benchmark.sh now has several new flags, including `--gail` and `--airl`.
* Bash scripts in experiments/ no longer support Mac OS.
  * Because I chose `getopt` (GNU/Linux variant) for flag parsing in `gail_benchmark.sh`.
* Removed some stray references to `auto.pkl` that should be `final.pkl`.

`algorithms/adversarial.py`
* Accept `expert_trajectories` rather than a `rollout_path`.
* Also delete the `n_expert_demos` param. (Caller is expected to handle this logic).

`scripts/analyze.py`
* Added new script `analyze.py` to generate a DataFrame from sacred logs.

`scripts/config/train_adversarial.py`
* Add `rollout_hint` field to help automatically generate the correct rollout path.

`scripts/{expert_demos,train_adversarial}.py`
* Automatically write a symlink to the sacred file observer directory to `{log_dir}/sacred`.
  * In earlier stages of this PR, I used the symlink to help with `tests/generate_data.sh`, but I found a better solution in the `--file_storage tests/data/imit_benchmark/` Sacred option.
  * Leaving this in here since it seems like it could be handy.

`tests/data/imit_benchmark/`
* Add new test data (Sacred `.json`s)
* This data can be regenerated by `tests/generate_data.sh`


shwang:
The GAIL table lives! (Partial run)
https://gist.github.com/shwang/75d70e13a2dc6e773ee621f33e40c866

experiments/imit_benchmark.sh
```diff
@@ -0,0 +1,84 @@
...
+    --file_storage)
+      extra_options+="--file_storage $2 "
+      shift 2
+      ;;
+    --)
+      shift
+      break
+      ;;
+    *)
+      echo "Parsing error" >&2
```
AdamGleave: Consider changing to `echo "Unrecognized flag: $1" >&2`

experiments/imit_benchmark.sh
```diff
@@ -0,0 +1,84 @@
...
+      ;;
+    --log_root)
+      LOG_ROOT="$2"
+      shift 2
+      ;;
+    --file_storage)
+      extra_options+="--file_storage $2 "
+      shift 2
+      ;;
+    --)
```
AdamGleave: When is this needed? It's a common idiom when you want to take some variable-length positional (non-flag) arguments, but we don't seem to use this?

src/imitation/algorithms/adversarial.py
```diff
@@ -313,9 +312,8 @@ def _n_steps_if_not_none(n_steps):
 
 
 def init_trainer(env_id: str,
-                 rollout_path: str,
+                 expert_trajectories: Sequence[rollout.Trajectory],
```
AdamGleave: :+1: much simpler, algorithms shouldn't be responsible for file I/O

src/imitation/scripts/analyze.py
```diff
@@ -0,0 +1,95 @@
...
+      scripts.
+    run_name: If provided, then only analyze results from Sacred directories
+      associated with this run name. `run_name` is compared against the
+      "experiment.name" key in `run.json`.
+    skip_failed_runs: If True, then filter out runs where the status is FAILED.
+    csv_output_path: If provided, then save a CSV output file to this path.
+    verbose: If True, then print the dataframe.
+
+  Returns:
+    A list of dictionaries used to generate the analysis DataFrame.
```
AdamGleave: Think it returns a DataFrame not a dictionary?

src/imitation/scripts/analyze.py
```diff
@@ -0,0 +1,95 @@
...
+    source_dir: A directory containing Sacred FileObserver subdirectories
+      associated with the `train_adversarial` Sacred script. Behavior is
+      undefined if there are Sacred subdirectories associated with other
+      scripts.
+    run_name: If provided, then only analyze results from Sacred directories
+      associated with this run name. `run_name` is compared against the
+      "experiment.name" key in `run.json`.
+    skip_failed_runs: If True, then filter out runs where the status is FAILED.
+    csv_output_path: If provided, then save a CSV output file to this path.
+    verbose: If True, then print the dataframe.
```
AdamGleave: I believe Sacred prints the return value of `main` anyway -- is this necessary? (Perhaps it gets truncated by default?)

src/imitation/scripts/analyze.py
```diff
@@ -0,0 +1,95 @@
...
+    if imit_stats is not None and expert_stats is not None:
+      # Assume that `result.imit_stats` and `result.expert_stats` are
+      # formatted correctly.
+      row["expert_return_summary"] = _make_return_summary(expert_stats)
+      row["imit_return_summary"] = _make_return_summary(imit_stats, "monitor_")
+      row["imit_expert_ratio"] = (imit_stats["monitor_return_mean"] /
+                                  expert_stats["return_mean"])
+
+  df = pd.DataFrame(rows)
+  df.to_csv(csv_output_path)
```
AdamGleave: What if this is `None`?

src/imitation/scripts/config/parallel.py
```diff
@@ -8,40 +8,37 @@
...
 
-import imitation.util as util
-
 parallel_ex = sacred.Experiment("parallel")
 
 
 @parallel_ex.config
 def config():
   inner_experiment_name = "expert_demos"  # The experiment to parallelize
+  inner_run_name = "DEFAULT"  # CLI --name option. Used for analysis grouping.
```
AdamGleave: Might be better if it defaulted to a random UUID, rather than to a fixed value? (We'll end up with a lot of things with "DEFAULT" because we forgot to set a run name.)

src/imitation/scripts/config/parallel.py
```diff
@@ -52,15 +49,57 @@ def debug_log_root():
...
   inner_experiment_name = "expert_demos"
+  outer_experiment_name = "example-cartpole"
+  n_seeds = 2
   search_space = {
     "config_updates": {
-      "seed": tune.grid_search([0, 1]),
       "init_rl_kwargs": {
-        "learning_rate": tune.grid_search([3e-4, 2e-4]),
-        "nminibatches": tune.grid_search([16, 32]),
+        "learning_rate": tune.grid_search([3e-4 * x for x in (1/3, 1/2, 1, 2)]),
```
AdamGleave: This is quite a narrow range to search for LR over. Normally you'd search in a log-uniform space, so e.g. `1e-x for x in range(2,4)`. If you're confident this is close to the right learning rate then fine to do a granular search.

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -88,22 +91,23 @@ def plots():
   plot_interval = 10
 
 
-# Standard Gym env configs
-
 @train_ex.named_config
 def acrobot():
   env_name = "Acrobot-v1"
+  rollout_hint = "acrobot"
```
AdamGleave: Why not just store rollouts under `env_name` rather than the named config name / `rollout_hint`? This seems like it'd allow us to get rid of a lot of code.

src/imitation/util/sacred.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2019 Google LLC.
```
AdamGleave: Unclear if Google really deserves copyright here. Your code looks very different from my Gist. Question of whether you just read it for inspiration, or iteratively edited it to reach this point.

If we think Google does deserve copyright here, then should put the full Apache 2.0 license header here. You should also add your name: it's not just a Google work.

Otherwise, I think it can be MIT licensed like the rest of this codebase.

src/imitation/util/sacred.py
```diff
@@ -0,0 +1,94 @@
...
+) -> List[str]:
+  """Walks through a directory tree, returning paths to filtered subdirectories.
+
+  Args:
+    root_dir: The start of the directory tree walk.
+    filter_fn: A function with takes a directory path and returns True if
+      we should include the directory path in this function's return value.
+    nested_ok: If False, then error if in the return value, one of the
+      directory paths is a subdirectory of another.
+  Returns:
```
AdamGleave: Should have new line between Args and Returns section

src/imitation/scripts/analyze.py
shwang: Looks like `None` 'overloads' `to_csv` to return a string. But that's not the intended behavior anyways. Adding an `if` statement.

src/imitation/scripts/analyze.py
shwang: `rows` is actually a list of `OrderedDict` right now.

experiments/imit_benchmark.sh
shwang: I changed the error message (because I don't expect us to ever reach that line), but I'm pretty sure that unrecognized flags error out on the `TEMP=$(getopt ...)` line instead.

Hard for me to verify right now because the `getopt` on my Mac is different from the `getopt` installed on Linux distriubtions.

experiments/imit_benchmark.sh
shwang: I"m pretty sure that GNU getopt's output denotes the end of its options section with a `--`. This while loop was copied from example code (don't remember source) that seemed like idiomatic usage of GNU getopt.

I don't know how otherwise to break the while loop.

(Note that GNU/Linux `getopt` is different from Posix `getopt` and the builtin bash command `getopts`).

src/imitation/util/sacred.py
shwang: Seems mostly like inspiration at this point. The "check that no filtered directory is a prefix of a filtered directory" was copied directly from your code however.

src/imitation/scripts/config/train_adversarial.py
shwang: If I don't do this, then the bash scripts need to contain a mapping from `env_config_name` => `env_name`.

src/imitation/scripts/analyze.py
shwang: Yes, it gets truncated by default

src/imitation/scripts/analyze.py
shwang: I changed it to return a `DataFrame`

src/imitation/scripts/config/parallel.py
shwang: Good idea, added UUID

experiments/imit_benchmark.sh
shwang: Yes, on Linux, an unrecognized flag errors out on the `getopt` line:
```
steven@astar:~/imitation$ ./experiments/imit_benchmark.sh --hello
getopt: unrecognized option '--hello'
```

shwang: Ah, but we would in fact reach the `*` case if `getopt` recognizes a flag that our `case`-statement doesn't. Changing to your suggestion.

experiments/imit_benchmark.sh
shwang: This is the canonical way to break out of `getopt`, according to these tutorials (hard to google search for because `getopt` is so overloaded and has POSIX and GNU/Linux versions)
https://dustymabe.com/2013/05/17/easy-getopt-for-a-bash-script/
https://www.codebyamir.com/blog/parse-command-line-arguments-using-getopt

src/imitation/scripts/config/parallel.py
shwang: Noted, changing all the learning rates to search over the`np.logspace(3e-6, 1e-1, num=3)`. Keeping the number of searches low because, isn't an actual experiment, just an example that I can use to check if things are working properly.

src/imitation/util/sacred.py
shwang: Removing copyright

PR #106, opened 2019-10-07
Add transfer learning script

shwang:
Added a new transfer learning script, `experiments/transfer_learn_benchmark.sh`, which is used to train and benchmark PPO2 experts trained on reward models generated by AIRL and GAIL.

The transfer learning scripts needs reward models, so `experiments/download_models.sh` (formerly `experiments/download_experts.sh`) also downloads `reward_models/{airl,gail}` from S3 now.

- [ ] I haven't generated/uploaded reward models for `AIRL` yet (though these rewards are more important). So right now `download_models.sh` only downloads gail rewards.


experiments/download_experts.sh
```diff
@@ -4,5 +4,7 @@
 SCRIPT_DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd)"
 PROJECT_DIR="$(dirname ${SCRIPT_DIR})"
 EXPERT_MODELS_DIR=${PROJECT_DIR}/expert_models
+REWARD_MODELS_DIR=${PROJECT_DIR}/reward_models
```
AdamGleave: Consider moving all data files to a single subdirectory? So e.g. `data/expert_models`, `data/reward_models`. This will let you decouple the data from source code location more easily in the future: e.g. if you were doing distributed training, you might want to have a shared data output directory mounted via NFS on multiple workers. This would also simplify `.gitignore`, syncing and reduce directory clutter. Would make file names longer though.

experiments/imit_benchmark.sh
```diff
@@ -30,11 +31,11 @@ while true; do
       shift
       ;;
     --gail)
-      extra_configs+="gail "
+      ALGORITHM="gail"
```
AdamGleave: Feels a little odd this is a no-op since `ALGORITHM` is already set to gail, but it probably is better UI to keep the flag.

experiments/imit_benchmark.sh
```diff
@@ -84,4 +86,7 @@ pushd ${LOG_ROOT}/parallel
 find . -name stdout | sort | xargs tail -n 15 | grep -E '==|\[result\]'
 popd
 
+echo "[Optional] Upload new reward models to S3 (replacing old ones) using:"
+echo "aws s3 sync --delete '${LOG_ROOT}' s3://shwang-chai/public/reward_models/${ALGORITHM}/"
```
AdamGleave: :+1: I like the self-documentation here :)

experiments/transfer_learn_benchmark.sh
```diff
@@ -0,0 +1,74 @@
...
+TIMESTAMP=$(date --iso-8601=seconds)
+REWARD_MODELS_DIR=reward_models
+LOG_ROOT="output/train_experts/${TIMESTAMP}"
+RESULTS_FILE="results.txt"
+ALGORITHM="gail"
+extra_configs=""
+
+
+SEEDS="0 1 2"
+ENVS="cartpole mountain_car acrobot half_cheetah "
```
AdamGleave: Why is this a different set of environments to `train_experts.sh`? (I'm guessing this is a GAIL vs AIRL thing?)

experiments/transfer_learn_benchmark.sh
```diff
@@ -0,0 +1,74 @@
...
+    --gail)
+      ALGORITHM="gail"
+      shift
+      ;;
+    --airl)
+      ALGORITHM="airl"
+      shift
+      ;;
+    --run_name)
+      extra_options+="--name $2 "
```
AdamGleave: Where is this variable used?

experiments/transfer_learn_benchmark.sh
```diff
@@ -0,0 +1,74 @@
...
+      ;;
+    *)
+      echo "Unrecognized argument $1"
+      exit 1
+      ;;
+  esac
+done
+
+echo "Writing logs in ${LOG_ROOT}"
+parallel -j 25% --header : --results ${LOG_ROOT}/parallel/ --colsep , --progress \
```
AdamGleave: Is `--colsep ,` needed here (I don't see a CSV)?

AdamGleave: I didn't know about `--progress`, will remember that :)

experiments/imit_benchmark.sh
shwang: Another option would be to force the user to choose GAIL or AIRL flags. Keeping as is for now.

experiments/download_experts.sh
shwang: Good idea, I'm also restructuring tests/data/cartpole_0 to tests/data/expert_models/cartpole_0 to mirror this

experiments/transfer_learn_benchmark.sh
shwang: Good catch, I'm missing the custom AIRL envs. I think I removed them previously for debugging purposes

shwang: Hmmm actually I think what happened is that this is meant to be exactly the same as the environments used in `imit_benchmark_config.csv`. Which only has GAIL environments right now.

experiments/transfer_learn_benchmark.sh
shwang: This is used for grouping TF runs and maybe analysis tables in the future (more specifically, the sacred utils directory finder can filter on run name). Adding a comment for this.

experiments/transfer_learn_benchmark.sh
shwang: No CSV, thanks

experiments/transfer_learn_benchmark.sh
shwang: (also adding extra_options to the parallel command so that it is actually used)

experiments/transfer_learn_benchmark.sh
shwang: I made `train_experts` and `imit_benchmark` have the same environments (superset of GAIL and AIRL environments).

`imit_benchmark_config.csv`, which is used to generate the reward models needed by this transfer learning script, originally had GAIL environments only. This was because I was confused about some missing experiment specifications from the AIRL paper (number of demonstrations used in particular I think).

To add the swimmer, maze, and custom/disabled ant environments into the imit_benchmark CSV, I just winged it and chose some arbitrary settings. Thinking that I can be more careful about these settings later.

AdamGleave: Seems like a good solution!

PR #111, opened 2019-10-28
Support custom arguments to VecNormalize

AdamGleave:
Use case that prompted this was wanting to normalize reward (since learnt reward models can have arbitrary scale) but not observations of an environment.

Shouldn't merge until https://github.com/hill-a/stable-baselines/pull/525 is merged (or at least looks like it'll be merged).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
         'scikit-learn>=0.21.2',
         # FIXME: Use stable release instead of tracking master once
         # commit 9a760542 is released.
-        'stable-baselines @ git+https://github.com/hill-a/stable-baselines.git',
+        'stable-baselines @ git+https://github.com/hill-a/stable-baselines.git@vec-normalize-pickle',
         'jax!=0.1.37',
         'jaxlib~=0.1.20',
         # sacred==0.7.5 build is broken without pymongo
--- a/setup.py
+++ b/setup.py
@@ -27,9 +27,8 @@ setup(
         'ray[debug]==0.7.4',
         'tqdm',
         'scikit-learn>=0.21.2',
-        # FIXME: Use stable release instead of tracking master once
-        # commit 9a760542 is released.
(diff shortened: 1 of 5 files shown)


shwang:
Summary: 

We want to allow reward normalization without observation normalization. The main change is adding a ~`vecnorm_kwargs` argument to `expert_demos` script which gets passed into SB's `VecNormalize` constructor.

This necessitates a change to our `VecNormalize` serialization procedure (why do we need to serialize the environment wrapper `VecNorm`? Because we always need to save `VecNorm` along with pickled policies, otherwise policy is operating on the wrong observation space.), which Adam solved totally by serializing the entire VecNormalize. Serializing VecNormalized required an upstream change making SB's VecNormalize picklable.

The new `VecNormalize` serialization procedure deprecates the current save format of all of our expert policies.

setup.py
```diff
@@ -29,7 +29,7 @@
         'scikit-learn>=0.21.2',
         # FIXME: Use stable release instead of tracking master once
         # commit 9a760542 is released.
-        'stable-baselines @ git+https://github.com/hill-a/stable-baselines.git',
+        'stable-baselines @ git+https://github.com/hill-a/stable-baselines.git@vec-normalize-pickle',
```
shwang: Looks like 9a760542 ready (so should update `FIXME`), but should keep at 'stable-baselines @ git+https://github.com/hill-a/stable-baselines.git' since `vec-normalize-pickle` isn't released yet.

src/imitation/scripts/config/expert_demos.py
```diff
@@ -15,6 +15,7 @@ def expert_demos_defaults():
   num_vec = 8  # Number of environments in VecEnv
   parallel = True  # Use SubprocVecEnv (generally faster if num_vec>1)
   normalize = True  # Use VecNormalize
+  normalize_kwargs = dict()
```
shwang:
```suggestion
  normalize_kwargs = dict()  # kwargs for `VecNormalize`
```

PR #116, opened 2019-11-06
Add experiments/ smoke tests and Fix nonunique imit_benchmark.sh log dirs

### LINKED ISSUE #62: GAIL end-to-end integration test

shwang:
Resolves #62 by adding a smoke test `test_experiments.py`, which checks whether the `--fast` version of each script in the pipeline can execute without crashing.

After running the script, I found and resolved the following bug:

I've accidentally been using the same log dirs for different imitation benchmark `parallel` jobs because I thought that the log dir depends only on (1) the environment name and (2) the seed. Aside from causing confusing behavior when reading reward models, this also led to `FileExistsError`s after the PR that started building symlinks to Sacred logs inside the log dir.

This PR adds another directory to the log dir path that makes the log dir path also depend on (3) the number of expert demonstrations. This is sufficient to give each `parallel` job a unique log dir path.

Now that the `data/reward_models/` directory hierarchy differentiates between reward models trained on different number of expert demonstrations, `transfer_learn_benchmark.sh` reads config directly from `imit_benchmark_config.csv` so that it tries out every available reward model.

(Note that this path change has no effect on the analysis imit-table-generating script, which is mostly directory-hierachy-agnostic, and reads (1), (2), and (3) directly from Sacred logs.)

Other small changes:
* Install `parallel` in `before_install` section of .travis.yml. Stop installing OpenMPI.
* Moved sacred test data into `tests/data/imit_benchmark_sacred`.
* Add `--all` option to `tests/generate_test_data.sh`. Without `--all`, skip generating RL experts (expensive).
* Add `--all` option to `experiments/download_models.sh`. Without `--all`, skip downloading 800MB of sacred logs, TensorBoard files, etc.
* `experiments/imit_benchmark.sh`: Only generate final checkpoint (used to generate checkpoints for every 5 steps).


shwang:
Works on my computer. But on Travis I get the following error when loading reward models:
```
AssertionError: Unused attributes in these objects (the attributes exist in the checkpoint but were not restored):
(root).None_dense0 (<tensorflow.python.layers.core.Dense object at 0x7fb980fd7e80>): ['OBJECT_CONFIG_JSON']
(root).None_dense1 (<tensorflow.python.layers.core.Dense object at 0x7fb980feae10>): ['OBJECT_CONFIG_JSON']
(root).None_dense_final (<tensorflow.python.layers.core.Dense object at 0x7fb99a015630>): ['OBJECT_CONFIG_JSON']
```

I'll regenerating test files and reinstalling Tensorflow later. (Maybe this was caused by a version mismatch of some sort).

experiments/download_models.sh
```diff
@@ -3,5 +3,52 @@
...
+    *)
+      echo "Unrecognized flag $1" >&2
+      exit 1
+      ;;
+  esac
+done
+
+FLAGS=""
+
+if not $ALL_MODE; then
```
AdamGleave: This does not seem like standard `bash`. `not` does not work on Linux at least; maybe this is a Mac thing?

AdamGleave: Try `[[ $ALL_MODE != "true" ]]`

experiments/download_models.sh
```diff
@@ -3,5 +3,52 @@
...
+    --exclude '*events.out.tfevents*' "
+fi
+
+if $DRY_RUN_MODE; then
+  FLAGS+='--dryrun '
+elif [[ -d ${DATA_DIR} ]]; then
+  rm -r ${DATA_DIR}
+fi
+
+command="aws s3 sync ${FLAGS} s3://shwang-chai/public/data/ ${DATA_DIR}"
```
AdamGleave: Can you check your S3 permissions?

```
$ aws s3 sync --dryrun s3://shwang-chai/public/data/ data/                                                                                       
fatal error: An error occurred (AccessDenied) when calling the ListObjects operation: Access Denied
```

experiments/download_models.sh
```diff
@@ -3,5 +3,52 @@
...
+eval $command  # Required for correct interpretation of ALL_MODE flags.
```
AdamGleave: I'm surprised this is needed. What were you trying before and what error did you see?

experiments/imit_benchmark.sh
```diff
@@ -72,9 +73,10 @@ parallel -j 25% --header : --results ${LOG_ROOT}/parallel/ --colsep , --progress
   ${ALGORITHM} \
   ${extra_configs} \
   {env_config_name} \
-  log_dir="${LOG_ROOT}/${ALGORITHM}/{env_config_name}_{seed}" \
+  log_dir="${LOG_ROOT}/{env_config_name}_{seed}/n_expert_demos_{n_expert_demos}" \
```
AdamGleave: Why do we not want `${ALGORITHM}` any longer? I guess this is not needed given log directory is unique thanks to timestamping (modulo race conditions).

experiments/train_experts.sh
```diff
@@ -15,28 +16,40 @@ OUTPUT_DIR="output/train_experts/${TIMESTAMP}"
...
-    fi
-    mkdir -p ${OUTPUT_DIR}
-    extra_configs+="rollout_save_n_episodes=50 "
-  fi
+TEMP=$(getopt -o fr -l fast,regenerate -- $@)
+if [[ $? != 0 ]]; then exit 1; fi
+eval set -- "$TEMP"
+
+while true; do
+  case "$1" in
```
AdamGleave: :+1: using `case` seems much cleaner

tests/generate_test_data.sh
```diff
@@ -3,23 +3,67 @@
...
+alias imit_benchmark="experiments/imit_benchmark.sh -f --file_storage=${sacred_dir}"
 shopt -s expand_aliases
-imit_benchmark --airl --run_name FOO
-imit_benchmark --gail --run_name BAR
-imit_benchmark --gail --run_name BAR
+
+imit_benchmark --gail --run_name BAR --log_root "${reward_models_dir}/gail"
+# Only need one reward model, so leave the rest in temporary directories.
+# The sacred logs are still saved to ${sacred_dir} and used for analysis test.
+imit_benchmark --gail --run_name BAR --log_root "$(mktemp -d)"
```
AdamGleave: Why do we run this twice (once with temp dir, once permanent directory)?

shwang:
> However it'd be nice to track modern versions of TensorFlow: what was the reason for avoiding 1.14? Was it IDSIA/sacred#493? 

Yes, it was the `__spec__` issue in  1.14. I think I'll change to TF 1.15 in this PR.

experiments/imit_benchmark.sh
shwang: The directory that was being uploaded to S3 ending up looking like `data/gail/gail`. Might be better to make a fix in the `aws s3` command instead. Taking a look...

AdamGleave: Probably a trailing `/` where there shouldn't be one (or vice versa), I've had this kind of problem with `rsync` before

experiments/download_models.sh
shwang: Thanks, making this change. I got confused while debugging

experiments/imit_benchmark.sh
shwang: `${ALGORITHM}` was never necessary for uniqueness (minus race conditions, yeah). There was also a problem where both this line and the S3 sync command explicitly appended `${ALGORITHM}/` to the path. To fix the double path `gail/gail`, I removed the `${ALGORITHM}/` from this line.

experiments/download_models.sh
shwang: The exclude flags don't seem to work properly without this set up. I'm guessing this is because they are all lumped together in one string and the `aws` parser somehow gets confused?

If I run this command directly, then I `aws` starts downloading excluded files like the monitor data:
```
(dryrun) download: s3://shwang-chai/public/data/expert_models/acrobot_0/monitor/mon000.monitor.csv to data/expert_models/acrobot_0/monitor/mon000.monitor.csv
(dryrun) download: s3://shwang-chai/public/data/expert_models/acrobot_0/monitor/mon001.monitor.csv to data/expert_models/acrobot_0/monitor/mon001.monitor.csv
(dryrun) download: s3://shwang-chai/public/data/expert_models/acrobot_0/monitor/mon002.monitor.csv to data/expert_models/acrobot_0/monitor/mon002.monitor.csv
```

tests/generate_test_data.sh
shwang: In `test_scripts`, `test_analyze_imitation` expects `${sacred_dir}` to contain sacred logs from 2 runs of `imit_benchmark --fast --run_name BAR`.

I'm been considering somehow moving this test data generation to `test/test_scripts.py` itself. In addition to grouping all the logic in one place, it would also leave me with fewer test files to generate and commit when make changes.

shwang: `experiments/imit_benchmark.sh -f` expects to find Cartpole and Pendulum reward models in `${reward_models_dir}/gail`. Don't need to generate this twice, so I discard the reward model from the second call in a temporary directory.

shwang: I removed the (confusing) repeated commands by generating Sacred logs within the relevant test.

experiments/download_models.sh
AdamGleave: I've made a commit that fixes this using `bash` arrays.

I'd generally avoid `eval` unless you need to use things like redirection, pipes. It just splits on whitespace, so will break if you ever had e.g. a directory with whitespace in it that you wanted to exclude.

shwang: Thanks!

shwang:
I'm now dynamically generating `tests/data/reward_models` instead of commit reward models to GH.

Ideally I'd do this for `tests/data/expert_models` too (by running `experiments/train_experts.sh --regenerate` before running any pytests), but some of the density baselines tests require expert Cartpole and Pendulum models. This takes way too long to train, and PPO2 sometimes gives a bad local minimum. (I expect this would add like 20 minutes to each Travis CI job.)

AdamGleave:
> I'm now dynamically generating `tests/data/reward_models` instead of commit reward models to GH.
Sweet!

> Ideally I'd do this for `tests/data/expert_models` too (by running `experiments/train_experts.sh --regenerate` before running any pytests), but some of the density baselines tests require expert Cartpole and Pendulum models. This takes way too long to train, and PPO2 sometimes gives a bad local minimum. (I expect this would add like 20 minutes to each Travis CI job.)
Yeah this doesn't seem worth it if it would slow down CI. If we're concerned about size of binary files we could try Git LFS or just sync these stuff down from S3.

PR #121, opened 2019-11-19
Migrate to CircleCI

AdamGleave:
- [X] Add CircleCI Config
  - [x] Delete .travis.yml config
  - [X] Add Dockerfile and build image, with MuJoCo support
  - [X] Parallelize tests using `pytest-shard`.

Should run significantly faster now.

No longer testing 3.6 so I've dropped support from that in `setup.py`. I don't feel strongly about this so we could also just test in 3.6 instead of 3.7. However I feel OK with only officially supporting latest release (3.8) plus last one.


Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN curl -o /usr/local/bin/patchelf https://s3-us-west-2.amazonaws.com/openai-sci-artifacts/manual-builds/patchelf_0.9_amd64.elf \
+    && chmod +x /usr/local/bin/patchelf
+
+ENV LANG C.UTF-8
+
+RUN    mkdir -p /root/.mujoco \
+    && curl -o mjpro150.zip https://www.roboti.us/download/mjpro150_linux.zip \
```
shwang: I've been using mujoco200 to build the expert and reward models currently on S3. No preference between the two right now, though I'll have to regenerate everything later on, which currently takes a few days. Is there a reason to favor 150 over 200?



(I noticed that you changed from 200 to 150 in the commit history.)

Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    && unzip mjpro150.zip -d /root/.mujoco \
+    && rm mjpro150.zip
+
+# Set the PATH to the venv before we create the venv, so it's visible in base.
+# This is since we may create the venv outside of Docker, e.g. in CI
+# or by binding it in for local development.
+ENV PATH="/imitation/venv/bin:$PATH"
+ENV LD_LIBRARY_PATH /usr/local/nvidia/lib64:/root/.mujoco/mjpro150/bin:${LD_LIBRARY_PATH}
+
+FROM base as python-req
```
shwang: I'm guessing that `base` is used here for caching purposes. Should `python-req` used for development, and `full` used for CI?

Not sure why `python-req` is useful to have as a separate build.

Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+
+WORKDIR /imitation
+# Copy over just setup.py and __init__.py (including version)
+# to avoid rebuilding venv when requirements have not changed.
+COPY ./setup.py /imitation/setup.py
```
shwang: Consider changing `dst` to `./setup.py` here and in the rest of the COPY lines to take advantage of the WORKDIR instruction.

Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+COPY ./src/imitation/__init__.py /imitation/src/imitation/__init__.py
+COPY ./ci/build_venv.sh /imitation/ci/build_venv.sh
+# mjkey.txt needs to exist for build, but doesn't need to be a real key
+RUN    mkdir -p /imitation/src/imitation \
```
shwang: This `mkdir` might be unnecessary, given that we are already successfully copying a file into `/imitation/src/imitation/__init__.py`

Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+    && touch /root/.mujoco/mjkey.txt \
+    && ci/build_venv.sh \
+    && rm -rf $HOME/.cache/pip
+
+FROM python-req as full
+
+# Delay copying (and installing) the code until the very end
+COPY . /imitation
```
shwang: Maybe also take advantage of the WORKDIR here too, but might look confusing
```suggestion
COPY . .
```

.gitignore
```diff
@@ -55,6 +54,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *.cover
+*.py,cover
```
shwang: Is this comma meant to be a `.`?

.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
...
+          name: codespell
+          command: codespell -I .codespell.skip --skip='*.pyc,tests/data/*,*.ipynb,*.csv' ${SRC_FILES}
+
+      - run:
+          name: sphinx
+          command: pushd docs/ && make clean && make html && popd
+
+      - run:
+          name: pytype
+          command: pytype ${SRC_FILES}
```
shwang: Why do we duplicate `ci/code_checks.sh` commands here instead of calling the script? (Maybe since this allows us check all of them even if the first check fails?)

.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
...
+            done
+          background: true
+
+      - run:
+          name: run tests
+          command: |
+            export DISPLAY=:0
+            pytest --cov=venv/lib/python3.7/site-packages/imitation --cov=tests \
+                   --junitxml=/tmp/test-reports/junit.xml \
+                   --shard-id=${CIRCLE_NODE_INDEX} --num-shards=${CIRCLE_NODE_TOTAL} \
```
shwang: Pretty neat how Circle CI has this sharding thing set up out-of-the-box

.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
...
+                    -vv tests/
+            mv .coverage .coverage.aprl
```
shwang: Prob want to use a different suffix than `aprl`

.circleci/config.yml
```diff
@@ -0,0 +1,120 @@
...
+
+      - run:
+          name: install dependencies
+          # MUJOCO_KEY is defined in a CircleCI context
+          # Do some sanity checks to make sure key works
+          command: |
+            curl -o /root/.mujoco/mjkey.txt ${MUJOCO_KEY}
+            md5sum /root/.mujoco/mjkey.txt
+            # Only create venv if it's not been restored from cache
+            # TODO(adam): build_venv
```
shwang: Is this done?
```suggestion
```

Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+# This is since we may create the venv outside of Docker, e.g. in CI
+# or by binding it in for local development.
+ENV PATH="/imitation/venv/bin:$PATH"
+ENV LD_LIBRARY_PATH /usr/local/nvidia/lib64:/root/.mujoco/mjpro150/bin:${LD_LIBRARY_PATH}
+
+FROM base as python-req
+
+WORKDIR /imitation
+# Copy over just setup.py and __init__.py (including version)
+# to avoid rebuilding venv when requirements have not changed.
```
shwang: Looks like we are always building a virtualenv at `venv` via `RUN ... ci/build_venv.sh`, so I'm not sure if this comment applies.

Dockerfile
AdamGleave: Gym currently only supports MuJoCo 1.5. They seem to have added this restriction recently: https://github.com/openai/gym/blob/master/setup.py

Dockerfile
AdamGleave: The intention here is:
  - `base` is used in a CI build.
  - `python-req` is used for local development, where you can bind the source code using `-v` from outside Docker.
  - `full` is used if you want to deploy the finished product, e.g. to Ray or to run tests on a local machine.

The reason I'm using `base` rather than `full` in a CI build is to avoid having to rebuild the Docker image for each build. This is possible but it tends to be slow and is quite a pain to set up. Instead we rebuild the `venv` leveraging CircleCI's built-in caching support.

AdamGleave: (I've added comments documenting this now.)

Dockerfile
AdamGleave: Thanks, done

Dockerfile
AdamGleave: you're right, removed

.gitignore
AdamGleave: I was just copying from https://github.com/github/gitignore/blob/master/Python.gitignore

Apparently it's right: https://github.com/github/gitignore/pull/3131

.circleci/config.yml
AdamGleave: So we have a pretty GUI that gives a breakdown of each stage and timing for it.

If we don't care for that, it'd be more DRY to just call the script.

.circleci/config.yml
AdamGleave: fixed, thanks

Dockerfile
AdamGleave: Docker will cache previously built layers when the `Dockerfile` has not changed and all files are up to date

Dockerfile
AdamGleave: I'm not sure if WORKDIR persists across build stages. But I think it's probably too confusing anyway.

shwang: Sounds good. 

WORKDIR seems to persist forever based on the docs.

.circleci/config.yml
shwang: Keeping pretty GUI seems ok for now. Don't think we will need to update code_checks.sh often.

Dockerfile
shwang: Ah, gotcha, so the `RUN ... build_venv.sh` is only executed if the previous layers have changed.

shwang:
Wow, 4min only for tests! (Used to be almost 20min per job)

Dockerfile
shwang: Yikes, it looks like MuJoCo 2.0 is missing contact forces in the Ant and Humanoid environments. 
https://github.com/openai/gym/issues/1541

shwang:
updated README instructions to match setup.py (supports python 3.8 and 3.7)

Dockerfile
AdamGleave: Ouch, that's a pretty bad bug!

PR #125, opened 2019-11-22
Add ability to save policies with BC

qxcv:
This allows the behavioural cloning class to save and reload policies. It'd be nice if there were some unified way of saving policies, but I think that would require upstream changes to stable baselines (e.g. to _always_ save the kwargs policies were created with so that they can be reconstructed).


AdamGleave:
Stable Baselines places the onus of loading/saving in the RL algorithm, i.e. they have to override `load()` and `save()` in `BaseRLModel`. It might be nice to follow that kind of interface, we could use `BaseRLModel._save_to_file` and `BaseRLModel._load_from_file`, but it's not *that* helpful.

AdamGleave:
Also you need to merge a recent version of `origin/master` into this branch (it's old enough it doesn't have the CircleCI config).

tests/test_bc.py
```diff
@@ -1,22 +1,50 @@
...
+  # just to change the values a little
+  bc_trainer.train(n_epochs=1)
+  var_values = bc_trainer.sess.run(bc_trainer.policy_variables)
+  bc_trainer.save_policy(pol_path)
+  with tf.Session() as sess:
+    # just make sure it doesn't die
+    with tf.variable_scope('new'):
+      bc.BCTrainer.reconstruct_policy(pol_path)
+    new_vars = tf.get_collection(
+        tf.GraphKeys.TRAINABLE_VARIABLES, scope='new')
```
shwang: Probably cleaner to use use `reconstructed_policy.policy_variables`. (Not sure if this is possible.)

setup.py
```diff
@@ -61,6 +62,7 @@
             'isort',
             'jupyter',
             'pytype',
+            'codespell',
```
shwang: Merge note: I moved `codespell`, `pytype` to TESTS_REQUIRE on master.

AdamGleave:
@shwang @qxcv are we blocked on anything here or should someone merge this?

tests/test_bc.py
shwang: (Was mistaken here, as `reconstruct_policy` returns a policy not a BCTrainer)

shwang:
Forgot about this, sorry. Will merge after tests pass

PR #130, opened 2019-11-26
adversarial: Use VecNormalized training reward and update plots

shwang:
`imitation.adversarial.trainer` never applied VecNormalize to rewards or observations during PPO2 training. This means that before this patch, PPO2 for the
GAN's generator was training with an unnormalized reward.

As an example of why this is bad, we have a dozen or so runs showing that training an expert in PPO2+MountainCar achieves minimum episode reward when observations are not normalized.

* Adding a new instance variable `AdversarialTrainer.env_train_norm` which has VecNormalized reward. `AdversarialTrainer.env_train` remains even though it isn't used for training GAIL/ARIL because it is useful for plotting training reward in `scripts/train_adversarial.py`.

* Plotting logic in the `train_adversarial` script needed an update after this change because rollouts no longer function properly on the three rollout environments `trainer.{,train_,test_}venv` without reapplying `trainer.train_venv_norm`'s normalization wrapper.
  * I fixed up some tests in plots in the meantime

### DIFF
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -103,16 +103,18 @@ class AdversarialTrainer:
     self._sess.run(tf.global_variables_initializer())
 
     if debug_use_ground_truth:
-      self.venv_train = self.venv_test = self.venv_train_norm = self.venv
+      self.venv_train = self.venv_test = self.venv
     else:
       reward_train = partial(
           self.discrim.reward_train,
           gen_log_prob_fn=self._gen_policy.action_probability)
+
       self.venv_train = reward_wrapper.RewardVecEnvWrapper(
           self.venv, reward_train)
       self.venv_test = reward_wrapper.RewardVecEnvWrapper(
           self.venv, self.discrim.reward_test)
-      self.venv_train_norm = VecNormalize(self.venv_train)
+
+    self.venv_train_norm = VecNormalize(self.venv_train)
 
     if gen_replay_buffer_capacity is None:
       gen_replay_buffer_capacity = 20 * self._n_disc_samples_per_buffer
@@ -282,8 +284,13 @@ class AdversarialTrainer:
     assert n_gen == len(gen_acts)
     assert n_gen == len(gen_next_obs)
 
+    # Normalize expert observations to match generator observations.
+    assert isinstance(self.venv_train_norm, VecNormalize)
+    expert_obs_norm = self.venv_train_norm._normalize_observation(
+      expert_sample.obs)
+
     # Concatenate rollouts, and label each row as expert or generator.
-    obs = np.concatenate([expert_sample.obs, gen_obs])
+    obs = np.concatenate([expert_obs_norm, gen_obs])
     acts = np.concatenate([expert_sample.acts, gen_acts])
     next_obs = np.concatenate([expert_sample.next_obs, gen_next_obs])
     labels = np.concatenate([np.zeros(n_expert, dtype=int),
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -103,16 +103,18 @@ class AdversarialTrainer:
     self._sess.run(tf.global_variables_initializer())
 
     if debug_use_ground_truth:
+      # Would use an identity reward fn here, but RewardFns can't see rewards.
+      self.reward_train = self.reward_test = None
       self.venv_train = self.venv_test = self.venv
     else:
-      reward_train = partial(
+      self.reward_train = partial(
           self.discrim.reward_train,
           gen_log_prob_fn=self._gen_policy.action_probability)
-
+      self.reward_test = self.discrim.reward_test
       self.venv_train = reward_wrapper.RewardVecEnvWrapper(
-          self.venv, reward_train)
+          self.venv, self.reward_train)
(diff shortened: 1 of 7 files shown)


shwang:
No need to update `expert_demos.py` (which loads these reward models during transfer learning) because it already applies a VecNorm on top of the reward wrapper.

shwang:
@AdamGleave this is ready for another review

src/imitation/algorithms/adversarial.py
```diff
@@ -281,8 +284,13 @@ def _build_disc_feed_dict(self, *,
     assert n_gen == len(gen_acts)
     assert n_gen == len(gen_next_obs)
 
+    # Normalize expert observations to match generator observations.
+    assert isinstance(self.venv_train_norm, VecNormalize)
+    expert_obs_norm = self.venv_train_norm._normalize_observation(
```
AdamGleave: `_normalize_observation` as a side-effect updates `self.obs_rms`.

Maybe we should change Stable Baselines to (a) make `_normalize_observation` part of public API (document, rename to remove trailing `_`) and (b) take a boolean flag for whether `obs_rms` should be updated?

Alternative is to write our own helper method that does this, or toggle `self.venv_train_norm.training` (but that's ugly!)

src/imitation/scripts/train_adversarial.py
```diff
@@ -257,8 +278,17 @@ def plot_disc_loss(self):
...
     self._savefig("plot_fight_loss_disc", self.show_plots)
 
-  def add_data_ep_reward(self, env, name):
+  def add_data_ep_reward(self, epoch):
     """Calculate and record average episode returns."""
+    if epoch in self.ep_reward_X:
+      # Don't calculate ep reward twice.
+      return
+    self.ep_reward_X.append(epoch)
+    self._add_data_ep_reward(self.venv_norm_obs, "Ground Truth Reward")
```
AdamGleave: We do rollouts once for each reward function? This feels inefficient: the policy is the same in all cases. It's also higher variance, since the rollouts will be different in each case.

Can we do a rollout just once in the ground truth, and then compute the returns under each of the models?

src/imitation/algorithms/adversarial.py
shwang: Wrote a helper method for now -- I'll make an issue in SB to make normalize_observation public.

src/imitation/scripts/train_adversarial.py
shwang: Good point. I've rewritten this section to use the same rollouts repeatedly.

src/imitation/algorithms/adversarial.py
shwang: https://github.com/hill-a/stable-baselines/issues/602

src/imitation/scripts/train_adversarial.py
```diff
@@ -284,22 +277,32 @@ def add_data_ep_reward(self, epoch):
...
+                                   (self.trainer.reward_test, "Test Reward")]:
+      if reward_fn is None:
+        trajs_rand, trajs_gen = trajs_rand_orig_rew, trajs_gen_orig_rew
+      else:
+        trajs_rand = [util.rollout.rewrite_rewards_traj(t, reward_fn)
+                      for t in trajs_rand_orig_rew]
+        trajs_gen = [util.rollout.rewrite_rewards_traj(t, reward_fn)
+                     for t in trajs_gen_orig_rew]
+
+      gen_ret = util.rollout.rollout_stats(trajs_gen)["return_mean"]
```
AdamGleave: I'm wondering if we can compute this more directly? If I understand correctly, the reason we need trajectories here is because that's what `rollout_stats` expects, but we only actually need the returns to compute these metrics.

How about modifying `rewrite_reward_trajs` to just return `new_rew` rather than a new `Trajectory`? If we called this `compute_rewards`, we could then compute the mean returns directly:
```
gen_ret = np.mean([sum(compute_rewards(t)) for t in trajs_gen_orig_rew])
```

If we wanted things other than just the mean, then it might make sense to instead factor out the summary statistics generating functionality from `rollout_stats` from constructing the various `traj_descriptors`.

src/imitation/util/rollout.py
```diff
@@ -51,6 +52,13 @@ def unwrap_traj(traj: Trajectory) -> Trajectory:
   return res
 
 
+def rewrite_rewards_traj(traj: Trajectory, reward_fn: RewardFn) -> Trajectory:
```
AdamGleave: Per my comment above not sure if this is the right atomic operation, or if we should just return the `new_rews`.

src/imitation/util/util.py
```diff
@@ -71,6 +71,14 @@ def make_env(i):
     return DummyVecEnv(env_fns)
 
 
+@contextlib.contextmanager
+def vec_norm_disable_training(vec_norm: VecNormalize):
```
AdamGleave: Can you add a TODO to remove this once hill-a/stable-baselines#602 is resolved?

src/imitation/scripts/train_adversarial.py
shwang: Good point, there's no need to regenerate the entire Trajectory and compute every stat. I've modified `rewrite_reward_trajs` to return new rewards only and calculate mean rewards only.

PR #133, opened 2019-11-30
Miscellaneous GAIL improvements and refactoring

qxcv:
This PR includes a set of changes that make GAIL more flexible & easier to use. The DAgger PR (#128) and the BC PR (#125) should be merged before this one, since the branch I'm pulling from is based on both of those. New changes here:

- Allows a custom discriminator model constructor to be passed to GAIL.
- Allows control over where `make_summary_writer` writes its logs & makes `AdversarialTrainer` responsible for passing in that directory.
- Moves the unique output directory generation code that was in `make_summary_writer` into `init_trainer` so that the resulting output directory can be used for logs, and not just TB summaries.
- Removes the four `build_` methods from `DiscrimNet` (`build_{train_reward,test_reward,summaries,disc_loss}`, IIRC) and replaces them with one `build_graph` method. Also merges the two `build_` methods in `AdversarialTrainer`.
- Adds an SB `Logger` to `AdversarialTrainer` that records discriminator stats at each update. At the moment this only works for GAIL; the extension to AIRL is straightforward, but I don't have time to test it manually, so I haven't done it myself.

Originally I was going to refactor `DiscrimNet` entirely so that it passes around Keras models instead of construction functions and kwargs, but I don't have time to do so at the moment (I've added it to the wishlist in #31).


src/imitation/summaries.py
```diff
@@ -1,27 +1,10 @@
...
-
-  tf.logging.info("building summary directory at " + run_dir)
-  if not os.path.exists(run_dir):
-    os.makedirs(run_dir)
-
-  summary_writer = tf.summary.FileWriter(run_dir, graph=graph)
+def make_summary_writer(log_dir, graph=None):
+  tf.logging.info("building summary directory at " + log_dir)
+  os.makedirs(log_dir, exist_ok=True)
+  summary_writer = tf.summary.FileWriter(log_dir, graph=graph)
```
shwang: Could we just get rid of this file, (either move the function into `util.py` or better yet, just copy the three lines into AdversarialTrainer?) I don't see `make_summary_writer` adding much value especially now that the `_run{}/` logic is externalized or outdated.

src/imitation/rewards/discrim_net.py
```diff
@@ -308,33 +354,67 @@ def act_ph(self):
...
+    correct_vec = tf.equal(bin_is_generated_pred, bin_is_generated_true)
+    acc = tf.reduce_mean(tf.cast(correct_vec, tf.float32))
+    expert_acc = tf.reduce_sum(tf.cast(
+      tf.logical_and(bin_is_expert_true, correct_vec), tf.float32)) \
+        / tf.cast(tf.maximum(1, n_expert), tf.float32)
+    generated_acc = tf.reduce_sum(tf.cast(
+      tf.logical_and(bin_is_generated_true, correct_vec), tf.float32)) \
+        / tf.cast(tf.maximum(1, n_generated), tf.float32)
+    label_dist = tf.distributions.Bernoulli(logits=self._discrim_logits)
+    entropy = tf.reduce_mean(label_dist.entropy())
```
shwang: Thanks for adding all of these summary stats, they look like useful items for hp tuning / debugging

src/imitation/rewards/discrim_net.py
```diff
@@ -273,26 +273,72 @@ def _load(cls, directory):
...
 
 
+DiscrimNetBuilder = Callable[[tf.Tensor, tf.Tensor],
+                             Tuple[util.LayersDict, tf.Tensor]]
+"""Type alias for function that takes an observation and action tensor and
+produces a tuple containing (1) a list of used TF layers, and (2) output logits
+for GAIL."""
+
+
+class build_mlp_discrim_net:
```
shwang: What sort of changes do we need to add in order to allow `Sacred` to modify the hidden sizes? Prior to this PR, it's possible via some very nested dictionary definitions like `init_trainer_kwargs....hidden_sizes = (32, 32)`.

Oh I suppose that we could later add a `hidden_sizes` top-level config to `scripts.config.train_adversarial`, and then have `scripts.config.train_adversarial` automatically convert that to a `build_mlp_discrim_net`.

src/imitation/algorithms/adversarial.py
```diff
@@ -146,13 +156,44 @@ def train_disc(self, n_steps=10, **kwargs):
...
+      for k, v in fetched['train_stats'].items():
+        self._disc_logger.log(k, v)
+        stat_dict_accum.setdefault(k, []).append(v)
+      self._disc_logger.dumpkvs()
+
+    # return only averaged stats (useful for, e.g., updating progress bars)
+    mean_stats = {
+      k: np.mean(v) for k, v in stat_dict_accum.items()
+    }
+    return mean_stats
```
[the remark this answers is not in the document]
qxcv: I do, but I won't need it once #135 is removed. It's fine to take out.

src/imitation/algorithms/adversarial.py
```diff
@@ -352,10 +395,28 @@ def init_trainer(env_name: str,
     init_rl_kwargs: Keyword arguments passed to `init_rl`,
         used to initialize the RL algorithm.
   """
+  # come up with an informative, unique name for this expt
+  log_prefix = log_dir if log_dir is not None else 'output'
+  date_str = datetime.datetime.now().strftime('%Y-%m-%dT%H:%M')
+  method_name = 'gail' if use_gail else 'airl'
+  log_base = os.path.join(
+    log_prefix, f'{method_name}_{env_name}_{date_str}')
```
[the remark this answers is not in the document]
qxcv: Sorry, didn't realise. I wasn't using `scripts.train_adversarial()` because I wanted to control the training loop, hence the date string here. It would be good to move all the unique experiment identifier generation stuff to one place, though (ideally up the stack from this code, or at least in a place where it can be used by all algorithms).

src/imitation/algorithms/adversarial.py
```diff
@@ -352,10 +395,28 @@ def init_trainer(env_name: str,
...
+  while True:
+    # find a short, unique suffix
+    full_log_dir = log_base + f"_run{i:02d}/"
+    try:
+      os.makedirs(full_log_dir, exist_ok=False)
+      break
+    except FileExistsError:
+      pass
+    i += 1
+  log_dir = full_log_dir
```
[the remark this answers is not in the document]
qxcv: Yep, that's fair. Maybe `log_dir` (or `full_log_dir`, or whatever) should be a mandatory argument instead of a kwarg? If we can't have a sensible automatic default that works for all use cases then we should not have a default at all.

src/imitation/rewards/discrim_net.py
qxcv: Yep, I think Sacred-friendly stuff should be pushed to `train_adversarial()`; anyone using these internal classes will probably want more direct control.

src/imitation/summaries.py
qxcv: Sure, will do when I next push to this branch.

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
qxcv: (I just added a comment saying you can remove it; I'll let you do the honours when you merge #135)

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
qxcv: Fixed by forcing caller to pass in `log_dir` in full.

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
qxcv: Fixed by forcing caller to pass in `log_dir` in full.

src/imitation/rewards/discrim_net.py
qxcv: Oh, I take it back; I can just let the user pass extra kwargs for `build_discrim_net` straight in. This doesn't fully solve the problem of specifying networks from the command line (e.g. what if you want to replace CNN with MLP?), but I'm going to declare that as out of scope for this PR.

shwang: >  can just let the user pass extra kwargs for build_discrim_net straight in.

If it's fine with you, I can make this change before we merge this PR.

qxcv: I already did change it, see https://github.com/HumanCompatibleAI/imitation/pull/133/files/#diff-c976e486405c3d9c2bb8d4bbd2aac4f1R315

shwang: Oh yeah, just saw that thanks

src/imitation/rewards/discrim_net.py
```diff
@@ -273,26 +273,76 @@ def _load(cls, directory):
...
+        observation_space: observation space for this environment.
+        action_space: action space for this environment.
+        build_discrim_net: a callable that takes an observation input tensor
+            and action input tensor as input, then computes the logits
+            necessary to feed to GAIL. When called, the function should return
+            *both* a `LayersDict` containing all the layers used in
+            construction of the discriminator network, and a `tf.Tensor`
+            representing the desired discriminator logits.
+        build_discrim_net_kwargs: optional extra keyword arguments for
+            `build_discrim_net()`.
```
shwang: Now that we have the `build_discrim_net_kwargs` pattern, could we make `build_mlp_discrim_net` into a function rather than a class? 

Not sure if `pytype` would be happy with Callable + arbitrary kwargs. (It looks like newer versions of Python will have a more precise `Protocol` type for defining callable types: https://stackoverflow.com/a/57840786/1091722)

qxcv:
Done! Sorry for the delay, I forgot that I still had a change to make.

PR #134, opened 2019-12-05
train_adversarial: Standardize training with total_timesteps

shwang:
This PR now uses `total_timesteps` like in Stable Baselines for `AdversarialTrainer.train_{gen,disc}()` and make changes to Sacred config files so that PPO2's `n_steps` argument is automatically calculated from `n_batch` and `num_vec_env`.

The training procedure in `scripts/train_adversarial` is changed so that each disc and gen "epoch" uses one `n_batch` of samples. (Previously, we used `n_disc_steps_per_epoch` gradient descent steps of batch size `n_disc_samples_per_buffer` in the disc step and `n_gen_steps_per_epoch` total_timesteps in the gen step.) This mirrors the Baselines and Stable Baselines TRPO/GAIL implementation.

- [x] Remove hacky "callback-to-generator"

Merge #135 first and change base branch to master.


src/imitation/algorithms/adversarial.py
```diff
@@ -157,33 +156,79 @@ def _log_context(self, name):
...
+      step, _ = self._sess.run([self._global_step, self._disc_train_op],
+                               feed_dict=fd)
+      if self._init_tensorboard and step % 20 == 0:
+        self._summarize(fd, step)
+
+  def train_gen(self, total_timesteps: int, callback=None):
+    """Trains the generator (via PPO2) to maximize the discriminator loss.
+
+    Args:
+      total_timesteps: The number of transitions to sample from self.venv_train.
```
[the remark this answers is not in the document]
shwang: > Note these are from the same environment (I think when we spoke last you indicated they were separate environments, but I may have been misremembering).

I made them separate environments on my hp tuning branch. Will make a PR for this fix later.

> I wonder if it'd be better to just record the timesteps taken during self.gen_policy.learn though? This would be more efficient

I suppose the easiest way to retain the rollouts is to use a callback that reads `locals['obs'], locals['actions']` and loads them into a buffer. This would assume PPO2 is the underlying RL algorithm though (note that the `batch_size` => `n_steps` logic in the config file also assumes PPO2).

Do you think that environment steps are a large chunk of run time? (I haven't thought about this before).

src/imitation/algorithms/adversarial.py
```diff
@@ -281,17 +308,15 @@ def _build_disc_feed_dict(self, *,
       tf.logging.debug("_build_disc_feed_dict: No generator rollout "
                        "parameters were "
                        "provided, so we are generating them now.")
-      gen_sample = self._gen_replay_buffer.sample(
```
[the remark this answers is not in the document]
shwang: Wanted to match SB/Baselines behavior. Agree now that it would nice to be able to set this hyperparameter

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -59,6 +62,18 @@ def train_defaults():
   rollout_hint = None  # Used to generate default rollout_path
 
 
+@train_ex.config
+def timesteps(n_epochs, batch_size):
```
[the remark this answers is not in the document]
shwang: I think it will be useful in hp tuning to be able to change the batch size while keeping the number of updates (via n_epochs) the same. Making a change right now to allow specifying either `total_timesteps` or `n_epochs`.

src/imitation/algorithms/adversarial.py
```diff
@@ -157,33 +156,79 @@ def _log_context(self, name):
...
+
+    Args:
+        total_timesteps (int): The number of transitions to sample from
+          `self.venv_train`. (More precisely, this is an upper bound. The
+          actual number of transition samples used is
+          `total_timesteps // self.n_batch * self.n_batch`).
+    """
+    n_epochs = total_timesteps // self.batch_size
+    for i in tqdm(range(n_epochs), desc="Adversarial train"):
+      self.train_disc(self.batch_size)
```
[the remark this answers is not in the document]
shwang: OK, added split `batch_size` config param into `{gen,disc}_batch_size`.

src/imitation/algorithms/adversarial.py
```diff
@@ -157,33 +156,79 @@ def _log_context(self, name):
...
+    each of the expert and generator replay buffers.
+
     Args:
-        n_steps (int): The number of training steps.
-        gen_obs (np.ndarray): See `_build_disc_feed_dict`.
-        gen_acts (np.ndarray): See `_build_disc_feed_dict`.
-        gen_next_obs (np.ndarray): See `_build_disc_feed_dict`.
+      total_timesteps: The number of transitions to sample from self.venv_train.
     """
+    n_epochs = total_timesteps // self.batch_size
```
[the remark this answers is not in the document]
shwang: Added disc_batch_size param

src/imitation/algorithms/adversarial.py
```diff
@@ -157,33 +156,79 @@ def _log_context(self, name):
...
-        fd = self._build_disc_feed_dict(**kwargs)
+      for _ in tqdm(range(n_epochs), desc="Adversarial train discriminator"):
+        fd = self._build_disc_feed_dict()
         step, _ = self._sess.run([self._global_step, self._disc_train_op],
                                  feed_dict=fd)
         if self._init_tensorboard and step % 20 == 0:
           self._summarize(fd, step)
 
-  def train_gen(self, n_steps=10000):
+  def train_disc_on_samples(self, **kwargs):
```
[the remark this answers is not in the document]
shwang: Renamed to `train_disc_step`, and `train_disc` calls this as a helper now. `train_disc_step` is used by `test_trainer` to repeatedly train the discriminator on a particular set of samples.

I'm thinking about refactoring this `train_disc_step` and the "build disc feed dict" helper method to take in a `rollout.Transition` instead. Would simplify the `none_count == 3` logic a bit.

src/imitation/scripts/config/train_adversarial.py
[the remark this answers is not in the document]
AdamGleave: Sorry for slow response. This seems odd to me: algorithms will almost always do better with more `total_timesteps`. Normally you choose some fixed `total_timesteps` budget, and search for hyperparameters that do best given that budget.

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
AdamGleave: I'd guess collecting rollouts is the majority of the time taken by the algorithm, but haven't profiled it. Will be environment dependent obviously, but note that environment rollouts also involve policy inference, which is naturally slow and hard to parallelize.

src/imitation/scripts/config/train_adversarial.py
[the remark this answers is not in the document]
shwang: OK, I'll get rid of the `n_epochs` option

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
shwang: As you might have seen already, I collected rollouts via a wrapper in #144

src/imitation/algorithms/adversarial.py
```diff
@@ -85,7 +90,10 @@ def __init__(self,
     self._sess = tf.get_default_session()
     self._global_step = tf.train.create_global_step()
 
-    self._n_disc_samples_per_buffer = n_disc_samples_per_buffer
+    assert disc_batch_size % 2 == 0
```
AdamGleave:
```suggestion
    assert disc_batch_size % 2 == 0, "discriminator batch size must be even (equal split between generator and expert samples)"
```

src/imitation/algorithms/adversarial.py
```diff
@@ -147,65 +160,104 @@ def gen_policy(self) -> BaseRLModel:
     """Policy (i.e. the generator) being trained."""
     return self._gen_policy
 
-  def train_disc(self, n_steps=10, **kwargs):
+  def train_disc(self, total_timesteps: Optional[int] = None) -> None:
```
AdamGleave: I think `batch_size` might be a clearer name for this than `total_timesteps`. When I hear "total" I think complete number of timesteps, across all training.

AdamGleave: `n_samples` is another option, since we're not really taking new timesteps in the environment.

src/imitation/algorithms/adversarial.py
```diff
@@ -147,65 +160,104 @@ def gen_policy(self) -> BaseRLModel:
...
-        gen_next_obs (np.ndarray): See `_build_disc_feed_dict`.
+      total_timesteps: The number of transitions to sample from the generator
+        replay buffer and the expert demonstration dataset. (Half of the
+        samples are from each source).
+        By default, `self.disc_batch_size`. The number of optimizer updates
+        performed is `total_timesteps // self.n_disc_minibatch`.
+    """
+    if total_timesteps is None:
+      total_timesteps = self.disc_batch_size
+    n_epochs = total_timesteps // self.disc_batch_size
```
AdamGleave: should check `total_timesteps` is divisible by `self.disc_batch_size`?

src/imitation/algorithms/adversarial.py
```diff
@@ -147,65 +160,104 @@ def gen_policy(self) -> BaseRLModel:
...
+    assert n_epochs >= 1, "should have at least one update"
+    for _ in range(n_epochs):
```
AdamGleave: Is there a difference between minibatches in different epochs? Seems like `self.disc_batch_size` doesn't really matter excepts as a default for `total_timesteps`.

src/imitation/algorithms/adversarial.py
```diff
@@ -147,65 +160,104 @@ def gen_policy(self) -> BaseRLModel:
...
+    """Trains the generator (via PPO2) to maximize the discriminator loss.
+
+    After the end of training populates the generator replay buffer (used in
+    discriminator training) with `self.disc_batch_size` transitions.
+
+    Args:
+      total_timesteps: The number of transitions to sample from
+        `self.venv_train_norm` during training. By default,
+        `self.gen_batch_size`.
+      callback: Callback argument to the Stable Baselines `RLModel.learn()`
```
AdamGleave:
```suggestion
      callback: Callback argument to the Stable Baselines `BaseRLModel.learn()`
```

src/imitation/algorithms/adversarial.py
```diff
@@ -147,65 +160,104 @@ def gen_policy(self) -> BaseRLModel:
...
       # It should work, but might incur unnecessary overhead: e.g. in PPO2
       # a new Runner instance is created each time. Also a hotspot for errors:
       # algorithms not tested for this use case, may reset state accidentally.
-      self._gen_policy.learn(n_steps, reset_num_timesteps=False)
+      self.gen_policy.learn(total_timesteps=total_timesteps,
+                            reset_num_timesteps=False,
+                            callback=callback)
       self._populate_gen_replay_buffer()
 
+  def train(self, total_timesteps: int) -> None:
```
AdamGleave: I like this function but is it used anywhere?

src/imitation/algorithms/adversarial.py
```diff
@@ -147,65 +160,104 @@ def gen_policy(self) -> BaseRLModel:
...
+    for _ in tqdm(range(n_epochs), desc="Adversarial train"):
+      self.train_disc()
+      self.train_gen()
+
   def _populate_gen_replay_buffer(self) -> None:
     """Generate and store generator samples in the buffer.
 
     More specifically, rolls out generator-policy trajectories in the
-    environment until `self._n_disc_samples_per_buffer` obs-act-obs samples are
+    environment until `self.disc_batch_size` obs-act-obs samples are
```
AdamGleave:
```suggestion
    environment until `self.disc_batch_size // 2` obs-act-obs samples are
```

src/imitation/algorithms/adversarial.py
```diff
@@ -147,65 +160,104 @@ def gen_policy(self) -> BaseRLModel:
...
-            Adam optimizer steps to perform.
-        n_gen_steps_per_epoch (int): The number of generator training steps
-            during each epoch. (ie, the timesteps argument in in
-            `policy.learn(timesteps)`).
-    """
-    for i in tqdm(range(n_epochs), desc="AIRL train"):
-      self.train_disc(**_n_steps_if_not_none(n_disc_steps_per_epoch))
-      self.train_gen(**_n_steps_if_not_none(n_gen_steps_per_epoch))
-      logger.dumpkvs()
+    gen_samples = util.rollout.generate_transitions(
```
AdamGleave: Is this OK in terms of not resetting the environment?

src/imitation/algorithms/adversarial.py
```diff
@@ -243,73 +294,47 @@ def _build_disc_train(self):
...
-      gen_obs = gen_sample.obs
-      gen_acts = gen_sample.acts
-      gen_next_obs = gen_sample.next_obs
-    elif none_count != 0:
-      raise ValueError("Gave some but not all of the generator params.")
-
-    # Sample expert training batch from replay buffer.
-    expert_sample = self._exp_replay_buffer.sample(
-        self._n_disc_samples_per_buffer)
+      gen_samples: Same as in `train_disc_step`.
```
AdamGleave: :+1: good change

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -15,11 +15,16 @@
 @train_ex.config
 def train_defaults():
   env_name = "CartPole-v1"  # environment to train on
-  n_epochs = 50
+
+  # Set `total_timesteps` to None when using `n_epochs`.
```
AdamGleave: Per my previous comment I'm still confused why we need `n_epochs`

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -40,24 +45,62 @@ def train_defaults():
...
+                                      gen_replay_buffer_size):
+  init_trainer_kwargs = dict(
+    trainer_kwargs=dict(
+      n_disc_minibatch=n_disc_minibatch,
+      gen_replay_buffer_capacity=gen_replay_buffer_size,
+      disc_batch_size=disc_batch_size,
+    ))
+
+
+@train_ex.config
```
AdamGleave: Don't get why we need this.

src/imitation/util/rollout.py
```diff
@@ -137,6 +137,7 @@ def min_episodes(n: int) -> GenTrajTerminationFn:
   """
   def f(trajectories: Sequence[Trajectory]):
     return len(trajectories) >= n
+  assert n >= 1
```
AdamGleave: weakly prefer this as first line in the method (validate input before you use it)

src/imitation/util/rollout.py
```diff
@@ -153,6 +154,7 @@ def min_timesteps(n: int) -> GenTrajTerminationFn:
   def f(trajectories: Sequence[Trajectory]):
     timesteps = sum(len(t.obs) - 1 for t in trajectories)
     return timesteps >= n
+  assert n >= 1
```
AdamGleave: ibid

src/imitation/scripts/train_adversarial.py
```diff
@@ -159,12 +151,13 @@ def train(_run,
       visualizer = None
 
     # Main training loop.
+    n_epochs = total_timesteps // trainer.gen_batch_size
+
     for epoch in tqdm.tqdm(range(1, n_epochs+1), desc="epoch"):
```
AdamGleave: Why not use `trainer.train`?

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
AdamGleave: :+1:

src/imitation/algorithms/adversarial.py
shwang: fair point, this is conceptually different from what `total_timesteps` usually means. `n_samples` sounds good to me.

src/imitation/algorithms/adversarial.py
shwang: I think this would be too strict (Stable Baselines just does `total_timesteps // batch_size` without warning for what it's worth).

At the very least I'll add some stuff to the documentation noting when we are rounding down total_timesteps.

src/imitation/algorithms/adversarial.py
shwang: It's only used in two places: Here in `train_disc` (and therfore indirectly in `learn()`) and in `_build_feed_dict` for the default sample size.

src/imitation/algorithms/adversarial.py
shwang: No, this isn't referenced outside of test_trainer.py

src/imitation/algorithms/adversarial.py
shwang: No, but this environment reset is resolved by #144

src/imitation/scripts/train_adversarial.py
shwang: `visualizer` needs two callbacks, one after `train_gen` and another after `train_disc`. I suppose I could merge these two blocks of code by adding these two callbacks into `train()`?

Also would add the `save(trainer, ...` to this hypothetical callback.

shwang: Another way to merge this code with `train()` would be to move the `Visualizer` into `Adversarial` and use a callback for the checkpointing `save(trainer, ...`.

src/imitation/algorithms/adversarial.py
shwang: In my head I'm thinking of `train_gen` and `train_disc` like I think of `BaseRLModel.learn(total_timesteps)`. You can stick any positive int inside there you want.

Might not be the right model

src/imitation/algorithms/adversarial.py
AdamGleave: Looking over this code again I'm confused what's going on.

If `total_timesteps == self.disc_batch_size`, then we have 1 epoch of `self.n_disc_minibatch` minibatches of `self.disc_batch_size // self.n_disc_minibatch` size. So far, so good.

But if e.g. `total_timesteps == 2 * self.disc_batch_size`, then we have 2 epochs of `self.n_disc_minibatch` minibatches of `2 * self.disc_batch_size // self.n_disc_minibatch` size. This seems weird: the mini batch size should never change.

I'd scrap this epoch vs minibatch distinction. There's no meaningful difference here: you don't do anything at the start of each epoch. So there are three parameters (any two of which determine the third): batch size, minibatch size and number of minibatches. The question is which ones should the user provide as configuration parameters?

My inclination is to let the user specify batch size and minibatch size. Loosely, mini-batch size you want to set based on how noisy the gradient is, whereas batch size you want to set based on how much training time the discriminator should be able to get. So you can vary them somewhat independently. But I'd be OK with other choices too.

shwang: Oops, yeah that should be `minibatch_size = self.disc_batch_size // self.n_disc_minibatch`

shwang: Setting the minibatch size directly sounds good to me. (My original motivation was to mirror PPO2's hyperparameters)

src/imitation/algorithms/adversarial.py
AdamGleave: `train_gen` and `train_disc` are really just performing a single training step though, whereas Stable Baselines `train()` is meant to train for many epochs. Rounding from 1 million to 999424 isn't a big deal, but 2000 to 1024 is.

src/imitation/algorithms/adversarial.py
AdamGleave: I think batch vs minibatch makes more sense in the case of PPO, where the *batch* is collecting a fresh dataset and the *minibatch* are gradient updates. But we're not doing any data collection here.

shwang: Made a change that sets minibatch size directly rather than using `n_disc_minibatch`.

src/imitation/scripts/train_adversarial.py
shwang: I got rid of `AdversarialTrainer.train` in favor of using this loop (alternative was to use multiple callbacks in `AdversarialTrainer.train()`)

src/imitation/algorithms/adversarial.py
shwang: ok, made this strict.

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -178,7 +162,7 @@ def hopper():
 def humanoid():
   env_name = "Humanoid-v2"
   rollout_hint = "humanoid"
-  n_epochs = 2000
+  total_timesteps = 4e6
```
AdamGleave: My guess is Humanoid will need more timesteps than this. Certainly Humanoid will need more than 2x that of HalfCheetah, but maybe we're being very generous with HalfCheetah.

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -263,7 +247,7 @@ def fast():
 # Shared settings
 
 ant_shared_locals = dict(
-    n_epochs=2000,
+    timesteps=3e7,
```
AdamGleave: This is a lot of timesteps! but maybe we need it. weird Humanoid gets less timesteps than this.

shwang:
Thanks for the reviews, Adam.

Agree that we can wait until hp tuning to check the timesteps.

shwang:
I accidentally `git merge`d this branch (`timesteps`) into `accumulating_log` #135. This resulted in this branch automatically closing. To make review easier for `accumulating_log`, I hard resetted back to before this accidental merge.

Because this `timesteps` PR is now permanently in a "already merged" state, I will merge back into `accumulating_log` after that that is approved and finally merge `accumulating_log` into master.

I meant to propagate fixes up from master like `master` -> `accumulating_log` -> `timesteps`.

PR #135, opened 2019-12-07
Accumulating loggers

shwang:
Add mean-accumulating log context for tracking discrim and generator stats together even though they are trained serially and with different numbers of kv logs.

- [x] See how this looks on our Sacred scripts
- [x] ~~Probably should forward `_AccumulatingLogger.logkv_mean` as well.~~


src/imitation/util/logger.py
```diff
@@ -0,0 +1,126 @@
...
+def _sb_logger_configure_replacement(*args, **kwargs):
+  raise RuntimeError("Shouldn't call stable_baselines.logger.configure "
+                     "once imitation.logger.configure() has been called")
+
+
+def _sb_logger_reset_replacement():
+  raise NotImplementedError("reset() while using imitation.configure()")
+
+
+_in_accumul_context = False
```
AdamGleave: Consider encapsulating state in a class rather than as globals. I think this would make things easier to read, and would also support the use case of having multiple of these imitation-specific loggers (not sure when we'd need this yet but may come up).

How about making a new class e.g. `HierarchicalLogger` that has e.g. `self.current : Logger` and `self.cached_logger` attributes and implements the same abstract interface as `Logger`? Methods like `logkv` etc you'd just forward to `self.current`. You could then make `accumulate_means` an instance method that switches `self.current`.

`configure()` could then set an instance of this class as `sb_logger.Logger.CURRENT` and you wouldn't have to update this again.

tests/test_logger.py
```diff
@@ -0,0 +1,105 @@
...
+                    "no_context": [1],
+                    }
+
+  _compare_csv_lines(osp.join(tmpdir, "progress.csv"), expect_default)
+  _compare_csv_lines(
+    osp.join(tmpdir, "raw", "gen", "progress.csv"), expect_raw_gen)
+  _compare_csv_lines(
+    osp.join(tmpdir, "raw", "disc", "progress.csv"), expect_raw_disc)
+
+  # Part Two:
```
AdamGleave: maybe add a comment for Part One too if there's a Part Two? :)

tests/test_logger.py
```diff
@@ -0,0 +1,105 @@
+from collections import defaultdict
```
AdamGleave: :+1: these are good tests

tests/test_logger.py
```diff
@@ -0,0 +1,105 @@
...
+import os.path as osp
+
+import stable_baselines.logger as sb_logger
+
+import imitation.util.logger as logger
+
+
+def _read_csv_lines(lines) -> dict:
```
AdamGleave: why are we rolling our own rather than using https://docs.python.org/3/library/csv.html ?

src/imitation/algorithms/adversarial.py
```diff
@@ -47,42 +48,49 @@ def __init__(self,
...
         debug_use_ground_truth: If True, use the ground truth reward for
-            `self.train_env`.
-            This disables the reward wrapping that would normally replace
-            the environment reward with the learned reward. This is useful for
-            sanity checking that the policy training is functional.
+          `self.train_env`.
+          This disables the reward wrapping that would normally replace
+          the environment reward with the learned reward. This is useful for
+          sanity checking that the policy training is functional.
+        use_custom_log: If True, then automatically enter
```
AdamGleave: When do we not want to do this?

src/imitation/util/logger.py
```diff
@@ -0,0 +1,126 @@
...
+               folder,
+               output_formats,
+               *,
+               subdir: str):
+    """Like Logger, except also accumulates logkv_mean on default logger."""
+    super().__init__(folder, output_formats)
+    self.subdir = subdir
+
+  def logkv(self, key, val):
+    raw_key = os.path.join("raw", self.subdir, key)
```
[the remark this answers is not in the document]
shwang: @qxcv In case this is helpful for quickly understanding which prefixes are used where, `test_logger.py::test_hard` shows the logger in action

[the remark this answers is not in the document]
qxcv: The motivation is that you need different prefixes for averaged stuff vs. not averaged stuff, right? If so the current arrangement seems fine to me: un-averaged data ends up in `raw/`, and averaged stats end up in `mean/`.

(It's really unfortunate that we can't have something like a relational database, where we just write everything to a file with appropriate foreign keys & then do all our averaging during post-processing. I still can't think of a way to get that with CSV, though, unless we're willing to limit ourselves to only using Pandas for analysis & throw away support for stuff like viskit.)

tests/test_logger.py
shwang: Switched to using python's csv library. thanks

src/imitation/algorithms/adversarial.py
shwang: The custom logger as of yet errors if `imitation.util.logger.configure(...)` isn't called ahead of time. Don't want `train()` to error by default because of logging.

Another thing we could do instead of keeping this `use_custom_log` flag is to check inside AdversarialTrainer whether the custom logger is configured, and only enter the accumulate_mean contexts when custom logger is enabled.

We could also consider making `init_trainer` automatically configure the custom logger. Actually I think this is the best solution, as `init_trainer` already takes in `log_dir` and configures `Monitor` logs via `make_vec_env`.

shwang: Thoughts on the second solution?

src/imitation/util/logger.py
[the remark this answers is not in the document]
AdamGleave: My point was that this particular logger *only* has keys `raw/{self.subdir}`, with the master logger having the `mean/...` keys. Are you imagining concatenating all the CSVs together? I'm not sure that'll work however we set it up.

src/imitation/util/logger.py
```diff
@@ -0,0 +1,126 @@
...
+      default_log = sb_logger.Logger.DEFAULT
+      folder = os.path.join(default_log.dir, "raw", subdir)
+      os.makedirs(folder, exist_ok=True)
+      output_formats = _build_output_formats(folder, _format_strs)
+      result = cls(folder, output_formats, subdir=subdir)
+      cls._cached_loggers[subdir] = result
+      return result
+
+
+def _build_output_formats(folder, format_strs):
```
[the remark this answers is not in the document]
shwang: > generally easier to read source files when code is in a topological sort of the dependency DAG
👍 Will keep this in mind

src/imitation/util/logger.py
```diff
@@ -0,0 +1,126 @@
+import contextlib
+import os
+from typing import Dict, Optional, Sequence
+
+from stable_baselines.common.misc_util import mpi_rank_or_zero
```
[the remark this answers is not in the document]
shwang: #119 (Stable Baselines GAIL/TRPO) uses MPI. Though I think it's a good idea to remove `mpi_rank_or_zero` for now.

Will think about whether to use this logger if/when I continue work on #119.

src/imitation/util/logger.py
```diff
@@ -0,0 +1,126 @@
...
+import stable_baselines.logger as sb_logger
+
+
+def _sb_logger_configure_replacement(*args, **kwargs):
+  raise RuntimeError("Shouldn't call stable_baselines.logger.configure "
+                     "once imitation.logger.configure() has been called")
+
+
+def _sb_logger_reset_replacement():
+  raise NotImplementedError("reset() while using imitation.configure()")
```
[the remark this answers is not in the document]
shwang: changed to RuntimeError with more similar error message

src/imitation/util/logger.py
shwang: Nice, that does seem like it would be cleaner.

src/imitation/algorithms/adversarial.py
AdamGleave: Second solution sounds good to me.

We should make sure `AdversarialTrainer` fails fast if the user doesn't have the right logger, e.g. add an assert statement somewhere. Don't want to only find this out during training.

src/imitation/util/logger.py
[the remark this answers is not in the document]
shwang: @AdamGleave Why would we need to concatenate data from the raw CSVs together? Isn't the main point of having the means logging so that we can sidestep comparing logs with conflicting timesteps?

src/imitation/util/logger.py
shwang: @AdamGleave Why would we need to concatenate data from the raw CSVs together? Isn't the main point of having the means logging so that we can sidestep comparing raw logs with conflicting timesteps?

src/imitation/util/logger.py
```diff
@@ -0,0 +1,126 @@
...
+    subdir_name: Chooses the logger subdirectories and temporary logger.
+  """
+  global _in_accumul_context
+  assert _configured
+  assert not _in_accumul_context
+  _in_accumul_context = True
+
+  try:
+    sb_logger.Logger.CURRENT = _AccumulatingLogger.from_subdir(subdir_name)
+    yield
```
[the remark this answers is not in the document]
shwang: Don't think that we should yield the `_AccumulatingLogger`. In my eyes this is a private class that the user shouldn't be messing with. Instead they should be interacting with `sb_logger` as usual.

I suppose that a helpful thing about yielding `_AccumulatingLogger` is that you wouldn't need to import `sb_logger` in addition to `imitation.util.logger` if you wanted to `logkv` yourself inside the `accumulate_means` context that you just made (right now we don't `sb_logger.logkv` in AdversarialTrainer because PPO2 does all of this).

We could avoid needing both imports in this case by simply importing all the `logkv`, ..., `debug`, `info`, `warning` utility methods from `sb_logger` straight into `imitation.util.logger`.

Was this the `yield` use case you were thinking about?

src/imitation/util/logger.py
shwang: done.

shwang: I made a `HierarchicalLogger` that I think mostly does what you described. I struggled for a little bit figuring out where to put the formerly global `_format_strs` variable (needed to make sure that logging outputs are consistent with the default logger when we switch to "raw" loggers). Turns out that format_str fits pretty well as a local variable within HierarchicalLogger, especially if HierarchicalLogger instantiates the new AccumulatingLogger.

src/imitation/algorithms/adversarial.py
shwang: Ok, moved util.logger.configure into `init_trainer` and out of the `train_adversarial` Sacred script. Also added a fail fast as the first statement of AdversarialTrainer.__init__.

src/imitation/util/logger.py
```diff
@@ -0,0 +1,200 @@
...
+
+  def __init__(self,
+               format_strs: Sequence[str],
+               *,
+               mean_logger: sb_logger.Logger,
+               subdir: str):
+    """Like Logger, except also accumulates logkv_mean on the mean_logger.
+
+    Args:
+      format_strs: An list of output format strings. For details on available
```
AdamGleave:
```suggestion
      format_strs: A list of output format strings. For details on available
```

src/imitation/util/logger.py
```diff
@@ -0,0 +1,200 @@
...
+        output formats see `stable_baselines.logger.make_output_format`.
+      mean_logger: A background logger that is used to keep track of log
+        means.
+      subdir: Used to build the logging directory. Also used in the logging
+        prefix for every key written to both this logger and `mean_logger`.
+    """
+    self.subdir = subdir
+    self.mean_logger = mean_logger
+    folder = os.path.join(self.mean_logger.dir, "raw", subdir)
```
AdamGleave: Should mention in the docstring that we write output in subdirectories of `mean_logger`

src/imitation/util/logger.py
```diff
@@ -0,0 +1,200 @@
...
+    output_formats = _build_output_formats(folder, format_strs)
+    os.makedirs(folder, exist_ok=True)
+    super().__init__(folder, output_formats)
+
+  def logkv(self, key, val):
+    raw_key = os.path.join("raw", self.subdir, key)
```
AdamGleave: I think I've raised this point before, but don't think we ever got to an agreement, so reopening the discussion.

I still don't understand why we rewrite the keys here. This instance is *only* used for keys of `self.subdir`, so everything logged will be prefixed with `raw/{self.subdir}`. Is this really what we want? The only use case I can see for this is if you wanted to concatenate data from loggers of different subdirs, but we know in general this won't be possible.

src/imitation/util/logger.py
```diff
@@ -0,0 +1,200 @@
...
+    super().logkv(raw_key, val)
+
+    mean_key = os.path.join("mean", self.subdir, key)
```
AdamGleave: Is the `"mean"` prefix here necessary? Won't everything in this logger be means? Or can it also be logged to directly, in which case maybe consider renaming to `root_logger` or `parent_logger`?

src/imitation/util/logger.py
```diff
@@ -0,0 +1,200 @@
...
+
+    `self.current_logger` is higher priority than `default_logger` when it
+    is not None. At initialization, `self.current_logger = None`. Use
+    the `self.accumulate_means()` context to temporarily set
+    `self.current_logger` to an `_AccumulatingLogger`.
+
+    Args:
+      default_logger: The logger to forward logging requests to when
+        `self.current_logger` is None.
+      format_strs: An list of output format strings that should be used by
```
AdamGleave:
```suggestion
      format_strs: A list of output format strings that should be used by
```

src/imitation/util/logger.py
```diff
@@ -0,0 +1,200 @@
...
+    self.format_strs = format_strs
+    super().__init__(folder=self.default_logger.dir, output_formats=None)
+
+  @contextlib.contextmanager
+  def accumulate_means(self, subdir: str):
+    """Temporarily use an _AccumulatingLogger as the current logger.
+
+    Args:
+        subdir: A string key for the _AccumulatingLogger which determines
+          its `folder` and logging prefix. All `_AccumulatingLogger` are cached,
```
AdamGleave:
```suggestion
          its `folder` and logging prefix. All `_AccumulatingLogger` instances are cached,
```

src/imitation/util/logger.py
```diff
@@ -0,0 +1,200 @@
...
+    format_strs: An list of output format strings. For details on available
+      output formats see `stable_baselines.logger.make_output_format`.
+  """
+  os.makedirs(folder, exist_ok=True)
+  output_formats = [sb_logger.make_output_format(f, folder)
+                    for f in format_strs]
+  return output_formats
+
+
+class _AccumulatingLogger(sb_logger.Logger):
```
AdamGleave: Do we need this class, or can we just use `sb_logger.Logger` directly?

It seems the key difference is that `logkv` also calls `logkv_mean` in `self.mean_logger`. But could we just change `logkv` in `_HierarchicalLogger` to do this directly?

src/imitation/util/logger.py
AdamGleave: Maybe the use case you had in mind was the `stdout` output format, where it'll be hard to know which logger things are coming from without prefixing the keys?

src/imitation/algorithms/adversarial.py
```diff
@@ -362,6 +369,8 @@ def init_trainer(env_name: str,
     init_rl_kwargs: Keyword arguments passed to `init_rl`,
         used to initialize the RL algorithm.
   """
+  util.logger.configure(folder=osp.join(log_dir, 'generator'),
```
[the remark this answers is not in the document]
shwang: That's right, thanks for catching this

src/imitation/scripts/train_adversarial.py
```diff
@@ -169,6 +165,9 @@ def train(_run,
         visualizer.add_data_disc_loss(False, epoch)
 
       trainer.train_gen(n_gen_steps_per_epoch)
```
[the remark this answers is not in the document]
shwang: [Answered here](https://github.com/HumanCompatibleAI/imitation/pull/134#discussion_r362671284). TLDR because there isn't a callback inside `train()` and in fact we would need two different callbacks if we wanted to add `visualizer` data the same way that we do above.

src/imitation/util/logger.py
shwang: I think it would be helpful in Tensorboard to be able to have the raw results in a separate plot from the mean results since they have different timescales. (This is the only reason, actually)

shwang: Haven't done hardcore hp tuning yet, especially with these changes so don't know if this actually important

shwang: You need to have distinct keys (not just distinct loggers) to group the raw and mean statistics into different plots in Tensorboard

src/imitation/util/logger.py
shwang: yes I think they can be merged (I'm fine either way)

src/imitation/util/logger.py
shwang: The `_AccumulatingLogger` only uses this logger instance to log means, so the attribute name `self.mean_logger` within `_AccumulatingLogger` seems appropriate to me

shwang: > Is the "mean" prefix here necessary? 

Discussion in [answer to previous comment](https://github.com/HumanCompatibleAI/imitation/pull/135#discussion_r362684318).

src/imitation/util/logger.py
shwang: done

src/imitation/util/logger.py
shwang: (Will revisit this after we make a decision on whether to keep the "raw/sub_dir" and "mean/..." logging key prefixes.)

shwang: Merged the classes

src/imitation/scripts/train_adversarial.py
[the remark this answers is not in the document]
shwang: train was removed in #134

src/imitation/util/logger.py
```diff
@@ -0,0 +1,189 @@
...
+        will safely append to logs written in this folder rather than
+        overwrite.
+    """
+    if self.current_logger is not None:
+      raise RuntimeError("Nested `accumulate_means` context")
+
+    if subdir in self._cached_loggers:
+      logger = self._cached_loggers[subdir]
+    else:
+      folder = os.path.join(self.mean_logger.dir, "raw", subdir)
```
AdamGleave: `self.mean_logger` does not exist? (Consider using refactor feature in PyCharm or the like which should automatically rename these?)

src/imitation/util/logger.py
```diff
@@ -0,0 +1,189 @@
...
+      self.current_logger = None
+
+  def logkv(self, key, val):
+    if self.current_logger is not None:
+      raw_key = os.path.join("raw", self.subdir, key)
+      super().logkv(raw_key, val)
+
+      mean_key = os.path.join("mean", self.subdir, key)
+      self.mean_logger.logkv_mean(mean_key, val)
+      return self.current_logger
```
AdamGleave: why are we returning the logger here?

src/imitation/util/logger.py
```diff
@@ -0,0 +1,189 @@
...
+    else:
+      return self.default_logger
```
AdamGleave: shouldn't we also actually log something here?

src/imitation/util/logger.py
```diff
@@ -0,0 +1,189 @@
...
+      self._cached_loggers[subdir] = logger
+
+    try:
+      self.current_logger = logger
+      yield
+    finally:
+      self.current_logger = None
+
+  def logkv(self, key, val):
+    if self.current_logger is not None:
```
AdamGleave: can't we reuse `self._logger` here to avoid some of the checks?

shwang: No, because we need to write to two different loggers (it isn't clear from this garbled commit sorry)

src/imitation/util/logger.py
```diff
@@ -0,0 +1,197 @@
...
+      self._subdir = subdir
+      yield
+    finally:
+      self.current_logger = None
+      self._subdir = None
+
+  def logkv(self, key, val):
+    if self.current_logger is not None:
+      assert self._subdir is not None
+      raw_key = os.path.join("raw", self._subdir, key)
```
qxcv: Maybe this is bike-shedding a little, but is it really necessary to have the `raw/{subdir}/` prefix on all of the entries in the `self.current_logger` log? As I understand it we'll end up with a bunch of log files in which _all log entries_ start with `raw/subdir/`, which would be redundant given that you can infer what kind of entries they are from the directory hierarchy (i.e. the raw log file is already in a `raw/subdir` folder).

(overall I still like the design!)

src/imitation/util/logger.py
```diff
@@ -0,0 +1,197 @@
...
+      self.default_logger.logkv_mean(key, val)
+
+  @property
+  def _logger(self):
+    if self.current_logger is not None:
+      return self.current_logger
+    else:
+      return self.default_logger
+
+  def logkv_mean(self, key, val):
```
qxcv: Will this nest correctly? e.g. say I was in a discriminator context (`"disc"`) and I did `.logkv_mean("foo", <value>)` `J` times, did a `.dumpkvs()`, then did `.logkv_mean()` `J` more times, and repeated that process `K` times. If I looked at the raw logs, I'd expect them to contain `K` entries of the form `raw/disc/foo: <mean of J values>`, while in the logs for the parent, I'd expect one entry of the form `mean/disc/foo: <mean of J*K values>`. As I understand it, the current implementation will write straight through to the "raw" child logger without putting anything in the parent, and without writing any key prefixes. Perhaps it's safer to remove this implementation & raise `NotImplementedError()` instead.

src/imitation/util/logger.py
```diff
@@ -0,0 +1,197 @@
...
+    self._logger.logkv_mean(key, val)
+
+  def dumpkvs(self):
+    self._logger.dumpkvs()
+
+  def log(self, *args, **kwargs):
+    self._logger.log(*args, **kwargs)
```
qxcv: Same comment as above; this won't do the expected name prefixing. Perhaps just prohibit calls to this with a ` NotImplementedError`?

src/imitation/util/logger.py
```diff
@@ -0,0 +1,197 @@
...
+
+  After this function is called, `stable_baselines.logger.{configure,reset}()`
+  are replaced with stubs that raise RuntimeError.
+
+  Args:
+      folder: Argument from `stable_baselines.logger.configure`.
+      format_strs: An list of output format strings. For details on available
+        output formats see `stable_baselines.logger.make_output_format`.
+  """
+  sb_logger.configure = _sb_logger_configure_replacement
```
qxcv: Not clear what problem this is preventing. Add comment maybe?

src/imitation/util/logger.py
qxcv: Oh, on reflection the correct thing to do is probably raise `NotImplementedError` when `self.current_logger` is not `None`, else pass through to `self.default_logger`.

src/imitation/util/logger.py
shwang: Adam thought this was weird too, we had a comment thread about this [here](https://github.com/HumanCompatibleAI/imitation/pull/135#discussion_r362351049).

Reasons for redundant-looking prefix:
(1) So that in Tensorboard plots, mean plots are automatically separated from raw plots (which have a different x axis). (TB plots are grouped by key name). 
(2) To distinguish between raw and mean keys in `stdout` logging (otherwise they look identical).

(1) is my primary motivation. Adam was convinced by (2). Though now that I think about it, it might be better to just turn off raw stdout logs and show mean stdout logs only.

src/imitation/util/logger.py
shwang: This is just a sanity check, don't want a gotcha from someone calling `stable_baselines.logger.configure` after `imitation.util.logger.configure`

src/imitation/util/logger.py
shwang: I haven't thought too carefully about the inside-context `logkv_mean` case.

I'm a bit hesitant about raising `NotImplementedError` while inside the generator context. What if `gen_policy.learn()` (ie some arbitrary Stable Baselines algorithm) wants to call `logkv_mean` for some reason?.

shwang: > As I understand it, the current implementation will write straight through to the "raw" child logger without putting anything in the parent, and without writing any key prefixes.

This understanding seems correct, thanks for catching this

shwang: @qxcv 
How does this solution sound? Because we don't want to error out if some third-party package calls `log` or `logkv_mean` inside the context, and `HierachicalLogger` only claims to modify `logkv` behavior, we can just
(1) redirect `log`, `logkv_mean` to the default logger, unchanged and without prefixes and
(2) note this prominently in the documentation?

src/imitation/util/logger.py
shwang: Response in `https://github.com/HumanCompatibleAI/imitation/pull/135#discussion_r364526909`

shwang:
Thanks for the review. I've pushed my proposed fixes here

PR #142, opened 2019-12-19
Remove SciPy version pin now wheels fixed

AdamGleave:
https://github.com/scipy/scipy/issues/11237 has been closed


AdamGleave:
Tests seem to be failing because of https://github.com/hill-a/stable-baselines/pull/609 being merged. @shwang is this fixed in another PR, or should we make a new one?

shwang:
@AdamGleave I'll add a new PR, thanks for pointing this out

shwang:
Codecov is not liking the sharded pytest. I looked through the containers pytest logs -- an interesting thing I noticed is that two tests that are showing as not covered, `test_buffer_sample_errors` and `test_policy_om_reasonable_mdp` only show up in container 2's logs. (Does not show up in container 0 and 1's "run tests" logs).

Another interesting thing is that the `def test_policy_om_reasonable_mdp` line shows 2 hits even though there are 3 containers. This seems to suggest that container 2's code cov data is getting skipped.

shwang:
image

PR #144, opened 2019-12-31
adversarial: Reuse PPO2 samples in generator replay buffer

shwang:
Use a custom `BufferingWrapper` to record the generator policy replay buffer samples in `AdversarialTrainer` (now uses the samples that the underlying RL algorithm (e.g. PPO2) uses.)

We used to directly rollout samples from the policy in between iterations of RL training, but this would result in unnecessary `VecEnv` resets and bias training/logging (`rollout.generate_trajectories` begins with a reset).

Another solution I tried was making an clone of `AdversarialTrainer.venv_train_norm` just for generating transition samples. This solution was less trivial than I thought (because of syncing `VecNormalizeWrapper`s and Monitor logging), and using `BufferingWrapper` should save compute.


src/imitation/algorithms/adversarial.py
```diff
@@ -133,12 +134,17 @@ def __init__(self,
...
     if gen_replay_buffer_capacity is None:
       gen_replay_buffer_capacity = 20 * self.gen_batch_size
     self._gen_replay_buffer = buffer.ReplayBuffer(gen_replay_buffer_capacity,
                                                   self.venv)
-    self._populate_gen_replay_buffer()
+    gen_samples = rollout.generate_transitions(
+      self.gen_policy,
+      self.venv_train_norm,
+      n_timesteps=self.gen_batch_size // 2)
+    self._gen_replay_buffer.store(gen_samples)
```
AdamGleave: Is this just to make sure the replay buffer isn't empty and so can be sampled immediately? I wonder if we can avoid this by doing `train_gen` before `train_disc`?

src/imitation/util/buffering_wrapper.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Args:
+      venv: The wrapped VecEnv.
+      error_premature_reset: Error if `reset()` is called on this wrapper
+        and there are saved samples that haven't yet been accessed.
+    """
+    super().__init__(venv)
+    self.error_on_premature_reset = error_on_premature_reset
+    self._did_reset = False
+    self.obs_list = []
```
AdamGleave: Weakly prefer these being `None` (unless it causes pytype to start complaining) since must call `reset()` before `step()`

src/imitation/util/buffering_wrapper.py
```diff
@@ -0,0 +1,75 @@
...
+        "BufferingWrapper reset() before samples were accessed")
+    obs = self.venv.reset(**kwargs)
+    self.obs_list = [obs]
+    self.acts_list = []
+    self.rews_list = []
+    self.dones_list = []
+    self._did_reset = True
+    return obs
+
+  def step(self, actions):
```
AdamGleave: This will never get called if consumer uses `step_async` and `step_wait` directly.

Think should introduce `step_async` that just appends to `self.acts_list`. Then the rest of this method can be `step_wait`.

src/imitation/util/buffering_wrapper.py
```diff
@@ -0,0 +1,75 @@
+import numpy as np
+from stable_baselines.common.vec_env import VecEnv, VecEnvWrapper
+
+from imitation.util import rollout
+
+
+class BufferingWrapper(VecEnvWrapper):
```
AdamGleave: It feels like this is doing something very similar to `imitation.util.rollout`. Can we combine them in some way to avoid code duplication?

One approach would be to use `BufferingWrapper` inside `imitation.util.rollout`, to store the trajectories.

Currently there's an implementation difference that `BufferingWrapper` stores transitions and `rollout` works on the basis of trajectories, but we could change `BuffferingWrapper` to be trajectories (just port over code from `rollout.generate_trajectories`).

tests/test_buffering_wrapper.py
```diff
@@ -0,0 +1,100 @@
...
+
+    obs = venv.reset()
+    np.testing.assert_array_equal(obs, [0, 0])
+    expect_obs_list.append(obs)
+
+    for t in range(1, 10):
+      acts = np.random.random(size=(2,))
+      expect_acts_list.append(acts)
+
+      obs, *_ = venv.step(acts)
```
AdamGleave: test `step_async`, `step_wait` version (if that works, just `step()` will work too)

tests/test_buffering_wrapper.py
```diff
@@ -0,0 +1,100 @@
...
+      np.testing.assert_array_equal(obs, [t, t])
+      expect_obs_list.append(obs)
+
+    samples = venv.pop_transitions()
+    np.testing.assert_array_equal(samples.obs,
+                                  np.concatenate(expect_obs_list[:-1]))
+    np.testing.assert_array_equal(samples.acts,
+                                  np.concatenate(expect_acts_list))
+    np.testing.assert_array_equal(samples.next_obs,
```
AdamGleave: check rewards are correct too?

src/imitation/algorithms/adversarial.py
shwang: Yes, that would work. But not prepopulating the replay buffer has another problem where we have to either
(1) error if the user calls`train_disc` is called before `train_gen` or 
(2) prepopulate the replay buffer if `train_disc` is called before `train_gen`

In my head (2) seems like the preferable solution to me in which case we might as well simplify the code by prepopulating upon `__init__`.

src/imitation/util/buffering_wrapper.py
shwang: As you noted it could be tricky to use the current `rollout.Transition`-based BufferingWrapper within `generate_trajectories`. I'm wondering right now if it's important that BufferingWrapper is Transition rather than Trajectory-based.

If a typical trajectory has many more timesteps than `buffer.capacity // n_vec` then at the very least this would lead to a `ValueError("Not enough capacity to store data.")`.

It could also be troublesome I think if there's epochs in the beginning of training where the buffer isn't updated.

shwang: Looking through `_TrajectoryAccumulator` a second time, I think it actually would be a straightforward refactor to allow `BufferingWrapper` to either `pop_trajectories` or `pop_transitions`.

We would use `TrajectoryAccumulator` inside `BufferingWrapper`, and simply do something similar to `finish_trajectory` on every unfinished intermediate trajectory data structure if the user calls `pop_transitions`.

tests/test_buffering_wrapper.py
shwang: I'll do something about terminal observations as well

src/imitation/util/buffering_wrapper.py
AdamGleave: The transitions vs trajectory thing is a pain. It seems the main issue is that we want to return trajectory segments that are not complete? Calling `finish_trajectory` prematurely seems fine.

src/imitation/algorithms/adversarial.py
shwang: We decided to error if `train_disc` is called before `train_gen` is called.

shwang:
Keeping base branch at `timesteps` for now because the number of commits that are shown are this page is ungodly (due to the squash merge of `timesteps` into `master`).

shwang:
Following @AdamGleave's suggestion to reduce code duplication between these two classes, I put a `TrajectoryAccumulator` inside `BufferingWrapper`. In fact there was a huge bug in my previous implementation involving getting incorrect `next_obs` that was most easily solved by using `TrajectoryAccumulator` with a few modifications.

I tried putting `BufferingWrapper` inside `generate_trajectory`. It was really messy and not an improvement. However, I was able to factored out the inner (`step()`) loop from `generate_trajectory` and `BufferingWrapper` into a new method `TrajectoryAccumulator.add_steps_and_auto_finish(...) -> List[Trajectory]`.

tests/test_buffering_wrapper.py
```diff
@@ -0,0 +1,215 @@
...
+    extra_pop_timesteps: By default, we only call `pop_*()` after `n_steps`
+      calls to `step()`. For every unique positive `x` in `extra_pop_timesteps`,
+      we also call `pop_*()` after the `x`th call to `step()`. All popped
+      samples are concatenated before validating results at the end of this
+      test case. All `x` in `extra_pop_timesteps` must be in range(1, n_steps).
+      (`x == 0` is not valid because there are no transitions to pop at timestep
+      0).
+  """
+  if not n_steps >= 1:
+    raise ValueError(n_steps)
```
shwang: Missing coverage on this line. I should change into `assert` or use `# nocov?something`.

src/imitation/util/rollout.py
```diff
@@ -117,32 +130,68 @@ def finish_trajectory(self, key: Hashable = None) -> Trajectory:
...
+      key: np.stack(arr_list, axis=0)
+      for key, arr_list in out_dict_unstacked.items()
     }
     traj = Trajectory(**out_dict_stacked)
+    assert traj.rews.shape[0] == traj.acts.shape[0] == traj.obs.shape[0] - 1
     return traj
 
-  def add_step(self, step_dict: Dict[str, np.ndarray], key: Hashable = None):
-    """Add a single step to the partial trajectory identified by `key`.
+  def add_steps_and_auto_finish(self,
```
qxcv: This method is coupled to `VecEnv` from stable baselines, but the rest of the class is not. Perhaps it would be better to have this as a standalone function?

src/imitation/util/buffering_wrapper.py
```diff
@@ -0,0 +1,90 @@
...
+    self._trajectories.extend(finished_trajs)
+    self.n_transitions += self.num_envs
+    return obs, rews, dones, infos
+
+  def _finish_partial_trajectories(self) -> List[rollout.Trajectory]:
+    """Finishes and returns partial trajectories in `self._traj_accum`."""
+    trajs = []
+    for i in range(self.num_envs):
+      # Check that we have any transitions at all.
+      n_transitions = len(self._traj_accum.partial_trajectories[i]) - 1
```
qxcv: Not clear what is going on here. Comment?

shwang: Ok, added a comment describing why `n_transitions` is calculated this way.

src/imitation/util/rollout.py
shwang: Don't agree with this point, as I think that `VecEnv` is a common use case throughout this repository and this file, and this method (or otherwise copying the code inside the method) is necessary for using VecEnvs with `TrajectoryAccumulator`.

PR #147, opened 2020-01-11
discrim_net: Fix accidental AIRL no-op softmax

### LINKED ISSUE #138: Strange log_softmax in AIRL policy train reward
Firstly, thanks for this nice implementations here!
I wonder if you can help me to understand log_softmax part in **DiscrimNetAIRL.build_policy_train_reward**.
```
R(s,a) = logD(s,a) - log(1-D(s,a)) = f(s,a) - log π(a|s)
```
When it comes to the comments above, I got it.

But the implemented code here:
```python
self._log_D = tf.nn.log_softmax(self.reward_net.reward_output_train)
self._log_D_compl = tf.nn.log_softmax(self.log_policy_act_prob_ph)
return self._log_D - self.entropy_weight * self._log_D_compl
```
Why don't you just return these:
```python
return self.reward_net.reward_output_train - self.entropy_weight * self.log_policy_act_prob_ph
```

For now, log_softmax looks to add some extra term like `-(logsumexp(f(s,a)) - logsumexp(log π(a|s)))`. Is this some reducing variance trick or else?
Please shed a light on me.

shwang:
Resolves #138

### DIFF
--- a/src/imitation/rewards/discrim_net.py
+++ b/src/imitation/rewards/discrim_net.py
@@ -237,20 +237,28 @@ class DiscrimNetAIRL(DiscrimNet):
         labels=self.labels_ph,
         logits=self._presoftmax_disc_logits,
         name="discrim_loss",
-    )
-
-    self._policy_test_reward = self.reward_net.reward_output_test
+    )  # (None,)
 
     # Construct generator reward:
     # \[\hat{r}(s,a) = \log(D_{\theta}(s,a)) - \log(1 - D_{\theta}(s,a)).\]
     # This simplifies to:
     # \[\hat{r}(s,a) = f_{\theta}(s,a) - \log \pi(a \mid s).\]
     # This is just an entropy-regularized objective.
-    self._log_D = tf.nn.log_softmax(self.reward_net.reward_output_train)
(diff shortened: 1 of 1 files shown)


AdamGleave:
Also would be good to have a quick before-and-after benchmark. Nothing fancy: can be in just one or two environments. Would be nice to know how much this bug actually mattered...

qxcv:
I can't tell at a glance whether this is correct. Why not just use `self.reward_net.reward_output_train - self.entropy_weight * self.log_policy_act_prob_ph`?

shwang:
> I'm also confused as to why there are `log_softmax` calls in here at all. Isn't this a binary classifier?

I didn't realize this also decomposed to train_reward + ent_bonus back when I coded this up. `log_softmax` seemed like a quick way to get a logsumexp without numerical instability.

shwang:
Using the direct approach now, I'm running a before-and-after comparison.

AdamGleave:
should we merge this?

shwang:
> should we merge this?

Sounds good, I'll post the before-and-after plots in here later today for reference.

PR #148, opened 2020-01-11
scripts: Ray compatibility fixes

shwang:
Quick fixes:
1. Fix bug where total_timesteps was float
2. Provide working directory to train_adversarial so that rollout path can be automatically located

### DIFF
--- a/src/imitation/scripts/config/train_adversarial.py
+++ b/src/imitation/scripts/config/train_adversarial.py
@@ -15,7 +15,7 @@ train_ex = sacred.Experiment("train_adversarial", interactive=True)
 @train_ex.config
 def train_defaults():
   env_name = "CartPole-v1"  # environment to train on
-  total_timesteps = 1e5  # Num of environment transitions to sample
+  total_timesteps = int(1e5)  # Num of environment transitions to sample
 
   n_expert_demos = None  # Num demos used. None uses every demo possible
   n_episodes_eval = 50  # Num of episodes for final mean ground truth return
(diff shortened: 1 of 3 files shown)


src/imitation/scripts/config/train_adversarial.py
```diff
@@ -47,6 +47,7 @@ def train_defaults():
   checkpoint_interval = 5  # num epochs between checkpoints (<=0 disables)
   init_tensorboard = False  # If True, then write Tensorboard logs.
   rollout_hint = None  # Used to generate default rollout_path
+  working_dir_hint = ""  # Automatically set to working dir by `parallel.py`
```
AdamGleave: I'm a bit confused by the notion of a working directory here. I can imagine two possible designs, both reasonable:
  1. Rollouts can be in arbitrary locations, but you assume a particular directory structure within that. In this case, I'd favour having a config attribute e.g. `rollout_root`, which is currently assumed to be `{working_dir_hint}/data/expert_models`.
  2. All outputs are in a single directory. In this case, let's specify this root, which is currently assumed to be `{working_dir_hint}/data`. We should also then change it so that `log_root` is some subdirectory of this.

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -15,7 +15,7 @@
 @train_ex.config
 def train_defaults():
   env_name = "CartPole-v1"  # environment to train on
-  total_timesteps = 1e5  # Num of environment transitions to sample
+  total_timesteps = int(1e5)  # Num of environment transitions to sample
```
AdamGleave: Why do we need to do the conversion here? Might be simpler to cast to `int` inside `train_adversarial` so only have to do it once.

src/imitation/scripts/config/train_adversarial.py
shwang: I was imagining design (1). 

Replaced `working_dir_hint` a new config attribute `data_dir = "data/"`. `parallel.py` now manipulates this variable instead, and this is used to build the default rollout path.

src/imitation/scripts/config/train_adversarial.py
shwang: fair, I've added an int cast as the first statement of train_adversarial. (Could also change the type of `total_timesteps` to `Union[float, int]` or something, but that felt excessive.)

PR #152, opened 2020-01-15
Rename .codecov.yml->codecov.yml

AdamGleave:
This is recommended in [latest docs](https://docs.codecov.io/docs/codecov-yaml#section-can-i-name-the-file-codecov-yml) of CodeCov, although they previously suggested either was OK. Seems not to be a problem for `imitation`, but I found in some other projects the `.codecov.yml` config wasn't taking effect, so best to fix it defensively.


AdamGleave:
Amusingly this seems to have broken the codecov config...

AdamGleave:
I think this actually broke something. No YAML is showing up in the build tab for this commit https://codecov.io/gh/HumanCompatibleAI/imitation/commit/7826ad3be48cddcdf68e4ee896c6930b3dbbad91/build but it does in commits in other branches e.g. https://codecov.io/gh/HumanCompatibleAI/imitation/commit/1b8731837d6311a57a959933fbf16c30ba0ee17c/build

I have no idea *why* since this works in other repos when it did not work with `.codecov.yml`...

I'm going to try deleting and then re-adding the file.

AdamGleave:
This PR seems to just be breaking things even though it worked better in other repos, so am closing. We should revisit if we have problems with `codecov` in the future.

PR #153, opened 2020-01-15
setup.py: Add mujoco extras

shwang:


AdamGleave:
Looks like this is failing because of https://github.com/pypa/pip/issues/4957

It's been open for a year and a half so not expecting a speedy resolution to it.

Can't think of any particularly good workaround. You could dynamically determine whether the dependency should be `gym[classic_control]` or `gym[classic_control,mujoco]` and then add the relevant one to `install_requires`. But that's pretty fragile.

shwang:
OK, seems messier than it's worth to me (and I have less vested interest now that my workstation is configured).

I'll close this for the time being and add a note to the README

PR #154, opened 2020-01-15
train_adversarial: Change checkpoint default, fix logic

shwang:
Now by default only save the final checkpoint. (per 5 epochs was too often, and I've only used the final epoch ever).

Also modifies meaning of <0 to allow disabling any checkpointing.


shwang:
> Not sure why the experiment test is failing, rerunning in case it's flaky.

Maybe something weird is going on with the run-time generated checkpoint that is used for transfer_learn_benchmark.sh. Looking into this...

PR #155, opened 2020-01-17
train_adversarial: Remove custom plotting

shwang:
I decided to remove custom plotting because:
(1) Now that we plot disc loss and other things in Tensorboard, the custom plotter is almost entirely redundant.
(2) Also there seems to be a bug in master (maybe related to our complicated reward/observation normalization copying scheme) that is under-reporting episode rewards in our custom plots. EG: TensorBoard shows Cartpole reaching 500, but our custom plots have reward less than 200.

Also moves the training loop to AdversarialTrainer.train() since we only need a simple callback after this simplification.

### DIFF
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -229,12 +229,6 @@ class AdversarialTrainer:
   def eval_disc_loss(self, **kwargs) -> float:
     """Evaluates the discriminator loss.
 
-    The generator rollout parameters of the form "gen_*" are optional,
-    but if one is given, then all such parameters must be filled (otherwise
-    this method will error). If none of the generator rollout parameters are
-    given, then a rollout with the same length as the expert rollout
-    is generated on the fly.
-
     Args:
       gen_samples (Optional[rollout.Transitions]): Same as in `train_disc_step`.
       expert_samples (Optional[rollout.Transitions]): Same as in
@@ -272,7 +266,7 @@ class AdversarialTrainer:
 
   def train(self,
             total_timesteps: int,
-            callback: Optional[Callable[[bool], None]] = None,
+            callback: Optional[Callable[[int], None]] = None,
             ) -> None:
     """Alternates between training the generator and discriminator.
 
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -1,12 +1,13 @@
 from functools import partial
 import os
-from typing import Optional, Sequence
+from typing import Callable, Optional, Sequence
 from warnings import warn
 
 import numpy as np
 from stable_baselines.common.base_class import BaseRLModel
 from stable_baselines.common.vec_env import VecEnv, VecNormalize
 import tensorflow as tf
+import tqdm
 
 import imitation.rewards.discrim_net as discrim_net
 from imitation.rewards.reward_net import BasicShapedRewardNet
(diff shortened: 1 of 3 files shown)


shwang:
Possible hazard: These custom plots helped us figure out that our Stable Baselines plots were faulty, letting them go, especially before we get them in a working state that corroborates the stable baselines plots can be sort of testy.

However, we can tell from the form of the Tensorboard plots that the early-end-of-episode bug from before is fixed. For example, the 500 reward of the gray Cartpole run would be impossible under the previous regime because the batch size doesn't accommodate such a long episode:

image

AdamGleave:
> Possible hazard: These custom plots helped us figure out that our Stable Baselines plots were faulty, letting them go, especially before we get them in a working state that corroborates the stable baselines plots can be sort of testy.

I would like to understand why they disagree. No need to fix the bug if it's hard, but not knowing what's wrong is concerning.

src/imitation/algorithms/adversarial.py
```diff
@@ -225,6 +226,26 @@ def train_disc_step(self, *,
         logger.logkv(k, v)
       logger.dumpkvs()
 
+  def eval_disc_loss(self, **kwargs) -> float:
+    """Evaluates the discriminator loss.
+
+    The generator rollout parameters of the form "gen_*" are optional,
```
AdamGleave: there's only one keyword argument beginning with `gen_*` documented below? confused

src/imitation/algorithms/adversarial.py
```diff
@@ -243,35 +264,41 @@ def train_gen(self, total_timesteps: Optional[int] = None, callback=None):
 
     with logger.accumulate_means("gen"):
       self.gen_policy.set_env(self.venv_train_norm_buffering)
-      # TODO(adam): learn was not intended to be called for each training batch
```
AdamGleave: if we're removing this warning comment, feel like we should also point to the fixed branch of stable baselines in dependency (or just wait until it gets merged to master)

src/imitation/algorithms/adversarial.py
```diff
@@ -243,35 +264,41 @@ def train_gen(self, total_timesteps: Optional[int] = None, callback=None):
...
-    Args:
-      gen_samples (Optional[rollout.Transitions]): Same as in `train_disc_step`.
-      expert_samples (Optional[rollout.Transitions]): Same as in
-        `train_disc_step`.
-
-    Returns:
-      The total cross-entropy error in the discriminator's classification.
+  def train(self,
+            total_timesteps: int,
+            callback: Optional[Callable[[bool], None]] = None,
```
AdamGleave: why is this `bool`? should it not be `int`?

src/imitation/algorithms/adversarial.py
```diff
@@ -243,35 +264,41 @@ def train_gen(self, total_timesteps: Optional[int] = None, callback=None):
...
-    is generated on the fly.
-
-    Args:
-      gen_samples (Optional[rollout.Transitions]): Same as in `train_disc_step`.
-      expert_samples (Optional[rollout.Transitions]): Same as in
-        `train_disc_step`.
-
-    Returns:
-      The total cross-entropy error in the discriminator's classification.
+  def train(self,
```
AdamGleave: :+1: nice we have more of the training logic in this class now

shwang:
> No need to fix the bug if it's hard, but not knowing what's wrong is concerning.

That's fair. I'll take another look into this...

src/imitation/algorithms/adversarial.py
shwang: I was just thinking about making setup.py point to the Runner patch (had an error earlier this week where I installed SB without my patch).

src/imitation/algorithms/adversarial.py
shwang: Hmm I thought I had deleted this part of the documentation a while ago. Maybe I made a bad merge sometime back.

src/imitation/algorithms/adversarial.py
shwang: Looks like it's fine for us to go ahead and merge the Runner patch after your review+approval. Let's merge this PR after that change.

shwang:
Actually, plots are fine!

I did find a custom plotting bug involving stale observation normalization though. Didn't seem to make a big difference on reporting cartpole before and after fixing it.

image

image

AdamGleave:
> Actually, plots are fine!
> 
> I did find a custom plotting bug involving stale observation normalization though. Didn't seem to make a big difference on reporting cartpole before and after fixing it.
> 
> image
> 
> image

I assume the custom plots are for the green-blue and dark-orange seeds in the TB plot? If so agree looks like they match closely (might want to reduce smoothing in TB to allow more direct comparison).

AdamGleave:
One of the `codecov` uploads failed on CI, I'm rerunning that job, think checks should be green after that.

shwang:
> I assume the custom plots are for the green-blue and dark-orange seeds in the TB plot?

Yes, IIRC it's green-blue and dark-orange.

PR #156, opened 2020-01-17
discrim: Share logging logic between GAIL & AIRL

shwang:
Completes @qxcv's TODO note from #133 to log training stats for AIRL. Does this by moving logging logic from `DiscrimNetGAIL`  to `DiscrimNet`, and requiring that a dependency `self._disc_logits` be assigned by the end of `DiscrimNet.build_graph()`.

I noticed some inconsistencies between discriminator-related variable names (most notably in build graph, `disc_loss` vs `disc_logits`) so I renamed all variables `discrim => disc`. With one exception: anything that contains `discrim_net` I left unchanged.

(Only looking for a review from one of you two, looping Sam in as a reviewer because he made this TODO)

### DIFF
--- a/src/imitation/rewards/discrim_net.py
+++ b/src/imitation/rewards/discrim_net.py
@@ -2,7 +2,7 @@ from abc import ABC, abstractmethod
 import collections
 import os
 import pickle
-from typing import Callable, Iterable, Optional, Tuple
+from typing import Callable, Dict, Iterable, Optional, Tuple
 
 import gym
 import numpy as np
@@ -33,32 +33,20 @@ class DiscrimNet(serialize.Serializable, ABC):
     # TODO(adam): this is only used by AIRL; sad we have to always include it
     self._log_policy_act_prob_ph = tf.placeholder(
         shape=(None,), dtype=tf.float32, name="log_ro_act_prob_ph")
+
+    self._disc_loss = None  # type: tf.Tensor
+    self._policy_train_reward = None  # type: tf.Tensor
+    self._policy_test_reward = None  # type: tf.Tensor
+    self._disc_logits = None  # type: tf.Tensor
+
     self.build_graph()
 
-    # check that required attributes have been set
-    required_attrs = [
-      '_disc_loss', '_policy_train_reward', '_policy_test_reward',
-    ]
-    for req_attr in required_attrs:
-      assert hasattr(self, req_attr), \
-        f"required attr .{req_attr} not added by {type(self)}" \
-        ".build_graph()"
+    assert self._disc_loss is not ...
+    assert self._policy_train_reward is not ...
+    assert self._policy_test_reward is not ...
+    assert self._disc_logits is not ...
 
-  @property
-  def disc_loss(self):
-    return self._disc_loss  # pytype: disable=attribute-error
-
-  @property
-  def policy_train_reward(self):
-    return self._policy_train_reward  # pytype: disable=attribute-error
-
-  @property
-  def policy_test_reward(self):
-    return self._policy_test_reward  # pytype: disable=attribute-error
-
-  @property
-  def train_stats(self):
-    return self._train_stats
+    self._add_default_train_stats()
 
   @abstractmethod
   def build_graph(self):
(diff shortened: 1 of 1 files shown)


src/imitation/rewards/discrim_net.py
```diff
@@ -228,17 +289,20 @@ def build_graph(self):
...
         labels=self.labels_ph,
-        logits=self._presoftmax_disc_logits,
-        name="discrim_loss",
+        logits=_presoftmax_disc_logits,
+        name="disc_loss",
     )  # (None,)
 
+    softmax_disc_logits = tf.nn.log_softmax(_presoftmax_disc_logits)
+    self._disc_logits = softmax_disc_logits[:, 0] - softmax_disc_logits[:, 1]
+
```
shwang: Quick sanity check: It's necessary to do a `log_softmax` before assigning `self._disc_logits` because we need the normalizing constant, correct?

AdamGleave:
@qxcv: deferring to you on this one, but lmk if you don't have time and I'll take a look.

qxcv:
Okay, I'll review first thing tomorrow morning.

src/imitation/rewards/discrim_net.py
qxcv: `log_softmax` shouldn't change the result, but it is unnecessary because in the next line the normaliser cancels out. Specifically we have:

```
norm_const = sum(exp(_presoftmax_disc_logits), axis=1)

softmax(_presoftmax_disc_logits) = [exp(_presoftmax_disc_logits[0]) / norm_const, exp(_presoftmax_disc_logits[1]) / norm_const]

softmax_disc_logits = log(softmax(_presoftmax_disc_logits))
                    = stack([_presoftmax_disc_logits[:, 0] - log(norm_const), _presoftmax_disc_logits[:, 1] - log(norm_const)], axis=1)

_disc_logits = softmax_disc_logits[:, 0] - softmax_disc_logits[:, 1]
             = _presoftmax_disc_logits[:, 0] - log(norm_const) - (_presoftmax_disc_logits[:, 1] - log(norm_const))
             = presoftmax_disc_logits[:, 0] - _presoftmax_disc_logits[:, 1]
```

src/imitation/rewards/discrim_net.py
```diff
@@ -75,6 +63,79 @@ def build_graph(self):
...
+    Also creates a `tf.summary.histogram` for the `disc_logits`
+    argument.
+
+    Params:
+      disc_logits: A rank-1 Tensor containing binary logit for expert
+        (positive value) versus generator (negative values).
+    """
+    tf.summary.histogram("disc_logits", self._disc_logits)
+    bin_is_generated_pred = self._disc_logits > 0
```
qxcv: Is this the right way around for AIRL? With the current AIRL code, `_disc_logits = self.reward_net.reward_output_train, self.log_policy_act_prob_ph`, so a high discriminator value suggests that the sample is from an expert, rather than the imitator. In GAIL I believe it's reversed, so high discriminator output is an indicator of fakeness (in which case this code does make sense—see eqn. (16) & exposition after it in https://arxiv.org/pdf/1606.03476.pdf to figure out which way sign should go for GAIL). It is worth running both AIRL & GAIL using these stats to verify that they move in the right direction in both cases.

src/imitation/rewards/discrim_net.py
```diff
@@ -33,32 +33,20 @@ def __init__(self):
...
 
-  @property
-  def policy_train_reward(self):
-    return self._policy_train_reward  # pytype: disable=attribute-error
+    self.build_graph()
 
-  @property
-  def policy_test_reward(self):
-    return self._policy_test_reward  # pytype: disable=attribute-error
+    assert self._disc_loss is not ...
```
qxcv: Why ellipsis and not `None`? Is this a `typing` convention?

src/imitation/rewards/discrim_net.py
```diff
@@ -75,6 +63,79 @@ def build_graph(self):
...
+      tf.logical_and(bin_is_generated_true, correct_vec), tf.float32))
+    _n_gen_or_1 = tf.cast(tf.maximum(1, n_generated), tf.float32)
+    generated_acc = _n_pred_gen / _n_gen_or_1
+
+    label_dist = tf.distributions.Bernoulli(logits=self._disc_logits)
+    entropy = tf.reduce_mean(label_dist.entropy())
+
+    self._train_stats.update([
+      # basic xent loss
+      ('disc_xent', tf.reduce_mean(self._disc_loss)),
```
qxcv: Is using `_disc_loss` always  correct? What if there's a regulariser or something in the loss? Might be safer just to use `label_dist.cross_entropy(self._disc_logits)` in this part (but also make sure the full discriminator loss is logged somewhere, maybe under a name like `disc_loss` instead).

src/imitation/rewards/discrim_net.py
shwang: Thanks for writing this out, makes sense

src/imitation/rewards/discrim_net.py
shwang: This is an error from when I was messing around with trying to get PyCharm inspection to be happy with my types

shwang: Hmmm looks like I had already fixed a day before you made this comment.
You seem to have missed https://github.com/HumanCompatibleAI/imitation/pull/156/commits/a6e6f55e0fa10b207ddf45b145e12fdf518b1127 where I also made some minor log name changes that you might want to see.

src/imitation/rewards/discrim_net.py
shwang: Ok, I am explicitly constructing the xent now.

src/imitation/rewards/discrim_net.py
qxcv: Oops, ignore the above comment! I believe got the GAIL paper the wrong way around. Looks like the policy in GAIL is treating `\log D(s,a)` as a cost, so it tries to drive it _down_. I misread it originally as a reward function. Interpretation of `\log D(s,a)` is consistent with eqn (14) in the paper.

shwang: `label_ph` always forces expert (0.0) and generated (1.0), and we see at least half-decent training in both algorithms.

In the definition of DiscrimNet.reward_train:

https://github.com/HumanCompatibleAI/imitation/blob/0331f0edb1f3e85f3d86bfd5ad89027656cef305/src/imitation/rewards/discrim_net.py#L121

So I don't think that AIRL/GAIL are flipped in our implementation. (Continuing to look at this more thoroughly now).

qxcv: Okay, thanks for checking that it works anyway. Sorry for the misleading comment earlier.

PR #157, opened 2020-01-17
setup.py: Install SB branch with Runner patch

shwang:
We need hill-a/stable-baselines#639 because we call `learn()` repeatedly, sometimes with low timesteps.

Using a fork of `fix_low_timesteps_learn` instead of the real deal in case I break the real deal during merge or code review later.


shwang:
Seems like the SB PR will be merged soon. Putting this one on hold

PR #158, opened 2020-01-17
AdversarialTrainer.train_gen: Don't env reset due to `gen_policy.set_env()`

shwang:
After https://github.com/hill-a/stable-baselines/pull/639 every call to `self.gen_policy.set_env(venv)` leads to a (delayed) reset of the environment.

This patch moves a `set_env` call in AdversarialTrainer from the `learn` method to `__init__`, ~~and adds optional detection of accidental changes to `self.gen_policy.env`.~~

~~It's possible that the tests in the PR will not pass until the aforementioned SB PR is merged.~~

edit: We decided that we don't want these tests, as they won't actually detect most venv resets.

### DIFF
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -241,7 +241,8 @@ class AdversarialTrainer:
     fd = self._build_disc_feed_dict(**kwargs)
     return np.mean(self._sess.run(self.discrim.disc_loss, feed_dict=fd))
 
-  def train_gen(self, total_timesteps: Optional[int] = None, callback=None):
+  def train_gen(self, total_timesteps: Optional[int] = None,
+                learn_kwargs: Optional[dict] = None):
     """Trains the generator to maximize the discriminator loss.
 
     After the end of training populates the generator replay buffer (used in
@@ -251,16 +252,18 @@ class AdversarialTrainer:
       total_timesteps: The number of transitions to sample from
         `self.venv_train_norm` during training. By default,
         `self.gen_batch_size`.
-      callback: Callback argument to the Stable Baselines `RLModel.learn()`
+      learn_kwargs: kwargs for the Stable Baselines `RLModel.learn()`
         method.
     """
     if total_timesteps is None:
       total_timesteps = self.gen_batch_size
+    if learn_kwargs is None:
+      learn_kwargs = {}
 
     with logger.accumulate_means("gen"):
       self.gen_policy.learn(total_timesteps=total_timesteps,
                             reset_num_timesteps=False,
-                            callback=callback)
+                            **learn_kwargs)
       gen_samples = self.venv_train_norm_buffering.pop_transitions()
       self._gen_replay_buffer.store(gen_samples)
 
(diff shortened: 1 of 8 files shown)


src/imitation/algorithms/adversarial.py
```diff
@@ -87,6 +88,19 @@ def __init__(self,
           This disables the reward wrapping that would normally replace
           the environment reward with the learned reward. This is useful for
           sanity checking that the policy training is functional.
+        error_on_unexpected_policy_change: If True, then enable training
```
shwang: Not sure if we actually want these errors to be optional. Leaning towards nonoptional errors because the silent resets can be so tricky to detect.

In the future if there's actually a use case for this sort of modification we can maybe add this back in.

AdamGleave: I'm OK making these errors mandatory (i.e. removing the keyword argument and just always having the check). In this case should move some of the docstring into a comment near the `assert` statement since this is useful context to have.

src/imitation/algorithms/adversarial.py
```diff
@@ -259,11 +269,17 @@ def train_gen(self, total_timesteps: Optional[int] = None, callback=None):
       callback: Callback argument to the Stable Baselines `RLModel.learn()`
         method.
     """
+    if self.error_on_unexpected_policy_change:
+      if self.gen_policy.env is not self.venv_train_norm_buffering:
+        raise ValueError("Unexpected change to self.gen_policy.env")
+      if self.venv_train_norm_buffering.n_resets > 1:
```
AdamGleave: This check doesn't seem to quite match the description in the docstring. In particular, I could still reset (maybe indirectly, by calling a `rollout` method) on `self.venv_train` or `self.venv_train_norm` and I don't think `self.venv_train_norm_buffering` would observe this.

In general not sure it's tractable to detect this. If we're really paranoid user should pass in a thunk to construct a `VecEnv`, and then we construct it internally to `AdversarialTrainer` and do not expose it publicly.

src/imitation/algorithms/adversarial.py
```diff
@@ -259,11 +269,17 @@ def train_gen(self, total_timesteps: Optional[int] = None, callback=None):
       callback: Callback argument to the Stable Baselines `RLModel.learn()`
         method.
     """
+    if self.error_on_unexpected_policy_change:
```
AdamGleave: Why does this check belong in `eval_disc_loss`?

src/imitation/algorithms/adversarial.py
shwang: >  I could still reset (maybe indirectly, by calling a rollout method) on self.venv_train or self.venv_train_norm and I don't think self.venv_train_norm_buffering would observe this.

That's a really good point

src/imitation/algorithms/adversarial.py
shwang: this is actually in `train_gen`

shwang:
We decided that the tests aren't worth keeping, and we should probably construct a private venv inside AdversarialTrainer sometime in the future to prevent accidental external resets.

I removed the tests from the PR and now we just have the pre-emptive bug fix for hill-a/stable-baselines#639.

PR #160, opened 2020-01-18
Bypass out-of-sync Gym registry in SubprocVecEnv by resolving EnvSpec

shwang:
We ran into an unusual problem when using `imitation.scripts.parallel` to parallelize training of our custom ant environment. The subprocesses created by `imitation.util.make_vec_env` reported that all our custom environments weren't registered, but when we ran `imitation.scripts.parallel`. During debuggins, we found that it didn't help to forcibly reregister the environments in scope of the stack trace before the inner worker function to SubprocVecEnv. (Even forcibly reregistering environments inside `make_vec_env` itself did not suffice)

The most likely explanation is that the Python forkserver used by `SubprocVecEnv` was somehow prematurely initialized by `ray` so that new processes forked from the forkserver didn't include our custom environments inside the Gym registry. (edit: We are actually quite unsure if this is actually the cause of the bug. THis is just a hypothesis)

h/t: @AdamGleave for the idea to resolve the custom environment's `EnvSpec` before initializing the `SubprocVecEnv`.

### DIFF
--- a/src/imitation/util/util.py
+++ b/src/imitation/util/util.py
@@ -46,8 +46,12 @@ def make_vec_env(env_name: str,
       seed: The environment seed.
       parallel: If True, uses SubprocVecEnv; otherwise, DummyVecEnv.
       log_dir: If specified, saves Monitor output to this directory.
-      max_episode_steps: If specified, wraps VecEnv in TimeLimit wrapper with
-          this episode length before returning.
+      max_episode_steps: If specified, wraps each env in a TimeLimit wrapper
+          with this episode length. If not specified and `max_episode_steps`
+          exists for this VecEnv in the Gym registry, uses the registry value
+          for every TimeLimit wrapper (this is the default behavior when calling
+          `gym.make`). Otherwise the environments are passed into the VecEnv
+          unwrapped.
   """
   # Resolve the spec outside of the subprocess first, so that it is available to
   # subprocesses running `make_env` via automatic pickling.
--- a/src/imitation/util/util.py
+++ b/src/imitation/util/util.py
@@ -47,10 +47,10 @@ def make_vec_env(env_name: str,
       log_dir: If specified, saves Monitor output to this directory.
       max_episode_steps: If specified, wraps each env in a TimeLimit wrapper
           with this episode length. If not specified and `max_episode_steps`
-          exists for this VecEnv in the Gym registry, uses the registry value
-          for every TimeLimit wrapper (this is the default behavior when calling
-          `gym.make`). Otherwise the environments are passed into the VecEnv
-          unwrapped.
+          exists for this `env_name` in the Gym registry, uses the registry
+          `max_episode_steps` for every TimeLimit wrapper (this automatic
+          wrapper is the default behavior when calling `gym.make`). Otherwise
+          the environments are passed into the VecEnv unwrapped.
   """
   # Resolve the spec outside of the subprocess first, so that it is available to
   # subprocesses running `make_env` via automatic pickling.
--- a/src/imitation/util/util.py
+++ b/src/imitation/util/util.py
@@ -70,6 +70,8 @@ def make_vec_env(env_name: str,
 
     if max_episode_steps is not None:
       env = TimeLimit(env, max_episode_steps)
+    elif (spec.max_episode_steps is not None) and not spec.tags.get('vnc'):
+      env = TimeLimit(env, max_episode_steps=spec.max_episode_steps)
 
     # Use Monitor to record statistics needed for Baselines algorithms logging
     # Optionally, save to disk
--- a/src/imitation/util/util.py
+++ b/src/imitation/util/util.py
@@ -50,17 +50,18 @@ def make_vec_env(env_name: str,
           this episode length before returning.
   """
   # Resolve the spec outside of the subprocess first, so that it is available to
-  # subprocesses via automatic pickling.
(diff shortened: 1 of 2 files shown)


shwang:
Confirmed that the new test fails before this PR.

AdamGleave:
> We ran into an unusual problem when using `imitation.scripts.parallel` to parallelize training of our custom ant environment. The subprocesses created by `imitation.util.make_vec_env` reported that all our custom environments weren't registered, but when we ran `imitation.scripts.parallel`. During debuggins, we found that it didn't help to forcibly reregister the environments in scope of the stack trace before the inner worker function to SubprocVecEnv. (Even forcibly reregistering environments inside `make_vec_env` itself did not suffice)
> 
> The most likely explanation is that the Python forkserver used by `SubprocVecEnv` was somehow prematurely initialized by `ray` so that new processes forked from the forkserver didn't include our custom environments inside the Gym registry.

Just to flag I'm still not sure about this particular explanation, although I agree with the overall conclusion that it's a bad interaction between `forkserver` and `ray`. In particular my (imperfect) understanding of `forkserver` is that it's a fresh process, like with `spawn` start method, but for efficiency it only has to be created once and then `fork()` inside the forkserver can cheaply create new (clean) Python processes. Some evidence in favour of this is that using `spawn` also failed, even though those processes are being created JIT.

Admittedly this doesn't explain why it works in Ray local mode. My guess is that the key difference is that when using Ray the worker is executing a pickled function rather than native code. This plausibly would cause a difference in how the helper method is pickled and sent to the `SubprocVecEnv` worker, although I've not confirmed this.

(We probably don't need to get to the bottom of this if the `EnvSpec` thing works, but I wanted to put this in writing in case we need to debug a similar issue in the future.)

src/imitation/util/util.py
```diff
@@ -49,15 +49,29 @@ def make_vec_env(env_name: str,
...
+  # Resolve the spec outside of the subprocess first, so that it is available to
+  # subprocesses via automatic pickling.
+  spec = gym.spec(env_name)
+
   def make_env(i, this_seed):
-    env = gym.make(env_name)
+    # Previously, we directly called `gym.make(env_name)`.
+    #
+    # That direct approach was problematic (especially in combination with Ray)
+    # because the forkserver from which subprocesses are forked might have been
```
AdamGleave: Are we sure this is what's going on? It's fine as a hypothesis but I don't want to immortalize in a comment something we're not sure about.

AdamGleave: I'm 80% confident the `forkserver` has a very minimal state, since:
  1. The Python `multiprocessing` docs say "No unnecessary resources are inherited".
  2. The [code](https://github.com/python/cpython/blob/master/Lib/multiprocessing/forkserver.py) seems to use `spawn` to start the forkserver, which starts a new Python process. 
  3. This [blog](http://www.bnikolic.co.uk/blog/python/parallelism/2019/11/13/python-forkserver-preload.html) states " Note that children retain a copy of the forkserver state. This state is intended to be relatively simple, but it is possible to adjust this through the multiprocess API through the set_forkserver_preload() method."

With that said, there does seem to be some logic to preload modules (by default, I think, the `__main__` module i.e. the entrypoint to the script). So IIUC, `forkserver` is intended to execute the code of the parent process (i.e. all imports), but should not execute an `if __name__ == '__main__'`-guarded block (Python [docs](https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods) explicitly state need to have this guard).

Seems plausible that starting `forkserver` in a Ray worker messes with the autodetection of what to import.

AdamGleave: If we can't get to the bottom of this, may be better to leave the comment vague and cite this PR.

shwang: Agree that we should be vague about the cause in the comments and note the PR.

src/imitation/util/util.py
```diff
@@ -49,15 +49,30 @@ def make_vec_env(env_name: str,
       max_episode_steps: If specified, wraps VecEnv in TimeLimit wrapper with
           this episode length before returning.
```
AdamGleave:
```suggestion
          this episode length before returning. Otherwise, defaults to `max_episode_steps` for `env_name` in the Gym registry.
```

src/imitation/util/util.py
```diff
@@ -49,15 +49,30 @@ def make_vec_env(env_name: str,
...
+    env = spec.make()
+
     # Seed each environment with a different, non-sequential seed for diversity
     # (even if caller is passing us sequentially-assigned base seeds). int() is
     # necessary to work around gym bug where it chokes on numpy int64s.
     env.seed(int(this_seed))
 
     if max_episode_steps is not None:
       env = TimeLimit(env, max_episode_steps)
+    elif (spec.max_episode_steps is not None) and not spec.tags.get('vnc'):
```
AdamGleave:
```suggestion
    elif spec.max_episode_steps is not None:
```

`'vnc'` check was removed in https://github.com/openai/gym/pull/1626

src/imitation/util/util.py
shwang: I expanded this comment a bit more in https://github.com/HumanCompatibleAI/imitation/pull/160/commits/365c1772e790e1956cf6d08c765ccda6a2a3d1cd. Wanted to note that the gym registry total timesteps thing is default behavior for `gym.make`.

PR #166, opened 2020-02-05
rewards.serialize: Load training rewards

shwang:
Adds two new reward type keys to the reward serialize registry: "DiscrimNet_train" and "DiscrimNet_test".

"DiscrimNet_test" is an alias for "DiscrimNet". "DiscrimNet_train" loads the train reward, and this requires a path to the generator directory. The caller can use the

### DIFF
--- a/src/imitation/rewards/serialize.py
+++ b/src/imitation/rewards/serialize.py
@@ -1,10 +1,14 @@
 """Load serialized reward functions of different types."""
 
 import contextlib
-from typing import Callable, ContextManager, Iterator
+from functools import partial
+from typing import Callable, ContextManager, Iterator, Optional
+from pathlib import Path
+import pickle
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


AdamGleave:
I'm a bit confused by the `train` vs `test` distinction here. In AIRL the `test` variant was without shaping. But in `DiscrimNet` that distinction doesn't really make sense.

If I understand correctly, the `gen_path` is only needed for AIRL discriminator network?

shwang:
Original motivation was to plot the actual training rewards for AIRL in heatmaps like this:

This doesn't seem important right now, but might revisit later if we want this or related things.
00000 (copy)

AdamGleave:
> Original motivation was to plot the actual training rewards for AIRL in heatmaps like this:

The heatmaps are a nice debugging technique, we should keep them around :)

PR #167, opened 2020-02-06
adversarial: Fix AIRL ent_bonus, buffer unnormalized gen samples

shwang:
Ent bonus for AIRL was wrong because the observation input to `gen_policy.action_probability()` needed to be normalized but was not.

Looking at before/after plots for cartpole and mountain car it's not clear that we are doing better than before in cartpole AIRL (maybe worse). Mountain car ARIL seems to be doing better.

Here are the [before plots](https://www.notion.so/chaiberkeley/Before-PRs-156-161-2c7970b81de242e8a8859169245c2782).

Cartpole training curve before/after (before is first image)
image

image

Mountain car training curve before/after:
(No longer has any flatlining -200 policies. Performance seems to improve more steadily?)

image
image
(An "after" mountain training curve with more timesteps coming soon)

EDIT: Adding more after plots for both envs.

Cart Pole
image

Mountain Car
image

### DIFF
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -282,7 +282,7 @@ class AdversarialTrainer:
                             reset_num_timesteps=False,
                             **learn_kwargs)
 
-    gen_samples = self.venv_train_norm_buffering.pop_transitions()
+    gen_samples = self.venv_train_norm.pop_transitions()
     self._gen_replay_buffer.store(gen_samples)
 
   def train(self,
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -135,9 +135,9 @@ class AdversarialTrainer:
       self.venv_test = reward_wrapper.RewardVecEnvWrapper(
           self.venv, self.reward_test)
 
-    self.venv_train_norm = VecNormalize(self.venv_train)
-    self.venv_train_norm_buffering = BufferingWrapper(self.venv_train_norm)
-    self.gen_policy.set_env(self.venv_train_norm_buffering)
+    self.venv_train_buffering = BufferingWrapper(self.venv_train)
+    self.venv_train_norm = VecNormalize(self.venv_train_buffering)
+    self.gen_policy.set_env(self.venv_train_norm)
 
     if gen_replay_buffer_capacity is None:
       gen_replay_buffer_capacity = 20 * self.gen_batch_size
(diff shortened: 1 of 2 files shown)


AdamGleave:
Also it makes sense the results only change slightly, tweaking entropy coefficient has always had minor impacts on performance in my experience, so I don't think the entropy bonus is that critical for AIRL performance.

shwang:
>  think to strictly be correct we should be storing unnormalized samples in `self._gen_replay_buffer` too (I think a question of moving the `BufferWrapper` earlier)

Agreed.

shwang:
I'll add a comment

shwang:
I started buffering unnormalized generator transitions instead of normalized generator transitions as you suggested and now AIRL cartpole looks super stable!
image


Mountain car AIRL seems to be much worse somehow?
image


(The reason we didn't buffer unnormalized generator transitions until now is that it wasn't possible before I implemented `BufferWrapper`, and by that time I didn't realize this was an option.)

shwang:
Cartpole GAIL (still) looks lovely:
image

MC GAIL still is stuck:
image

shwang:
Toying with AIRL entropy weight in MC and CP:

(Note that entropy weight has opposite effects on the performance of these two environments)

Mountain Car w/ ent_weight=0.0:

image

ent_weight=0.5:
image

ent_weight=1.0 (default):
image

CartPole w/ ent_weight=0.0:
image

ent_weight=0.5:
image

ent_weight=1.0:
image

qxcv:
That's bizarre. One difference between the envs is that in CartPole you want long episodes (so positive rewards are better), whereas in mountain car you want short episodes (so negative rewards are better). That is consistent with the results above. Don't take it too seriously, though; a priori one might also expect that MountainCar benefits more from entropy than CartPole, since MountainCar has harder exploration, and is (probably?) more resistant to control noise.

qxcv:
I had a quick skim of the code. Didn't see anything super suspicious except this (which you might already be aware of?).

```#!python
# in adversarial.py:142, it seems like buffer is big enough to cause problems
# due to running average shifts
if gen_replay_buffer_capacity is None:
    gen_replay_buffer_capacity = 20 * self.gen_batch_size
```

shwang:
> I had a quick skim of the code. Didn't see anything super suspicious except this (which you might already be aware of?).
> 
> ```
> # in adversarial.py:142, it seems like buffer is big enough to cause problems
> # due to running average shifts
> if gen_replay_buffer_capacity is None:
>     gen_replay_buffer_capacity = 20 * self.gen_batch_size
> ```

@qxcv: The Sacred script automatically sets `gen_replay_buffer_capacity = gen_batch_size`. This probably isn't a good default for people manually instantiating `AdversarialTrainer`.

qxcv:
Setting `gen_replay_buffer_capacity = gen_batch_size` is probably not a big deal. I haven't personally seen much advantage to actually using a real replay buffer in my own experiments. It's unclear whether Justin used a non-trivial replay buffer for his original AIRL experiments anyway.

Also I originally flagged the `gen_replay_buffer_capacity = 20 * self.gen_batch_size` because it might have messed with the running average reward normalisation mechanism. After talking to Adam, I no longer think that this is likely to be a big deal. It seems like Stable Baselines' normalisation code takes average/std reward _over all of history_, and not just an exponentially weighted moving average, so after a few updates the average reward is going to be updating very, very slowly anyway.

PR #168, opened 2020-02-06
Support non-PPO2 RL algorithms in `expert_demos`

AdamGleave:
It's intended to be able to support arbitrary Stable Baselines algorithms in `expert_demos`. However, there are several issues that make it hard to use off-policy algorithms that this PR resolves:
  - `init_rl_kwargs` has default values that *cannot be deleted* due to https://github.com/IDSIA/sacred/issues/238. This PR worksaround that by only adding the default values if `init_rl_kwargs` is an empty dict.
  - The log callbacks get called each `callback` and the `step` is assumed to be no more than 5 digits long. This is problematic for off-policy algorithms where `callback` is executed every timestep of training. This PR makes the log frequency configurable, and allows up to 9 digit long steps.
  - Added custom loading logic for SAC which has some non-standard policy format.

### DIFF
--- a/src/imitation/scripts/expert_demos.py
+++ b/src/imitation/scripts/expert_demos.py
@@ -81,6 +81,8 @@ def rollouts_and_policy(
           reward_type to load the reward model. For more information, see
           `imitation.rewards.serialize.load_reward`.
 
+      log_interval: The number of training updates in between custom
+          logging callbacks.
       rollout_save_interval: The number of training updates in between
           intermediate rollout saves. If the argument is nonpositive, then
           don't save intermediate updates.
--- a/src/imitation/scripts/expert_demos.py
+++ b/src/imitation/scripts/expert_demos.py
@@ -37,6 +37,7 @@ def rollouts_and_policy(
   reward_type: Optional[str],
   reward_path: Optional[str],
 
+  log_interval: int,
   rollout_save_interval: int,
   rollout_save_final: bool,
   rollout_save_n_timesteps: Optional[int],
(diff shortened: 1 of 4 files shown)


src/imitation/scripts/expert_demos.py
```diff
@@ -37,6 +37,7 @@ def rollouts_and_policy(
   reward_type: Optional[str],
   reward_path: Optional[str],
 
+  log_interval: int,
```
shwang: Missing param doc in docstring

shwang:
Looks like tests are failing because SAC needs a Box actions space. (Mountain Car and Cartpole both have discrete action space). Pendulum and Continuous Mountain Car would work here.

shwang:
<pre>tests/test_policies.py::test_serialize_identity[False-model_cfg2-CartPole-v0] <font color="#C4A000">SKIPPED</font><font color="#4E9A06">                                                                                                 [ 38%]</font> 
tests/test_policies.py::test_serialize_identity[False-model_cfg2-MountainCarContinuous-v0] <font color="#4E9A06">PASSED                                                                                     [ 44%]</font>
tests/test_policies.py::test_serialize_identity[False-model_cfg2-Pendulum-v0] <font color="#CC0000">FAILED                                                                                                  [ 50%]</font>                                     
tests/test_policies.py::test_serialize_identity[True-model_cfg2-CartPole-v0] <font color="#C4A000">SKIPPED</font><font color="#CC0000">   [ 88%]</font>
tests/test_policies.py::test_serialize_identity[True-model_cfg2-MountainCarContinuous-v0] <font color="#CC0000">FAILED [ 94%]</font>
tests/test_policies.py::test_serialize_identity[True-model_cfg2-Pendulum-v0] <font color="#CC0000">FAILED    [100%]</font> 
</pre>

shwang:
For the identical rollouts test, SAC+MountainCarContinuous is ok for unnormalized policy only. Pendulum fails in both unnormalized and normalized cases. Something is probably wrong with policy serialization right now.

I'm going to move on to fixed-length episode wrappers for now, but I can take another look at this later.

AdamGleave:
> For the identical rollouts test, SAC+MountainCarContinuous is ok for unnormalized policy only. Pendulum fails in both unnormalized and normalized cases. Something is probably wrong with policy serialization right now.
> 
> I'm going to move on to fixed-length episode wrappers for now, but I can take another look at this later.

Thanks for helping out on this! I can take a look later in the week too (whenever I move back to interpretability for reward learning, which is what triggered this PR). Let's just ping each other before we start to avoid any duplication.

AdamGleave:
I would like this feature still, but it'll need to be re-implemented in SB3 anyway, so closing for now.

PR #172, opened 2020-02-22
AdvTrainer: Minimum default gen buffer size

shwang:
Change the default generator buffer size to be `self.gen_batch_size` for consistency with Sacred script. (Used to be `20 * self.gen_batch_size`).

Also fixes a bug that disallows training generator alone for more than `len(replay_buffer)` timesteps, because this would put more samples into the buffer than the buffer has capacity.

### DIFF
--- a/src/imitation/util/buffer.py
+++ b/src/imitation/util/buffer.py
@@ -299,7 +299,6 @@ class ReplayBuffer:
     trans_dict = transitions._asdict()
     if lengths[0] > self.capacity and truncate_ok:
       trans_dict = {k: v[-self.capacity:] for k, v in trans_dict.items()}
-        trans_dict[k] = v[-self.capacity:]
 
     self._buffer.store(trans_dict)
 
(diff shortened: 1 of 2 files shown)


src/imitation/util/buffer.py
```diff
@@ -277,19 +277,31 @@ def sample(self, n_samples: int) -> rollout.Transitions:
...
         ValueError: The arguments didn't have the same length.
     """
     lengths = [len(arr) for arr in transitions]
     if len(set(lengths)) != 1:
       raise ValueError("Arguments must have the same length.")
-    self._buffer.store(transitions._asdict())
+
+    trans_dict = transitions._asdict()
+    if lengths[0] > self.capacity and truncate_ok:
+      for k, v in trans_dict.items():
```
AdamGleave: This could be just a dict comprehension?

```suggestion
      trans_dict = {k: v[-self.capacity:] for k, v in trans_dict.items()}
```

PR #174, opened 2020-03-09
Replace squeeze by flatten to fix (1,) shape case in DiscrimNet.reward_train

pedrofreire:
If your `batch_size == 1` (e.g. if your `VecEnv` has only one copy of the environment during training), then
`log_act_prob.shape == () != (1,) == (batch_size,)`, since squeeze will transform a `(1,)`
shape into a `()` shape - and thus the shape assert will fail.
AFAIK, replacing it by `.reshape(-1)` should fix the problem while maintaining same behavior in other cases.

### DIFF
--- a/src/imitation/rewards/discrim_net.py
+++ b/src/imitation/rewards/discrim_net.py
@@ -200,7 +200,8 @@ class DiscrimNet(serialize.Serializable, ABC):
         The rewards. Its shape is `(batch_size,)`.
     """
     del steps
-    log_act_prob = gen_log_prob_fn(observation=obs, actions=act, logp=True).reshape(-1)
+    log_act_prob = np.reshape(
+      gen_log_prob_fn(observation=obs, actions=act, logp=True), (-1,))
 
     n_gen = len(obs)
     assert obs.shape == next_obs.shape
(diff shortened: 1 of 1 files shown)


src/imitation/rewards/discrim_net.py
```diff
@@ -196,13 +196,12 @@ def reward_train(
...
         `log_act_prob[i]` is equal to the generator's log probability of
         choosing `act[i]` given `obs[i]`.
-        `np.squeeze(log_act_prob)` has shape `(batch_size,)`.
     Returns:
         The rewards. Its shape is `(batch_size,)`.
     """
     del steps
-    log_act_prob = np.squeeze(
-      gen_log_prob_fn(observation=obs, actions=act, logp=True))
+    log_act_prob = np.reshape(
```
AdamGleave: Does `np.flatten` do what we want here? That seems more canonical than `np.reshape(..., (-1,))`.

pedrofreire: Oh, that's right. Replaced `reshape` with `flatten`.

shwang:
idk why checks aren't showing up here. Manually ran on my machine and everything passed.

AdamGleave:
> idk why checks aren't showing up here. Manually ran on my machine and everything passed.

I think checks are configured to only run on branches of the original repo, not forks (this is for security since otherwise people can steal our sweet API secrets).

We may want to revisit this if we have external collaborators in the future, although pedro should have write access to the main repo.

pedrofreire:
> We may want to revisit this if we have external collaborators in the future, although pedro should have write access to the main repo.

It seems I don't have write access, that's why I made a fork :)

AdamGleave:
> It seems I don't have write access, that's why I made a fork :)

OK, I've given you write access now (thought CHAI members had it by default, seems not).

PR #175, opened 2020-03-18
Import benchmark envs in Sacred Scripts

shwang:
This allows us to run experiments using benchmark environments from script CLI.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,9 +3,6 @@ import src.imitation  # pytype: disable=import-error
 
 TF_VERSION = '>=1.15.0,<2.0'
 TESTS_REQUIRE = [
-    # TODO(adam): use PyPi rather than Git once we make first release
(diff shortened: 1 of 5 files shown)


AdamGleave:
Do we really want to force all consumers of this library to install `benchmark-environments`?

Maybe we should have a `'scripts'` section in `extras_require` for this, and just not install the scripts by default? Or you could do:
```
try:
    import benchmark_environments.classic_control  # noqa: F401
except ImportError:
    pass
```
and then the dependency is still optional.

shwang:
The try-except solution seems good to me.

shwang:
Benchmark environments doesn't seem to a big install (compared to our other requirements like TF), but looks like it requires the mujoco version of gym (and therefore py mujoco, and a mujoco key). We definitely don't want to require a mujoco key.

PR #176, opened 2020-03-19
Baselines GAIL traj conversion utility

shwang:
Simple script for converting our `rollout.pkl` into the `rollout.npz` format used by openai/baselines GAIL.

### DIFF
--- a/experiments/convert_traj.py
+++ b/experiments/convert_traj.py
@@ -27,24 +27,24 @@ def convert_trajs_to_sb(trajs: List[rollout.Trajectory]) -> dict:
 
 
 def main():
-    parser = argparse.ArgumentParser()
-    parser.add_argument("src_path", type=str)
-    parser.add_argument("dst_path", type=str)
-    args = parser.parse_args()
+  parser = argparse.ArgumentParser()
+  parser.add_argument("src_path", type=str)
+  parser.add_argument("dst_path", type=str)
+  args = parser.parse_args()
 
-    src_path = Path(args.src_path)
-    dst_path = Path(args.dst_path)
+  src_path = Path(args.src_path)
+  dst_path = Path(args.dst_path)
 
-    assert src_path.is_file()
-    with open(src_path, "rb") as f:
-      src_trajs = pickle.load(f)  # type: List[rollout.Trajectory]
+  assert src_path.is_file()
+  with open(src_path, "rb") as f:
+    src_trajs = pickle.load(f)  # type: List[rollout.Trajectory]
 
-    dst_trajs = convert_trajs_to_sb(src_trajs)
-    os.makedirs(dst_path.parent, exist_ok=True)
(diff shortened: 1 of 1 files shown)


experiments/convert_traj.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python
+"""
+Convert trajectories from `imitation` format to openai/baselines GAIL format.
+"""
+
+from typing import List
+
+import numpy as np
+
+from imitation.util import rollout as ro_util
```
AdamGleave: Why `from x import y as z`?

[Google style guide](https://google.github.io/styleguide/pyguide.html#224-decision) says:
> Use from x import y as z if two modules named y are to be imported or if y is an inconveniently long name.

I don't think there's a name clash nor is rollout particularly long

experiments/convert_traj.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def convert_trajs_to_sb(trajs: List[ro_util.Trajectory]) -> dict:
+  """Converts Trajectories into the dict format used by Stable Baselines GAIL.
+  """
```
AdamGleave: One-liner docstrings do not normally have a newline before closing `"""`

experiments/convert_traj.py
```diff
@@ -0,0 +1,47 @@
...
+  return dict(
+    acs=trans.acts,
+    rews=trans.rews,
+    obs=trans.obs,
+    ep_rets=np.array([np.sum(t.rews) for t in trajs]),
+  )
+
+
+if __name__ == "__main__":
+    import argparse
```
AdamGleave: why are these being imported here and not at the top-level of the module?

experiments/convert_traj.py
```diff
@@ -0,0 +1,47 @@
...
+    import os
+    from pathlib import Path
+    import pickle
+    parser = argparse.ArgumentParser()
```
AdamGleave: Put the body in a separate function, e.g.`main()`. Right now all these variables are module-level variables. https://google.github.io/styleguide/pyguide.html#317-main

experiments/convert_traj.py
shwang: Doesn't fit in 80 characters here

shwang:
Ah, sorry for all the style things. I wasn't careful with this PR!

experiments/convert_traj.py
```diff
@@ -0,0 +1,51 @@
...
+  trans = rollout.flatten_trajectories(trajs)
+  return dict(
+    acs=trans.acts,
+    rews=trans.rews,
+    obs=trans.obs,
+    ep_rets=np.array([np.sum(t.rews) for t in trajs]),
+  )
+
+
+def main():
```
AdamGleave: This is indented 4-spaces whereas the rest of the file is 2-spaces. I think we are using 2-spaces for this project?

PR #177, opened 2020-03-19
Add mountain car reward heatmap and reward-vs-time plotters

shwang:
Putting the Heatmap / reward vs time plotting code from my ipynb into a python file.

Not sure if `experiments/` is the best place to put this. Maybe `imitation.scripts` is better since the user might want to import these plotting functions into another file.

While translating the code from one format to another, I found some bugs, related yet again to VecNormalize. In this case it seems like my reward functions might not have actually received normalized inputs.

So I'd expect heatmaps to look different upon a new run.


experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
+"""Heatmaps and reward plotting code for debugging MountainCar."""
+
+from functools import partial
```
AdamGleave: "Use import statements for packages and modules only, not for individual classes or functions." https://google.github.io/styleguide/pyguide.html#22-imports

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+  Required for evaluating AIRL reward at arbitrary (s, a) points.
+  """
+  env = gym.make("MountainCar-v0")
+  obs = np.array(obs)
+  acts = np.array(acts)
+  next_obs = []
+  for ob, act in zip(obs, acts):
+    assert ob.shape == (2,)
+    env.reset()
+    env.env.state = ob
```
AdamGleave:
```suggestion
    env.unwrapped.state = ob
```

I think this is assuming the `env` is wrapped exactly once (presumably by a `TimeLimitWrapper`)? `env.unwrapped` will be more robust than `env.env`

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+    next_ob, *_ = env.step(act)
```
AdamGleave:
```suggestion
    next_ob = env.step(act)[0]
```

Don't have strong feelings on this, the current method is more visible (`[0]` is easily missed), but it does seem a bit indirect.

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+    next_obs.append(next_ob)
+  return np.array(next_obs)
+
+
+def make_policy_rollout(policy_path, env_name="MountainCar-v0",
+                        ) -> List[util.rollout.Trajectory]:
+  """Load policy from path and return a list of Trajectories."""
+  venv = DummyVecEnv([lambda: gym.make(env_name)])
+  with load_policy("ppo2", str(policy_path), venv) as gen_policy:
```
AdamGleave: Think it should be callees responsibility to pass a `str`-type `policy_path`, not this functions responsibility to convert it. (I might be missing a use case though.)

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+    title: Custom title.
+    heatmap: Whether to plot the heatmap.
+    filter_trans_by_act: If True, then filter out transitions from
+      `gen_trajs` and `exp_trajs` that don't use action `act` before
+      scatterplotting.
+  """
+  assert 0 <= act <= 2
+
+  obs_vec, acts_vec = [], []
+  for p in np.arange(MC_POS_MIN, MC_POS_MAX + 1e-5, step=pos_step_size):
```
AdamGleave: How about specifying the number of data points for position and velocity, and then using `np.linspace`? This will guarantee the endpoints get included, and avoid needing the `+ 1e-5` trick.

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+  R = reward_fn(obs_vec, acts_vec, next_obs_vec, None)
+
+  df = pd.DataFrame()
+  df["position"] = np.round(obs_vec[:, 0], 5)
+  df["velocity"] = np.round(obs_vec[:, 1], 5)
+  df["reward"] = R
+  df = df.pivot("velocity", "position", "reward")
+
+  def convert_to_rc(pos, vel):
+    x = (np.array(pos) - MC_POS_MIN) / pos_step_size + 0.5
```
AdamGleave: I don't understand why this conversion is necessary. Why can't you plot using the `pos` and `vel` coordinates directly? `matplotlib` can handle the conversion between data coordinates and pixels.

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+    trans = util.rollout.flatten_trajectories(trajs)
+    obs = trans.obs
+    if filter_trans_by_act:
+      obs = obs[trans.acts == act]
+    pos, vel = obs[:, 0], obs[:, 1]
+    return convert_to_rc(pos, vel)
+
+  plt.figure()
+  if heatmap:
+    sns.heatmap(df)
```
AdamGleave:
```suggestion
    sns.heatmap(df, ax=ax)
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+
+  def convert_to_rc_filtered(trajs):
+    trans = util.rollout.flatten_trajectories(trajs)
+    obs = trans.obs
+    if filter_trans_by_act:
+      obs = obs[trans.acts == act]
+    pos, vel = obs[:, 0], obs[:, 1]
+    return convert_to_rc(pos, vel)
+
+  plt.figure()
```
AdamGleave:
```suggestion
fig, ax = plt.subplots()
```
Less error prone to explicitly pass axis around. Most important if you're working with multiple figures.

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+  if heatmap:
+    sns.heatmap(df)
+
+  if mark_goal:
+    goal_r, _ = convert_to_rc(MC_GOAL_POS, 0)
+    plt.axvline(x=goal_r, linestyle='--',
```
AdamGleave:
```suggestion
    ax.axvline(x=goal_r, linestyle='--',
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+                label=f"goal state (pos={MC_GOAL_POS})")
+  if exp_trajs is not None:
+    X, Y = convert_to_rc_filtered(exp_trajs)
+    plt.scatter(X, Y, marker="o", label="expert samples", alpha=0.2)
```
AdamGleave:
```suggestion
    ax.scatter(X, Y, marker="o", label="expert samples", alpha=0.2)
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+  if gen_trajs is not None:
+    X, Y = convert_to_rc_filtered(gen_trajs)
+    plt.scatter(X, Y, marker="o", c="yellow", label="policy samples",
```
AdamGleave:
```suggestion
    ax.scatter(X, Y, marker="o", c="yellow", label="policy samples",
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+                alpha=0.2)
+
+  act_names = ["left", "neutral", "right"]
+  if title is None:
+    title = f"Action {act_names[act]}"
+  plt.title(title)
```
AdamGleave:
```suggestion
  ax.set_title(title)
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+  if legend_on:
+    plt.legend(loc="center left", bbox_to_anchor=(0, 1.3))
```
AdamGleave:
```suggestion
    ax.legend(loc="center left", bbox_to_anchor=(0, 1.3))
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+    pos_step_size: float = 0.1,
+    velocity_step_size: float = 1e-2,
+    mark_goal: bool = True,
+    gen_trajs: Optional[List[util.rollout.Trajectory]] = None,
+    exp_trajs: Optional[List[util.rollout.Trajectory]] = None,
+    legend_on: bool = True,
+    title: bool = None,
+    heatmap: bool = True,
+    filter_trans_by_act: bool = True,
+) -> None:
```
AdamGleave:
```suggestion
) -> plt.Figure:
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+    plt.scatter(X, Y, marker="o", c="yellow", label="policy samples",
+                alpha=0.2)
+
+  act_names = ["left", "neutral", "right"]
+  if title is None:
+    title = f"Action {act_names[act]}"
+  plt.title(title)
+  if legend_on:
+    plt.legend(loc="center left", bbox_to_anchor=(0, 1.3))
+
```
AdamGleave:
```suggestion
    return fig
```

Clearer if more functional style. (You can do things like create multiple heatmaps, put the figures in a dict, save them later...)

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+    exp_trajs: A list of exp trajectories to scatterplot on
+      top of the heatmap.
+    legend_on: Whether to plot the legend.
+    title: Custom title.
+    heatmap: Whether to plot the heatmap.
+    filter_trans_by_act: If True, then filter out transitions from
+      `gen_trajs` and `exp_trajs` that don't use action `act` before
+      scatterplotting.
+  """
+  assert 0 <= act <= 2
```
AdamGleave: Magic number.

How about introducing a constant `NUM_ACTS = 3` in the module, and then `assert 0 <= act <= NUM_ACTS - 1`?

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+
+  act_names = ["left", "neutral", "right"]
+  if title is None:
+    title = f"Action {act_names[act]}"
+  plt.title(title)
+  if legend_on:
+    plt.legend(loc="center left", bbox_to_anchor=(0, 1.3))
+
+
+def batch_reward_heatmaps(
```
AdamGleave: I think it's unnecessary that this function both generates the heatmaps and saves them.

What if I want to call this function from a Jupyter notebook and just show the heatmaps inline?

I suggest splitting up this functionality. Have one function return a dict of filenames to `plt.Figure`s, and another function that saves them.

See e.g. https://github.com/HumanCompatibleAI/evaluating-rewards/blob/master/src/evaluating_rewards/analysis/visualize.py#L80 for an example of a save function

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+          Generator trajectories
+          are dynamically generated from generator checkpoints.
+  """
+  checkpoints_dir = Path(checkpoints_dir)
+  for checkpoint_dir in sorted(checkpoints_dir.iterdir()):
+    vec_normalize_path = checkpoint_dir / "gen_policy" / "vec_normalize.pkl"
+    discrim_path = checkpoint_dir / "discrim"
+    policy_path = checkpoint_dir / "gen_policy"
+
+    with open(vec_normalize_path, "rb") as f:
```
AdamGleave: I suggest moving this after `gen_trajs`, and adding a comment to the effect that `make_policy_rollout` automatically loads normalization statistics. (For a while I was puzzled by why we weren't using `vec_normalize` anywhere in the rollouts.)

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+      vec_normalize = pickle.load(f)  # type: VecNormalize
+    vec_normalize.training = False
+
+    gen_trajs = make_policy_rollout(policy_path)
+
+    reward_fn_ctx = load_reward("DiscrimNet", discrim_path, venv="unused")
```
AdamGleave: Why `"unused"` rather than `None`?

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+  rew = reward_fn(norm_obs, acts, norm_next_obs, steps)
+  if norm_reward:
+    rew = vec_normalize.normalize_reward(rew)
+  if verbose:
+    print("rew normed min:", rew.min())
+    print("rew normed max:", rew.max())
+  return rew
+
+
+def build_norm_reward_fn(*, reward_fn, vec_normalize, **kwargs) -> RewardFn:
```
AdamGleave: Also it should be moved earlier in the code. Imagine if the code is a DAG, we want the source file to be a topological sort. (In practice there can be cycles which break this.)

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+      vec_normalize = pickle.load(f)  # type: VecNormalize
+    vec_normalize.training = False
+
+    gen_trajs = make_policy_rollout(policy_path)
+
+    reward_fn_ctx = load_reward("DiscrimNet", discrim_path, venv="unused")
+    with reward_fn_ctx as reward_fn:
+      norm_rew_fn = build_norm_reward_fn(reward_fn=reward_fn,
+                                         vec_normalize=vec_normalize)
+      for i in range(3):
```
AdamGleave: Magic number. Using `NUM_ACTS` suggested earlier would fix this.

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
+"""Heatmaps and reward plotting code for debugging MountainCar."""
```
AdamGleave: Can you call out which functions are useful as an "entry point"? It seems a lot of them are helper methods.

You could either make the helper methods start with `_`, or just document which methods people should be looking at in the module docstring.

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+        os.makedirs(save_path.parent, exist_ok=True)
+        plt.savefig(save_path)
+        print(f"saved to {save_path}")
+
+
+def plot_reward_vs_time(
+    trajs_dict: Dict[str, List[util.rollout.Trajectory]],
+    reward_fn: RewardFn,
+    colors: Sequence[str] = ("tab:blue", "tab:orange"),
+):
```
AdamGleave:
```suggestion
) -> plt.Figure:
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+  """Plots rewards received by many trajectories from many agents over time.
+
+  Args:
+      trajs_dict: Dictionary mapping rollout labels (e.g. "expert" or
+        "gen policy") to rollouts associated with those labels.
+      reward_fn: Reward function for evaluating rollout rewards.
+      colors: Custom colors for plotted rewards.
+  """
+  plt.clf()
```
AdamGleave:
```suggestion
  fig, ax = plt.subplots()
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+      Y.extend(rews)
+    if i < len(colors):
+      color = colors[i]
+    else:
+      color = None
+    plt.plot(X, Y, alpha=1.0, c=color, label=traj_name)
+  plt.xlabel("timestep")
+  plt.ylabel("test reward")
+  plt.legend()
+  plt.show()
```
AdamGleave:
```suggestion
  return fig
```

Do not assume what the callee wants to do. Maybe they want to save it, maybe display it to a user, ...

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+    for traj in traj_list:
+      T = len(traj.rews)
+      X.extend(range(T))
+      rews = reward_fn(traj.obs[:-1], traj.acts, traj.obs[1:], None)
+      Y.extend(rews)
+    if i < len(colors):
+      color = colors[i]
+    else:
+      color = None
+    plt.plot(X, Y, alpha=1.0, c=color, label=traj_name)
```
AdamGleave:
```suggestion
    ax.plot(X, Y, alpha=1.0, c=color, label=traj_name)
```

experiments/mountain_car_plots.py
```diff
@@ -0,0 +1,269 @@
...
+  plt.xlabel("timestep")
+  plt.ylabel("test reward")
+  plt.legend()
+  plt.show()
+
+
+def _reward_fn_normalize_inputs(obs: np.ndarray,
```
AdamGleave: This function should be moved earlier in the code. Imagine the function call graph is a DAG: then the source code should be a topological sort of this DAG.

src/imitation/util/reward_wrapper.py
```diff
@@ -1,11 +1,11 @@
 """Common wrapper for adding custom reward values to an environment."""
 import collections
-from typing import Callable
+from typing import Callable, Optional
 
 import numpy as np
 from stable_baselines.common import vec_env
 
-RewardFn = Callable[[np.ndarray, np.ndarray, np.ndarray, np.ndarray],
+RewardFn = Callable[[np.ndarray, np.ndarray, np.ndarray, Optional[np.ndarray]],
```
AdamGleave: What are you imagining the semantics being here? Some reward functions ignore steps, but others depend on it and throw an error if not provided?

experiments/mountain_car_plots.py
shwang: Oh nice, didn't know about this one!

experiments/mountain_car_plots.py
shwang: Seaborn heatmap (at the way I set it up) plots the corners of each heatmap square on the integer lattice.

To make this clearer in the code I've renamed `covert_to_rc.*` to `convert_obs_to_heatmap_coords`. Didn't find a quick way to make `plt.imshow` or `sns.heatmap` center their heatmap squares around the actual datapoint, but `pcolor` looks promising.

https://matplotlib.org/gallery/images_contours_and_fields/pcolor_demo.html#sphx-glr-gallery-images-contours-and-fields-pcolor-demo-py

This seems like something worth looking into because it could simplify the code and remove the pandas/seaborn dependencies.

src/imitation/util/reward_wrapper.py
shwang: This change is unnecessary, thanks for pointing it out

shwang: I'm passing in `steps=np.arange(len(obs))` in heatmap reward functions now.

experiments/mountain_car_plots.py
shwang: Changed this guy to `steps=np.arange(...)`. Thanks

experiments/mountain_car_plots.py
shwang: This suggestion becomes necessary now that I've moved from `seaborn.heatmap` to `matplotlib.pcolor(X, Y, Z)`.

experiments/mountain_car_plots.py
shwang: Moved to `pcolor`

experiments/mountain_car_plots.py
shwang: Agree that this would be better more modular

shwang:
Thanks for the review. I've unsuccessfully attempted before to figure out how to gracefully use the OOP interface to matplotlib, so it was good to learn a cleaner way to manage these plots.

It's a pain for me to manually check whether plots are working so I'm going to write a smoke test of some sort. Ask for another round of review after then.

experiments/mountain_car_plots.py
shwang: Ended up getting rid of this helper function, since it was only used by `batch_reward_heatmaps`.

experiments/mountain_car_plots.py
shwang: I either moved helper methods out of the file (into `.rewards.common`), or prepended `_` to make functions private. Every public function here is now a potential Figure-generating entry point for analyzing mountain car rewards.

shwang:
I moved some reward-related helper functions and `util.reward_wrapper.RewardFn` into a new submodule `imitation.rewards.common`.

To make the functions importable and testable, I moved mountain_car_plots.py from `experiments/` to a new submodule `imitation.analysis.mountain_car_plots`. (Also considered `imitation.scripts`, but this isn't really a script right now, more like a collection of plotting tools.)

This should be ready for another round of review now.

shwang:
Somehow, this branch is set off the Sacred race condition that is fixed by https://github.com/IDSIA/sacred/pull/473 every time I reran tests.

Maybe this is bad luck on my part (in one of the reruns, `train_experts.sh` was the breaking script, in another rerun `train_experts.sh` was fine but `transfer_learn...sh` was broken.); maybe this is because CircleCI somehow changed.

shwang:
I think it would be really helpful for me if Sacred configs were picklable. This would allow us to upgrade to a version of Sacred that doesn't run into the FileStorageObserver race condition, and let me use the same version of Sacred that is used by the Minecraft project.

Asked on https://github.com/IDSIA/sacred/pull/508 if all they need is just dropping `pyrsistent.PMap`.

src/imitation/analysis/mountain_car_plots.py
```diff
@@ -0,0 +1,221 @@
...
+    vec_normalize_path = checkpoint_dir / "gen_policy" / "vec_normalize.pkl"
+    discrim_path = checkpoint_dir / "discrim"
+    policy_path = checkpoint_dir / "gen_policy"
+
+    if n_gen_trajs > 0:
+      # `load_policy` automatically loads VecNormalize for policy evaluation.
+      with policies_serialize.load_policy(
+            "ppo2", str(policy_path), venv) as gen_policy:
+        gen_trajs = rollout.generate_trajectories(
+          gen_policy, venv, sample_until=rollout.min_episodes(50))
```
AdamGleave:
```suggestion
          gen_policy, venv, sample_until=rollout.min_episodes(n_gen_trajs))
```

src/imitation/rewards/common.py
```diff
@@ -0,0 +1,46 @@
...
+                                norm_reward: bool = True,
+                                ) -> np.ndarray:
+  """Combine with `functools.partial` to create an input-normalizing RewardFn.
+
+  Args:
+    reward_fn: The reward function that normalized inputs are evaluated on.
+    vec_normalize: Instance of VecNormalize used to normalize inputs and
+     rewards.
+    norm_reward: If True, then also normalize reward before returning.
+  Returns:
```
AdamGleave:
```suggestion

  Returns:
```

PR #178, opened 2020-04-15
MonitorPlus: Rebrand as RolloutInfoWrapper

shwang:
MonitorPlus added raw observations and rewards to the
`info['episode']` dict generated by its parent class Monitor.

PPO2 and other Stable Baselines algorithms hold on to the 100 most
recent values of `info['episode']`.

For environment with long episodes and large observations, using
MonitorPlus and PPO2 used an unnecessarily large amount of memory
(up to 1.50 GB per episode). This patch reduced memory usage by
converting MonitorPlus into a separate wrapper RolloutInfoWrapper
which uses `info['rollout']` instead.


shwang:
Lint should pass after #179

src/imitation/util/rollout.py
```diff
@@ -30,22 +31,54 @@ class Trajectory(NamedTuple):
   infos: Optional[List[dict]]
 
 
+class RolloutInfoWrapper(gym.Wrapper):
+  """Add the entire episode's rewards and observations to `info` at episode end.
+
+  Whenever done=True, `info["rollouts"]` is a dict with keys
```
AdamGleave, line 37:
```suggestion
  Whenever done=True, `info["rollout"]` is a dict with keys
```

PR #180, opened 2020-04-29
Make Transitions and Trajectory data classes

### LINKED ISSUE #95: TransitionNoRew type
Right now a number of functions take input for transitions containing everything except reward (`RewardFn`, some AIRL/GAIL code). We could introduce an additional type like `Transition` but missing the reward, and then rewrite `RewardFn` and others to take this as input.

I'm not sure on the right inheritance hierarchy here: maybe have a single abstract named tuple `BaseTransition`, and then `TransitionNoRew` that inherits from this (without adding extra fields) and `Transition` that inherits and adds the `rew` field.

I suspect as we dig into the code there may be additional differences that get thrown up -- e.g. some places take `done` and `steps` and some don't. I think we should probably unify that code though rather than 

An alternative design is to just use `Transition` everywhere, and make the `rew` field optional, but I think there's a meaningful semantic difference here (e.g. `rollout_stats` needs `Transition` not `TransitionNoRew`).

AdamGleave:
Convert `Transitions` and `Trajectory` from named tuples to [data classes](https://docs.python.org/3/library/dataclasses.html) and add variants with and without reward.

Specifically, I introduce a `TransitionsNoRew` and `TrajectoryNoRew` base class, and add `TransitionsWithRew` and `TrajectoryWithRew` subclasses. We need dataclasses for this, not named tuples, since named tuples can't be subclassed (an m-tuple is not a subclass of an n-tuple even if m>n). I've declared them in a new file `imitation.util.data` since `imitation.util.rollout` was getting a bit unwieldy and in principle they could be generated by means other than rollouts.

The reward and imitation learning algorithms work with `TrajectoryNoRew` and `TransitionsNoRew`. `pytype` static analysis will therefore check that they do not depend on leaked reward information. This also lets us learn from data sources that do not include rewards.

However, most of `imitation.util.rollout` operates on `TrajectoryWithRew` and `TransitionsWithRew`. This reward information is useful in some cases, e.g. to evaluate a policy on the ground-truth reward. If it's unnecessary it can just be ignored as before.

For backward compatibility, I've added a `load()` method to `imitation.util.rollout` for sequences of trajectories, which will convert old-form pickles to this new form.

Closes #95.


src/imitation/util/data.py
```diff
@@ -0,0 +1,121 @@
...
+import dataclasses
+import os
+import pickle
+from typing import List, NamedTuple, Optional, Sequence
+
+import numpy as np
+import tensorflow as tf
+
+
+@dataclasses.dataclass(frozen=True)
```
shwang, line 12: Oh nice, I didn't know that you could freeze dataclasses

src/imitation/util/rollout.py
```diff
@@ -63,64 +47,36 @@ def step(self, action):
...
   Args:
-    traj: A Trajectory generated from `RolloutInfoWrapper`-wrapped Environments.
+    traj: A trajectory generated from `RolloutInfoWrapper`-wrapped Environments.
 
   Returns:
     A copy of `traj` with replaced `obs` and `rews` fields.
   """
   ep_info = traj.infos[-1]["rollout"]
-  res = traj._replace(obs=ep_info["obs"], rews=ep_info["rews"])
+  # no need to replace next_obs since that is never changed by VecEnv auto-reset
```
shwang, line 66: I found this comment confusing because `TrajectoryWithRew` doesn't have `next_obs`.

Is the comment saying here that it may seem like there is an edge case involving terminal observations, but this is actually sidestepped?

src/imitation/util/data.py
```diff
@@ -0,0 +1,121 @@
...
+  """Loads a sequence of trajectories saved by `save()` from `path`."""
+  # TODO(adam): remove backwards compatibility logic eventually (2021?)
+  from unittest import mock
+  with mock.patch(
+    'imitation.util.rollout.Trajectory',
+    new=_TrajectoryBackwardCompatible,
+    create=True,
+  ):
+    with open(path, "rb") as f:
+      trajectories = pickle.load(f)
```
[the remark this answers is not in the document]
AdamGleave: Yeah, if it's an old-style trajectory it'll try to import `imitation.util.rollout.Trajectory` and this will be replaced with `_TrajectoryBackwardCompatible`. Things work out OK since the new `Trajectory` is `imitation.util.data` so there's no naming clash -- otherwise it'd be impossible to load. (Long-term we should maybe consider a custom format given how fragile pickling can be.)

src/imitation/util/rollout.py
AdamGleave: Hm, I think maybe I thought this was for transitions? I'll just delete this comment, I think it's confusing.

PR #181, opened 2020-04-29
Use black for code formatting

AdamGleave:
- Add black to CircleCI checks and `ci/code_checks.sh` script for local linting.
  - Refomat code using `black`.
  - Also reformat docstrings with [docformatter](https://github.com/myint/docformatter) since it turns out going from 2-space to 4-space really messes up docstrings...
  - Manually fix flake8 errors.

Should not be any semantic changes here.


AdamGleave:
Something is wrong with `codecov`: it's claiming my patch to `tests` only has 84.92% of diff hit, but also says we have 100% of test LOC run. So, I think we can ignore this.

shwang:
As we discussed on Slack there were some lingering spacing errors in docstrings. Went through and fixed them in the latest commit.

AdamGleave:
> As we discussed on Slack there were some lingering spacing errors in docstrings. Went through and fixed them in the latest commit.

Thanks! This isn't the most glamorous job but it is important -- I know it would annoy me every time I work with the code base otherwise :)

PR #183, opened 2020-05-04
Fix early-termination bias in trajectory rollouts

AdamGleave:
Currently, we collect rollouts greedily in multiple environments simultaneously using a `VecEnv`, terminating the process when a condition is reached, such as a minimum number of episodes or timesteps. Any environments where an episode is still in progress -- it has not terminated and been added to the list of trajectories -- will be discarded.

This introduces a bias: longer-running episodes are more likely to be interrupted, and so less likely to be sampled. This effect is negligible when the number of episodes being sampled is much larger than the number of environments, but can be very large when the number of episodes is small.

This PR fixes this bias by, once sufficient data has been collected, continue to sample from other environments until the end of their episode. Provided all environments are i.i.d. and there is no systematic bias across episodes within an environment, this results in each trajectory being sampled i.i.d. 

The number of trajectories sampled is variable, and may overshoot the minimum condition. Furthermore, the *order* in which the trajectories are returned is non-random: shorter episodes finish and are added to the list earlier. This bias can be significant when the callee truncates the returned data to the desired length, so to avoid this the PR also shuffles the trajectories before returning.

I've added some tests although would welcome suggestions for improvement. They're more testing the implementation than the objective, but testing the objective (unbiased samples) seems hard to do in a deterministic manner and we do not want flaky tests.

### DIFF
--- a/src/imitation/util/rollout.py
+++ b/src/imitation/util/rollout.py
@@ -254,6 +254,7 @@ def generate_trajectories(
     sample_until: GenTrajTerminationFn,
     *,
     deterministic_policy: bool = False,
+    rng: np.random.RandomState = np.random,
 ) -> Sequence[data.TrajectoryWithRew]:
     """Generate trajectory dictionaries from a policy and an environment.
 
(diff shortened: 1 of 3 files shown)


AdamGleave:
Thanks for the review!

PR #184, opened 2020-05-08
Handle episode termination in reward models

AdamGleave:
This PR:
  - Makes `RewardNetShaped` and subclasses set next potential to be fixed to zero when the episode terminates. 0 is required in the undiscounted case for convergence guarantees (where it can be chosen WLOG).
  - `RewardNet` and subclasses can optionally use episode termination directly for prediction; also make inclusion of state, action and next state configurable.
  - Replaces `steps` (that was not used anywhere as far as I can tell) in `common.RewardFn` with `dones`.
  - Refactors MLP building code into `imitation.util.networks` out of various other files.

### DIFF
--- a/src/imitation/rewards/reward_net.py
+++ b/src/imitation/rewards/reward_net.py
@@ -334,13 +334,10 @@ class BasicShapedRewardNet(RewardNetShaped, serialize.LayersSerializable):
     """A shaped reward network with simple, default settings.
 
     With default parameters this RewardNet has two hidden layers [32, 32]
-    for the reward shaping phi network, and a linear function approximator
-    for the theta network. These settings match the network architectures for
-    continuous control experiments described in Appendix D.1 of the
-    AIRL paper.
+    for the theta network and reward shaping phi network.
 
-    This network flattens inputs. So it isn't suitable for training on
-    pixel observations.
+    This network is feed-forward and flattens inputs, so is a poor choice for
+    training on pixel observations.
     """
 
     def __init__(
--- a/src/imitation/rewards/reward_net.py
+++ b/src/imitation/rewards/reward_net.py
@@ -173,17 +173,29 @@ class RewardNetShaped(RewardNet):
 
         with tf.variable_scope("phi_network"):
             res = self.build_phi_network(self.obs_inp, self.next_obs_inp)
+            # end_potential is the potential when the episode terminates.
+            if discount_factor == 1.0:
+                # If undiscounted, terminal state must have potential 0.
+                end_potential = tf.constant(0.0)
+            else:
+                # Otherwise, it can be arbitrary, so make a trainable variable.
+                end_potential = tf.Variable(
+                    name="end_phi", shape=(), dtype=tf.float32, initial_value=0.0
+                )
+                self._layers.update(end_potential=end_potential)
             self._old_shaping_output, self._new_shaping_output, phi_layers = res
+            self._layers.update(**phi_layers)
 
         with tf.variable_scope("f_network"):
-            alive = 1 - self.done_inp
+            new_shaping = (
+                self.done_inp * end_potential
+                + (1 - self.done_inp) * self._new_shaping_output
+            )
             self._shaped_reward_output = (
                 self._theta_output
-                + self._discount_factor * alive * self._new_shaping_output
+                + self._discount_factor * new_shaping
                 - self._old_shaping_output
             )
-
-        self._layers.update(**phi_layers)
 
     @property
     def reward_output_train(self):
--- a/src/imitation/rewards/reward_net.py
+++ b/src/imitation/rewards/reward_net.py
@@ -1,7 +1,7 @@
 """Constructs deep network reward models."""
 
 from abc import ABC, abstractmethod
-from typing import Iterable, Optional, Tuple
+from typing import Iterable, Optional, Sequence, Tuple
 
 import gym
 import tensorflow as tf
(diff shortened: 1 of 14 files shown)


src/imitation/util/networks.py
```diff
@@ -0,0 +1,150 @@
...
+    hid_sizes: Optional[Iterable[int]] = None,
+    **kwargs: dict,
+) -> Tuple[tf.Tensor, LayersDict]:
+    """Builds an MLP depending on specified inputs.
+
+    All specified inputs will be flattened and then concatenated. They are then
+    applied using `sequential` to an MLP built using `build_mlp`.
+
+    Arguments:
+      hid_sizes: Number of units at each hidden layer. Default is [], i.e. linear.
```
shwang, line 128: The part about the default number of units is incorrect. It's actually (32, 32) right now.

Make `hid_sizes` a required argument with default (32, 32) or `()`, but still note in the docstring that an empty list means linear?

src/imitation/rewards/reward_net.py
```diff
@@ -221,50 +251,6 @@ def build_summaries(self):
...
-
-        inputs = [tf.layers.flatten(x) for x in inputs]
-        inputs = tf.concat(inputs, axis=1)
-        theta_mlp = util.build_mlp(hid_sizes=hid_sizes, name="reward", **kwargs)
-        theta_output = util.sequential(inputs, theta_mlp)
-
-        return theta_output, theta_mlp
-
-
 class BasicRewardNet(RewardNet, serialize.LayersSerializable):
```
shwang, line 254: In the doc string it says that we are using a linear function approximator, but I think the default right now is actually (32, 32). WE should either actually use linear function approx by passing in `hid_size=[]` or update the docstring.

src/imitation/rewards/reward_net.py
```diff
@@ -148,25 +165,38 @@ def __init__(
...
             res = self.build_phi_network(self.obs_inp, self.next_obs_inp)
+            # end_potential is the potential when the episode terminates.
+            if discount_factor == 1.0:
+                # If undiscounted, terminal state must have potential 0.
+                end_potential = tf.constant(0.0)
+            else:
+                # Otherwise, it can be arbitrary, so make a trainable variable.
+                end_potential = tf.Variable(
+                    name="end_phi", shape=(), dtype=tf.float32, initial_value=0.0
+                )
```
shwang, line 184: Why train an arbitrary potential for the absorbing state instead of just using zero?

Seems like if this value is positive, then it could encourage early episode termination in non-fixed-length environment.

src/imitation/util/networks.py
```diff
@@ -0,0 +1,150 @@
...
+    layer = tf.layers.Dense(
+        1, kernel_initializer=initializer, name=f"{name}_dense_final"
+    )  # type: tf.layers.Layer
+    layers[f"{name}_dense_final"] = layer
+
+    return layers
+
+
+def sequential(inputs: tf.Tensor, layers: LayersDict,) -> tf.Tensor:
+    """Applies a sequence of layers to an input."""
```
shwang, line 43: Until we have OrderedDict typing, could you an `assert isinstance(layers, OrderedDict)` here?

src/imitation/util/networks.py
AdamGleave: Fixed the docstring to note the default (you're right this was in existing code I copied & pasted).

I wanted to make it required rather than optional. The problem is it's used in a couple of different places (reward network and discriminator network) with None defaults, so we'd have to update those to have some default. I'm not sure whether it's better to have the default in one place or let the networks have their own defaults.

src/imitation/rewards/reward_net.py
AdamGleave: Good catch, fixed. I've also removed the bit about the AIRL paper since we no longer match that I guess.

src/imitation/util/networks.py
shwang: Ah that's right, the others are passing in None.

src/imitation/rewards/reward_net.py
AdamGleave: Normally potential shaping in the discounted case can represent adding a constant to the reward. If $\Phi(s) = k$ then each reward is increased by $(\gamma - 1)k$. If we fix potential shaping to be zero at the absorbing state it's removing the ability to add constant offsets to potential, therefore restricting the class of potential shaping we can represent.

It's plausible we don't need this degree of freedom and restricting it will make it easier to learn -- but it's just one extra parameter so I'd be surprised if it makes much difference.

In variable-horizon environments being able to learn to add a positive or negative constant to the reward seems useful. Sure, if it learns the wrong sign it'll hurt performance; but presumably it'd tend to learn the right sign? There's still the issue that episode termination leaks information so any results are dubious for evaluation, but I don't see a way to fix that.

src/imitation/util/networks.py
AdamGleave: On reflection I think I should just drop the warning comment and ignore this. We're assuming Python >3.7 in `setup.py` and dictionaries are guaranteed to be insertion ordered from 3.7 and above: https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6

shwang: Oh wow, OrderedDict becomes much less important then.

https://stackoverflow.com/questions/50872498/will-ordereddict-become-redundant-in-python-3-7

PR #186, opened 2020-05-08
Refactor data-related modules out of util

AdamGleave:
`imitation.util` was getting a bit overwhelming with over 10 submodules, so I've split out those that are data related into `imitation.data`. Some moderate rearrangement within those submodules, but nothing too drastic.

I also got rid of the `from imitation.util.util import *` in `imitation.util.__init__` since this was making the code hard to follow in places, and updated callsites.

### DIFF
--- a/ci/code_checks.sh
+++ b/ci/code_checks.sh
@@ -23,5 +23,9 @@ if [ "$skipexpensive" != "true" ]; then
   popd
 
   echo "Type checking"
+  # Tell pytype Python path explicitly using -P: otherwise it gets lost
+  # when the package is installed in editable mode (`pip -e .`). Note CI
+  # checks still run using full install (`pip .`) which is more robust
+  # but cumbersome for a Git commit hook.
   pytype -P src/:. ${SRC_FILES[@]}
 fi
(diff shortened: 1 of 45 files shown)


shwang:
@qxcv: Are you using `steps` for RewardFn right now? (I think that you added it in way back)

qxcv:
No

ci/code_checks.sh
```diff
@@ -23,5 +23,5 @@ if [ "$skipexpensive" != "true" ]; then
   popd
 
   echo "Type checking"
-  pytype ${SRC_FILES[@]}
+  pytype -P src/:. ${SRC_FILES[@]}
```
shwang, line 26: What problem does adding `src/` and `.` to the PYTHONPATH explicitly fix?

AdamGleave: When I install in editable (developer) mode, i.e. `pip -e .`, `pytype` non-deterministically fails to detect a significant fraction of modules. It works on other commits, so something in this PR breaks its detection in some way, but I've been unable to figure out what.

I'll add a comment to this effect.

PR #187, opened 2020-05-08
Clean up for PyPI Upload

### LINKED ISSUE #120: Upload to PyPI?
Does someone want to upload this package to PyPI? There doesn't seem to be an "imitation" package on PyPI yet, so it makes sense to reserve the name for this one :smile:

(possibly @shwang is the best person to manage PyPI uploads?)

AdamGleave:
- Add info to `setup.py`
  - Remove stray `src/__init__.py` file (FYI @shwang this was why `pytype` was breaking, oops)
  - Update README.md, especially code contribution guidelines.
  - Add PyPI badge to `README.md`; this doesn't work yet, but will once I upload it to PyPI and I'd rather not trash the commit history with two PRs for this.

I've done an upload to test PyPI and it looks OK, so probably checking the `README.md` makes sense is most important part.

Closes #120

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -3,7 +3,6 @@ coverage:
     project:
       default: false
       main:
-        target: 80%
         paths:
           - "src/imitation/"
           - "!src/imitation/envs/examples/"
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![CircleCI](https://circleci.com/gh/HumanCompatibleAI/imitation.svg?style=svg)](https://circleci.com/gh/HumanCompatibleAI/imitation)
+[![Documentation Status](https://readthedocs.org/projects/imitation/badge/?version=latest)](https://imitation.readthedocs.io/en/latest/?badge=latest)
 [![codecov](https://codecov.io/gh/HumanCompatibleAI/imitation/branch/master/graph/badge.svg)](https://codecov.io/gh/HumanCompatibleAI/imitation)
 
 # Imitation Learning Baseline Implementations
--- a/README.md
+++ b/README.md
@@ -29,13 +29,22 @@ View Tensorboard with `tensorboard --logdir output/`.
 
 
 # Contributing
-  * Follow the [Google Python Style Guide](http://google.github.io/styleguide/pyguide.html). Examples of Google-style
-docstrings can be found [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
-  * Add units tests covering any new features, or bugs that are being fixed.
-  * PEP8 guidelines with line width 80 and 2-space indents are enforced by `ci/lint.sh`,
-which is automatically run by Travis CI.
-  * Static type checking via `pytype` is automatically run in `ci/type_check.sh`.
-  * Code coverage is automatically enforced by CodeCov.
-    The exact coverage required by CodeCov depends on the previous
-    code coverage %. Files in `imitation/{examples,scripts}/` have no
-    coverage requirements.
+
+All PRs should include unit tests for any new features, using `pytest`.
+We use `pytype` for static type checking. Please add type annotations where possible.
+It is OK to omit annotations when it would make the code significantly more complex.
+
+Code coverage is automatically enforced by CodeCov.
+The exact coverage required by CodeCov depends on the previous code coverage %.
+Files in `imitation/{examples,scripts}/` have no coverage requirements.
+
+Please follow a coding style of:
+  * PEP8, with line width 88.
+  * Use the `black` autoformatter.
+  * Follow the [Google Python Style Guide](http://google.github.io/styleguide/pyguide.html) unless
+    it conflicts with the above. Examples of Google-style docstrings can be found
+    [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).
+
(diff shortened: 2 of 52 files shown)


qxcv:
The substantive changes look good to me. However, when trying to install, I ran into the following issues:

- Currently requires Python >=3.7. Is this actually necessary? It's not a huge deal, but it would be nice to have something that works with the default Python shipped in Ubuntu 18.04 (at least for the next 6-12 months, anyway).
- Also requires Python <=3.7: `pip` (shipped with Conda + Python 3.8) can't find Ray 0.7.4 for some reason: `ERROR: Could not find a version that satisfies the requirement ray[debug]==0.7.4 (from imitation==0.1a0) (from versions: 0.8.4, 0.8.5)`. Pretty sure this is because 0.7.4 was released in September, and there were no Python 3.8 wheels back then.
- When using Python 3.7.7 in a fresh Conda environment, everything installs (via `pip`), but with the following warning: `ERROR: gym 0.17.2 has requirement cloudpickle<1.4.0,>=1.2.0, but you'll have cloudpickle 1.4.1 which is incompatible.`. I can't see any package that actually requires `cloudpickle>1.4.0`, so this is probably just a case of `pip` sucking at dependency resolution. Neverthless, it might be an idea to put `cloudpickle<1.4.0` in this package until OpenAI updates Gym?

qxcv:
On [install order](https://pip.pypa.io/en/stable/reference/pip_install/?highlight=topological#installation-order):

> As of v6.1.0, pip installs dependencies before their dependents, i.e. in “topological order.” This is the only commitment pip currently makes related to order. While it may be coincidentally true that pip will install things in the order of the install arguments or in the order of the items in a requirements file, this is not a promise.

…so there probably aren't any reliable escape hatches to make pip do the right thing.

AdamGleave:
Thanks for the review @qxcv.

> * Currently requires Python >=3.7. Is this actually necessary? It's not a huge deal, but it would be nice to have something that works with the default Python shipped in Ubuntu 18.04 (at least for the next 6-12 months, anyway).

We're using `dataclasses` which were only introduced in Python 3.7, sadly.

> * Also requires Python <=3.7: `pip` (shipped with Conda + Python 3.8) can't find Ray 0.7.4 for some reason: `ERROR: Could not find a version that satisfies the requirement ray[debug]==0.7.4 (from imitation==0.1a0) (from versions: 0.8.4, 0.8.5)`. Pretty sure this is because 0.7.4 was released in September, and there were presumably no Python 3.8 wheels back then. Could fix by upgrading Ray, or just by asking them to upload more wheels for their old releases stuck_out_tongue

Good catch, this is annoying. @shwang can you remember why we chose 0.7.4 -- do more recent versions not work? I'm also inclined to make this an optional dependency -- as far as I can see this is only used in `scripts/parallel.py`, which many potential consumers may not need it.

> * When using Python 3.7.7 in a fresh Conda environment, everything installs (via `pip`), but with the following warning: `ERROR: gym 0.17.2 has requirement cloudpickle<1.4.0,>=1.2.0, but you'll have cloudpickle 1.4.1 which is incompatible.`. I can't see any package that actually requires `cloudpickle>1.4.0`, so this is probably just a case of `pip` sucking at dependency resolution. Neverthless, it might be an idea to put `cloudpickle<1.4.0` in the `install_requires` for this package until OpenAI updates Gym? (OTOH maybe there's some way to make `pip` do the right thing by moving Gym either before or after all the other dependencies? Not really sure because `pip`'s install order has always been black magic to me.)

Are we sure `gym` has a good reason for this? The PR that introduced this range is https://github.com/openai/gym/pull/1836/files back in March and 1.4.0 was only released 12 days ago. I'm inclined to ignore it unless we see breakage -- cloudpickle doesn't have a changelog but if following semantic versioning this shouldn't have breaking changes.

AdamGleave:
Ray 0.7.5 was released Sep 24 and our Ray commit https://github.com/HumanCompatibleAI/imitation/commit/a4d42f80a82f05b82d919508d5044ea3310e16f0 was on Sep 19th, so I think 0.7.4 is just the last tested-working version, rather than us having a particular reason to pin to an older version.

I just tried it with the latest version and `tests/test_scripts.py` still passes. Certainly possible there's a gremlin we'll run into, but most failure modes I can think of for Ray would manifest immediately. Given this I've updated `setup.py` to permit any of the 0.8.x series.

I think TensorFlow also doesn't have 3.8 wheels for the version range we want, though. I imagine they'll add that next bugfix release for 1.15.x though, and hopefully we will transition to PyTorch with Stable Baselines eventually.

qxcv:
> I think TensorFlow also doesn't have 3.8 wheels for the version range we want, though. I imagine they'll add that next bugfix release for 1.15.x though, and hopefully we will transition to PyTorch with Stable Baselines eventually.

Good catch on TF, but this issue suggests they won't be releasing Python 3.8 wheels for TF 1.15.X:

https://github.com/tensorflow/tensorflow/issues/37461

Guess this package is Python 3.7-only until we switch to PyTorch :man_shrugging:

AdamGleave:
> Good catch on TF, but this issue suggests they won't be releasing Python 3.8 wheels for TF 1.15.X:
> 
> [tensorflow/tensorflow#37461](https://github.com/tensorflow/tensorflow/issues/37461)
> 
> Guess this package is Python 3.7-only until we switch to PyTorch

Ah that's a shame :( Porting to PyTorch shouldn't be that hard although I'm not sure when/if I'll have the bandwidth.

Issue #109, opened 2019-10-15
Upgrade to Sacred 0.8.x


AdamGleave:
Our code currently breaks with anything >0.7.4 because they made config objects non-picklable. We can workaround this via a config attribute to disable their `ReadOnly*`, or by using a helper method. 

See https://github.com/IDSIA/sacred/issues/499 and https://github.com/IDSIA/sacred/pull/508 for some discussion

AdamGleave:
Keep running into problems with parallelism with Sacred 0.7.4: https://github.com/IDSIA/sacred/pull/503

Fixed in 0.8.x, so slightly higher priority to upgrade (although I can work around it for now)

shwang:
@AdamGleave: It seems to me that our serialization errors in `imitation` might not actually be coming from implicitly pickling Configs (as I think you might have mentioned in one of the threads related to this Sacred issue). I didn't realize this until today, but the latest version of Sacred (v0.8) under regular Sacred settings, never passes `dict` into argument-capturing functions any more. It always passes in `ReadOnlyDict`.

At least in `imitation` it looks like the pickle error comes from me trying to forward a `ReadOnlyDict` (which I used to think is just a `dict`) into Ray subprocesses.

shwang:
IIRC in the Adversarial Policies repo you actually pass around pickled configs. For `imitation` it might be sufficient to just convert every `ReadOnlyDict` into a `dict`. (This would be messy though, as we'd have to recursively "unfreeze" the data structure)

AdamGleave:
> @AdamGleave: It seems to me that our serialization errors in `imitation` might not actually be coming from implicitly pickling Configs (as I think you might have mentioned in one of the threads related to this Sacred issue). I didn't realize this until today, but the latest version of Sacred (v0.8) under regular Sacred settings, never passes `dict` into argument-capturing functions any more. It always passes in `ReadOnlyDict`.
> 
> At least in `imitation` it looks like the pickle error comes from me trying to forward a `ReadOnlyDict` (which I used to think is just a `dict`) into Ray subprocesses.

I'm a bit confused what you mean by implicitly pickling configs? My current understanding was the problem is that the arguments Sacred passes in aren't pickleable, and we want to pickle them when sending to other processes (e.g. for Ray). Apologies if I gave a misleading impression. My view is Sacred should not be changing the types of config arguments under the user's feet in this way, which is why I opened the issue, not that it's an insurmountable problem.

[...]

PR #194, opened 2020-06-01
Upgrade Sacred to v0.8.1

shwang:
This PR updates our setup script to install ~~a patched version of Sacred v0.8.1 with two merged PRs.~~ Sacred v0.8.1.

We were unable to upgrade Sacred to the latest version because of issued noted in #109.  I recently submitted a patch to Sacred that makes ReadOnly{Dict,List} picklable, https://github.com/IDSIA/sacred/pull/737.

While running our Ray jobs with this patched version of Sacred, I also found that Sacred experiments run on Raylets fail to close their stdout-capturing `tee`s for some reason, which lead to an unhandled TimeoutError. This issue is fixed by https://github.com/IDSIA/sacred/pull/740, which warns that the tees failed to close instead of crashing.

~~I'm choosing to use our own version instead of IDSIA/sacred for now because the maintainers have not been responsive on issues and PRs for a while now, and I'd like to use `imitation` along with `realistic_benchmarks`, which requires a more recent version of `sacred`.~~

To avoid messing up PyPI and until the PRs are merged by IDSIA to the main repo we've decided to just address the two issues above by converting ReadOnlyContainer to regular python types when necessary, and by using `--capture=sys` to avoid the Tee TimeoutError.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,8 @@ setup(
         "jax~=0.1.66",
         "jaxlib~=0.1.47",
         # Install from IDSIA/sacred@master once
-        # https://github.com/IDSIA/sacred/pull/737 is merged.
+        # https://github.com/IDSIA/sacred/pull/737 and
+        # https://github.com/IDSIA/sacred/pull/740 are merged.
         "sacred @ git+https://github.com/HumanCompatibleAI/sacred@0.8.1-with-imit-compat",  # noqa: E501
     ],
     tests_require=TESTS_REQUIRE,
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ setup(
         "stable-baselines~=2.10.0",
         "jax~=0.1.66",
         "jaxlib~=0.1.47",
-        "sacred",
+        "sacred~=0.8.1",
     ],
     tests_require=TESTS_REQUIRE,
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -55,9 +55,9 @@ setup(
         "stable-baselines~=2.10.0",
         "jax~=0.1.66",
         "jaxlib~=0.1.47",
-        # sacred==0.7.5 build is broken without pymongo
-        # sacred>0.7.4 have non-picklable config objects (see GH #109)
-        "sacred==0.7.4",
+        # Install from IDSIA/sacred@master once
+        # https://github.com/IDSIA/sacred/pull/737 is merged.
+        "sacred @ git+https://github.com/HumanCompatibleAI/sacred@0.8.1-with-imit-compat",  # noqa: E501
     ],
     tests_require=TESTS_REQUIRE,
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -55,10 +55,7 @@ setup(
         "stable-baselines~=2.10.0",
         "jax~=0.1.66",
         "jaxlib~=0.1.47",
-        # TODO(shwang): Install from IDSIA/sacred@master once
-        # https://github.com/IDSIA/sacred/pull/737 and
-        # https://github.com/IDSIA/sacred/pull/740 are merged.
-        "sacred @ git+https://github.com/HumanCompatibleAI/sacred@0.8.1-with-imit-compat",  # noqa: E501
+        "sacred",
     ],
     tests_require=TESTS_REQUIRE,
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ setup(
         "stable-baselines~=2.10.0",
         "jax~=0.1.66",
         "jaxlib~=0.1.47",
-        # Install from IDSIA/sacred@master once
+        # TODO(shwang): Install from IDSIA/sacred@master once
         # https://github.com/IDSIA/sacred/pull/737 and
         # https://github.com/IDSIA/sacred/pull/740 are merged.
         "sacred @ git+https://github.com/HumanCompatibleAI/sacred@0.8.1-with-imit-compat",  # noqa: E501
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -348,6 +348,8 @@ def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> Dict[str, 
             stat_value: np.generic = getattr(np, stat_name)(desc_vals)
             out_stats[f"{desc_name}_{stat_name}"] = stat_value.item()
 
+    for v in out_stats.values():
+        assert isinstance(v, (int, float))
     return out_stats
 
 
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -346,6 +346,9 @@ def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> Dict[str, 
     for desc_name, desc_vals in traj_descriptors.items():
         for stat_name in stat_names:
             stat_value: np.generic = getattr(np, stat_name)(desc_vals)
+            # Convert numpy type to float or int. The numpy operators always return
+            # a numpy type, but we want to return type float. (int satisfies
+            # float type for the purposes of static-typing).
             out_stats[f"{desc_name}_{stat_name}"] = stat_value.item()
 
     for v in out_stats.values():
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -314,7 +314,7 @@ def generate_trajectories(
     return trajectories
 
 
-def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> dict:
+def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> Dict[str, float]:
     """Calculates various stats for a sequence of trajectories.
 
     Args:
@@ -331,7 +331,7 @@ def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> dict:
         are only included if the `trajectories` contain Monitor infos.
     """
     assert len(trajectories) > 0
-    out_stats = {"n_traj": len(trajectories)}
+    out_stats: Dict[str, float] = {"n_traj": len(trajectories)}
     traj_descriptors = {
         "return": np.asarray([sum(t.rews) for t in trajectories]),
         "len": np.asarray([len(t.rews) for t in trajectories]),
@@ -345,8 +345,9 @@ def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> dict:
     stat_names = ["min", "mean", "std", "max"]
     for desc_name, desc_vals in traj_descriptors.items():
         for stat_name in stat_names:
-            stat_value = getattr(np, stat_name)(desc_vals)
-            out_stats[f"{desc_name}_{stat_name}"] = stat_value
+            stat_value: np.generic = getattr(np, stat_name)(desc_vals)
+            out_stats[f"{desc_name}_{stat_name}"] = stat_value.item()
+
     return out_stats
 
 
--- a/src/imitation/scripts/expert_demos.py
+++ b/src/imitation/scripts/expert_demos.py
@@ -121,6 +121,7 @@ def rollouts_and_policy(
 
         if init_tensorboard:
             sb_tensorboard_dir = osp.join(log_dir, "sb_tb")
+            init_rl_kwargs = dict(init_rl_kwargs)  # Cast sacred's ReadOnlyDict to dict.
             init_rl_kwargs["tensorboard_log"] = sb_tensorboard_dir
 
         venv = util.make_vec_env(
--- a/src/imitation/scripts/expert_demos.py
+++ b/src/imitation/scripts/expert_demos.py
@@ -238,5 +238,5 @@ def main_console():
     expert_demos_ex.run_commandline()
 
 
-if __name__ == "__main__":
+if __name__ == "__main__":  # pragma: no cover
     main_console()
--- a/src/imitation/scripts/expert_demos.py
+++ b/src/imitation/scripts/expert_demos.py
@@ -121,7 +121,8 @@ def rollouts_and_policy(
 
         if init_tensorboard:
             sb_tensorboard_dir = osp.join(log_dir, "sb_tb")
-            init_rl_kwargs = dict(init_rl_kwargs)  # Cast sacred's ReadOnlyDict to dict.
+            # Convert sacred's ReadOnlyDict to dict so we can modify on next line.
+            init_rl_kwargs = dict(init_rl_kwargs)
             init_rl_kwargs["tensorboard_log"] = sb_tensorboard_dir
 
         venv = util.make_vec_env(
--- a/src/imitation/scripts/expert_demos.py
+++ b/src/imitation/scripts/expert_demos.py
@@ -232,7 +232,7 @@ def rollouts_from_policy(
 
 
 def main_console():
-    observer = FileStorageObserver.create(osp.join("output", "sacred", "expert_demos"))
+    observer = FileStorageObserver(osp.join("output", "sacred", "expert_demos"))
     expert_demos_ex.observers.append(observer)
     expert_demos_ex.run_commandline()
 
--- a/src/imitation/scripts/parallel.py
+++ b/src/imitation/scripts/parallel.py
@@ -52,7 +52,9 @@ def parallel(
     # change the working directory for each Raylet.
     if sacred_ex_name == "train_adversarial":
         if "data_dir" not in base_config_updates:
-            base_config_updates["data_dir"] = os.path.join(os.getcwd(), "data/")
+            data_dir = os.path.join(os.getcwd(), "data/")
+            base_config_update = dict(base_config_updates)
+            base_config_update["data_dir"] = data_dir
 
     trainable = _ray_tune_sacred_wrapper(
         sacred_ex_name, run_name, base_named_configs, base_config_updates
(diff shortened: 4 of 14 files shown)


shwang:
In the end, I found it cleanest to used copy.deepcopy to "unfreeze" the `ReadOnlyDict`s in `parallel.py` anyways. Which is a shame considering that we could have foregone making ReadOnlyDict picklable if I figured this out in advance.

But https://github.com/IDSIA/sacred/pull/740 might have been necessary anyways.

shwang:
Various fixes made to make the PR pass CI:

* Added code coverage for entrypoints to each Sacred script after refactoring deprecated Observer initializations. 
* Moved an assertion out of Sacred config and into the associated Sacred experiment command because the assertion was preventing `print_config` from running properly.

src/imitation/scripts/expert_demos.py
```diff
@@ -121,6 +121,7 @@ def rollouts_and_policy(
 
         if init_tensorboard:
             sb_tensorboard_dir = osp.join(log_dir, "sb_tb")
+            init_rl_kwargs = dict(init_rl_kwargs)  # Cast sacred's ReadOnlyDict to dict.
```
AdamGleave, line 124: Is this necessary? Are we parallelizing things inside `expert_demos`?

src/imitation/scripts/parallel.py
```diff
@@ -47,15 +49,28 @@ def parallel(
...
     # Explicitly set `data_dir` if parallelizing `train_adversarial`.
-    # We need this to automatically find rollout pickles Ray will automatically
-    # change the working directory for each Raylet.
+    # We need this to automatically find rollout pickles because Ray
+    # sets a new working directory for each Raylet.
     if sacred_ex_name == "train_adversarial":
         if "data_dir" not in base_config_updates:
-            base_config_updates["data_dir"] = os.path.join(os.getcwd(), "data/")
+            data_dir = os.path.join(os.getcwd(), "data/")
+            base_config_update = dict(base_config_updates)
```
AdamGleave, line 66: Typo? `base_config_update` != `base_config_updates`. If these are meant to be different then please rename to something more clearly distinct.

src/imitation/scripts/parallel.py
```diff
@@ -47,15 +49,28 @@ def parallel(
...
+            base_config_update["data_dir"] = data_dir
 
     trainable = _ray_tune_sacred_wrapper(
-        sacred_ex_name, run_name, base_named_configs, base_config_updates
+        sacred_ex_name,
+        run_name,
+        copy.deepcopy(base_named_configs),
```
AdamGleave, line 72: Aren't `base_named_configs` just a list of strings? Can we just do `list(base_named_configs)`? I'm not sure I even understand why this is needed -- is `base_named_configs` a `FrozenList`? Good to add a comment explaining this.

src/imitation/scripts/parallel.py
```diff
@@ -70,7 +85,7 @@ def parallel(
     try:
         ray.tune.run(
             trainable,
-            config=search_space,
+            config=copy.deepcopy(search_space),
```
AdamGleave, line 88: Again, a comment explaining this would be useful.

src/imitation/scripts/parallel.py
```diff
@@ -115,16 +138,26 @@ def inner(config: dict, reporter) -> dict:
...
+        named_configs.extend(run_kwargs["named_configs"])
+        updated_run_kwargs["named_configs"] = named_configs
+
+        config_updates = {}
+        config_updates.update(base_config_updates)
+        config_updates.update(run_kwargs["config_updates"])
+        updated_run_kwargs["config_updates"] = config_updates
+
+        # Add other run_kwargs items to updated_run_kwargs.
+        for k, v in run_kwargs.items():
```
AdamGleave, line 156: To check my understanding: the semantics here are that `base_config_updates` should take priority over `run_kwargs` when both specify the same key?

src/imitation/scripts/train_adversarial.py
```diff
@@ -108,25 +109,22 @@ def train(
...
         assert len(expert_trajs) >= n_expert_demos
         expert_trajs = expert_trajs[:n_expert_demos]
 
-    expert_stats = rollout.rollout_stats(expert_trajs)
-
     with networks.make_session():
         if init_tensorboard:
             sb_tensorboard_dir = osp.join(log_dir, "sb_tb")
-            kwargs = init_trainer_kwargs
+            kwargs = copy.deepcopy(init_trainer_kwargs)
```
AdamGleave, line 127: Why is this here? I don't understand the need to have `copy.deepcopy` in script files (here or elsewhere). Non-pickleable objects will cause a problem before it even hits this line of code: they need to be addressed in the code that invokes Ray, not here.

tests/test_scripts.py
```diff
@@ -6,24 +6,37 @@
...
+    expert_demos,
+    parallel,
+    train_adversarial,
+)
+
+ALL_SCRIPTS_MODS = [analyze, eval_policy, expert_demos, parallel, train_adversarial]
+
+
+@pytest.mark.parametrize("script_mod", ALL_SCRIPTS_MODS)
+def test_main_console(script_mod):
```
AdamGleave, line 31: :+1:

setup.py
```diff
@@ -55,9 +55,10 @@ def get_readme() -> str:
         "stable-baselines~=2.10.0",
         "jax~=0.1.66",
         "jaxlib~=0.1.47",
-        # sacred==0.7.5 build is broken without pymongo
-        # sacred>0.7.4 have non-picklable config objects (see GH #109)
-        "sacred==0.7.4",
+        # TODO(shwang): Install from IDSIA/sacred@master once
+        # https://github.com/IDSIA/sacred/pull/737 and
+        # https://github.com/IDSIA/sacred/pull/740 are merged.
+        "sacred @ git+https://github.com/HumanCompatibleAI/sacred@0.8.1-with-imit-compat",  # noqa: E501
```
AdamGleave, line 61: We can't upload to PyPi if our requirements are not in PyPi, so I'm reluctant to make this change on a non-temporary basis.

src/imitation/scripts/parallel.py
```diff
@@ -47,15 +49,28 @@ def parallel(
...
+            data_dir = os.path.join(os.getcwd(), "data/")
+            base_config_update = dict(base_config_updates)
+            base_config_update["data_dir"] = data_dir
 
     trainable = _ray_tune_sacred_wrapper(
-        sacred_ex_name, run_name, base_named_configs, base_config_updates
+        sacred_ex_name,
+        run_name,
+        copy.deepcopy(base_named_configs),
+        copy.deepcopy(base_config_updates),
```
AdamGleave, line 73: Again, a comment explaining this would be useful.

src/imitation/scripts/train_adversarial.py
shwang: The next line modifies `kwargs`, which is illegal without converting to `dict`. I chose to "defensively" deep copy in case for some reason in the future any code downstream (including in the imitation learning algorithm) wants to modify any dicts or lists inside the dict.

src/imitation/scripts/parallel.py
shwang: :+1: this is a typo.

src/imitation/scripts/parallel.py
shwang: No, it's actually the opposite. Because `base_config_updates` gets `update`-applied first, and then `run_kwargs["named_configs"]` overwrites this with the second `update`.

Looking back at the docstring, this doesn't seem well explained.
- [ ] Better docs for base_*_config

src/imitation/scripts/expert_demos.py
shwang: Need this in order to modify `tensorboard_log` in the next line.

src/imitation/scripts/parallel.py
shwang: I added in the `deepcopy` to check if non-forked Sacred would work, and left it in for a confused reason that probably isn't worth elaborating on.

non-forked Sacred didn't work because it would deterministically fail on the issue fixed by https://github.com/IDSIA/sacred/pull/740.

src/imitation/scripts/parallel.py
shwang: (see explanation above in https://github.com/HumanCompatibleAI/imitation/pull/194#discussion_r434180592)

src/imitation/scripts/parallel.py
shwang: (See https://github.com/HumanCompatibleAI/imitation/pull/194#discussion_r434180592)

shwang: Looking back `deepcopy` + ReadOnlyDict is pretty esoteric knowledge. Agreed that this would have done better with a comment at least explaining what `deepcopy` does here

setup.py
shwang: This is a reasonable objection, I forgot that PyPI has this requirement

shwang:
@AdamGleave thanks for your thorough review!

Especially with the PyPI consideration in mind I agree that it isn't a good idea to point to the Sacred fork as a requirement.

I think I'll do the thing you suggested where I install a personal `imitation` branch when I'm in a python environment that also needs the latest version of Sacred.

AdamGleave:
> I think I'll do the thing you suggested where I install a personal `imitation` branch when I'm in a python environment that also needs the latest version of Sacred.

I would like to upgrade to Sacred 0.8.1 still...

With `deepcopy`, is the only blocking issue the `tee` bug? I'm more optimistic about some kind of fix there making it's way into Sacred eventually (whereas the pickle issue I think is more a difference of opinion between us and the maintainers). Is there a way we can workaround it? It seems to be very dependent on platform, so if it doesn't break unit tests on CI, we can maybe just add a warning in the README with a link to your fork...

shwang:
Oops, I missed your most recent comment! Looking into CI+Sacred 0.8.1 stuff.

shwang:
Tried it out, and in the first CI run one of the parallel_ex test cases and a expert_demos_ex case failed on the Tee error.

https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/402/workflows/33b80186-3b6c-4594-82b3-76071abab6da/jobs/836/parallel-runs/2/steps/2-107

AdamGleave:
> Tried it out, and in the first CI run one of the parallel_ex test cases and a expert_demos_ex case failed on the Tee error.
> 
> https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/402/workflows/33b80186-3b6c-4594-82b3-76071abab6da/jobs/836/parallel-runs/2/steps/2-107

I think this can probably be worked around by changing the [capture mode](https://sacred.readthedocs.io/en/stable/experiment.html#capturing-stdout-stderr). The nuclear option is to just set capture mode to `no`, but I think setting it to `sys` would also work. It'll miss some print statements from libraries etc but I don't think we really need these? Certainly not for tests.

I'd ideally still like to fix the root cause, but I think Sacred maintainers would be receptive to any PRs on the `tee` issue we might come up with.

shwang:
The `sys` capture mode sounds appropriate for tests. This could be a fixture in the test_scripts.py file, and we could add a little comment that the fixture can be removed for more detailed error logging and stuff.

shwang:
"sys" seems to fix things on my end, at least with the manual tests I"ve managed so far, setting `-C sys` from the command line.

Unfortunately, AFAIK using `SETTINGS["CAPTURE_MODE"] = ...` is a no-op because this setting takes 2nd priority over the command line option `-C`, which is by default set to `"fd"` on Unix systems.

shwang:
Looking into whether this is a workaround for this...

shwang:
https://github.com/IDSIA/sacred/issues/742

src/imitation/data/rollout.py
```diff
@@ -345,8 +345,11 @@ def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> dict:
     stat_names = ["min", "mean", "std", "max"]
     for desc_name, desc_vals in traj_descriptors.items():
         for stat_name in stat_names:
-            stat_value = getattr(np, stat_name)(desc_vals)
-            out_stats[f"{desc_name}_{stat_name}"] = stat_value
+            stat_value: np.generic = getattr(np, stat_name)(desc_vals)
+            out_stats[f"{desc_name}_{stat_name}"] = stat_value.item()
+
+    for v in out_stats.values():
+        assert isinstance(v, (int, float))
```
shwang, line 352: These return values need to be non-numpy because otherwise they don't return properly as run-results from expert_demos_ex, etc.

I could add this add as a comment to the code itself if that seems helpful.

src/imitation/scripts/expert_demos.py
shwang: Modifying the comment to be explicit why we are ~"casting" to dict here.

src/imitation/scripts/parallel.py
shwang: Added better docs.

setup.py
```diff
@@ -55,9 +55,7 @@ def get_readme() -> str:
         "stable-baselines~=2.10.0",
         "jax~=0.1.66",
         "jaxlib~=0.1.47",
-        # sacred==0.7.5 build is broken without pymongo
-        # sacred>0.7.4 have non-picklable config objects (see GH #109)
-        "sacred==0.7.4",
+        "sacred",
```
shwang, line 58: Should we lock to 0.8.1?

src/imitation/scripts/parallel.py
shwang: This for loop adds all keys except "config_updates" and "named_configs" from the search_space into the run_kwargs. Hopefully the docs should clarify this a bit.

It's pretty messy getting the base_config_updates + config_updates in through the ray.tune API.

setup.py
AdamGleave: I'd rather not -- you're then forcing users of `imitation` to choose a particular `sacred` version. You probably want to have a minimum version though?

src/imitation/data/rollout.py
AdamGleave: Hmm, it might be worth mentioning briefly that the NumPy is returning a NumPy type even for scalars, but I don't think we need to mention the scripts since this is in a non-script file.

src/imitation/scripts/parallel.py
```diff
@@ -104,6 +137,18 @@ def _ray_tune_sacred_wrapper(
     """
 
     def inner(config: dict, reporter) -> dict:
+        """Trainable function with the correct signature for `ray.tune`.
+
+        Args:
+            config: Keyword arguments for `ex.run()`, where `ex` is the
+                `sacred.Experiment` instance associated with `sacred_ex_name`.
+        """
+        # Set Sacred capture mode to "sys" because default "fd" option leads to error.
```
AdamGleave, line 146: I'd make this a TODO or SOMEDAY since hopefully one day the Sacred issue is resolved and we can remove this workaround

tests/test_scripts.py
```diff
@@ -6,24 +6,50 @@
...
+    expert_demos,
+    parallel,
+    train_adversarial,
+)
+
+ALL_SCRIPTS_MODS = [analyze, eval_policy, expert_demos, parallel, train_adversarial]
+
+
+@pytest.fixture(autouse=True)
+def sacred_capture_use_sys():
```
AdamGleave, line 32: Should probably also have a TODO/SOMEDAY on it since a workaround for a bug in Sacred.

tests/test_scripts.py
```diff
@@ -6,24 +6,50 @@
...
+
+    See https://github.com/IDSIA/sacred/issues/289."""
+    temp = sacred.SETTINGS["CAPTURE_MODE"]
+    sacred.SETTINGS.CAPTURE_MODE = "sys"
+    yield
+    sacred.SETTINGS.CAPTURE_MODE = temp
+
+
+@pytest.mark.parametrize("script_mod", ALL_SCRIPTS_MODS)
+def test_main_console(script_mod):
```
AdamGleave, line 43: Should we remove the `# pragma: no cover` lines then from the main console blocks?

src/imitation/scripts/parallel.py
```diff
@@ -27,35 +30,65 @@ def parallel(
...
+            CLI '--name' option on the inner experiment. Offline analysis jobs can use
+            this argument to group similar data.
+        search_space: A dictionary which can contain `ray.tune.grid_search` and is
+            passed as the `config` argument to `ray.tune.run()`. After the
+            `search_space` is transformed by Ray, it passed into
+            `sacred_ex.run(**run_kwargs)` as `run_kwargs` (`sacred_ex` is the Sacred
+            Experiment selected via `sacred_ex_name`). Usually `search_space` only has
+            the keys "named_configs" and "config_updates", but any parameter names
+            to `sacred.Experiment.run()` are okay.
+        base_named_configs: Default Sacred Run named configs. Any named configs
```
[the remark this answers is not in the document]
shwang: probably meant to just write "named configs" there.

tests/test_scripts.py
shwang: `main_console` never had `pragma: no cover`. Did you mean the `if __name__ == "__main__"` blocks?

`test_main_console` doesn't account for that, as it calls `main_console()` directly.

src/imitation/scripts/parallel.py
```diff
@@ -27,35 +30,65 @@ def parallel(
...
-      upload_dir: `upload_dir` argument to `ray.tune.run()`.
+        sacred_ex_name: The Sacred experiment to tune. Either "expert_demos" or
+            "train_adversarial".
+        run_name: A name describing this parallelizing experiment.
+            This argument is also passed to `ray.tune.run` as the `name` argument.
+            It is also saved in 'sacred/run.json' of each inner Sacred experiment
+            under the 'experiment.name' key. This is equivalent to using the Sacred
+            CLI '--name' option on the inner experiment. Offline analysis jobs can use
+            this argument to group similar data.
+        search_space: A dictionary which can contain `ray.tune.grid_search` and is
```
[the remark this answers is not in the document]
shwang: Updated comment

tests/test_scripts.py
shwang: If coverage for the __main__ block is something we want to have, I can add that test in a separate PR.

src/imitation/data/rollout.py
shwang: Added a comment to that effect.

PR #197, opened 2020-06-18
Update gail total_timesteps params

shwang:
Increase total_timesteps for Ant and Cartpole environments based on performance of recent GAIL runs.

Ant: https://docs.google.com/document/d/18gkeBEJ9zR2vByF8sMTz8OSdEKND1lBZHXmyjFMCRMw/edit#heading=h.ftga7aa8ppsp

CartPole: https://docs.google.com/document/d/18gkeBEJ9zR2vByF8sMTz8OSdEKND1lBZHXmyjFMCRMw/edit#heading=h.ws8sgs9dvin4

### DIFF
--- a/src/imitation/scripts/config/train_adversarial.py
+++ b/src/imitation/scripts/config/train_adversarial.py
@@ -121,7 +121,7 @@ def airl():
 MUJOCO_SHARED_LOCALS = dict(init_trainer_kwargs=dict(airl_entropy_weight=0.1,),)
 
 ANT_SHARED_LOCALS = dict(
-    total_timesteps=3e7,
+    total_timesteps=8e7,
     gen_batch_size=2048 * 8,
     disc_batch_size=2048 * 8,
     init_trainer_kwargs=dict(max_episode_steps=500,),  # To match `inverse_rl` settings.
@@ -142,6 +142,7 @@ def cartpole():
     env_name = "CartPole-v1"
     rollout_hint = "cartpole"
     init_trainer_kwargs = dict(scale=False,)
+    total_timesteps = 3e7
 
 
 @train_ex.named_config


AdamGleave:
It seems our default `total_timesteps = 1e5` which is rather low (I guess meant for easy interactive testing), and then we increase this on an ad-hoc basis for some but not all environments (e.g. `acrobot`, `mountain_car`, `pendulum`, `hopper`, `reacher`, `walker`, etc all seem to be left at defaults).

Can we clear this up a bit? I'd be inclined to think we should set `total_timesteps` for all or no environments.

shwang:
It seems super useful to have the default settings for any particular environment to cause GAIL/AIRL to reach expert, or as good as our algorithm can get.

To satisfy this w.r.t. total_timesteps, I can take the max over the required timesteps for every environment once I know this number. However, this would run unnecessarily long for most non-Mujoco environments.

Maybe this would be ok in conjuction with a callback that halts training once we reach expert performance.

shwang:
If we want to have a single setting of total_timesteps across all environments or not set it all then I'll close this PR for now.

@AdamGleave would you also imagine having a single set of hyperparameters across similar GAIL/AIRL environments?

AdamGleave:
Sorry I don't think I was very clear. My main issue was that a bunch of environments don't have `total_timesteps` specified and so fall back to the default. It feels weird for `with hopper` to get expert but `with acrobot` to stop after `1e5` timesteps with a bad policy. I'd suggest setting the other environments to some conservative large value if you've not determined the right length yet?

shwang:
Ok, I understand now. I'll soon have a good idea on a conservative default for `total_timesteps` as I do more GAIL rebenchmarking.

PR #198, opened 2020-06-18
DiscrimNetGAIL: Add negative rewards option

shwang:
Adds an parameter to DiscrimNetGAIL constructor that flips the training reward sign. Also sets train_adversarial's `mountain_car` named config to use negative training reward by default.

Finally, adds coverage for negative training reward in test_trainer, which is renamed to test_adversarial in this PR. `discrim_kwargs` everywhere is renamed to `discrim_net_kwargs` for consistency with `reward_net_kwargs`.

### DIFF
--- a/src/imitation/rewards/discrim_net.py
+++ b/src/imitation/rewards/discrim_net.py
@@ -360,22 +360,28 @@ class DiscrimNetGAIL(DiscrimNet, serialize.LayersSerializable):
         build_discrim_net: DiscrimNetBuilder = networks.build_and_apply_mlp,
         build_discrim_net_kwargs: Optional[dict] = None,
         scale: bool = False,
+        positive_rewards: bool = True,
     ):
         """Construct discriminator network.
 
         Args:
-          observation_space: observation space for this environment.
-          action_space: action space for this environment:
-          build_discrim_net: a callable that takes an observation input tensor
-            and action input tensor as input, then computes the logits
-            necessary to feed to GAIL. When called, the function should return
-            *both* a `LayersDict` containing all the layers used in
-            construction of the discriminator network, and a `tf.Tensor`
-            representing the desired discriminator logits.
-          build_discrim_net_kwargs: optional extra keyword arguments for
-            `build_discrim_net()`.
-          scale: should inputs be rescaled according to declared observation
-            space bounds?
+            observation_space: Observation space for this environment.
+            action_space: Action space for this environment:
+            build_discrim_net: A callable that takes an observation input tensor
+                and action input tensor as input, then computes the logits
+                necessary to feed to GAIL. When called, the function should return
+                *both* a `LayersDict` containing all the layers used in
+                construction of the discriminator network, and a `tf.Tensor`
+                representing the desired discriminator logits.
+            build_discrim_net_kwargs: Optional extra keyword arguments for
+                `build_discrim_net()`.
+            scale: If True, then normalize observations to numbers between 0 and 1 by
+                scaling them by observation space bounds.
+            positive_rewards: In the default formulation, all GAIL rewards are positive,
+                which can lead to episode-length maximizing behavior in environments
+                with variable-length episodes. Use False to flip the sign of the reward,
+                as the original authors did for their MountainCar-v0 results (a
+                variable-episode-length environment with only negative rewards).
         """
         # for serialisation
         args = dict(locals())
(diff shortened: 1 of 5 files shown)


shwang:
Waiting on other stuff to get merged (because Sacred config's discrim kwargs got revamped, I think) before completing this one

src/imitation/rewards/discrim_net.py
```diff
@@ -424,11 +431,14 @@ def build_graph(self):
...
-        )
 
         self._disc_loss = tf.nn.sigmoid_cross_entropy_with_logits(
             logits=self._disc_logits_gen_is_high,
             labels=tf.cast(self.labels_gen_is_one_ph, tf.float32),
         )
+
+        reward = tf.log_sigmoid(self._disc_logits_gen_is_high)
+        if self._positive_rewards:
+            reward *= -1
```
AdamGleave, line 442: I don't understand how this works. AFAIK we don't backpropagate through `reward` -- so won't flipping the sign not change the discriminator training, and so cause the policy to try to maximize dissimilarity from expert?

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -140,6 +140,9 @@ def cartpole():
 def mountain_car():
     env_name = "MountainCar-v0"
     rollout_hint = "mountain_car"
+    # Original paper used negative rewards, as this is a variable-length episode
+    # environment with only negative rewards.
+    discrim_net_kwargs = {"gail": {"positive_rewards": False}}
```
AdamGleave, line 145: If we're just trying to match the paper results to verify the implementation then... OK. But I do want to be clear this is a horrible hack. I think how our performance in the fixed-horizon version of MountainCar matches that of other GAIL implementations is likely to be much more informative. Should we not at least have a config for those?

src/imitation/rewards/discrim_net.py
shwang: Oh, I def coded their hack up wrong. Yeah, this is a pretty obvious mistake. But FWIW, we were stuck on GAIL mountain car flatlining at -200 performance for a month, and after adding in the hack, I got 1/5 seeds to reach expert and another seed to not be flatline.

Make sense given that even with an incorrectly coded version of the hack we now have negative living reward.

src/imitation/scripts/config/train_adversarial.py
shwang: Adding a config for seals:MountainCar-v0 seems good.

src/imitation/rewards/discrim_net.py
shwang: What we actually want is this:
image

shwang: OK, I'm super confused now, because looks like our default (before this PR) is equivalent to `self.favor_zero_expert_rewards is True`, which is what openai/imitation claims is better for MountainCar because it has negative rewards?

Yeah, our log sigmoid is negative rewards only.

AdamGleave: I haven't thought about this enough to understand what's going on, but it looks like before we had the reward being `-tf.log_sigmoid(self._disc_logits_gen_is_high)`. Now we have it being `tf.log_sigmoid(self._disc_logits_gen_is_high)` and then you flip the sign if `self._positive_rewards` is set.

So the default before the PR I don't think maps onto either of the GAIL options.

Is it possible we had a sign error this whole time? It'd explain why things work better when you flip the sign! :thinking:

shwang: Ok, now I'm doubly super confused....

How on earth is GAIL doing even passably well if we have the wrong sign. Will be taking a look into this in the paper, the paper's code, and our experiments

shwang: Haven't checked their code throughly yet, but I believe that their logits are "fake is high" whereas our logits are "low is high". Which would mean that before this PR, we were the equivalent of
`rewards_B = -thutil.logsigmoid(-scores_B)`.

shwang: `logsigmoid(-x) ?= log(1 - sigmoid(x))`? In which case we were actually following the else-clause correctly?

I need to brush up on my math...

shwang: I want to reverse the meaning of our logits so that the logits are consistent with the paper and other codebases

AdamGleave: Being consistent sounds good.

shwang:
Closing this because #207 is coming soon

PR #203, opened 2020-06-25
bc: Add logging, l2 loss, entropy weight and modify API

shwang:
Several updates to BCTrainer:

* Rename to BC
* Remove env param for BC.init, instead require obs and action spaces.
* Remove `test_policy` method which used `BC.env` and the policy to collect rollouts and then returns a dict of statistics on those rollouts. After `bc.ipynb` was deleted, nothing in the codebase references this, and this can be done via standard library calls as when working with Adversarial.
* Use formerly unused `optimizer_{cls,kwargs}`.
* Add ent loss, l2 loss and various logs.


src/imitation/algorithms/bc.py
```diff
@@ -116,8 +128,61 @@ def set_expert_dataset(
         assert expert_dataset.size() is not None
         self.expert_dataset = expert_dataset
 
+    def _build_tf_graph(self):
+        with tf.variable_scope("bc_supervised_loss"):
+            with tf.variable_scope("model") as model_scope:
+                self.policy_kwargs = dict(
```
qxcv, line 134: Should these be configurable?

src/imitation/algorithms/bc.py
```diff
@@ -116,8 +128,61 @@ def set_expert_dataset(
...
+                [None], name="ref_acts_ph"
+            )
+            self._entropy = tf.reduce_mean(self.policy.proba_distribution.entropy())
+            self._ent_loss = -self.ent_weight * self._entropy
+            self._neglogp = tf.reduce_mean(
+                self.policy.proba_distribution.neglogp(self._true_acts_ph)
+            )
+
+            _l2_norms = [
+                tf.reduce_mean(tf.norm(w)) for w in model_scope.trainable_variables()
```
qxcv, line 158: Use `tf.nn.l2_loss` (squared L2 norm), not `tf.norm`.

src/imitation/algorithms/bc.py
shwang: This is fixed in a later commit, thanks for pointing this out

src/imitation/algorithms/bc.py
```diff
@@ -132,63 +197,32 @@ def train(
         """
         assert self.batch_size >= 1
         samples_so_far = 0
+        batch_num = 0
         for epoch_num in trange(n_epochs, desc="BC epoch"):
             while samples_so_far < (epoch_num + 1) * self.expert_dataset.size():
```
qxcv, line 202: This is okay, but in future it would be nice to create a common batch iterator wrapper for datasets so that we can simplify this sort of code.

(hold off on this change; if we move to PyTorch then we can probably reuse their dataset abstractions)

shwang:
Thanks for the review, @qxcv. I believe the issues are now fixed.

shwang:
It would be cool to support some Stable Baselines Callbacks to this thing later as a wish list type thing. The `on_epoch_end` thing is not as flexible as I would like.

src/imitation/algorithms/bc.py
```diff
@@ -90,6 +92,17 @@ def __init__(
...
         self.policy_class = policy_class
+        policy_kwargs_defaults = dict(
+            ob_space=self.observation_space,
+            ac_space=self.action_space,
+            n_batch=None,
+            n_env=1,
+            n_steps=1000,
+        )
+        self.policy_kwargs = collections.ChainMap(
+            policy_kwargs or {}, policy_kwargs_defaults
```
qxcv, line 103: That's cool, I didn't realise this existed!

src/imitation/algorithms/bc.py
```diff
@@ -155,7 +161,8 @@ def _build_tf_graph(self):
             )
 
             _l2_norms = [
-                tf.reduce_mean(tf.norm(w)) for w in model_scope.trainable_variables()
+                tf.reduce_mean(tf.nn.l2_loss(w))
+                for w in model_scope.trainable_variables()
             ]
             self._l2_norm = tf.reduce_sum(_l2_norms)
```
qxcv, line 167: ~Maybe replace this with `reduce_mean` for consistency with the `reduce_mean` on line 164?~
**Edit:** actually, it would make more sense to replace the `reduce_mean` on line 164 with a `reduce_sum`. That's at least consistent with the `$\lambda \|\theta\|_2^2$` notation that people use in papers.

shwang: Double reduce sum makes sense

shwang:
OK, these issues should be fixed. Waiting on a Github approval before pushing

src/imitation/algorithms/bc.py
shwang: Ugh, actually I'm thinking maybe we should keep the mean on the inside, but only mean over the batch dimension?

If we use double reduce sum, then the L2 Loss scales with the batch size, which seems wrong, or at least annoying from a Tensorboard and hyperparam tuning perspective.

qxcv: These are just weights, right? There shouldn't be a batch dimension.

shwang: Oh, there isn't a batch index.

shwang: As it turns out, l2_loss returns a scalar anyways, so a reduction is not necessary inside the list comprehension.

PR #204, opened 2020-06-29
buffer: Add truncate_ok and capacity to methods

shwang:
Adds truncate_ok and capacity params to .store() and .from_data()
methods that previously lacked them.

Having truncate_ok and capacity is useful for constructing Buffers
`from_data()` when we want to use a larger capacity than the arrays
in the arguments have.


src/imitation/data/buffer.py
```diff
@@ -65,41 +65,56 @@ def __init__(
...
+    def from_data(
+        cls,
+        data: Dict[str, np.ndarray],
+        capacity: Optional[int] = None,
+        truncate_ok: bool = False,
+    ) -> "Buffer":
+        """Constructs and return a Buffer containing the provided data.
 
-        The returned Buffer is at full capacity and ready for sampling.
+        Shapes and dtypes are automatically inferred.
```
AdamGleave, line 76: Could you add one or two examples showing usages of `capacity` and `truncate_ok`? At first I was really confused why you need both. I think the use cases are:
  - Specify `capacity` and `truncate_ok=True` if you are specifying a smaller capacity than your data.
  - Specify `capacity` and `truncate_ok=False` if you are specifying a larger capacity than you intend your data to fill, and may call `store` subsequently to add more data.
  - Do not specify `capacity` if you want the buffer to be the same size as your data. In this case `truncate_ok` has no effect.

shwang: Added examples, looks like this.
image

src/imitation/data/buffer.py
```diff
@@ -65,41 +65,70 @@ def __init__(
...
                 in their shape, but should agree in the first axis.
+            capacity: The Buffer capacity. If not provided, then this is automatically
+                set to the size of the data, so that the returned Buffer is at full
+                capacity.
+            truncate_ok: Whether to error if `capacity` < the number of samples in
+                `data`. If False, then only store the last `capacity` samples from
+                `data` when overcapacity.
+
+        Examples:
+            `Buffer` with same capacity as arrays in `data`::
```
AdamGleave, line 89:
```suggestion
            In this example, suppose the arrays in `data` are length-1000.
            
            `Buffer` with same capacity as arrays in `data`::
```
I found it a bit ambiguous whether `10000` is larger capacity (assumes how big `data` is). Otherwise, example looks good :)

Issue #185, opened 2020-05-08
Stable format for serializing trajectories


AdamGleave:
Currently we (de)serialize trajectories using `imitation.data.types.{save,load}` using Python pickle. This makes it very sensitive to changes in the format, including even relocating modules.

We could write our own (de)serialization logic to make it more stable, and probably also save space on disk.

AdamGleave:
May actually be easiest to save/load `Transitions`, since that's just a few big NumPy arrays. Then write some code to convert between `Transitions` and `Trajectory`. Should not be hard -- the only thing that's not bijective is `info`, but (a) we usually don't want to serialize this anyway, and (b) we should probably add optional `infos` to `Transitions` anyway.

shwang:
Another reason `Transitions` doesn't map directly onto `List[Trajectory]` is that `Transitions` aren't guaranteed to be in timestep or trajectory order (at least as currently documented), so we can't just split them into different Trajectories at every done=True.

It's true that the way that we currently construct `Transitions` from `flatten_trajectories` preserves timestep and trajectory order.

shwang:
One idea is to save `List[Trajectory]` as a `npz`. `npz` supports Dict (we could make each list index an integer key in the dict form).

```python
{
1: dict(obs=..., acts=..., infos=...),
2: ...,

}
```

AdamGleave:
> Another reason `Transitions` doesn't map directly onto `List[Trajectory]` is that `Transitions` aren't guaranteed to be in timestep or trajectory order (at least as currently documented), so we can't just split them into different Trajectories at every done=True.
> 
> It's true that the way that we currently construct `Transitions` from `flatten_trajectories` preserves timestep and trajectory order.

OK, fair point. This seems easy enough to fix though -- we can just add an "in order" flag to `Transitions`, that is set when produced by `flatten_trajectories`, and refuse to convert if this flag is false. Or just have separate `save()` and `load()` functions for `Trajectory`, which internally map to/from `Transitions` and use that `save()` and `load()`, adding some extra metadata to prevent the user calling `Trajectory.load()` on something produced by `Transitions.save()`.

As a digression, I am still a bit confused why your transitions are out-of-order in MineRL by default. It seems like a desirable thing to do when training many algorithms (epoch order can lead to catastrophic forgetting), but it's odd in serialized data. Can't you just load them all, sort them and then save them in sorted order (then shuffle in-memory if/when desired)?

AdamGleave:
[...]

PR #210, opened 2020-07-11
.data.types.save: Fix docstring

shwang:
Had docstring arguments from `rollout.rollout_and_save`.

Maybe it's a good time to get rid of `types.save` and `types.load`? I'm not sure why we need these guys. At the very least, we don't need the backwards-compatibility code any more, I believe.


AdamGleave:
I think there is value in keeping these in case we change the format in the future.

I'd like to rewrite them in accordance with https://github.com/HumanCompatibleAI/imitation/issues/185 to have a standard serialization format, since Pickle is really fragile.

PR #211, opened 2020-07-11
Add TransitionsMinimal and use it in BC

shwang:
* Adds a subclass of `Transitions`, the dataclass `TransitionsMinimal` with field `obs`, `acts`, and `infos` (`infos` is new to Transitions types!).

  * Note that `infos` has type annotation `np.ndarray` rather than `Optional[Sequence[dict]]` like in `Trajectory`. This is to simplify `Dataset` and `Buffer` class implementations (this way we don't have to add weird check-for-None, allow-any-sequence edges cases when handling the infos field).

* `BC` now accepts `Dataset[TransitionsMinimal]` rather than `DictDataset`. We should probably make an analogous change for the pytorch implementations in #208 later.

* `rollout.flatten_trajectories{,_with_rew}(trajectories, transitions_cls)` were merged into a single function `rollout.flatten_trajectories(trajectories, transitions_cls)` that can also return `TransitionsMinimal` if `transitions_cls=TransitionsMinimal`.

* `TransitionsDictDatasetAdaptor[S]` type annotations now automatically infer the subclass of `TransitionsMinimal` that is returned by `.sample() -> S`, using a bounded TypeVar. (The bounded TypeVar is working now although it didn't before on my local computer when I first contributed `TransitionsDictDataset`. Maybe pytype upgrade fixed this?).

* Renames `imitation.data.dataset` module to `imitation.data.datasets`.


src/imitation/data/rollout.py
```diff
@@ -366,43 +376,57 @@ def mean_return(*args, **kwargs) -> float:
     return rollout_stats(trajectories)["return_mean"]
 
 
+T = TypeVar("T", bound=types.TransitionsMinimal)
+
+
 def flatten_trajectories(
```
AdamGleave, line 382: I don't get what we're trying to do here.

`TransitionsWithRew` is a `Transitions` is a `TransitionsMinimal` -- so why do we need `transitions_cls`? It's fine to return a `Transitions` and use it as a `TransitionsMinimal`. I guess it might take up some extra space?

I get the desire to not have both `flatten_trajectories` and `flatten_trajectories_with_rew`, but combining them doesn't seem to reduce code complexity overall, and it loses some typing guarantees (since we now have to do a run-time check that "rews" is present).

I may be missing something but right now I'd be in favor of reverting this change.

src/imitation/data/types.py
```diff
@@ -21,8 +21,8 @@ class Trajectory:
     acts: np.ndarray
     """Actions, shape (trajectory_len, ) + action_shape."""
 
-    infos: Optional[Sequence[dict]]
-    """A sequence of info dicts, length trajectory_len."""
+    infos: Optional[np.ndarray]
```
AdamGleave, line 24: `Sequence[dict]` seems like a much more natural type than `np.ndarray`, although admittedly `np.ndarray` can store `dict` objects. What is the advantage of `ndarray`?

src/imitation/data/types.py
```diff
@@ -69,26 +69,53 @@ def __post_init__(self):
...
-    Invariant: obs.dtype == next_obs.dtype.
+    by the agent when choosing action `acts[i]`. `obs[i]` is not required to
+    be from the timestep preceding `obs[i+1]`.
     """
 
     acts: np.ndarray
     """Actions. Shape: (batch_size,) + action_shape."""
 
+    infos: np.ndarray
+    """Optional array of info dicts. Shape: (batch_size,)."""
```
AdamGleave, line 92: If it is optional, then why is it not `Optional[np.ndarray]`?

Yes, this would require changing the `Dataset` implementations to handle optional elements in order to type check, but I think lying about the types is just going to shift the error to run time.

src/imitation/data/types.py
```diff
@@ -69,26 +69,53 @@ def __post_init__(self):
...
+
+        if self.infos is not None and len(self.infos) != len(self.obs):
+            raise ValueError(
+                "obs and infos must have same number of timesteps: "
+                f"{len(self.obs)} != {len(self.infos)}"
+            )
+
+
+@dataclasses.dataclass(frozen=True)
+class Transitions(TransitionsMinimal):
```
AdamGleave, line 116: I am a little weirded out that `TransitionsMinimal` does not contain `dones`, but I guess it makes sense that `next_obs` and `dones` would go together (since it's something you get *after* a transition).

tests/test_buffer.py
```diff
@@ -97,38 +97,49 @@ def test_replay_buffer(capacity, chunk_len, obs_shape, act_shape, dtype):
 
         dones = np.arange(i, i + chunk_len, dtype=np.int32) % 2
```
AdamGleave, line 98: If `infos` is optional, we should have a test case for it being `None`.

tests/test_data.py
```diff
@@ -50,21 +50,33 @@ def trajectory_rew(trajectory: types.Trajectory) -> types.TrajectoryWithRew:
 
 
 @pytest.fixture
-def transitions(
+def transitions_min(
     obs_space: gym.Space, act_space: gym.Space, length: int
+) -> types.TransitionsMinimal:
+    obs = np.array([obs_space.sample() for _ in range(length)])
+    acts = np.array([act_space.sample() for _ in range(length)])
+    infos = np.array([{}] * length)
```
AdamGleave, line 58: Again, if this is optional, we should have a test case for that.

src/imitation/data/types.py
AdamGleave: I think maybe "Optional" was just not meant to be there, since the rest of the code seems to assume it's always present.

In that case, what is the repercussion of making this mandatory? E.g. does it significantly inflate memory consumption? I think probably not in RAM (but should be tested), but will inflate serialized size a lot, so should we have some logic to optionally drop it (at least if it's all empty dicts) in `load`/`save`?

tests/test_data.py
```diff
@@ -128,23 +140,36 @@ def test_invalid_trajectories(
 
     def test_valid_transitions(
         self,
+        transitions_min: types.Transitions,
```
AdamGleave, line 143:
```suggestion
        transitions_min: types.TransitionsMinimal,
```

tests/test_data.py
```diff
@@ -128,23 +140,36 @@ def test_invalid_trajectories(
...
     ) -> None:
         """Checks trajectories can be created for a variety of lengths and spaces."""
+        assert len(transitions_min) == length
         assert len(transitions) == length
         assert len(transitions_rew) == length
 
     def test_invalid_transitions(
-        self, transitions: types.Transitions, transitions_rew: types.TransitionsWithRew,
+        self,
+        transitions_min: types.Transitions,
```
AdamGleave, line 155:
```suggestion
        transitions_min: types.TransitionsMinimal,
```

tests/test_data.py
```diff
@@ -390,7 +411,10 @@ def test_sample_sizes_and_types(
                 sample = ds.sample(n_samples)
                 assert isinstance(sample, type(trans))
                 for k, v in dataclasses.asdict(sample).items():
-                    trans_v: np.ndarray = trans_dict[k]
+                    if k == "infos":
```
AdamGleave, line 414: Could be simplified:

```python
trans_v = trans_dict[k]
if k == "info":
  trans_v = np.array(trans_v)
assert isinstance(trans_v, np.ndarray)
```

src/imitation/data/datasets.py
```diff
@@ -133,32 +133,35 @@ def sample(self, n_samples: int) -> Dict[str, np.ndarray]:
...
                 internally as a `dict` of copied arrays. Note that `sample` will return
-                the same type of `Transitions` as the type of this arg. In other words,
-                if `transitions` is an instance of `TransitionsWithRew`, then
-                `.sample()` also returns `TransitionsWithRew`.
+                the same subclass type of `TransitionsMinimal` as the type of this arg.
             dict_dataset_cls: `DictDataset` class to be adapted.
             dict_dataset_kwargs: Optional kwargs for initializing `DictDataset` class.
         """
         data_map: Dict[str, np.ndarray] = dataclasses.asdict(transitions)
+        data_map["infos"] = np.array(data_map["infos"])
```
AdamGleave, line 156: Why do we need this? Isn't it already an array?

src/imitation/data/rollout.py
shwang: I think I built this because I wanted a way to generate `TransitionsMinimal` from `List[Trajectory]`. Agreed that `Transitions` can just be substituted for `TransitionsMinimal`, so I'll probably revert this.

One way we could keep all static type checking guarantees with this merged function is using `@typing.overload`.

src/imitation/data/types.py
shwang: I forgot to update the docstring after changing my mind to use `infos: np.array` rather than `Optional[np.ndarray]`.

shwang: I was assuming that we would continuing using our default save format of Trajectories (I don't think I've ever saved Transitions), so I didn't consider the disk space consequences.

I could do a quick RAM benchmark.

src/imitation/data/types.py
shwang: I didn't try to come up with a principled reason to split up the fields between classes.

I stuck to what `BC` needed, and then added in `infos` afterwards because I'd need that for a MineRL-variant of `FrameStackTransformation` later on, and maybe other wrappers / datasets would want to stick metadata into all types of transition.

src/imitation/data/datasets.py
shwang: Unnecesary now, thanks for catching. This was leftover back from when `infos` could be a sequence, and I'd get an error from `infos[perm_idx]` inside `EpochOrderDictDataset`.

tests/test_data.py
shwang: I did this if-else structure to avoid a pytype error. I'll try again to see if it's okay now.

src/imitation/data/types.py
AdamGleave: Fair point, I guess it's OK if we just save trajectories. Although note that `TransitionsMinimal` is not convertable into `Trajectory`.

Would be interested in the RAM repercussion here. I think an empty object is surprisingly large (compared to e.g. an int).

src/imitation/data/types.py
shwang: There's logic in `EpochOrderDataset`, `Buffer`, and `rollout` that assumes ndarray. For example, `EpochOrderDataset` use numpy indexing to select a random sample: `self.data_map["infos"][perm_idx]`. `Buffer` saves its data in a ndarray, returns an ndarray, and will ask for the dtype of `infos` when allocating its underlying arrays used for ring buffers.

We could support `infos: Sequence[dict]` by adding edge cases in these files but it seems easier to just use `np.ndarray` like all the other fields in `Transitions`.

tests/test_data.py
shwang: Not optional

src/imitation/data/rollout.py
shwang: Reverted.

src/imitation/data/types.py
shwang: Saving list of empty dicts in various formats:

https://gist.github.com/shwang/a925b013b9f813c319f08e2fec344bd2
```
+------------------+---------+-----------------+
|   save_format    | n_dicts | disk usage (KB) |
+------------------+---------+-----------------+
|      pickle      |    0    |        0        |
|      pickle      |   300   |        0        |
|      pickle      |   700   |        0        |
|      pickle      |   1000  |        5        |
|      pickle      | 1000000 |       5860      |
|      pickle      | 5000000 |      29305      |
|      numpy       |    0    |        0        |
|      numpy       |   300   |        1        |
|      numpy       |   700   |        3        |
|      numpy       |   1000  |        5        |
|      numpy       | 1000000 |       5861      |
|      numpy       | 5000000 |      29306      |
| numpy_compressed |    0    |        0        |
| numpy_compressed |   300   |        0        |
| numpy_compressed |   700   |        1        |
| numpy_compressed |   1000  |        2        |
| numpy_compressed | 1000000 |       1483      |
| numpy_compressed | 5000000 |       7413      |
+------------------+---------+-----------------+
```

AdamGleave: I'm confused by this:

```
5000000/(29305*1024)
0.16662045725985328
```

Does an empty dict really take up less than a byte?

src/imitation/data/types.py
AdamGleave: My understanding was that `EpochOrderDataset` and `Buffer` both operate on `Dict[str, np.ndarray]` and we use `TransitionsDictDatasetAdaptor` to convert them to/from `types.Transitions`. So wouldn't the conversion code just need to lie in `TransitionsDictDatasetAdaptor`? Perhaps we think it's still not worth adding that code, but this feels quite different than needing to make changes in multiple classes.

What part of `rollout` depends on this? I think maybe `finish_trajectory` in `TrajectoryAccumulator` would need to be changed.

src/imitation/data/types.py
```diff
@@ -69,26 +69,53 @@ def __post_init__(self):
...
     """
 
     obs: np.ndarray
     """
     Previous observations. Shape: (batch_size, ) + observation_shape.
 
     The i'th observation `obs[i]` in this array is the observation seen
-    by the agent when choosing action `acts[i]`.
-
-    Invariant: obs.dtype == next_obs.dtype.
```
AdamGleave, line 86: Shouldn't this invariant be documented somewhere (in `Transitions` now, I guess?)

src/imitation/data/types.py
AdamGleave: I guess I'm OK with it being an `np.ndarray` since there don't seem to be any major downsides: see e.g. https://stackoverflow.com/questions/26768013/what-are-the-benefits-drawbacks-of-a-list-of-lists-compared-to-a-numpy-array-o https://stackoverflow.com/questions/26767694/what-are-the-benefits-drawbacks-of-a-list-of-lists-compared-to-a-numpy-array-o

But I'd still like to understand better what the problem with `Sequence[dict]` is here.

tests/test_data.py
```diff
@@ -128,23 +140,36 @@ def test_invalid_trajectories(
...
+                ValueError, match=r"obs and acts must have same number of timesteps:.*"
             ):
-                dataclasses.replace(trans, obs=trans.obs[:-1])
+                dataclasses.replace(trans, acts=trans.acts[:-1])
+            with pytest.raises(
+                ValueError, match=r"obs and infos must have same number of timesteps:.*"
+            ):
+                dataclasses.replace(trans, infos=[{}] * (length - 1))
+
+        """Checks input validation catches space and dtype related errors."""
```
AdamGleave, line 171: Shouldn't this docstring be at the top of the function?

src/imitation/data/types.py
shwang: The number you calculated is the inverse quantity, the number of dicts per byte.

```
+------------------+---------+-----------------+--------------------+
|   save_format    | n_dicts | disk usage (KB) |   bytes per dict   |
+------------------+---------+-----------------+--------------------+
|      pickle      |    0    |        0        |        nan         |
|      pickle      |   300   |        1        | 3.4766666666666666 |
|      pickle      |   700   |        3        | 4.918571428571428  |
|      pickle      |   1000  |        5        |       5.243        |
|      pickle      | 1000000 |       5860      |      6.001241      |
|      pickle      | 5000000 |      29305      |     6.0018482      |
|      numpy       |    0    |        0        |        nan         |
|      numpy       |   300   |        1        | 4.986666666666666  |
|      numpy       |   700   |        3        | 5.565714285714286  |
|      numpy       |   1000  |        5        |       5.696        |
|      numpy       | 1000000 |       5861      |      6.001696      |
|      numpy       | 5000000 |      29306      |     6.0019392      |
| numpy_compressed |    0    |        0        |        nan         |
| numpy_compressed |   300   |        0        | 3.0633333333333335 |
| numpy_compressed |   700   |        1        | 2.2885714285714287 |
| numpy_compressed |   1000  |        2        |       2.106        |
| numpy_compressed | 1000000 |       1483      |      1.518984      |
| numpy_compressed | 5000000 |       7413      |     1.5183624      |
+------------------+---------+-----------------+--------------------+

```

src/imitation/data/types.py
shwang: Agree that `TransitionsDictDatasetAdaptor` can handle the `EpochOrderDataset` case.

Places to add info conversions should we stick with Sequence[dict]:
  * `ReplayBuffer` accepts `Transitions` directly. (So need to convert infos to `array(dtype=object)`.)
  *  rollout: `np.concat` and `np.stack` in `finish_trajectory` and `flatten_trajectories_*`.
 
I think I had a cached belief that conversion is more onerous than it actually is because I was thinking about doing this with `Optional[Sequence[dict]]` instead, so I'd have to convert from None to array of empty dicts in several places.

src/imitation/data/types.py
AdamGleave: Oops, thanks for catching that, yes those results make more sense :) I find it weird the overhead increases superlinearly with the number of dicts, although it seems to asymptote at 6 bytes at least? I guess that's OK.

Is it reasonably fast, too? I seem to remember having some performance problems when pickling large numbers of Python objects, but they may have fixed this in recent versions.

src/imitation/data/types.py
AdamGleave: Thanks for the examples. I think I'm OK with either option now so you can make the final call.

src/imitation/data/types.py
shwang: Oh that's interesting. numpy loading is way faster than pickle loading, even when compressed.

https://gist.github.com/shwang/ae6c96170548929a21abbab939dc3ae4
```
 % python scrap2.py
+------------------+---------+-----------------+--------------------+------------------+------------------+
|   save_format    | n_dicts | disk usage (KB) |    bytes/n_dict    |  serialize_time  | deserialize_time |
+------------------+---------+-----------------+--------------------+------------------+------------------+
|      pickle      |    0    |        0        |        nan         | [7.51018524e-05] | [9.10758972e-05] |
|      pickle      |   300   |        0        |        0.0         | [7.00950623e-05] |   [0.00020194]   |
|      pickle      |   700   |        0        |        0.0         |   [0.00016189]   |    [0.000247]    |
|      pickle      |   1000  |        5        |       5.243        |   [0.00018287]   |   [0.00026011]   |
|      pickle      | 1000000 |       5860      |      6.001241      |   [0.44883609]   |   [0.41594386]   |
|      pickle      | 5000000 |      29305      |     6.0018482      |   [2.64626718]   |   [3.25023007]   |
|      numpy       |    0    |        0        |        nan         |   [0.0172019]    |   [0.00144839]   |
|      numpy       |   300   |        1        | 4.986666666666666  |   [0.00096893]   |   [0.00035596]   |
|      numpy       |   700   |        3        | 5.565714285714286  |   [0.00078297]   |   [0.00015378]   |
|      numpy       |   1000  |        5        |       5.696        |   [0.00078106]   |   [0.00014973]   |
|      numpy       | 1000000 |       5861      |      6.001696      |   [0.69544315]   |   [0.00032496]   |
|      numpy       | 5000000 |      29306      |     6.0019392      |   [3.1576438]    |   [0.00021815]   |
| numpy_compressed |    0    |        0        |        nan         |   [0.00091314]   |   [0.00016975]   |
| numpy_compressed |   300   |        0        | 3.0633333333333335 |   [0.00230289]   |   [0.00061703]   |
| numpy_compressed |   700   |        1        | 2.2885714285714287 |   [0.00221896]   |   [0.00024486]   |
| numpy_compressed |   1000  |        2        |       2.106        |   [0.00181007]   |   [0.00020194]   |
| numpy_compressed | 1000000 |       1483      |      1.518984      |   [1.39837718]   |   [0.00025392]   |
| numpy_compressed | 5000000 |       7413      |     1.5183624      |   [7.0082581]    |   [0.00024199]   |
+------------------+---------+-----------------+--------------------+------------------+------------------+
```

AdamGleave: Wow that's quite dramatic.

I'm a bit confused why the deserialize time for NumPy is almost *constant* -- surely we should start seeing some variable costs show up with `5000000` dicts? I assume costs are more linear if we stored e.g. a random array of numbers?

It's possible NumPy has some very efficient representation for repeated references to the same object. But it's also possible it has some hack that's screwing up the benchmark (e.g. notice if you're loading in something that was just saved?)

src/imitation/data/types.py
shwang: Accidental deletion. Thanks

src/imitation/data/types.py
shwang: Caching seems really likely. I'll try moving the file first before loading

shwang: Ok, making a lot more sense with caching. But still much much faster than pure pickle. Tried cloudpickle too, which is much slower than pickle.
https://gist.github.com/shwang/ae6c96170548929a21abbab939dc3ae4/edit
```
 % python scrap2.py
+------------------+---------+-----------------+--------------------+------------------+------------------+
|   save_format    | n_dicts | disk usage (KB) |    bytes/n_dict    |  serialize_time  | deserialize_time |
+------------------+---------+-----------------+--------------------+------------------+------------------+
|      pickle      |    0    |        0        |        nan         | [7.89165497e-05] | [9.08374786e-05] |
|      pickle      |   300   |        1        | 3.4766666666666666 | [8.48770142e-05] |   [0.00016999]   |
|      pickle      |   700   |        3        | 4.918571428571428  |   [0.00016689]   |   [0.00022817]   |
|      pickle      |   1000  |        5        |       5.243        |   [0.00022912]   |   [0.00041413]   |
|      pickle      | 1000000 |       5860      |      6.001241      |   [0.40788722]   |   [0.36054015]   |
|      pickle      | 5000000 |      29305      |     6.0018482      |   [2.32574725]   |    [3.259233]    |
|   cloudpickle    |    0    |        0        |        nan         |   [0.00019193]   | [7.82012939e-05] |
|   cloudpickle    |   300   |        0        | 2.0533333333333332 |   [0.00175905]   |   [0.00012112]   |
|   cloudpickle    |   700   |        1        | 2.0228571428571427 |   [0.00449204]   |   [0.00019622]   |
|   cloudpickle    |   1000  |        1        |       2.016        |   [0.00549197]   |   [0.00018096]   |
|   cloudpickle    | 1000000 |       1955      |      2.002284      |   [6.52187085]   |   [0.31458831]   |
|   cloudpickle    | 5000000 |       9776      |     2.0022764      |  [35.75733614]   |   [2.05795884]   |
|      numpy       |    0    |        0        |        nan         |   [0.03340602]   |   [0.00164199]   |
|      numpy       |   300   |        1        | 4.986666666666666  |   [0.00078702]   |   [0.00018001]   |
|      numpy       |   700   |        3        | 5.565714285714286  |     [0.0007]     |   [0.00017262]   |
|      numpy       |   1000  |        5        |       5.696        |   [0.00093722]   |   [0.00018883]   |
|      numpy       | 1000000 |       5861      |      6.001696      |   [0.52698088]   |   [0.00025892]   |
|      numpy       | 5000000 |      29306      |     6.0019392      |   [2.89590693]   |   [0.00024486]   |
| numpy_compressed |    0    |        0        |        nan         |   [0.0009737]    |   [0.00021696]   |
| numpy_compressed |   300   |        0        | 3.0633333333333335 |   [0.00159812]   |   [0.00035095]   |
| numpy_compressed |   700   |        1        | 2.2885714285714287 |   [0.00205636]   |   [0.00030589]   |
| numpy_compressed |   1000  |        2        |       2.106        |   [0.00208092]   |   [0.00024891]   |
| numpy_compressed | 1000000 |       1483      |      1.518984      |   [1.37592196]   |   [0.00024319]   |
| numpy_compressed | 5000000 |       7413      |     1.5183624      |   [7.09819221]   |   [0.00024414]   |
+------------------+---------+-----------------+--------------------+------------------+------------------+

```

shwang: Another interesting thing I noticed is that `np.load` has a `mmap_mode` param which can be super useful for huge arrays.

shwang: Oh wait, deserialize time is still the same. I was looking at the wrong column.

src/imitation/data/types.py
AdamGleave: `TransitionsMinimal` doesn't have `next_obs`, so I think this belongs in `Transitions` now instead

AdamGleave:
What's the plan for this PR? Are we merging or waiting until after the PyTorch port?

shwang:
Plan on wrapping this up this evening

PR #223, opened 2020-08-28
setup.py: Bump Stable-Baselines version

shwang:


shwang:
Failing because something funny going on with (1) `isort` and (2) Ray webgui keyword arguments.
Should be easy to fix

PR #225, opened 2020-08-28
Fix broken CI

shwang:
Fix CI errors due to changes in latest versions of isort, black, stable-baselines, and Ray.


AdamGleave:
> To help debug discrepancies between local and CI lint.

Printing versions seems a good idea in general, but for the particular use case of debugging note you can also SSH into CircleCI jobs.

PR #227, opened 2020-09-30
train_adversarial: Use correct eval environment

shwang:
Final evaluation environment used to be unnormalized, leading to
Sacred reporting worse final performance than actual.

### DIFF
--- a/src/imitation/scripts/train_adversarial.py
+++ b/src/imitation/scripts/train_adversarial.py
@@ -201,7 +201,7 @@ def train(
     results = {}
     sample_until_eval = rollout.min_episodes(n_episodes_eval)
     trajs = rollout.generate_trajectories(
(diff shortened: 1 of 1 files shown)


qxcv:
I'm a bit confused by this. I assume one of the environments has the ground truth reward ovewritten with the synthetic learnt reward. Which is which, though?

qxcv:
(whichever reward it is, why does it make sense to normalise it during evaluation?)

shwang:
The problem is that the observations aren’t normalized as the policy
expects during evaluation.

Whether the rewards are normalized is less relevant — monitor_rewards will
track the original rewards (although as we have discussed before this
notation for the statistics is pretty confusing).

qxcv (approved):
(after offline conversation: `venv_train_norm` uses shaped synthetic reward and an observation/reward/etc. normalisation wrapper)

Issue #215, opened 2020-07-22
Basic script usage (the example in README.md) fails


ejmichaud:
I am in the `port_clones_to_pytorch` branch: #208 
The following commands, as listed in the README are failing for me:
```
# Train PPO2 agent on cartpole and collect expert demonstrations
python -m imitation.scripts.expert_demos with cartpole
# Train AIRL on from demonstrations
python -m imitation.scripts.train_adversarial with cartpole airl
```
I have followed the installation instructions exactly and am running on `Ubuntu 18.04.4 LTS`. 
The first command (`expert demos`) displays the following error after training the policy:
```
INFO - root - Rollout stats: {'n_traj': 14, 'return_min': 203.0, 'return_mean': 322.64285714285717, 'return_std': 115.8352322068211, 'return_max': 500.0, 'len_min': 203, 'len_mean': 322.64285714285717, 'len_std': 115.8352322068211, 'len_max': 500}
INFO - root - Dumped demonstrations to output/expert_demos/CartPole-v1/20200722_155156_022b0f/rollouts/final.pkl.
INFO - root - Saved policy to output/expert_demos/CartPole-v1/20200722_155156_022b0f/policies/final
INFO - expert_demos - Result: {'n_traj': 57, 'return_min': 7.388836879311779, 'return_mean': 15.001208959872894, 'return_std': 3.078998671602082, 'return_max': 17.673824873098567, 'len_min': 215, 'len_mean': 441.29824561403507, 'len_std': 91.40174763431135, 'len_max': 500, 'monitor_return_min': 215.0, 'monitor_return_mean': 441.29824561403507, 'monitor_return_std': 91.40174763431135, 'monitor_return_max': 500.0}
INFO - expert_demos - Completed after 0:02:01
ERROR - expert_demos - Failed after 0:02:02!
Traceback (most recent calls WITHOUT Sacred internals):
  File "/home/eric/Code/miniconda3/envs/imitation/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/home/eric/Code/miniconda3/envs/imitation/lib/python3.7/subprocess.py", line 1019, in wait
    return self._wait(timeout=timeout)
  File "/home/eric/Code/miniconda3/envs/imitation/lib/python3.7/subprocess.py", line 1645, in _wait
    raise TimeoutExpired(self.args, timeout)
[...]

PR #228, opened 2020-10-01
Script fixes

### LINKED ISSUE #218: scripts/expert_demos.py fails if `rollout_save_path` doesn't contain `/`.
Should be a straightforward fix:

```
$ python -m imitation.scripts.expert_demos rollouts_from_policy --capture=sys with env_name=CartPole-v0 log_root="imitation_rollouts" policy_path=/Users/cody/Code/data/expert_models/cartpole_0/policies/final/ rollout_save_path="data/cartpole_rollout.pkl" rollout_save_n_episodes=15 rollout_save_n_timesteps=None
```

```
ERROR - expert_demos - Failed after 0:00:07!
Traceback (most recent calls WITHOUT Sacred internals):
  File "/Users/cody/.pyenv/versions/3.7.0/lib/python3.7/contextlib.py", line 74, in inner
    return func(*args, **kwds)
  File "/Users/cody/Code/imitation/src/imitation/scripts/expert_demos.py", line 233, in rollouts_from_policy
    rollout.rollout_and_save(rollout_save_path, policy, venv, sample_until)
  File "/Users/cody/Code/imitation/src/imitation/data/rollout.py", line 481, in rollout_and_save
    types.save(path, trajs)
  File "/Users/cody/Code/imitation/src/imitation/data/types.py", line 213, in save
    os.makedirs(os.path.dirname(path), exist_ok=True)
  File "/Users/cody/.pyenv/versions/3.7.0/lib/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
FileNotFoundError: [Errno 2] No such file or directory: ''
```

shwang:
Change assert statements in Sacred scripts and config files to
ValueErrors in Sacred scripts. Fix expert_demos bug where
`rollout_path` errored if argument did not contain "/".

Addresses part of #215. Resolves #218.

### DIFF
--- a/src/imitation/scripts/parallel.py
+++ b/src/imitation/scripts/parallel.py
@@ -65,12 +65,17 @@ def parallel(
         upload_dir: `upload_dir` argument to `ray.tune.run()`.
     """
     # Basic validation for config options before we enter parallel jobs.
-    for name in base_named_configs:
-        assert isinstance(name, str)
-    for k in base_config_updates:
-        assert isinstance(k, str)
-    assert isinstance(search_space["named_configs"], collections.abc.Sequence)
-    assert isinstance(search_space["config_updates"], collections.abc.Mapping)
+    if not isinstance(base_named_configs, collections.abc.Sequence):
+        raise ValueError("base_named_configs must be a Sequence")
+
+    if not isinstance(base_config_updates, collections.abc.Mapping):
+        raise ValueError("base_config_updates must be a Mapping")
+
+    if not isinstance(search_space["named_configs"], collections.abc.Sequence):
+        raise ValueError('search_space["named_configs"] must be a Sequence')
+
+    if not isinstance(search_space["config_updates"], collections.abc.Mapping):
+        raise ValueError('search_space["config_updates"] must be a Mapping')
 
     # Convert Sacred's ReadOnlyList to List because not picklable.
     base_named_configs = list(base_named_configs)
--- a/src/imitation/scripts/parallel.py
+++ b/src/imitation/scripts/parallel.py
@@ -66,16 +66,16 @@ def parallel(
     """
     # Basic validation for config options before we enter parallel jobs.
     if not isinstance(base_named_configs, collections.abc.Sequence):
-        raise ValueError("base_named_configs must be a Sequence")
+        raise TypeError("base_named_configs must be a Sequence")
 
     if not isinstance(base_config_updates, collections.abc.Mapping):
-        raise ValueError("base_config_updates must be a Mapping")
+        raise TypeError("base_config_updates must be a Mapping")
 
     if not isinstance(search_space["named_configs"], collections.abc.Sequence):
-        raise ValueError('search_space["named_configs"] must be a Sequence')
+        raise TypeError('search_space["named_configs"] must be a Sequence')
 
     if not isinstance(search_space["config_updates"], collections.abc.Mapping):
-        raise ValueError('search_space["config_updates"] must be a Mapping')
+        raise TypeError('search_space["config_updates"] must be a Mapping')
 
     # Convert Sacred's ReadOnlyList to List because not picklable.
     base_named_configs = list(base_named_configs)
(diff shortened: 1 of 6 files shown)


src/imitation/scripts/parallel.py
```diff
@@ -65,12 +65,17 @@ def parallel(
...
     """
     # Basic validation for config options before we enter parallel jobs.
-    for name in base_named_configs:
-        assert isinstance(name, str)
-    for k in base_config_updates:
-        assert isinstance(k, str)
-    assert isinstance(search_space["named_configs"], collections.abc.Sequence)
-    assert isinstance(search_space["config_updates"], collections.abc.Mapping)
+    if not isinstance(base_named_configs, collections.abc.Sequence):
+        raise ValueError("base_named_configs must be a Sequence")
```
qxcv, line 69: Pedantic: these should be `TypeError`s.

src/imitation/scripts/train_adversarial.py
```diff
@@ -117,20 +121,43 @@ def train(
...
+    if gen_batch_size % num_vec != 0:
+        raise ValueError(
+            f"num_vec={num_vec} must evenly divide gen_batch_size={gen_batch_size}."
+        )
 
-    logging.info("Logging to %s", log_dir)
-    logger.configure(log_dir, ["tensorboard", "stdout"])
-    os.makedirs(log_dir, exist_ok=True)
-    sacred_util.build_sacred_symlink(log_dir, _run)
+    allowed_keys = {"shared", "gail", "airl"}
```
qxcv, line 129: Why do this here instead of letting Python error out later?

src/imitation/scripts/train_adversarial.py
```diff
@@ -140,26 +167,20 @@ def train(
...
-    # TODO(shwang): Let's get rid of init_rl later on?
-    # It's really just a stub function now.
-
     # if init_tensorboard:
     #     tensorboard_log = osp.join(log_dir, "sb_tb")
     # else:
     #     tensorboard_log = None
 
+    # TODO(shwang): Let's get rid of init_rl later on?
+    # It's really just a stub function now.
```
qxcv, line 176: AFAICT the only reason why we still have `init_rl()` is because it has a hack to get around SB3 issue #109. Maybe merge this comment with my comment below to make it clear that we need to do: (fix 109) -> (remove init_rl)?

tests/test_scripts.py
```diff
@@ -136,18 +137,51 @@ def test_train_adversarial(tmpdir):
...
-            named_configs=named_configs,
-            config_updates=config_updates,
+            named_configs=base_named_configs,
+            config_updates=base_config_updates.new_child(dict(algorithm="BAD_VALUE")),
+        )
+
+    with pytest.raises(ValueError, match=".*BAD_VALUE.*"):
+        train_adversarial.train_ex.run(
+            named_configs=base_named_configs,
+            config_updates=base_config_updates.new_child(
```
qxcv, line 159: This is a cool idiom!

qxcv (commented):
A few minor comments, otherwise good.

src/imitation/scripts/train_adversarial.py
shwang: Fail fast before parallelization via Ray, where error logs become super annoying.

qxcv: Okay, sounds reasonable.

src/imitation/scripts/train_adversarial.py
shwang: There seem to be two separate problems here, only one of which is related to #109.

The first is that Sacred transformed dict into `ReadOnlyDict` which errors on edit within Stable Baselines.

The second is that Stable Baselines can reconfigure our custom logger leading to problems (not immediately clear to me which problem would occur, maybe it would just overwrite our logger).
image

shwang: It doesn't seem to me that init_rl contains any hack to mitigate the first problem with ReadOnlyDict, unless I'm mistaken that the first and second problem are mostly unrelated.

qxcv: I was referring to issue 109 in the SB3 repo, not in this repo.

Issue #79, opened 2019-08-14
Consistent interface for algorithms & single training script


AdamGleave:
Currently we don't have any CLI script for behavioral cloning or the density baseline. I envisage this codebase as being particularly useful in being able to rapidly benchmark against a wide variety of algorithms. For this to be possible, we need to have (as far as is possible) a consistent interface between different imitation learning algorithms, similarly to how Stable Baselines has a similar interface for all RL algorithms implemented. Ideally we'd then also have `imitation.scripts.train` work for all of them, although there'll clearly need to be som algorithm-specific configs.

shwang:
![](https://github.trello.services/images/mini-trello-icon.png) [Consistent interface for algorithms & single training script](https://trello.com/c/rJAUb8aD/5-consistent-interface-for-algorithms-single-training-script)

shwang:
We now have a SB3-like `.train(total_timesteps)` interface for GAIL, AIRL, DAgger, and BC. Also we have Sacred scripts for each of these algorithms, though they are not unified into a single `train`.

AdamGleave:
We've made a lot of progress since this issue was first opened but there's still room for improvement so I'm leaving this open for now.

ernestum:
Right now we can train:

Algo | script| comments
-------|---------|--------------
`DAgger`, `BC` | `train_imitation` | BC is a "hidden feature" by [setting use_dagger=False](https://github.com/HumanCompatibleAI/imitation/blob/64b0a7359c96242c12e9a48cc643f4e4be80599c/src/imitation/scripts/train_imitation.py#L162)
`MCE IRL` | - | do we need a script for this?
`AIRL`, `GAIL` | `train_adversarial` | They being thrown together into one script results in lots of weird constructs such as an [`algorithm_specific` configuration](https://github.com/HumanCompatibleAI/imitation/blob/64b0a7359c96242c12e9a48cc643f4e4be80599c/src/imitation/scripts/config/train_adversarial.py#L33) that are merged with the config using [dubious hacks](https://github.com/HumanCompatibleAI/imitation/blob/64b0a7359c96242c12e9a48cc643f4e4be80599c/src/imitation/scripts/train_adversarial.py#L38) and [configuration being modified during runtime](https://github.com/HumanCompatibleAI/imitation/blob/64b0a7359c96242c12e9a48cc643f4e4be80599c/src/imitation/scripts/train_adversarial.py#L145).
[...]

PR #231, opened 2020-10-06
GAIL: Use DataLoader interface for loading expert demonstrations

shwang:
Part 2 of 2 continuing from #224.

# Changes:
## AdversarialTrainer:
* Remove `disc_batch_size` and `disc_minibatch_size` in favor of `expert_batch_size` and `n_disc_updates_per_round` parameter.
  * `disc_batch_size` used to describe number of discriminator transitions to load from the expert buffer, times 2. Its replacement, `expert_batch_size`, describes the number of transitions that each batch from the expert DataLoader must contain.
  * `disc_mini_batch_size`, inspired by a similar parameter from Stable Baselines' `PPO2`, used to determine the number of discriminator updates that would occur every "round", specifically `disc_batch_size // disc_minibatch_size`. Its replacement, `n_disc_updates_per_round` explicitly determines this number of updates.
    * [x] ~~`n_disc_updates_per_round` is proposed but is not implemented yet.~~
* Merge `train_disc` and `train_disc_step` into a single method `train_disc` that performs a single discriminator update.
* `train_disc` accepts optional expert and generator samples as dict, not `Transitions`.  (To match DataLoader interface, which yields dict)

## Dataset:
* Completely deleted in favor of `DataLoader` interface. Bye!

## util:
* Add `endless_iter` utility for endlessly loading from `DataLoader.__iter__` without StopIteration at the end of every data "epoch".


src/imitation/util/util.py
```diff
@@ -135,3 +135,28 @@ def helper(obj):
         return obj
 
     return helper
+
+
+def endless_iter(iterable: Iterable) -> Generator:
```
qxcv, line 140: IIRC this is `itertools.chain.from_iterable(itertools.repeat(iterable))`.

shwang: Oh nice. Although not quite the same in that `endless_iter` will sanity check against repeating an iterator (which would lead to hanging).

shwang:
[DataLoader interface](https://trello.com/c/2XqpcJUh/6-dataloader-interface)

shwang:
Probably want to rename `n_discrim_steps_per_turn` to `..._per_round`. Let me know if you've got a better name for this parameter.

shwang:
This PR should be ready for review now. Thanks in advance.

I think that @AdamGleave might not be familiar with the intricacies of pytorch DataLoaders yet, but he'll have some thoughts on the new `n_discrim_steps_per_round` parameter or on the overall usability of the updated API.

shwang:
The current plan for merging is to combine this branch with #224 into a mega-DataLoader-migration branch before merging into `master`.

src/imitation/algorithms/adversarial.py
```diff
@@ -305,29 +312,31 @@ def train(
...
         sampled from the environment to exceed `total_timesteps`.
 
         Params:
           total_timesteps: An upper bound on the number of transitions to sample
-            from the environment during training.
+              from the environment during training.
           callback: A function called at the end of every epoch which takes in a
-            single argument, the epoch number. Epoch numbers are in
-            `range(total_timesteps // self.gen_batch_size)`.
+              single argument, the epoch number. Epoch numbers are in
```
AdamGleave, line 325: We're still calling them epochs here...

src/imitation/algorithms/adversarial.py
```diff
@@ -305,29 +312,31 @@ def train(
...
-        n_epochs = total_timesteps // self.gen_batch_size
-        assert n_epochs >= 1, (
+        n_turns = total_timesteps // self.gen_batch_size
+        assert n_turns >= 1, (
             "No updates (need at least "
             f"{self.gen_batch_size} timesteps, have only "
             f"total_timesteps={total_timesteps})!"
         )
-        for epoch in tqdm.tqdm(range(0, n_epochs), desc="epoch"):
+        for epoch in tqdm.tqdm(range(0, n_turns), desc="turn"):
```
AdamGleave, line 334: s/epoch/turn?

src/imitation/algorithms/adversarial.py
```diff
@@ -350,43 +359,70 @@ def _torchify_with_space(
...
+            )
 
-        if expert_samples is None:
-            expert_samples = self._expert_dataset.sample(n_gen)
-        n_expert = len(expert_samples.obs)
+        # Copy items and ensure Mapping argument is in mutable form.
+        expert_samples = dict(expert_samples)
+        gen_samples = dict(gen_samples)
+
+        # Convert applicable Tensor values to Numpy.
```
AdamGleave, line 399:
```suggestion
        # Convert applicable Tensor values to NumPy.
```

src/imitation/algorithms/bc.py
```diff
@@ -62,7 +61,7 @@ def __init__(
         *,
         policy_class: Type[policies.BasePolicy] = base.FeedForward32Policy,
         policy_kwargs: Optional[Mapping[str, Any]] = None,
-        expert_dataloader: Optional[BCDataLoaderDucktype] = None,
+        expert_dataloader: Optional[types.DataLoaderInterface] = None,
```
AdamGleave, line 64: Why do we insist on a dataloader here but accept `types.Transitions` for `Adversarial`?

Where possible would be nice to move towards greater harmonization of their interfaces to simplify benchmarking different algorithms on the same environment.

src/imitation/algorithms/bc.py
```diff
@@ -145,6 +150,9 @@ def _calculate_loss(self, obs, acts) -> Tuple[th.Tensor, Dict[str, float]]:
             stats_dict: Statistics about the learning process to be logged.
 
         """
+        obs = th.as_tensor(obs, device=self.device)
```
AdamGleave, line 153: What happens if these were originally tensors? Do we end up taking gradients through them (so we might change the observations to be closer to the policy rather than the other way around)?

src/imitation/data/buffer.py
```diff
@@ -4,10 +4,10 @@
 import numpy as np
 from stable_baselines3.common import vec_env
 
-from imitation.data import datasets, types
+from imitation.data import types
 
 
-class Buffer(datasets.Dataset[Dict[str, np.ndarray]]):
+class Buffer:
```
AdamGleave, line 10: +1 on deleting datasets (always nice to have less code!) -- but it seems nice to highlight that these classes do implement a common interface. e.g. both have `from_data`, `store`, `sample` differing only in types. Perhaps we can make a new (generic) abstract class for this?

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -115,7 +116,7 @@ def airl():
 ANT_SHARED_LOCALS = dict(
     total_timesteps=3e7,
     max_episode_steps=500,  # To match `inverse_rl` settings.
-    algorithm_kwargs=dict(shared=dict(disc_batch_size=2048 * 8)),
+    algorithm_kwargs=dict(shared=dict(expert_batch_size=2048 * 4)),
```
AdamGleave, line 119: why do we specify it as `2048 * 4` rather than `8192`?

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -115,7 +116,7 @@ def airl():
...
     gen_batch_size=2048 * 8,
```
AdamGleave, line 120: id.

src/imitation/util/util.py
AdamGleave: Since you have an `iterable` (not an iterator) I think you can do something like:

```python
try:
    next(iter(iterable)))
except StopIteration:
    raise RuntimeError(f"iterable {iterable} had no elements to iterate over.")
return itertools.chain.from_iterable(itertools.repeat(iterable))
```

AdamGleave (commented):
Overall API seems easy enough to use. I like how we were able to eliminate `datasets` -- the less code the better :) Few minor suggestions for improvements to docs, etc.

src/imitation/util/util.py
shwang: The current implementation also accounts for a nondeterministically-lengthed `__iter__` which could be empty on the 3rd call etc. But then again, if it were empty only sometimes, then we wouldn't run into the "hanging on infinite empty Iterables" problem.

I'll probably replace this with @AdamGleave's solution since it's a bit simpler. will think about this a bit more.

tests/test_adversarial.py
```diff
@@ -27,19 +28,20 @@ def test_train_disc_small_expert_data_warning(tmpdir, _algorithm_cls):
...
-        _algorithm_cls(
+    with pytest.raises(ValueError, match="expert_batch_size"):
+        trainer = _algorithm_cls(
             venv=venv,
             expert_data=small_data,
+            expert_batch_size=21,
             gen_algo=gen_algo,
             log_dir=tmpdir,
         )
+        trainer.close()
```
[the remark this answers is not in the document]
shwang: These `close()` calls are also unnecessary. I experienced some weird slowdowns when parallelizing Pytorch-GAIL runs, so I assumed it was multiprocessed DataLoaders not cleaning up. But by default DataLoader is single-processed.

@qxcv thinks that the slowdown due to pytorch by default spawning a process for every core, or something like that.

src/imitation/algorithms/adversarial.py
```diff
@@ -57,53 +59,80 @@ def __init__(
...
-              `self.train_env`.
-              This disables the reward wrapping that would normally replace
-              the environment reward with the learned reward. This is useful for
-              sanity checking that the policy training is functional.
+                `self.train_env`.
+                This disables the reward wrapping that would normally replace
+                the environment reward with the learned reward. This is useful for
+                sanity checking that the policy training is functional.
         """
+        # TODO(shwang): Allow Transitions (`expert_data`) to have length less than
```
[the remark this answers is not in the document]
shwang: I was just thinking for convenience reasons, so that the user wouldn't error on the length of the Transitions passed in.

src/imitation/algorithms/adversarial.py
```diff
@@ -57,53 +59,80 @@ def __init__(
...
+        #  set has sufficient samples.
+
         assert (
             logger.is_configured()
         ), "Requires call to imitation.util.logger.configure"
         self._global_step = 0
         self._disc_step = 0
+        self.n_disc_updates_per_turn = n_disc_updates_per_turn
+
+        assert expert_batch_size > 0
```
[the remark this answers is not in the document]
shwang: ugh, I think I would use `ValueError` more automatically if it didn't create a coverage hole 😂

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
shwang: But yeah, looking at this again, it makes sense to me that this would just lead to wasted compute.

src/imitation/scripts/config/train_adversarial.py
shwang: I don't remember. I'll convert this

src/imitation/algorithms/bc.py
shwang: Hm, I wasn't considering that case. I think that when an observation comes from a DataLoader, it's almost never expected to have a gradient on it. Thoughts on whether DataLoader can or ever should return a Tensor with parents, @qxcv?

shwang: According to [the docs](https://pytorch.org/docs/stable/generated/torch.as_tensor.html), `th.as_tensor` will copy the computational graph if the Tensor can have a gradient.

src/imitation/algorithms/adversarial.py
```diff
@@ -182,48 +211,26 @@ def _gen_log_action_prob_from_unnormalized(
...
 
         Args:
-          gen_samples: Transition samples from the generator policy. If not
-            provided, then take `self.disc_batch_size // 2` samples from the
-            generator replay buffer. Observations should not be normalized.
-          expert_samples: Transition samples from the expert. If not
-            provided, then take `n_gen` expert samples from the expert
-            dataset, where `n_gen` is the number of samples in `gen_samples`.
-            Observations should not be normalized.
+            expert_samples: Transition samples from the expert in dictionary form.
```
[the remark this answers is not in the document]
shwang: Thanks, should contain `self.expert_batch_size` samples. (Although this is not strictly necessary, we could also just require that there is an equal number of samples as `gen_samples`).

src/imitation/algorithms/bc.py
shwang: I'd support allowing `types.Transititions` as a convenience type here. I'm pretty sure I had implemented this choice at some point; don't remember why I removed it.

shwang: Oh, that's because in `BC` we are missing a batch size variable which is needed to complete the default `DataLoader` constructor.

I think I'll add in a  "default batch size" class constant for this.

shwang: Support for Transitions argument type added

src/imitation/data/buffer.py
shwang: I don't think this effort is worthwhile right now, especially since among the two classes we are only using `ReplayBuffer` and in only one place.

src/imitation/util/util.py
shwang: I ended up coding Adam's solution except using a `yield from` instead of a `return`. Because Generator has a `.close()` method, but `Iterator` doesn't. I'm questioning this decision because it's not clear to me right now that `.close()` actually adds any value for its complexity right now.

shwang: Finally decided to use `return` and get rid of all the `.close()` related clean ups, because they didn't seem to be necessary for this iterator. Will add them back in if when working with MineRL, our DataLoader seems to need this clean up explicitly.

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
shwang: removed comment.

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
shwang: converted to ValueError

shwang:
Addressed @AdamGleave's comments or left questions. Waiting on a look over from @qxcv!

src/imitation/algorithms/bc.py
AdamGleave: My read is it does preserve the computational graph? "otherwise a new Tensor will be returned with **computational graph retained** if data Tensor has requires_grad=True" (emphasis mine) But I'm not sure.

src/imitation/data/buffer.py
AdamGleave: OK, I don't feel strongly about this.

src/imitation/util/util.py
AdamGleave: I guess the argument for `close` would be if it depended on resources that need to be explicitly cleared up (will not be garbage collected): e.g. if you're computing rollouts from an expert policy, you might need to explicitly destroy the policy when no longer needed. My impression is PyTorch does a better job of being GC-compatible than Tensorflow though so this may not be needed.

src/imitation/algorithms/bc.py
```diff
@@ -112,27 +121,45 @@ def __init__(
...
-    def set_expert_dataloader(self, expert_dataloader: BCDataLoaderDucktype) -> None:
-        """Set the expert dataloader, which yields batches of obs-act pairs.
+    def set_expert_data_loader(
+        self,
+        expert_data: Union[types.DataLoaderInterface, types.TransitionsMinimal],
+    ) -> None:
+        """Set the expert data_loader, which yields batches of obs-act pairs.
 
-        Changing the expert dataloader on-demand is useful for DAgger and other
+        Changing the expert data_loader on-demand is useful for DAgger and other
```
qxcv, line 137: Should this be `expert_data_loader`?

src/imitation/algorithms/bc.py
qxcv: The dataloader shouldn't return tensors with `requires_grad=True`, but I don't think there's much cost to `.deatch()`ing the tensor if that's a worry.

qxcv: (TBH I've never bothered doing this in the Torch code I write unless I expect it to be genuinely ambiguous to whoever is calling the function)

src/imitation/algorithms/bc.py
```diff
@@ -192,7 +222,7 @@ def train(
         samples_so_far = 0
         batch_num = 0
         for epoch_num in trange(n_epochs, desc="BC epoch"):
```
qxcv, line 224: It would be very cool if you could refactor this to run for a fixed # of batches instead of a fixed # of epochs (like the GAIL training code does). The concept of an "epoch" is a bit awkward when you only have datasets of a few dozen or a few hundred time steps :(

(don't feel pressured to do it this PR, though)

tests/test_adversarial.py
```diff
@@ -77,43 +110,65 @@ def trainer(_algorithm_cls, _parallel: bool, tmpdir: str, _convert_dataset: bool
...
-    return _algorithm_cls(
+    trainer = _algorithm_cls(
         venv=venv,
         expert_data=expert_data,
+        expert_batch_size=expert_batch_size,
         gen_algo=gen_algo,
         log_dir=tmpdir,
     )
 
+    yield trainer
```
qxcv, line 121: Maybe close the `venv` explicitly at the end?

qxcv (approved):
A few minor comments, otherwise LGTM. Thanks for doing this!

src/imitation/algorithms/bc.py
shwang: I ran some scratch code and it looks like `as_tensor` as currently implemented simply returns the argument if it's a Tensor and the device doesn't change.

So in this case it's the exact same computational graph. But I'd imagine that you would have to copy the computational graph if you changed devices?

shwang: Anyways, I'm not sure if there's something I should be commenting on or adding a `detach` to here. I don't believe that this is a case that folks will easily run into, and if they did they probably are being intentional about it.

tests/test_adversarial.py
shwang: Sounds good

shwang: Ugh, actually it doesn't really matter for this case since it's just CartPole. Whatever, I'll add it anyways

src/imitation/algorithms/bc.py
shwang: I think epochs is very natural for supervised learning on a fixed dataset. Probably save this for #79 if we want to think about unifying the algorithms' interfaces.

src/imitation/algorithms/adversarial.py
```diff
@@ -305,31 +285,33 @@ def train(
...
         sampled from the environment to exceed `total_timesteps`.
 
         Params:
           total_timesteps: An upper bound on the number of transitions to sample
-            from the environment during training.
-          callback: A function called at the end of every epoch which takes in a
-            single argument, the epoch number. Epoch numbers are in
-            `range(total_timesteps // self.gen_batch_size)`.
+              from the environment during training.
+          callback: A function called at the end of every round which takes in a
```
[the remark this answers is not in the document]
shwang: It's round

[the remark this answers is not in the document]
shwang: I'm pretty sure that I found all the "turn"s and replaced them, though I didn't think up a comprehensive regex for grepping.

src/imitation/algorithms/adversarial.py
```diff
@@ -305,31 +285,33 @@ def train(
...
-        n_epochs = total_timesteps // self.gen_batch_size
-        assert n_epochs >= 1, (
+        n_rounds = total_timesteps // self.gen_batch_size
+        assert n_rounds >= 1, (
             "No updates (need at least "
             f"{self.gen_batch_size} timesteps, have only "
             f"total_timesteps={total_timesteps})!"
         )
-        for epoch in tqdm.tqdm(range(0, n_epochs), desc="epoch"):
+        for r in tqdm.tqdm(range(0, n_rounds), desc="round"):
```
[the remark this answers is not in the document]
shwang: Yeah piping tqdm is not a great look. I can introduce a gated utility.

`os.isatty` worked fine here:

image

src/imitation/algorithms/adversarial.py
```diff
@@ -146,24 +175,24 @@ def __init__(
...
+    def _next_expert_batch(self) -> dict:
+        return next(self._endless_expert_iterator)
+
+    def close(self):
+        """Cleans up memory-intensive resources."""
+        self._endless_expert_iterator.close()
 
     @property
     def gen_batch_size(self) -> int:
+        # TODO(shwang): Sad that this basically hard-codes a dependency on PPO2.n_steps
```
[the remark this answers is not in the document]
shwang: Saving this TODO for a future PR (since if this is bug, it wasn't introduced by this PR), and making an Issue.

src/imitation/algorithms/adversarial.py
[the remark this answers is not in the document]
shwang: Making an issue for this instead of putting this in the current PR.

shwang:
Superceded by #236

src/imitation/algorithms/bc.py
AdamGleave: What use case do we have in mind for these already being tensors? My inclination is to call `detach()` unless there's actually a case where we want to take gradients through observations.

shwang: `pytorch.util.data.DataLoader` will automatically convert np.ndarrays from its Torch Dataset argument to `Tensor`.

I'm fine with detaching, making that change now.

PR #233, opened 2020-10-07
Add imitation.contrib.data_loader_wrappers

shwang:


shwang:
As a refresher (since it's been a while), `GAIL` and `BC` will soon operate on pytorch-style DataLoaders which yield dictionaries with keys like [[["obs", "acts", "infos"], "next_obs"], "rews"].

In the API, I define a abstract class `DataLoaderWrapper` which modifies each data batch yielded from the `DataLoader` that it wraps. I also define a function `wrap_dataset_with_env_wrappers(data_loader, env)` which applies the `DataLoaderWrappers` corresponding to each Wrapper on the `env` on top of `data_loader`.

There's two ways to pair a `DataLoaderWrapper` with a `gym.Wrapper` instance so that `wrap_dataset_with_env_wrappers` works properly.

1. Any `gym.Wrapper` can subclass the `DefinesDataLoaderWrapper` mixin to define its corresponding `DataLoaderWrapper` via `DefinesDataLoaderWrapper.apply_data_loader_wrapper`.

2. The abstract class `BidirectionalTransform` contains placeholders for simple transformations over the {obs, act, next_obs, ...} and a simple psuedo-inverse transformation over actions. These transformations are enough to define a simple Gym Wrapper `BidirectionalGymWrapper` and its corresponding `BidirectionalDataLoaderWrapper`. This style of `DataLoaderWrapper`-`gym.Wrapper` pairing is meant to be used for the simplest Gym Wrappers (e.g. affine transformation on observations) where it will reduce code duplication.

shwang:
[Finish](https://trello.com/c/p5RPknx4/4-finish)

src/imitation/contrib/data_loader_wrappers.py
```diff
@@ -0,0 +1,211 @@
...
+    def __init__(self, data_loader: Iterable[S]):
+        self.data_loader = data_loader
+
+    @abc.abstractmethod
+    def __iter__(self) -> Iterator[T]:
+        pass
+
+
+class BidirectionalTransform(abc.ABC):
+    @abc.abstractmethod
```
AdamGleave, line 25: If they are pure functions, why are they instance methods? Alternatives:
  - Static methods
  - NamedTuple of functions
(You could be sneaky and keep state with either of these, but you couldn't do it unintentionally, whereas if I'm writing a class it's quite natural to use `self` to store things.)

src/imitation/contrib/data_loader_wrappers.py
```diff
@@ -0,0 +1,211 @@
...
+        transform: BidirectionalTransform,
+    ):
+        self.data_loader = data_loader
+        self.transform = transform
+        super().__init__(data_loader)
+
+    def __iter__(self) -> dict:
+        for trans in self.data_loader:
+            result = self.transform.forwards(trans)
+            result["acts"] = self.transform.backwards_act(trans["acts"])
```
AdamGleave, line 97: It's confusing that actions and observations have to be transformed in different directions. I think this at least deserves a warning in the documentation, but perhaps we can make this more intuitive.

Is there any time we actually want to use `forwards` by itself (without `backwards_act`)?

src/imitation/contrib/data_loader_wrappers.py
```diff
@@ -0,0 +1,211 @@
...
+            yield result
+
+
+class DefinesDataLoaderWrapper(abc.ABC, Generic[S, T]):
```
AdamGleave, line 101:
```suggestion
class DataLoaderWrapperMixIn(abc.ABC, Generic[S, T]):
```

AdamGleave, line 101: So this mix-in is really just defining a function that should be present?

src/imitation/contrib/data_loader_wrappers.py
```diff
@@ -0,0 +1,211 @@
...
+        self.transform = transform
+        super().__init__(env)
+
+    def reset(self):
+        obs = self.env.reset()
+        return self.forward_obs(obs)
+
+    def step(self, action: Any) -> Tuple[Any, float, dict]:
+        act = self.transform.backwards_act(action)
+        return self.env.step(act)
```
AdamGleave, line 143: This is missing a `forwards` step (for reward, observations, etc)?

src/imitation/contrib/data_loader_wrappers.py
```diff
@@ -0,0 +1,211 @@
...
+
+    Intended for use as an abstract mix-in in a subclass of `gym.Wrapper` so that the
+    `gym.Wrapper` can be processed by `wrap_data_loader_with_env_wrappers`.
+    """
+
+    @abc.abstractmethod
+    def apply_dataset_wrapper(
+        self,
+        data_loader: types.DataLoaderInterface,
+    ) -> DataLoaderWrapper[S, T]:
```
AdamGleave, line 112: When would we want to do something other than applying a `BidirectionalTransform` here?

AdamGleave (commented):
I've added some more detailed comments. Overall this seems reasonable to me. The API is quite involved -- so I'd err towards simplifying where possible. One possible simplification is just having separate `forwards_obs`, `forwards_acts`, and `forwards_rew`. This would prohibit some entangled preprocessing -- but do we have a concrete use case in mind for this? Note we already require that `obs` can be forwarded on its own, and `acts` backwarded on its own -- so it doesn't feel like there are that many degrees of freedom left.

I still have the concern that defining a `gym.Wrapper` not via a transformation has the potential to lead to extremely difficult to detect bugs, if the `apply_data_loader_wrapper` does something different to the transformation that `step` and `reset` do. Are we sure this saves time? It seems like in many cases `apply_data_loader_wrapper` would have to re-implement the transformation `step` performs anyway, so not sure it's actually any easier. If we do keep this feature I'd advocate for adding automated tests to sanity check the wrappers are internally consistent on some random inputs.

 It occurs to me that things like `gym.ObservationWrapper`, `gym.RewardWrapper`, `gym.ActionWrapper` could potentially be automatically handled (note `gym.ActionWrapper` even includes a `reverse_action` -- how convenient): we can infer a `BidirectionalTransform` from them I think.

Often we wrap at the `VecEnv` level, e.g. `VecNormalize`, so we'd need things to handle that too. But that seems like a detail: whatever we decide for `Env` should generalize naturally to `VecEnv` I think.

src/imitation/contrib/data_loader_wrappers.py
shwang: Yes, and it used in an `isinstance(_, DefinesDataLoaderWrapper)` call in `wrap_data_load_with_env_wrappers()`. To check which wrappers can apply wrappers.

src/imitation/contrib/data_loader_wrappers.py
shwang: that's right, thanks

src/imitation/contrib/data_loader_wrappers.py
shwang: Ah they aren't actually pure. It's more like they are pure with respect to initialization variables. For example, we something like a `CameraObservationScalingWrapper.__init__(self, scale_factor)` which multiplies a key inside dict obs by `scale_factor`.

I'll figure out how to express this better in the docs.

src/imitation/contrib/data_loader_wrappers.py
shwang: I prefer `DefinesDataLoaderWrapper` over `DataLoaderWrapperMixIn` because I think the second one doesn't remind the API reader what the MixIn is for. (And there are a lot of abstractions to remember here).

We could also do `DefinesDataLoaderWrapperMixin` but that's kind of long. /shrug

src/imitation/contrib/data_loader_wrappers.py
shwang: Something like FrameStack, if we ever decide to support that. Leaving the API open to allowing non-BidirectionalTransform for now.

shwang:
Disentangling each individual value (obs, rew, act) seems like a good idea.

Agree that the definitions of `ObservationWrapper.observation` and `RewardWrapper.reward` should be sufficient to construct `DataLoaderWrapper`.

Interesting that ActionWrapper already has an ~abstract `reverse_action` method. It doesn't look like anything in Gym uses the `reverse_action` method though, and this method is not implemented in most ActionsWrappers, so we can't rely on this.

shwang:
I'd like to keep the disentangled API batched because I think this will speed up preprocessing a lot for simple scaling operations over a batch size of say 512.

In other words we'll be able to do `return rews * 2` rather than have the CPU iterate over every float in `rews`, multiply each value, concatenate them together into a big rews. And then iterate over each of the values in `acts` and `obs` to do an identity operation.

shwang:
Maybe it would be better to use `in/out`, or even better, `up/down` instead of `forwards/backwards`. Much less ambiguous in my head to think about going up and down the stack of Env or DataLoader wrappers.

shwang:
Oh this is so much more intuitive.

src/imitation/contrib/data_loader_wrappers.py
shwang: I settled for `ApplyDataLoaderWrapperMixin` (prioritizing clarity over length)

src/imitation/contrib/data_loader_wrappers.py
AdamGleave: To make sure I understand, `FrameStack` can't be supported in `BidirectionalTransform` because it needs to be stateful? This would also require that the data loader returns transitions in sequence, I guess.

src/imitation/contrib/data_loader_wrappers.py
AdamGleave: I see, so really we're saying two calls to a given *instance* with the same variables should return the same output? No mutable state, in a way.

AdamGleave:
Keeping things batched is fine. There's no loss of generality: you can always just make a singleton list and unwrap things. (For the same reason, I tend to always use `VecEnv` rather than `Env`, since I usually end up eventually wanting to use `VecEnv` so easier to just do it everywhere.)

`up`/`down` seems good, though we should make clear in the docs whether the "stack" of wrappers starts at the top or bottom!

src/imitation/contrib/data_loader_wrappers.py
shwang: Agree that this is weird. I think that "needing to apply an oppositely-named action transformation method" is no longer a problem now that I've switched to up/down, but will add a comment if that isn't the case.

src/imitation/contrib/data_loader_wrappers.py
shwang: Yes.

I'll probably write in the docs that this ABC is recommended for wrappers with that property.

shwang:
Renaming `BidirectionalTransform` to simply `Transform`, because this is sorter and we are actually only "bidirectional" for actions.

shwang:
Thanks for the reviews; the new module is certainly looking less convoluted now! (Or maybe I'm just getting more familiar with the abstractions now 😝). I addressed the comments (or left TODOs in places where I think it's a "we don't need it yet" feature) and added some basic tests.

I think I've reached a point with this contrib module that I can start using modified MineRL datasets with them. Going to focus on polishing the GAIL/BC-DataLoader PRs now.

AdamGleave:
> Thanks for the reviews; the new module is certainly looking less convoluted now! (Or maybe I'm just getting more familiar with the abstractions now ). I addressed the comments (or left TODOs in places where I think it's a "we don't need it yet" feature) and added some basic tests.
> 
> I think I've reached a point with this contrib module that I can start using modified MineRL datasets with them. Going to focus on polishing the GAIL/BC-DataLoader PRs now.

OK, let me know if you want me to take another review. Having some real-world experience with the API with MineRL also seems good for validating the API.

shwang:
This looks generally good to me -- I will probably merge soon, without additional review.

AdamGleave:
The rebase seems to have destroyed history so I'm not sure what you've changed recently? But from a quick skim it seems reasonable. I'd be inclined to replace the "XXX: " with a TODO referencing a GitHub issue though.

shwang:
On the status of this PR -- we didn't actually end up using this interface for BASALT baselines or development, but it could be useful to finish this feature another time (since reversible transformations like this are common in imitation learning)

PR #237, opened 2020-10-15
pytype: Use multithreading

shwang:


AdamGleave:
CircleCI Docker containers have all the host CPUs visible, but we're throttled to a smaller number (2 by default), so I think `-j auto` will do the wrong thing.

In https://github.com/HumanCompatibleAI/evaluating-rewards/blob/master/.circleci/config.yml I declare an environment variable `NUM_CPUS` and then use it throughout for things like this (`pytype`, `pylint`, `pytest`, ...)

shwang:
Ah, thanks for the pointer. Added an env var; will merge after tests pass

Issue #190, opened 2020-05-09
Basic docs / quickstart tutorial


AdamGleave:
We have a Sphinx build and I've set up [readthedocs](https://imitation.readthedocs.io/), but right now our documentation is very unstructured. We actually have pretty good docstrings on individual classes, but no intro guide, or other structure that would help new users.

Not that high priority: I think it's more important to focus on making the algorithms solid and possibly doing some refactoring. But once we have that a small tutorial would make a big difference to adoption.

araffin:
as discussed with @Miffyli , I would be for integrating some doc (presentation of the lib and example) directly in SB3 doc (as it is done for the zoo).
I have a plan to do the same for offline algos implemented in https://github.com/takuseno/d3rlpy (see issue https://github.com/takuseno/d3rlpy/issues/5).
Both repos are nice extension to SB3.

AdamGleave:
Sounds good -- let us know if we can help with this, by e.g. providing examples of `imitation`

araffin:
>Sounds good -- let us know if we can help with this, by e.g. providing examples of imitation

Would be good if someone from your team could to a PR to add a page with presentation and examples, as I already have other doc to write ;)
(see https://github.com/DLR-RM/stable-baselines3/issues/166)

AdamGleave:
@shwang do you have the bandwidth to do this?

Miffyli:
A suggestion: Add quick example close to top in README to demonstrate the ease of use (in minimal case) outside the premade script runs. I found the BC implementation here very convenient to use with my experiments!

shwang:
I wrote up a quickstart blurb over here as per @Miffyli's suggestion.

https://gist.github.com/shwang/aba1711ffec8a83da4a315ab10c6cd62

I'm thinking that with some editing, this could comprise half of the SB3 docs page alongside a written description of the project?

araffin:
>I'm thinking that with some editing, this could comprise half of the SB3 docs page alongside a written description of the project?

yes, look good, thanks =)
I would also add a one sentence description of each algorithm implemented and a link to the paper.
Also, once the imitation doc is online, we will link to it (so people can take a look at the available parameters).

shwang:
Imitation docs are actually [already online](https://imitation.readthedocs.io/en/latest/imitation.algorithms.html#module-imitation.algorithms.adversarial) in a rudimentary form, but the default Sphinx layout makes it super hard to read.

[...]

PR #238, opened 2020-10-21
Add quickstarts + related tests, create CONTRIBUTING.md

shwang:
We also have Dagger and SDE, but thinking we can save those for later, or put them in an `examples/` script.

Contributes to work on #190.

Other changes (described in commit messages) include:
* changing `str` typing to `Pathlike` typing in `util.logger`
* updating test rollouts to new format so that we can use `pickle.load`.
* Get the density baselines working again (a few import changes) and add a fast mode by default (for CI and user exploration).


README.md
```diff
@@ -7,48 +7,45 @@
...
-Currently we have implementations of [AIRL](https://arxiv.org/abs/1710.11248) and 
-[GAIL](https://arxiv.org/abs/1606.03476), and intend to add more in the future.
+Currently we have implementations of Behavioral Cloning, [DAgger](https://arxiv.org/pdf/1011.0686.pdf) (with synthetic examples), [Adversarial Inverse Reinforcement Learning](https://arxiv.org/abs/1710.11248), and [Generative Adversarial Imitation Learning](https://arxiv.org/abs/1606.03476).
 
 ### To install:
 ```
-conda create -n imitation python=3.7
-conda activate imitation
-pip install -e '.[dev]'  # install `imitation` in developer mode
+git clone http://github.com/HumanCompatibleAI/imitation
```
AdamGleave, line 14: We also are on PyPi, right? I'd be inclined to make that the first suggestion for how to install, then say "if you want to get the latest release, you can also install from Git via ..."

README.md
```diff
@@ -7,48 +7,45 @@
...
+Follow instructions to install [mujoco\_py v1.5 here](https://github.com/openai/mujoco-py/tree/498b451a03fb61e5bdfcb6956d8d7c881b1098b5#install-mujoco).
 
-### To run:
-```
-# Train PPO agent on cartpole and collect expert demonstrations
-python -m imitation.scripts.expert_demos with cartpole
-# Train AIRL on from demonstrations
-python -m imitation.scripts.train_adversarial with cartpole airl
+
+## Sacred CLI Quickstart:
```
AdamGleave, line 24: Most people won't know what Sacred is. How about just "CLI Quickstart" as heading, then we can have an intro sentence saying:

```
We provide several CLI scripts as a front-end to the algorithms implemented in `imitation`. These use [Sacred](https://github.com/idsia/sacred) for configuration and replicability.
```

README.md
```diff
@@ -7,48 +7,45 @@
...
+python -m imitation.scripts.train_adversarial with fast gail cartpole rollout_path=quickstart/rl/rollouts/final.pkl
+
+# Train AIRL from demonstrations. Tensorboard logs saved in output/ (default log directory).
+python -m imitation.scripts.train_adversarial with fast airl cartpole rollout_path=quickstart/rl/rollouts/final.pkl
 ```
-View Tensorboard with `tensorboard --logdir output/`.
+Tips:
+  * Remove the "fast" option from the above runs to run training to completion.
+  * `python -m imitation.scripts.expert_demos print_config` will list Sacred script options, which are documented
+  in each script's docstrings.
```
AdamGleave, line 39: This makes it sound like Sacred is scraping the docstrings when it's not. Perhaps separate the two sentences. "These configuration options are documented in each script's docstrings".

README.md
```diff
@@ -7,48 +7,45 @@
...
 
+For more information configuring on Sacred CLI options, see [Sacred docs](https://sacred.readthedocs.io/en/stable/).
```
AdamGleave, line 41:
```suggestion
For more information on how to configure Sacred CLI options, see [Sacred docs](https://sacred.readthedocs.io/en/stable/).
```

examples/quickstart.sh
```diff
@@ -0,0 +1,8 @@
+# Train PPO agent on cartpole and collect expert demonstrations. Tensorboard logs saved in `quickstart/rl/`
```
AdamGleave, line 1: Do we need this file? It's not referenced anywhere, and seems to duplicate the README. There is something nice about being able to actually *run* it, but not sure it's worth duplication. (Sadly no include directive in Markdown.)

src/imitation/data/types.py
```diff
@@ -15,6 +15,8 @@
 
 T = TypeVar("T")
 
+Pathlike = Union[str, bytes, pathlib.Path]
```
AdamGleave, line 18: This is confusing with `os.PathLike`. Some possibilities:
  - `Pathish`
  - `CouldBePath`
  - `PathCastable`

AdamGleave: To make matters complicated, we probably want to actually allow `pathlib.PathLike` too! `typeshed` uses (for Python >=3.6):
```python
AnyPath = Union[str, bytes, PathLike[str], PathLike[bytes]]
```

Yuck! But perhaps copying that is best?

https://github.com/python/typeshed/blob/c41f3e9bf946c5cb2fa53ed98380dc56977f82e8/stdlib/2and3/_typeshed/__init__.pyi#L52

tests/test_examples.py
```diff
@@ -0,0 +1,49 @@
...
+SH_PATHS = EXAMPLES_DIR.glob("*.sh")
+NB_PATHS = EXAMPLES_DIR.glob("*.ipynb")
+PY_PATHS = EXAMPLES_DIR.glob("*.py")
+
+
+@pytest.mark.parametrize("nb_path", NB_PATHS)
+def test_run_example_notebooks(nb_path):
+    """Smoke test ensuring that example notebooks run without error.
+
+    The `pytest_notebook` package also includes regression test functionality against
```
AdamGleave, line 18: FYI I have been using that regression functionality (just using `--nb-test` CLI option and default config). It works reasonably well but I have found it breaks for some annoying reasons. E.g. IIRC the notebook format changes periodically as Python versions get upgraded which will break it, until you rerun the notebook and commit to Git again. But it did catch a real breakage once too.

tests/test_examples.py
```diff
@@ -0,0 +1,49 @@
+import pathlib
+import subprocess
+
+import pytest
+from pytest_notebook import execution, notebook
+
+EXAMPLES_DIR = pathlib.Path("examples/")
```
AdamGleave, line 7: I'm a bit nervous of using relative paths. We don't want to use e.g. `pkgresource` because `examples/` isn't installed. But we could use a path relative to the *source* file, which should work wherever you run `pytest` from?

https://stackoverflow.com/a/3718923/14507224 gives an example of this

examples/quickstart.sh
AdamGleave: Ah I see you added a test to check the snipped hasn't diverged. That removes my concern re: code duplication. But we should probably still mention the files existence from the README?

examples/quickstart.py
```diff
@@ -0,0 +1,54 @@
...
+    trajectories = pickle.load(f)
+
+# Convert List[types.Trajectory] to an instance of `imitation.data.types.Transitions`.
+# This is a more general dataclass containing unordered
+# (observation, actions, next_observation) transitions.
+transitions = rollout.flatten_trajectories(trajectories)
+
+venv = util.make_vec_env("CartPole-v1")
+
+with tempfile.TemporaryDirectory(prefix="quickstart") as tempdir:
```
AdamGleave, line 25: It is a bit unfortunate that the temporary directory is destroyed when training is finished. We could either be bad and make a temporary directory and not clean it up. Or add some "press enter to exit this script, at which point the result will be deleted" blocking wait at the end.

examples/quickstart.py
```diff
@@ -0,0 +1,54 @@
+import pathlib
```
AdamGleave, line 1: I like this script! It really emphasizes how similar the interfaces of the different algorithms are (for simple use cases anyway) which is great.

Maybe add a one-liner module docstring in case someone just stumbles across this script without seeing the README?

examples/density_baseline_demo.ipynb
```diff
@@ -0,0 +1,252 @@
+{
```
AdamGleave, line 1: Worth mentioning this file exists in the README so people actually look at it?

examples/density_baseline_demo.ipynb
```diff
@@ -0,0 +1,252 @@
...
+     "text": [
+      "Using cuda device\n"
+     ],
+     "output_type": "stream"
+    }
+   ],
+   "source": [
+    "env_name = 'Pendulum-v0'\n",
+    "env = util.make_vec_env(env_name, N_VEC)\n",
+    "rollouts = types.load(\"../tests/data/expert_models/pendulum_0/rollouts/final.pkl\")\n",
```
AdamGleave, line 81: See comments for `tests/test_examples.py` re: relative paths. Bugs me less for notebooks though which are meant to be hacky.

examples/density_baseline_demo.ipynb
```diff
@@ -0,0 +1,252 @@
...
+    "imitation_trainer = util.init_rl(env, learning_rate=3e-4, n_steps=2048)\n",
+    "density_trainer = db.DensityTrainer(env,\n",
+    "                                    rollouts=rollouts,\n",
+    "                                    imitation_trainer=imitation_trainer,\n",
+    "                                    density_type=db.STATE_ACTION_DENSITY,\n",
+    "                                    is_stationary=False,\n",
+    "                                    kernel='gaussian',\n",
+    "                                    kernel_bandwidth=0.2,  # found using divination & some palm reading\n",
```
AdamGleave, line 89: Look forward to seeing your SIGBOVIK paper on divination for hyperparameter tuning :)

AdamGleave (approved):
Thanks for this addition -- will make it much easier for new people to start, and I appreciate the examples actually being tested.

Note I've made a few commits in this PR to improve the Sphinx docs (seemed like a good place for it) -- I suggest reviewing my commits.

I've made some minor comments to address, but apart from that LGTM -- feel free to merge once those are addressed (or tag me for review again if there's something that needs further discussion).

examples/quickstart.py
shwang: Oops, I forgot this that cleaned up the tmpdir. I think it's fine to leave the tmpdir (it will be auto-cleaned by the system later anyways).

README.md
shwang: That sounds good. I'm inclined to make a new PyPi release immediately after this PR, so that the documentation will actually match, in that case.

shwang: - [ ] @shwang PyPI release

examples/density_baseline_demo.ipynb
shwang: At least 0.5 years ago, Jupyter notebook always runs the notebook with the notebook's parent directory as the working directory (I tried in vain to configure this to be the project root instead). So this should be more likely to be fine than the `tests/test_examples.py` case.

shwang:
@AdamGleave: Thanks for figuring out the Sphinx stuff and the review. Be submitting a version of this to SB3 docs shortly.

README.md
shwang: done.

PR #243, opened 2020-10-24
BC: Add n_batches end condition, modify callback

shwang:
Adds `n_batches` parameter to BC.train(), as requested by @qxcv,
while retaining the `tqdm` progress bar.

Refactoring changed the location in the code where an epoch end
can be straightfowardly detected, so the callback was moved inside
a different function. Accordingly, I removed the `locals` argument
to the callback, which probably wasn't great form in the first place.

We could consider using SB3-style callbacks in the future.


src/imitation/algorithms/bc.py
```diff
@@ -52,6 +52,94 @@ def __call__(self, _):
...
+            data_loader: An iterable over data dicts, as used in `BC`.
+            n_epochs: The number of epochs to iterate through in one call to
+                __iter__. Exactly one of `n_epochs` and `n_batches` should be provided.
+            n_batches: The number of batches to iterate through in one call to
+                __iter__. Exactly one of `n_epochs` and `n_batches` should be provided.
+            on_epoch_end: A callback function without parameters to be called at the
+                end of every epoch.
+        """
+        if n_epochs is not None and n_batches is None:
+            self.mode = "epochs"
```
AdamGleave, line 77: Nit: since `self.mode` can only take two values legally, can we make this explicitly -- with an `Enum`, or just making it a boolean flag (e.g. `self.use_epochs`)?

tests/test_bc.py
```diff
@@ -79,10 +79,16 @@ def test_weight_decay_init_error(venv):
         )
 
 
+def test_train_end_cond_error(trainer: bc.BC, venv):
+    with pytest.raises(ValueError, match="exactly one.*n_epochs"):
```
AdamGleave, line 83: Add test case for providing both None too?

AdamGleave (approved):
Two minor changes requested but otherwise LGTM, no need to re-review unless any substantial changes made.

src/imitation/algorithms/bc.py
shwang: Made a boolean flag

qxcv (approved):
LGTM. Stuffing all the progress/statistics logic into `EpochOrBatchIteratorWithProgress` feels a bit weird, since it makes it hard to, e.g., disable `tqdm`, or include other statistics like loss in the progress bar. I think it's okay for now though.

PR #244, opened 2020-10-24
Adversarial.gen_batch_size: Fix vec scaling bug

### LINKED ISSUE #234: AdversarialTrainer: Is n_gen_batch_size off by factor of n_vec?
The definition of the `AdversarialTrainer.gen_batch_size` property implies that `n_steps == gen_batch_size`. Is this wrong? I'm think that this really should be `n_steps * n_vec == gen_batch_size` IIRC.

shwang:
Fixes #234.

Along the way I had to change some default batch sizes and training timesteps to prevent an "`AdversarialTrainer.train()` is a no-op because too few steps" error from firing. (This assertion used to be erroneously based on a lower `gen_batch_size`.)

### DIFF
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -178,7 +178,7 @@ class AdversarialTrainer:
 
     @property
     def gen_batch_size(self) -> int:
-        return self.gen_algo.n_steps * self.venv.num_envs
+        return self.gen_algo.n_steps * self.gen_algo.get_env().num_envs
 
     def train_disc(
         self,
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -178,9 +178,7 @@ class AdversarialTrainer:
 
     @property
     def gen_batch_size(self) -> int:
-        # TODO(shwang): Is `n_steps` really the generator batch size? Isn't the batch
-        #  size actually `n_steps` * `n_vec`? I'm surprised that this works.
-        return self.gen_algo.n_steps
+        return self.gen_algo.n_steps * self.venv.num_envs
 
     def train_disc(
         self,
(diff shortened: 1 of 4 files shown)


src/imitation/algorithms/adversarial.py
```diff
@@ -178,9 +178,7 @@ def _next_expert_batch(self) -> Mapping:
 
     @property
     def gen_batch_size(self) -> int:
-        # TODO(shwang): Is `n_steps` really the generator batch size? Isn't the batch
-        #  size actually `n_steps` * `n_vec`? I'm surprised that this works.
-        return self.gen_algo.n_steps
+        return self.gen_algo.n_steps * self.venv.num_envs
```
AdamGleave, line 181: Is there any guarantee that `self.venv` is the same `VecEnv` used in `self.gen_algo`? `gen_algo` is constructed by the user and they could feed something else in.

We could use `self.venv.get_env().num_envs` instead?

AdamGleave: Also, in general `n_steps` may not be defined and this semantically only really makes sense for `OnPolicyAlgorithm`, so I'd change the type signature in `AdversarialTrainer`, `GAIL` and `AIRL` accordingly. (Note off-policy algorithms will often take a single step, update the model, etc.)

AdamGleave (changes requested):
Well done catching this bug! I think the new computation is a bit fragile though so requesting some changes to make it more robust.

src/imitation/algorithms/adversarial.py
shwang: Did you mean to change the type signature to `OnPolicyAlgorithm` wherever applicable? I already did after we discussed the problem in #231.

https://github.com/HumanCompatibleAI/imitation/blob/ac4f4a50a928063477b2f9661312ed5d48c0b3a4/src/imitation/algorithms/adversarial.py#L453-L459

shwang: >  Is there any guarantee that self.venv is the same VecEnv used in self.gen_algo? 

In the __init__ function we call:

AdamGleave: I missed the `set_env`, does seem less concerning, but still more fool-proof to access directly.

AdamGleave (approved):
LGTM.

Sorry about the bogus comment re: `OnPolicyAlgorithm` -- I must have been looking at a stale version of the source code.

PR #248, opened 2020-11-06
Support generating videos in `eval_policy`

AdamGleave:
Adds support for generating video renderings of policies acting in environments with `eval_policy`.

I include a new `VideoWrapper` class (adapted from https://github.com/HumanCompatibleAI/adversarial-policies/blob/master/src/aprl/policies/wrappers.py). We could instead use `gym.Monitor` (not to be confused with `stable_baselines.bench.Monitor`, which we already use) but it (a) has a lot of features we don't need, and (b) always makes a new video file for each episode, which I find annoying (all episodes in one MP4 is something I want more often and I am too lazy to concatenate).

I wanted to test this, so I added a dummy X server in our Dockerfile, and then also enabled testing for rendering. I've pushed the new images to DockerHub.


AdamGleave:
Before merging this PR: get https://github.com/HumanCompatibleAI/seals/pull/44 merged and make new release then update `setup.py` here to point to new release. We're using some of `seals` new render testing functionality. (I could split this into a separate PR if desired.)

shwang:
Missing some diff coverage in VideoWrapper for `if not self.single_video:`, `VideoWrapper.close` (comes for free if `not self.single_video`).

shwang (changes requested):
Overall looks good; thanks for adding this in. Some minor comments, and would be nice for have a test for `VideoWrapper.single_video = False` so we can cover the entire diff.

src/imitation/envs/examples/model_envs.py
```diff
@@ -67,7 +67,7 @@ def make_obs_mat(n_states, is_random, obs_dim, rand_state=np.random):
     assert (
         obs_mat.ndim == 2 and obs_mat.shape[:1] == (n_states,) and obs_mat.shape[1] > 0
     )
-    return obs_mat
```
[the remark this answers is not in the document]
AdamGleave: Test suite errors because we added to `seals` a check for `dtype`. This is a consequence of upgrading the `seals` version to the GitHub master not rendering per se. I could split this into a separate PR if desired but it only affects a couple of places.

src/imitation/scripts/eval_policy.py
```diff
@@ -33,6 +38,17 @@ def step_wait(self):
         return ob
 
 
+def video_functor(log_dir: str):
```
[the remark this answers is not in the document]
AdamGleave: hmm, I actually think I'm abusing terminology here, and it should be a "higher-order function". I wanted to call it `video_wrapper` but that clashed with the file of the same name. I've renamed to `video_wrapper_factory`.

src/imitation/util/util.py
```diff
@@ -32,6 +41,7 @@ def make_vec_env(
     parallel: bool = False,
     log_dir: Optional[str] = None,
     max_episode_steps: Optional[int] = None,
+    post_wrappers: Optional[Sequence[Callable[[gym.Env, int], gym.Env]]] = None,
```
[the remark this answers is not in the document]
AdamGleave: done, good catch

AdamGleave (commented):
Thanks for the review! Agree with your comments, believe I've resolved everything. I've also added a test case for `single_videos` to improve coverage.

src/imitation/scripts/eval_policy.py
```diff
@@ -67,8 +85,10 @@ def eval_policy(
           TimeLimit so that they have at most `max_episode_steps` steps per
           episode.
       render: If True, renders interactively to the screen.
-      log_dir: The directory to log intermediate output to. (As of 2019-07-19
-          this is just episode-by-episode reward from bench.Monitor.)
+      render_fps: the target number of frames per second to render on screen.
```
shwang, line 88:
```suggestion
      render_fps: The target number of frames per second to render on screen.
```

src/imitation/scripts/eval_policy.py
```diff
@@ -67,8 +85,10 @@ def eval_policy(
...
+      videos: If True, saves videos to `log_dir`.
+      video_kwargs: keyword arguments passe through to `video_wrapper.VideoWrapper`.
```
shwang, line 90:
```suggestion
      video_kwargs: Keyword arguments passed through to `video_wrapper.VideoWrapper`.
```

src/imitation/util/util.py
```diff
@@ -47,6 +57,8 @@ def make_vec_env(
             `max_episode_steps` for every TimeLimit wrapper (this automatic
             wrapper is the default behavior when calling `gym.make`). Otherwise
             the environments are passed into the VecEnv unwrapped.
+        post_wrappers: If specified, iteratively wraps each environment with each
+            of the wrappers specified in the sequence.
```
shwang, line 61:
```suggestion
            of the wrappers specified in the sequence. The argument should be a Callable
            accepting two arguments, the Env to be wrapped and the environment index.
```

src/imitation/util/video_wrapper.py
```diff
@@ -0,0 +1,73 @@
...
+from gym.wrappers.monitoring import video_recorder
+
+from imitation.data import types
+
+
+class VideoWrapper(gym.Wrapper):
+    """Creates videos from wrapped environment by calling render after each timestep."""
+
+    def __init__(
+        self, env: gym.Env, directory: types.AnyPath, single_video: bool = True
```
shwang, line 15:
```suggestion
        self, env: gym.Env, directory: types.AnyPath, single_video: bool = True,
```

tests/test_scripts.py
```diff
@@ -77,7 +77,8 @@ def test_expert_demos_rollouts_from_policy(tmpdir):
 
 
 EVAL_POLICY_CONFIGS = [
-    {},
+    {"videos": True},
+    {"videos": True, "video_kwargs": {"single_video": True}},
```
shwang, line 81: Oh, looks like we should actually be using `single_video=False` to increase coverage, actually.
```suggestion
    {"videos": True, "video_kwargs": {"single_video": False}},
```

shwang (approved):
Few docstring nits. More significantly we still don't have diff coverage because I suggested the oppositely-argumented test case `single_video=True`.

Don't think this should require another review after we have diff coverage.

AdamGleave (commented):
Thanks for the review and catching my dumb typos and `single_video=True` rather than `False`. Diff coverage is fixed now. 

I made some changes to CircleCI config to increase size of executor since was running into OOM error. Oddly it's now running slower than before, despite more total CPUs (1 instance with 8 CPUs in parallel, compared to 3 instances with 2 CPUs). I dunno what's causing this, will try and dig into it tomorrow, but would welcome any ideas if anyone's seen this before.

AdamGleave:
Hmm, running the tests locally even with `-n 2` on a 16 core machine I'm seeing 100% CPU saturation at some points. So I suspect that each parallel instance of PyTorch may be trying to use all the available CPUs. There's probably some way to turn this off.

shwang:
I’ve had similar problems with parallelizing local tests on my workstation.

AdamGleave:
Works well now (test suite running in <3m) after I set PyTorch and JAX to only run single-threaded, so I think that was the main issue. We could probably get more speedups by tweaking settings slightly, but this is good enough for me. Merging.

PR #251, opened 2020-11-22
BC: Change policy_cls type add LR Sched.

shwang:
Adds LRScheduler functionality.

Also, switches `policy_cls` type to `Type[ActorCriticPolicy]`, which is guaranteed to have `policy.evaluate_actions` method. (`Policy` doesn't define this method).


shwang:
Note: By my understanding, the way this was previously coded with two separate optimizers made it so that the learning rate set over here had absolutely no bearing on BC. This seems like a potential hot spot for learning rate bugs, if someone thought that `policy_kwargs["lr_schedule"]` was a legitimate way to change the BC learning rate.

image

qxcv:
> At this point we can rely on policy.optimizer and no longer need to construct our own bc.optimizer.

The decision to use a separate optimiser was actually intentional, as discussed [in this thread](https://github.com/HumanCompatibleAI/imitation/pull/208#discussion_r453132396). Indeed, one reason was so that BC could be passed just a `BasePolicy`, and not an `ActorCriticPolicy`. I'd personally prefer to keep the separate optimizer in BC and give a warning if `policy_kwargs` contains a learning rate argument. In addition to the advantages mentioned in the other thread, using a separate optimiser would also mean that we don't need a separate `ConstantLRScheduler` that only exists to call the `.step()` method of the _real_ LR scheduler. Sorry if I miscommunicated this yesterday!

shwang:
This PR is stale, and `BC` has changed a lot since I opened this change. I'm going to close it for now, and assume that if ILR still uses LR scheduling, that the BC code for that will be merged in along with other ILR changes if at all.

PR #255, opened 2020-12-05
Remove device= kwarg from policies, update SB3

### LINKED ISSUE #254: 'device' in kwargs in bc causes error
ActorCriticPolicy which is the default policy doesn't have 'device' as input
 
https://github.com/HumanCompatibleAI/imitation/blob/7c2882697b92b1b93cc32b36d4d4bce8916cf880/src/imitation/algorithms/bc.py#L192-L197

https://github.com/HumanCompatibleAI/imitation/blob/7c2882697b92b1b93cc32b36d4d4bce8916cf880/src/imitation/algorithms/bc.py#L201-L202

https://github.com/DLR-RM/stable-baselines3/blob/e747e7e2b3f92a6650a7d69fd839544db4c5ffba/stable_baselines3/common/policies.py#L364-L383

qxcv:
Changes:

- No longer supplies the `device=` keyword argument to policies, since this was removed after SB3 0.8.
- Updates SB3 to 0.10.

Fixes #254.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ commands:
           # Build a wheel then install to avoid copying whole directory (pip issue #2195)
           command: |
             python setup.py sdist bdist_wheel
-            pip install --upgrade dist/imitation-*.whl
+            pip install --upgrade --force-reinstall --no-deps dist/imitation-*.whl
 
 jobs:
   lintandtype:
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ setup(
         "torch>=1.4.0",
         "tqdm",
         "scikit-learn>=0.21.2",
-        "stable-baselines3~=0.8.0",
+        "stable-baselines3~=0.10.0",
         "jax~=0.1.66",
         "jaxlib~=0.1.47",
         "sacred~=0.8.1",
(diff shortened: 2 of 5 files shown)


qxcv:
`unit-test` keeps failing on CircleCI with tracebacks involving this line:

```
/venv/lib/python3.7/site-packages/imitation/policies/serialize.py:45: in __init__
    device=policy.device,
```

However, this PR [specifically removes that line](https://github.com/HumanCompatibleAI/imitation/pull/255/files#diff-76e2df596b5bffc3dba4a86b7beef8400908f57fda746950875ff1cf9c23b172) from `serialize.py`. Any idea what might be going on?

AdamGleave:
> However, this PR [specifically removes that line](https://github.com/HumanCompatibleAI/imitation/pull/255/files#diff-76e2df596b5bffc3dba4a86b7beef8400908f57fda746950875ff1cf9c23b172) from `serialize.py`. I can't reproduce the test failures on my own machine. Any idea what might be going on?

Seems what's going on is that `pip` is not reinstalling `imitation` (we install a built wheel, rather than using developer mode, to make sure we test what end users will actually be using). It used to work with `--upgrade` but I think recent versions now require `--force-reinstall` if the version hasn't actually increased.

I figured this out by rerunning the job on CircleCI with SSH enabling, and poking around until I could replicate the failure. I'd highly recommend this for troubleshooting, it's quite opaque otherwise.

AdamGleave (approved):
LGTM

Issue #229, opened 2020-10-01
AIRL: double inputs preprocessing error


shwang:
AIRL is broken because its inputs are accidentally preprocessed twice by SB helper functions.

shwang:
This error has been around since the pytorch port.

AdamGleave:
Wow, nice catch. Are there any simple tests we can add that would catch issues like this?

shwang:
Should mention that @qxcv found this bug!

qxcv:
>  Should mention that @qxcv found this bug!

I also created the bug—a double achievement! :wink:

araffin:
can you link the issue? Is that the image processing issue?

qxcv:
This is not a bug in SB, I just made the mistake of applying the SB helper functions twice in AIRL (at least along some code paths).

shwang:
![](https://github.trello.services/images/mini-trello-icon.png) [AIRL bugfix](https://trello.com/c/6MAN0wuD/3-airl-bugfix)

andipeng:
where exactly is the bug? working through airl now and can't seem to find double processing of inputs

shwang:
@andipeng:

If I recall correctly:
`AIRL` observations, actions, and next_obs are processed once here in `AdversarialTrainer`:
https://github.com/HumanCompatibleAI/imitation/blob/cbcbbc78f7b191031bba30b4848e97b03377b07c/src/imitation/algorithms/adversarial.py#L321-L327

And then a second time here in `RewardNet._eval_reward` while evaluating the generator-policy-training reward:
https://github.com/HumanCompatibleAI/imitation/blob/cbcbbc78f7b191031bba30b4848e97b03377b07c/src/imitation/rewards/reward_nets.py#L164-L172

I'm 75% confident that this is what @qxcv was talking about to me in private conversation a few weeks ago?

Interestingly, there's a duplicate function [`DiscrimNet._eval_reward`](https://github.com/HumanCompatibleAI/imitation/blob/cbcbbc78f7b191031bba30b4848e97b03377b07c/src/imitation/rewards/discrim_nets.py#L161-L170), even though we manually determined that GAIL doesn't experience the same preprocessing duplication bug.

AdamGleave:
I did some more digging and I don't think they get preprocessed twice, although there's certainly duplicated code.

[...]

PR #256, opened 2020-12-09
Fix observation normalization

AdamGleave:
*Changes*

1) The current implementation of AdversarialTrainer only normalizes the `obs` component, not the `next_obs` component.

    Our base reward networks don't use `next_obs` by default, so this is not as bad as it seems, but we do use it for potential shaping. This is problematic since this inconsistent normalization makes it no longer be potential shaping, and does actually change the optimal policy.

2) `VecNormalize` was applied *after* reward wrapping. So `DiscrimNet.predict_train_reward` gets called on un-normalized observations. Yet we train the discriminator on normalized observations!

    I've swapped the order, so we normalize observations first. I introduce another `VecNormalize` to normalize (and clip) rewards but not observations after the wrapped reward.

3) I added keyword arguments that let you control whether to normalize observations or rewards. There are some environments in particular where normalizing observations is unnecessary or undesirable (e.g. Atari).

4) I deleted `venv_test` since it is not used anywhere and complicates the code. I think it is plausible that we should in fact be using it somewhere, it looks like `train_adversarial` used to use it for the final rollout stage but that got swapped to `venv_train` sometime during the port.

5) I also deleted `device` in `AdversarialTrainer` which is unused. Maybe we should make it do something, but if it does nothing its inclusion is just confusing.

*Benchmark*

Running `train_adversarial` with the default config on CartPole, I got:
  - Before the change
    - GAIL: Seed 1: 379 SD 33.5. Seed 2: 500 SD 0.0. Seed 3: 381 SD 22.8.
    - AIRL: Seed 1: 8.73 SD 0.54. Seed 2: 8.84 SD 0.49. Seed 3: 9.17 SD 0.76.
  - After the change
    - GAIL: Seed 1: 438 SD 26. Seed 2: 500 SD 0.0. Seed 3: 500.0 SD 0.0.
    - AIRL: Seed 1: 9.5 SD 0.70 (but it did peak at around ~100 part way through training). Seed 2: 497 SD 12.8. Seed 3: 18.52 SD 6.5.

Of course `CartPole-v1` is a stupid evaluation environment since if the discriminator happens to be the right sign then RL will learn the right policy. (I would not be surprised if this is why it works with AIRL some of the time given how high variance this is.) Still, the changes have certainly not made training performance any worse, and it seems like it may be a bit more stable.

I tried to get it working with `seals/CartPole-v0` which is fixed horizon but couldn't. Might just need more timesteps or other hyperparameters tweaked. Since even RL training on the ground-truth is a lot slower in that environment.

### DIFF
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -44,13 +44,14 @@ class AdversarialTrainer:
         n_disc_updates_per_round: int = 2,
         *,
         log_dir: str = "output/",
+        normalize_obs: bool = True,
+        normalize_reward: bool = True,
         disc_opt_cls: Type[th.optim.Optimizer] = th.optim.Adam,
         disc_opt_kwargs: Optional[Mapping] = None,
         gen_replay_buffer_capacity: Optional[int] = None,
         init_tensorboard: bool = False,
         init_tensorboard_graph: bool = False,
         debug_use_ground_truth: bool = False,
-        device: Union[str, th.device] = "auto",
     ):
         """Builds AdversarialTrainer.
 
@@ -81,6 +82,8 @@ class AdversarialTrainer:
             n_discrim_updates_per_round: The number of discriminator updates after each
                 round of generator updates in AdversarialTrainer.learn().
             log_dir: Directory to store TensorBoard logs, plots, etc. in.
+            normalize_obs: Whether to normalize observations with `VecNormalize`.
+            normalize_reward: Whether to normalize rewards with `VecNormalize`.
             disc_opt_cls: The optimizer for discriminator training.
             disc_opt_kwargs: Parameters for discriminator training.
             gen_replay_buffer_capacity: The capacity of the
@@ -153,17 +156,23 @@ class AdversarialTrainer:
 
         self.venv_buffering = wrappers.BufferingWrapper(self.venv)
         self.venv_norm_obs = vec_env.VecNormalize(
-            self.venv_buffering, norm_reward=False
+            self.venv_buffering,
+            norm_reward=False,
+            norm_obs=normalize_obs,
         )
 
         if debug_use_ground_truth:
             # Would use an identity reward fn here, but RewardFns can't see rewards.
-            self.venv_train = self.venv
+            self.venv_wrapped = self.venv_norm_obs
+            self.gen_callback = None
         else:
             self.venv_wrapped = reward_wrapper.RewardVecEnvWrapper(
                 self.venv_norm_obs, self.discrim.predict_reward_train
             )
-        self.venv_train = vec_env.VecNormalize(self.venv_wrapped, norm_obs=False)
+            self.gen_callback = self.venv_wrapped.log_callback
+        self.venv_train = vec_env.VecNormalize(
+            self.venv_wrapped, norm_obs=False, norm_reward=normalize_reward
+        )
 
         self.gen_algo.set_env(self.venv_train)
 
@@ -271,7 +280,9 @@ class AdversarialTrainer:
                 **learn_kwargs,
             )
             self._global_step += 1
-            self.venv_wrapped.log_callback(logger)
+
+            if self.gen_callback:
+                self.gen_callback(logger)
 
         gen_samples = self.venv_buffering.pop_transitions()
         self._gen_replay_buffer.store(gen_samples)
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -271,6 +271,7 @@ class AdversarialTrainer:
                 **learn_kwargs,
             )
             self._global_step += 1
+            self.venv_wrapped.log_callback(logger)
 
         gen_samples = self.venv_buffering.pop_transitions()
         self._gen_replay_buffer.store(gen_samples)
--- a/src/imitation/algorithms/adversarial.py
+++ b/src/imitation/algorithms/adversarial.py
@@ -23,17 +23,16 @@ class AdversarialTrainer:
     venv: vec_env.VecEnv
     """The original vectorized environment."""
 
+    venv_norm_obs: vec_env.VecEnv
+    """Like `self.venv`, but wrapped with `VecNormalize` normalizing the observations.
+
+    These statistics must be saved along with the model."""
+
     venv_train: vec_env.VecEnv
     """Like `self.venv`, but wrapped with train reward unless in debug mode.
 
     If `debug_use_ground_truth=True` was passed into the initializer then
     `self.venv_train` is the same as `self.venv`."""
-
-    venv_test: vec_env.VecEnv
-    """Like `self.venv`, but wrapped with test reward unless in debug mode.
-
-    If `debug_use_ground_truth=True` was passed into the initializer then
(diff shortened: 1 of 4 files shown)


AdamGleave:
OK, take those benchmarks with a gigaton of salt. I found that by flipping the sign in `DiscrimNet._eval_reward` I can make AIRL reliably get >300 and GAIL <20. Who knew the sign would be such an important hyperparameter!

I think we really need to benchmark in some fixed-horizon tasks to get anything meaningful. I wonder if we should even refuse to run on variable-horizon environments unless the user sets the `shoot_self_in_foot` flag to true?

AdamGleave:
Results from ```parallel --header : python -m imitation.scripts.train_adversarial with env_name=seals/CartPole-v0 rollout_path=data/expert_models/cartpole_0/rollouts/final.pkl total_timesteps=1e6 {algo} seed={seed} log_dir={algo}/{seed}/ ::: algo gail airl ::: seed 0 1 2 3 4``` on current master (https://github.com/HumanCompatibleAI/imitation/commit/23e7c150074f522715ef839d6609175076cb9d0f) and this PR (https://github.com/HumanCompatibleAI/imitation/pull/256/commits/343eb25c551909c92af50a6c6d0162d4b31e1cb1)

Reporting mean ground-truth return means and SDs (`monitor_return_mean` and `monitor_return_sd`) over 56 imitation trajectories. (So S.E. is about 1/7th the SD.)

master GAIL (before this change)

| Seed | Mean | SD    |
| ------- | -------- | ------ |
| 0       | 500.0  | 0.0   |
| 1       | 11.93  | 0.92 |
| 2       | 500.0  | 0.0   |
| 3       | 22.4    | 9.58 |
| 4       | 500.0  | 0.0   |

image


This PR GAIL

| Seed | Mean | SD    |
| ------- | -------- | ------ |
| 0       | 500     | 0.0   |
| 1       | 480.8  | 19.76 |
| 2       | 500.0  | 0.0   |
| 3       | 500.0  | 0.0   |
| 4       | 500.0  | 0.0   |

image

master AIRL

| Seed | Mean | SD    |
| ------- | -------- | ------ |
| 0       | 8.52    | 0.78 |
| 1       | 456.9  | 59.9 |
| 2       | 500.0  | 0.0   |
| 3       | 8.5      | 0.96 |
| 4       | 385.1  | 49.9 |

image

This PR AIRL

| Seed | Mean | SD    |
| ------- | -------- | ------ |
| 0       | 500.0  | 0.0   |
| 1       | 242.7  | 10.12 |
| 2       | 306.8  | 8.37   |
| 3       | 500.0  | 0.0   |
| 4       | 500.0  | 0.0   |

image

Full logs attached.

Changes seem to have significantly stabilized training, making the minimum mean return across seeds much higher. However, the maximum return achieved is unchanged at least in this simple environment.

Training is slow: basically flatlines for the first 100,000 timesteps. Reward oscillates during training. For GAIL the size of the oscillations seem to decrease over time. It looks like it actually oscillates more over time after this PR than before, but there's less variance between seeds. AIRL training curves look like more extreme version of GAIL. Most seeds don't work on master, all train somewhat with this PR, but one flatlines for 600,000 timesteps! Lots of oscillation, seems to be slow to decay and may have plateaued towards the end.

Overall seems like a significant improvement but I feel we should be able to do better, especially on a simple task like CartPole.

AdamGleave:
I may try and make some further improvements in other PRs but I think this is now ready for review @qxcv @shwang

shwang:
Great, I'm taking a look today

shwang (approved):
Thanks for these catches and the comparison runs -- good to see an improvement, sad to see that there were still lingering VecNormalization problems!

I've confirmed that the diff itself makes sense, that it matches the PR description, and verified the problems mentioned in the PR. 

I didn't check whether variable name changes match with things outside the diff, etc. because I'm swamped right now with grad apps stuff right now.

If it is helpful, I can give this PR more thorough look later.

shwang:
We discussed this a while ago, but to refresh, it seems like all this normalization logic would be automatically solved for us (except reward normalization), and serialization would be greatly simplified if there were an SB3 policy that included VecNorm-style normalization?

The batchnorm (mean=0, std=1) solution you [mentioned](https://github.com/HumanCompatibleAI/imitation/issues/229#issuecomment-742539096) seems like a promising approach.

src/imitation/algorithms/adversarial.py
```diff
@@ -23,18 +23,17 @@ class AdversarialTrainer:
...
+
+    These statistics must be saved along with the model."""
+
     venv_train: vec_env.VecEnv
     """Like `self.venv`, but wrapped with train reward unless in debug mode.
 
     If `debug_use_ground_truth=True` was passed into the initializer then
     `self.venv_train` is the same as `self.venv`."""
 
-    venv_test: vec_env.VecEnv
```
shwang, line 32: This should be fine to remove.

AdamGleave:
Thanks for the review, especially in these busy times!

> We discussed this a while ago, but to refresh, it seems like all this normalization logic would be automatically solved for us (except reward normalization), and serialization would be greatly simplified if there were an SB3 policy that included VecNorm-style normalization?

We have three places where normalization happens:
  1. Normalization of observations for the generator (policy)
  2. Normalization of observations for the discriminator (reward, for AIRL);
  3. Normalization of reward (from discriminator).

Making a policy wrapper to normalize observations would only handle the first case. We could make a separate discriminator wrapper that normalizes observations. I think this would be meaningfully different, though. Right now observation normalization statistics are only updated on-policy from fresh samples from the environment. A discriminator wrapper would presumably instead see samples we use during training, which would be 50% expert, and could be repeated (if sampled from a replay buffer).

My guess (~80%) is both approaches would achieve similar performance, with ~10% chance that one performs better than the other (but I don't know which!)

I discussed policy normalization wrappers with SB3 maintainers a while back and it seems unlikely to make it in. But of course we can implement it in `imitation` if we think it'd be simpler for our particular use case. I think I weakly favor this but it doesn't seem a priority (compared to e.g. improving algorithm performance or better docs)

> > The batchnorm (mean=0, std=1) solution you [mentioned](https://github.com/HumanCompatibleAI/imitation/issues/229#issuecomment-742539096) seems like a promising approach.

Thanks, hope to look into it soon!

qxcv (approved):
Also looks good to me, although I'm also sceptical that normalising observations with the running mean is more helpful than simpler alternatives (such as rescaling using the `Box` bounds, and/or using batch/layer norm).

AdamGleave:
> Also looks good to me, although I'm also sceptical that normalising observations with the running mean is more helpful than simpler alternatives (such as rescaling using the `Box` bounds, and/or using batch/layer norm).

Thanks for the review!

FYI I think a decent number of environments don't have Box bounds, e.g.:

```python
import gym
env = gym.make('HalfCheetah-v2')
print(env.observation_space.low)
```

gives:

```python
array([-inf, -inf, -inf, -inf, -inf, -inf, -inf, -inf, -inf, -inf, -inf,
       -inf, -inf, -inf, -inf, -inf, -inf])
```

Admittedly `VecNormalize` is a heavy weight solution to that problem. (You could also just renormalize based on static statistics, e.g. from the expert demos or a rollout of a random policy.)

PR #257, opened 2020-12-18
Migrate to new SB callback API

AdamGleave:
This PR replaces the old-style function callback used for SB `train` with the new-style class callback. This has the following semantic changes:
    - `policy_save_interval` now controls how many *timesteps* the policy gets saved, not the number of epochs. (The functional callback was de facto based on something approximately equal to timesteps, since it got called on each rollout timestep as well, but that was not the intention nor the old-style behaviour.)
    - `RewardVecEnvWrapper` callback now only gets called at the start of each rollout, which (at least for PPO) should correspond with each time the log gets printed. Previously it was getting called at each timestep.
    - I deleted the functionality to save rollouts except at the end of training. The way it was previously implemented I think could interfere with training, since it does rollouts in the *same* environment as training. To avoid this we'd need to use a separate evaluation environment, with `VecNormalize` statistics set up to match, which is possible but non-trivial. The functionality is disabled by default and I've not seen us ever use it, so it seemed easiest to delete it. But I can implement it properly if there's enough demand.

I haven't done any benchmarking but this should speed things up a bit, and more importantly avoid creating GBs of near-identical policy checkpoints each time we run this script :)

### DIFF
--- a/experiments/download_models.sh
+++ b/experiments/download_models.sh
@@ -49,4 +49,4 @@ elif [[ -d "${DATA_DIR}" ]]; then
   rm -r "${DATA_DIR}"
 fi
 
-aws s3 sync "${FLAGS[@]}" s3://shwang-chai/public/data/ "${DATA_DIR}"
+aws --no-sign-request s3 sync "${FLAGS[@]}" s3://shwang-chai/public/data/ "${DATA_DIR}"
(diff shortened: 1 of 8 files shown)


qxcv (approved):
LGTM

Deepakgthomas:
Hello @AdamGleave ,

Firstly, this repository looks really cool! 

I came across this page and wanted to know about your modifications to the rollout algorithm. I see that you wrote the following in your post, " _I deleted the functionality to save rollouts except at the end of training. The way it was previously implemented I think could interfere with training, since it does rollouts in the same environment as training. To avoid this we'd need to use a separate evaluation environment ..._" Could you please elaborate about how this modification helps the PPO algorithm? 

Thank you!

PR #263, opened 2021-01-12
logger: Fix behavior outside of accumulate_means

shwang:
Calling `logger.record` outside of a `with logger.accumulate_means`
context should lead to a `record`, not a `record_means` call.

h/t @decodyng.


shwang:
Something's breaking with jaxlib right now. I"ll give this a look

decodyng:
@shwang Does the jaxlib issue you mentioned only impact imitation-specific tests, or does it also mean I should wait until it's fixed to merge it into the branch used for ILR?

shwang:
Imitation-specific. Relevant Tests, lint, etc are passing on my local
machine.

shwang:
Problem should be fixed by #265

tests/test_logger.py
```diff
@@ -25,6 +25,7 @@ def _compare_csv_lines(csv_path: str, expect: dict):
 
 def test_no_accum(tmpdir):
     logger.configure(tmpdir, ["csv"])
+    sb_logger.record("A", -1)
```
shwang, line 28: Adding this line checks that the recorded "A": -1 is overwritten by  "A": 1 in the next line.

Previously, the observed value would be the mean of these two values (0) instead.

AdamGleave: This might be worth adding as a Python comment?

AdamGleave (approved):
LGTM

tests/test_logger.py
shwang: added

PR #265, opened 2021-01-22
Bump Jax version to avoid broken import

shwang:
Also print output of `pip freeze --all` in CircleCI.

See https://github.com/google/jax/issues/5374


AdamGleave (approved):
LGTM

PR #273, opened 2021-03-12
compute_train_stats: Fix logits passed in as proba

shwang:
Fixes an ~"negative probability out-of-bounds" error I ran into when I was training.

See: https://pytorch.org/docs/stable/distributions.html#bernoulli for func signature


AdamGleave (approved):
LGTM.

This seems like a badly designed function: `Bernoulli(probs, logits)` is illegal (can only specify one), you *have* to specify `logits` via keyword argument. Would have been better if it had been `Bernoulli(*, probs, logits)` to force people to always use keyword arguments. Oh well.

PR #276, opened 2021-04-02
Fix deprecated np.bool and np.int aliases

shwang:
np.bool and np.int are just aliases for bool and int. I'm getting
a deprecation warning on my machine, so doing a quick find and
replace.

https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

### DIFF
--- a/tests/test_buffer.py
+++ b/tests/test_buffer.py
@@ -69,7 +69,7 @@ def test_buffer(capacity, chunk_len, sample_shape):
 @pytest.mark.parametrize("chunk_len", [1, 4, 9])
 @pytest.mark.parametrize("obs_shape", [(), (1, 2)])
 @pytest.mark.parametrize("act_shape", [(), (5, 4, 4)])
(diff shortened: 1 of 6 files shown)


tests/test_buffer.py
```diff
@@ -95,8 +95,8 @@ def test_replay_buffer(capacity, chunk_len, obs_shape, act_shape, dtype):
         assert buf.size() == min(i, capacity)
         assert buf._buffer._idx == i % capacity
 
-        dones = np.arange(i, i + chunk_len, dtype=np.int32) % 2
-        dones = dones.astype(np.bool)
+        dones = np.arange(i, i + chunk_len, dtype=int32) % 2
```
shwang, line 98:
```suggestion
        dones = np.arange(i, i + chunk_len, dtype=np.int32) % 2
```

PR #278, opened 2021-04-15
make_vec_env: Don't wrap RolloutInfoWrapper by default

shwang:
This PR modifies `make_vec_env` to stop wrapping each Env with `RolloutInfoWrapper`, and modifies `expert_demos`.

This wrapper increases the size of the info dict substantially
because it appends all episode observatations to the into dict with
done=True. It is only used by the expert_demos script to save
original (unnormalized) observations and rewards.

### DIFF
--- a/src/imitation/analysis/mountain_car_plots.py
+++ b/src/imitation/analysis/mountain_car_plots.py
@@ -210,7 +210,7 @@ def plot_reward_vs_time(
         X = []
         Y = []
         for traj in trajs_list:
-            T = len(traj.obs)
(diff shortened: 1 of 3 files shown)


AdamGleave:
Do we need to introduce a separate flag? Can't we just use the existing `post_wrappers` flag?

shwang:
`post_wrappers` seems like a good solution here actually. (Forgot that it configures Env wrapper rather than VecEnv wrappers)

AdamGleave (approved):
LGTM

PR #286, opened 2021-06-10
Add a on_batch_end Callback

decodyng:
Add a `on_batch_end` Callback that works in an identical fashion to our `on_epoch_end` callback.


shwang (approved):
LGTM upon test pass (should be working now that #288 is merged)

PR #290, opened 2021-06-30
Error if Data Loader Ends Early

decodyng:
A simple fix to give a more clean and comprehensible error in the edge case where your data iterator ends prior to hitting the number of batches you have asked BC to train for. In the current state of the code, this situation results in an infinite loop of `epoch_num` incrementing upwards, with no explicit error

### DIFF
--- a/src/imitation/algorithms/bc.py
+++ b/src/imitation/algorithms/bc.py
@@ -112,6 +112,7 @@ class EpochOrBatchIteratorWithProgress:
             )
 
         with contextlib.closing(display):
+            prior_epoch_batches = batch_num
             while True:
                 update_desc()
                 for batch in self.data_loader:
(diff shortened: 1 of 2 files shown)


AdamGleave (commented):
Thanks for the PR. This seems like it does a bit more bookkeeping than is really needed. I think we can just do:

```
            while True:
               got_data = False
                update_desc()
                for batch in self.data_loader:
                        got_data = True
@@ -132,6 +133,13 @@ def update_desc():
                        display.update(1)
                        if batch_num >= self.n_batches:
                            return
                if not got_data:
                    raise AssertionError(
                        "Data loader returned no data -- did it reset correctly?"
                    )
```

WDYT?

decodyng:
I think something in this vein would make sense, though I'd prefer to update the error message to something like: 
```
raise AssertionError(
                        f"Data loader returned no data after {num_batches} batches, in epoch {num_epochs} -- did it reset correctly?"
                    )
```
Just to clarify, for someone who potentially only gets the error message and doesn't look at the full logs, that it wasn't (necessarily) the case that the Data Loader returned no data period, just that it ran out of data at some point before the full course of training had run.

AdamGleave:
> I think something in this vein would make sense, though I'd prefer to update the error message to something like:
> 
> ```
> raise AssertionError(
>                         f"Data loader returned no data after {num_batches} batches, in epoch {num_epochs} -- did it reset correctly?"
>                     )
> ```
> 
> Just to clarify, for someone who potentially only gets the error message and doesn't look at the full logs, that it wasn't (necessarily) the case that the Data Loader returned no data period, just that it ran out of data at some point before the full course of training had run.

Sure, a more detailed error message seems better.

AdamGleave (approved):
LGTM

PR #291, opened 2021-07-13
quickstart.sh: Various cleanup and ease-of-use improvements

shwang:
Before this PR, a "symlink already exists" error occurred when running `quickstart.sh` a second time,
due to `quickstart.sh` writing `train_expert_demos` output to the same directory and
`imitation.util.sacred.build_sacred_symlink` unsuccessfully to overwrite the Sacred symlink generated
by the first run.

This PR modifies `imitation.util.sacred.build_sacred_symlink` to successfully overwrite the symlink
by first removing it if it already exists.

Other minor changes made along the way:
  * Make examples/quickstart.sh an executable file.
  * Use `pathlib` instead of `os.path` in areas near the symlink fix in `imitation.util.sacred`.
  * Add `bash -e` flag to `tests/test_examples.sh::test_run_example_sh_scripts` so that example scripts
    can error on any script line during test.


src/imitation/util/sacred.py
```diff
@@ -67,24 +71,33 @@ def filter_subdirs(
...
+    # Path.symlink_to errors if the symlink already exists. In our case, we actually
+    # want to override the symlink to point to the most recent Sacred dir. The
+    # examples/quickstart.sh script fails without this check when run a second time.
+    if symlink_path.is_symlink():
+        symlink_path.unlink()
+
     # Use relative paths so we can mount the output directory at different paths
     # (e.g. when copying across machines).
-    os.symlink(target_path, symlink_path, target_is_directory=True)
+    symlink_path.symlink_to(target_path, target_is_directory=True)
```
AdamGleave, line 93: Potential race condition here if something else creates a symlink in the background, but I don't think that matters too much and can't see a way to avoid it using pathlib.

src/imitation/util/sacred.py
```diff
@@ -67,24 +71,33 @@ def filter_subdirs(
...
         return
-    symlink_path = os.path.join(log_dir, "sacred")
-    target_path = os.path.relpath(sacred_dir, start=log_dir)
+    symlink_path = pathlib.Path(log_dir, "sacred")
+    target_path = pathlib.Path(os.path.relpath(sacred_dir, start=log_dir))
+
+    # Path.symlink_to errors if the symlink already exists. In our case, we actually
+    # want to override the symlink to point to the most recent Sacred dir. The
+    # examples/quickstart.sh script fails without this check when run a second time.
+    if symlink_path.is_symlink():
```
AdamGleave, line 88: There's a case where this exists and is not a symlink, in which case we don't unlink it, and it'll fail on the next line. Is this WAI? If so, can you add a comment to that effect (e.g. that if it's not a symlink something other than this script created it and we want to fail)? If not, you could skip the if statement, and just do `symlink_path.unlink(missing_ok=True)`.

src/imitation/util/sacred.py
```diff
@@ -1,9 +1,13 @@
 import json
 import os
+import pathlib
 import warnings
 from typing import Any, Callable, List, NamedTuple, Union
 
-import sacred
```
AdamGleave, line 6: Why is `sacred_run` better than `sacred.run`?

AdamGleave (commented):
Thanks for implementing this, mostly looks good, left a few comments (mostly clarifying questions).

src/imitation/util/sacred.py
shwang: Yes, this is WAI. I'll note that in the comment

src/imitation/util/sacred.py
shwang: Hmm looks like I actually misread Google Style guide -- switching this back

AdamGleave (approved):
LGTM

PR #292, opened 2021-07-14
Move Jax dependencies to setup extra

shwang:
This PR moves the `jax` and `jaxlib` dependencies (used only by
`imitation.algorithms.tabular_irl`) from the main package
requirements to a new "jax" extra.

It also adds a warning to `tabular_irl.py` that is displayed when the
submodule is imported but `jax` cannot be imported, and adds
new documentation to the same submodule explaining how to install
the jax extra.

`tests/test_tabular.py` is also updated to skip tests when
Jax isn't installed.

`pip install jax` is not available for Windows machines, and this
change might make `imitation` Windows-compatible (haven't checked
personally.)

### DIFF
--- a/src/imitation/algorithms/tabular_irl.py
+++ b/src/imitation/algorithms/tabular_irl.py
@@ -7,17 +7,45 @@ PyTorch/TensorFlow, and some code for simple reward models.
 
 .. _PhD thesis:
     http://www.cs.cmu.edu/~bziebart/publications/thesis-bziebart.pdf
+
+.. note::
+    Our current implementation of MCE IRL uses Jax, which is not bundled with the
+    default ``pip`` installation of ``imitation`` because Jax is incompatible with
+    Windows. You can install the Jax dependencies used by MCE IRL via
+    ``pip install imitation[jax]``.
+
+    We are considering porting MCE IRL from Jax to PyTorch in a future release.
 """
 
 import abc
 import logging
+import warnings
 
-import jax
-import jax.experimental.stax as jstax
-import jax.numpy as jnp
-import jax.random as jrandom
 import numpy as np
 import scipy
+
+try:
+    # pytype: disable=import-error
+    import jax
+    import jax.experimental.stax as jstax
+    import jax.numpy as jnp
+    import jax.random as jrandom
+
+    # pytype: enable=import-error
+
+except ImportError:
+    msg = (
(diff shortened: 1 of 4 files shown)


shwang:
I'm not sure what would be the best way to add codecov for these dependencies -- add more CircleCI jobs that run on an virtualenv that doesn't include the `jax` extra? Maybe write a script that temporarily hides the `jax` modules a la https://stackoverflow.com/questions/16490055/temporarily-removing-a-python-module

I'm also OK with not covering this code for now.

src/imitation/algorithms/tabular_irl.py
```diff
@@ -7,18 +7,46 @@
...
+except ImportError:
+    msg = (
+        f"Failed to import module {__name__} because Jax dependency is not installed. "
+        "See module docstring for more information on installation and OS "
+        "compatibility."
+    )
+    # ImportWarning is more appropriate than UserWarning here, but ImportWarning
+    # has been ignored by default since Python 3.7:
+    # https://docs.python.org/3/library/devmode.html#effects-of-the-python-development-mode
+    warnings.warn(msg)
```
AdamGleave, line 45: Why do we want this to be a warning? Is there a use case where we need to import `tabular_irl` but not use any of the functions? I'm inclined to just re-raise ImportError, but including the more informative error `msg` above.

AdamGleave:
> I'm not sure what would be the best way to add codecov for these dependencies -- add more CircleCI jobs that run on an virtualenv that doesn't include the `jax` extra? Maybe write a script that temporarily hides the `jax` modules a la https://stackoverflow.com/questions/16490055/temporarily-removing-a-python-module
> 
> I'm also OK with not covering this code for now.

I'm OK with not testing the pretty error message / warning in the absence of JAX. If we do mess that code up, it'll still fail, just in a less obvious way -- not great, but way less bad then e.g. a subtle algorithm failure.

shwang:
I'll add a pragma no cover and explanation to the warning message

src/imitation/algorithms/tabular_irl.py
shwang: That sounds good to me. There's no reason to import this submodule passively (nothing other than `tests/` imports this module)

AdamGleave (approved):
LGTM

PR #294, opened 2021-07-15
util.sacred.filter_subdirs: Follow symlinks

shwang:
Recursively search through symlinks (previously ignored), like
those generated by `build_sacred_symlink`. This is useful for
building performance tables from directories that contain Sacred
symlinks but not the actual Sacred directory.


AdamGleave (approved):
LGTM

PR #296, opened 2021-07-15
reward_nets: Add easy hidden size configuration

shwang:
Previously, `base_reward_net` and `potential_net` only accepted
`nn.Module`. Now they also accept `Sequence[int]`, which can be
used to build a default MLP with particular hidden sizes.


AdamGleave (approved):
LGTM

PR #297, opened 2021-07-16
Rename `train_ex` to `train_adversarial_ex`

shwang:
In anticipation of a later PR that will add a `train_bc` experiment, rename `train_ex` to `train_adversarial_ex`.


AdamGleave (approved):
Rename seems fine although I'm wondering if we can add support for BC and AIRL/GAIL into single train script instead? Messaged you on slack about this.

PR #298, opened 2021-07-16
expert_demos, train_adversarial: Add seals configs

shwang:
Merge #297 first


shwang:
One of the Ray tests failed 2 out of the 4 times I launched it, likely due to low memory from parallel workers simultaneously launching memory-intensive tests. If this continue to be an issue, we could consider marking all of the high-memory jobs as `@pytest.mark.mem_intensive` and running these tests serially, on a separate CircleCI job.

Test: `tests/test_scripts.py::test_parallel_train_adversarial_custom_env`
Error: `ValueError: Attempting to cap object store memory usage at 62453760 bytes, but the minimum allowed is 78643200 bytes.
`
https://github.com/ray-project/ray/issues/5839 says that this is caused by low memory.

AdamGleave (approved):
LGTM

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -129,12 +129,26 @@ def cartpole():
...
 @train_adversarial_ex.named_config
 def mountain_car():
     env_name = "MountainCar-v0"
     rollout_hint = "mountain_car"
 
 
+@train_adversarial_ex.named_config
+def seals_mountain_car():
+    env_name = "seals/MountainCar-v0"
+    rollout_hint = "mountain_car"  # TODO: Use seals/MountainCar-v0 rollouts.
```
AdamGleave, line 149: good to add issue number or username to TODO so we can get more info about it in the future. e.g. `TODO(shwang): ...`

AdamGleave (approved):
LGTM

src/imitation/scripts/config/train_adversarial.py
shwang: thanks for catching, fixed

PR #299, opened 2021-07-17
Don't set policy's environment in generate_trajectories

ejnnr:
Previously, `generate_trajectories` could change the environment attached to the `policy` if it was a `BaseAlgorithm`. This seems like an unwanted side effect and I think it's unnecessary (`generate_trajectory` only uses `policy.predict`, which doesn't require the environment to be set).

I did keep the check whether observation/action space match, just put it into `generate_trajectories` directly instead of using `policy.set_env`.


AdamGleave (approved):
LGTM. I've triggered tests to run on CircleCI, can merge once they pass.

PR #304, opened 2021-08-05
Remove code coverage for Sacred config scripts

shwang:
Code coverage checks currently fail whenever we add a new
Sacred config or namedconfig function because the body of these
functions is never meant to be executed. This patch excludes the
files containing these functions from CodeCov checks.

(Example of failure: #298)

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -12,6 +12,7 @@ coverage:
         paths:
           - "src/imitation/envs/examples/"
           - "src/imitation/scripts/"
+          - "!src/imitation/scripts/config"
       tests:
         # Should not have dead code in our tests
         target: 100%
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -30,6 +30,7 @@ coverage:
           - "examples/"
           - "src/imitation/envs/examples/"
           - "src/imitation/scripts/"
+          - "!src/imitation/scripts/config"
       tests:
         target: 100%
         paths:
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -12,6 +12,7 @@ coverage:
         paths:
           - "src/imitation/envs/examples/"
           - "src/imitation/scripts/"
+          - "!src/imitation/scripts/config"
```
AdamGleave, line 15: Why do we need to remove it here? The target is fixed at 0% for auxiliary so it'll never fail the check. Seems useful to still at least calculate the code coverage change.

AdamGleave (changes requested):

.codecov.yml
shwang: Ah, that's right. I was confusing it with the project auxiliary

shwang:
I was a bit sloppy and merged in the codecov ignore changes from master with #298 (forgot that this PR should be merged first). The diff of this PR relative to master now only removes the superfluous ignore.

AdamGleave (approved):
LGTM

AdamGleave:
> I was a bit sloppy and merged in the codecov ignore changes from master with #298 (forgot that this PR should be merged first).

OK. Might want to consider using something like https://github.com/z0al/dependent-issues in the future to stop accidental merges? (Or just keeping the original branch clean.)

PR #305, opened 2021-08-06
bc: Option for hiding progress bar

shwang:
Add `progress_bar` parameter to `BC.train` which toggles
whether we show a tqdm progress bar.


AdamGleave (approved):
LGTM

PR #306, opened 2021-08-07
Deprecate old data format, update old test data, and provide data update script

shwang:
Start printing a DeprecationWarning when we call `imitation.data.load` on
an old-format demonstration pickle, and also print a script that can be
used to update that pickle file in place. Use that script to update
our old pickles in this repository.


src/imitation/scripts/convert_trajs_in_place.py
```diff
@@ -0,0 +1,42 @@
+"""
+This script takes as command-line input multiple paths to pickle files containing
+Sequence[imitation.types.TrajectoryWithRew] or Sequence[imitation.old_types.Trajectory].
+It overwrites each path with a new pickle where the demonstration data is saved as
+Sequence[TrajectoryWithRew].
```
AdamGleave, line 5: Is this an `imitation.types.TrajectoryWithRew`? Docstring currently makes it sound like it takes types A or B and converts them to C, but I think it's actually A or B and converts them to A?

AdamGleave (approved):
LGTM, thanks for adding test cases and the warning

src/imitation/scripts/convert_trajs_in_place.py
shwang: That makes sense -- expanding this reference

PR #307, opened 2021-08-07
train_bc script and associated changes to BC

shwang:
Adds a Sacred script for BC.

Other changes:
* Adds to BC.train the `log_rollouts_venv` and `log_rollouts_n_episodes` parameters, which can be used to log rollout statistics every `log_interval` updates.
  * I used these parameters for experiments earlier, but perhaps a callback would be the better choice here
* BC.save_policy: use `AnyPath` type instead of `str`


src/imitation/algorithms/bc.py
```diff
@@ -366,10 +375,23 @@ def train(
                 for stats in [stats_dict_it, stats_dict_loss]:
                     for k, v in stats.items():
                         logger.record(k, v)
+                # TODO(shwang): Maybe instead use a callback that can be shared between
+                # all algorithms' `.train()` for generating rollout stats.
+                if log_rollouts_venv is not None and log_rollouts_n_episodes > 0:
```
AdamGleave, line 380: This does seem like it might be a better solution. Does [EvalCallback](https://stable-baselines3.readthedocs.io/en/master/guide/callbacks.html#evalcallback) do what you need? It does log [episode reward and length](https://stable-baselines3.readthedocs.io/en/master/_modules/stable_baselines3/common/callbacks.html).

The flip-side is I think every other algorithm we have logs stuff like this without a callback? So perhaps it does make sense this is built into BC.

src/imitation/scripts/train_bc.py
```diff
@@ -0,0 +1,145 @@
...
+            statistics to stdout and Tensorboard.
+        log_rollouts_n_episodes: The number of rollout episodes generated for
+            training statistics every `log_interval` updates. If `venv` is None or
+            this argument is nonpositive, then no rollouts are generated.
+        n_episodes_eval: The number of final evaluation rollout episodes, if `venv` is
+            provided. These rollouts are used to generate final statistics saved into
+            Sacred results, which can be compiled into a table by
+            `imitation.scripts.analyze.analyze_imitation`.
+    """
+    assert action_space is not None
```
AdamGleave, line 67: `assert` is meant for internal consistency checks, not input validation; note they will be disabled when running with `-O`. raise ValueError instead if missing.

src/imitation/scripts/train_bc.py
```diff
@@ -0,0 +1,145 @@
...
+    assert observation_space is not None
+
+    log_dir = pathlib.Path(log_dir)
+    log_dir.mkdir(parents=True, exist_ok=True)
+    logging.info("Logging to %s", log_dir)
+
+    logger.configure(log_dir, ["tensorboard", "stdout"])
+    sacred_util.build_sacred_symlink(log_dir, _run)
+
+    assert expert_data_src_format in ("path", "trajectory")
```
AdamGleave, line 77: Redundant -- you check and raise ValueError later (good!)

src/imitation/scripts/train_bc.py
```diff
@@ -0,0 +1,145 @@
...
+    if expert_data_src_format == "path":
+        expert_trajs = types.load(expert_data_src)
+    elif expert_data_src_format == "trajectory":
+        # Convenience option for launching experiment from Python script with
+        # in-memory trajectories.
+        expert_trajs = expert_data_src
+    else:
+        raise ValueError(f"Invalid expert_data_src_format={expert_data_src_format}")
+
+    # Copied from scripts/train_adversarial -- refactor with "auto"?
```
AdamGleave, line 87: Should be a TODO comment?

src/imitation/scripts/train_bc.py
```diff
@@ -0,0 +1,145 @@
...
+        n_batches=n_batches,
+        log_interval=log_interval,
+        log_rollouts_venv=venv,
+        log_rollouts_n_episodes=log_rollouts_n_episodes,
+    )
+    model.save_policy(policy_path=pathlib.Path(log_dir, "final.th"))
+
+    print(f"Visualize results with: tensorboard --logdir '{log_dir}'")
+
+    # Later: Use auto env, auto stats thing with shared `env` and stats ingredient,
```
AdamGleave, line 122: Also should be a TODO comment.

tests/test_scripts.py
```diff
@@ -49,6 +61,14 @@ def test_main_console(script_mod):
         script_mod.main_console()
 
 
+def test_train_bc_main(tmpdir):
```
AdamGleave, line 64: This fails because the data file is missing -- you need to set the path to `CARTPOLE_TEST_POLICY_PATH`.

AdamGleave (approved):
Some minor changes needed but otherwise LGTM. Tests failing due to path error that should be fixed before merge.

src/imitation/scripts/train_bc.py
shwang: Makes sense, thanks

tests/test_scripts.py
shwang: ah that's right. Fixed!

src/imitation/algorithms/bc.py
shwang: EvalCallback does seem like a good fit -- adding it to the comment here. For better SB3 integration it could be nice to add Callbacks to every imitation alg's `train` someday.

PR #308, opened 2021-08-13
policies.serialize: For SB3 load/save use .zip

shwang:
We previously saved to a 'model.pkl' path instead of 'model.zip'
because this is what SB2 did. Since SB3, models have been saved
as ZIP files, not pickles, and we get confusing error messages
like "can't find 'model.pkl.zip'" if incorrectly we ask SB3 to load
from nonexistent 'model.pkl'.


shwang:
Oops, it looks like some tests/ bash scripts don't like the rename yet. Those should be quick spot fixes tomorrow that don't impact the overall PR

src/imitation/data/types.py
```diff
@@ -298,8 +298,9 @@ def save(path: str, trajectories: Sequence[TrajectoryWithRew]) -> None:
     """
     p = pathlib.Path(path)
     p.parent.mkdir(parents=True, exist_ok=True)
-    with open(path + ".tmp", "wb") as f:
+    tmp_path = "{path}.tmp"
```
shwang, line 301:
```suggestion
    tmp_path = f"{path}.tmp"
```

AdamGleave (approved):
LGTM

PR #309, opened 2021-08-13
Train dagger

shwang:


shwang:
Ah having the readthedocs.org hook here is really helpful. It helped me catch that the DAgger page was missing the new class!

src/imitation/algorithms/dagger.py
```diff
@@ -237,46 +275,55 @@ class DAggerTrainer:
...
         """
         # for pickling
-        self._init_args = locals()
-        self._init_args.update(bc_kwargs)
-        del self._init_args["self"]
-        del self._init_args["bc_kwargs"]
+        # self._init_args = locals()
+        # self._init_args.update(bc_kwargs)
+        # del self._init_args["self"]
+        # del self._init_args["bc_kwargs"]
```
shwang, line 304: I did some experimenting and it turned out that `self._init_args` wasn't necessary for getting the "pickle and load" `test_dagger.py` test to pass. Is this vestigal code, @qxcv?

src/imitation/algorithms/dagger.py
```diff
@@ -310,25 +358,39 @@ def _check_has_latest_demos(self):
                 f".get_trajectory_collector()"
             )
 
-    def _try_load_demos(self):
-        self._check_has_latest_demos()
         if self._last_loaded_round < self.round_num:
+            # TODO(shwang): Do we actually only want to load the most recent round's
+            #   data, or should we load all the data?
```
shwang, line 363: @qxcv: IIRC in DAgger we want to update BC on the entire aggregated dataset. However, it looks like we only load the most recent round's data here. Is this intentional?

If this is indeed the intended behvario, then we should change `self._last_loaded_round < self.round_num`  to `self._last_loaded_round != self.round_num`.

src/imitation/algorithms/dagger.py
```diff
@@ -391,30 +450,166 @@ def save_trainer(self) -> Tuple[str, str]:
...
+    def __init__(
+        self,
+        venv: vec_env.VecEnv,
+        log_dir: types.AnyPath,
+        expert_policy: policies.BasePolicy,
+        expert_trajs: Optional[List[types.Trajectory]] = None,
+        beta_schedule: Callable[[int], float] = None,
+        batch_size: int = 32,
+        bc_kwargs: Optional[dict] = None,
+    ):
```
shwang, line 494: Oops, missing `__init__` docstring here

src/imitation/data/rollout.py
```diff
@@ -124,8 +124,9 @@ def add_steps_and_auto_finish(
         zip_iter = enumerate(zip(acts, obs, rews, dones, infos))
         for env_idx, (act, ob, rew, done, info) in zip_iter:
             if done:
-                # actual obs is inaccurate, so we use the one inserted into step info
-                # by stable baselines wrapper
+                # When done[i] from VecEnv.step() is True, obs[i] is the first
+                # observation following reset() of the ith VecEnv, and
+                # info["terminal_observation"] is the actual final observation.
```
shwang, line 129: infos[i]["terminal_observation]

src/imitation/algorithms/dagger.py
```diff
@@ -391,30 +450,182 @@ def save_trainer(self) -> Tuple[str, str]:
...
+            logger.record("dagger/round_episode_count", round_episode_count)
+            logger.record("dagger/round_timestep_count", round_timestep_count)
+
+            # TODO(shwang): It looks like BC might start looping Tensorboard
+            #   back to x=0 with each new call to BC.train(). Consider adding a
+            #   `reset_tensorboard: bool = False` argument to BC.train() if this turns
+            #   out to be the case?
+
+            # `logger.dump` is called inside BC.train within the following fn call:
+            self.extend_and_update(bc_train_kwargs)
```
AdamGleave, line 630: I was initially confused this wasn't passing the data to `extend_and_update`. I now see we save the data to disk, then load it again. What's the point of this indirection? Checkpointing? Demonstrations bigger than RAM?

src/imitation/algorithms/dagger.py
```diff
@@ -310,25 +358,39 @@ def _check_has_latest_demos(self):
...
+        Note that these defaults set the `log_rollouts_venv` argument, meaning that
+        `BC.train` will periodically roll out episodes to log the average BC policy
+        return.
+        """
+        return dict(
+            n_epochs=4,
+            log_rollouts_venv=self.venv,
+        )
+
+    def extend_and_update(self, bc_train_kwargs: Optional[dict]) -> int:
```
AdamGleave, line 392: Why `dict` not `Mapping`?

src/imitation/scripts/train_dagger.py
```diff
@@ -0,0 +1,160 @@
...
+    # TODO(shwang): Copied from scripts/train_adversarial -- refactor with "auto"?
+    if n_expert_demos is not None:
+        if not len(expert_trajs) >= n_expert_demos:
+            raise ValueError(
+                f"Want to use n_expert_demos={n_expert_demos} trajectories, but only "
+                f"{len(expert_trajs)} are available."
+            )
+        expert_trajs = expert_trajs[:n_expert_demos]
+
+    # pytype is getting super confused about the type of expert_trajs and doesn't
```
AdamGleave: The problem is you made `DAggerTrainer` expect a `List` (when in fact any `Sequence` would do), so it actually didn't type check as you declared `expert_trajs` here to be a `Sequence`. I've fixed this now and removed the `# pytype: disable`. I'd in general suggest using the most generic type in parameter and return annotations to keep the API as assumption-free as possible.

tests/test_dagger.py
```diff
@@ -25,62 +29,146 @@ def test_beta_schedule():
 
 
 def test_traj_collector(tmpdir):
-    env = gym.make(ENV_NAME)
+    venv = vec_env.DummyVecEnv([lambda: gym.make(ENV_NAME)])
```
AdamGleave, line 32: A `venv` with a single environment is a bit of a degenerate case. Maybe use the existing `venv` fixture, and merhaps even make the number of environments parameterized?

tests/test_dagger.py
```diff
@@ -25,62 +29,146 @@ def test_beta_schedule():
...
     )
     collector.reset()
-    zero_action = np.zeros((), dtype="int")
-    obs, rew, done, info = collector.step(zero_action)
-    assert rew != 0
-    assert not done
-    assert isinstance(info, dict)
+    zero_acts = np.zeros((venv.num_envs,), dtype="int")
+    obs, rews, dones, infos = collector.step(zero_acts)
+    assert rews != 0
```
AdamGleave, line 46: I'm confused how this can work, normally `bool(np.array(...))` throws an error. Should this be `np.all(rews != 0)`?

tests/test_dagger.py
```diff
@@ -25,62 +29,146 @@ def test_beta_schedule():
...
+        )
     return dagger.DAggerTrainer(
-        env,
-        tmpdir,
-        beta_schedule,
-        optimizer_kwargs=dict(lr=1e-3),
+        venv,
+        scratch_dir=tmpdir,
+        beta_schedule=beta_schedule,
+        bc_kwargs={"optimizer_kwargs": dict(lr=1e-3)},
```
AdamGleave, line 100: Nit: I find it odd that there's two different ways of defining a dictionary used here.

tests/test_dagger.py
```diff
@@ -25,62 +29,146 @@ def test_beta_schedule():
...
+    venv,
+    beta_schedule,
+    expert_policy,
+    expert_trajs,
+):
+    return dagger.SimpleDAggerTrainer(
+        venv,
+        log_dir=tmpdir,
+        beta_schedule=beta_schedule,
+        bc_kwargs={"optimizer_kwargs": dict(lr=1e-3)},
```
AdamGleave, line 115: as line 100

AdamGleave (changes requested):
Thanks for the PR! Overall looks like definite improvement, I particularly like that we can now run DAgger experiments easily on synthetic data + have a script for it. As discussed outside GitHub, would be nice to eventually consolidate training scripts into one (or at least merge BC & DAgger which are very similar), but that does not need to take place in this PR.

I've left a variety of mostly minor comments; please request re-review when you've replied to / addressed them.

src/imitation/algorithms/dagger.py
shwang: Question for @qxcv

src/imitation/scripts/train_dagger.py
shwang: Oh nice, thanks for catching that!

tests/test_dagger.py
shwang: (probably bad port of test from Env to VecEnv)

shwang:
Thanks for your detailed review, Adam!

tests/test_dagger.py
shwang: Interesting edge-case here that allowed this to not error:
`array([1.], dtype=float32) != 0` => `array([True])`, and `bool(array[True]) is True`.

tests/test_dagger.py
shwang: Made both suggestions here

src/imitation/algorithms/dagger.py
```diff
@@ -310,25 +358,39 @@ def _check_has_latest_demos(self):
...
+                drop_last=False,
                 shuffle=True,
                 collate_fn=types.transitions_collate_fn,
             )
             self.bc_trainer.set_expert_data_loader(data_loader)
             self._last_loaded_round = self.round_num
 
-    def extend_and_update(self, **train_kwargs) -> int:
-        """Extend internal batch of data and train.
+    def _default_bc_train_kwargs(self) -> dict:
```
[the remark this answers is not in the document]
shwang: Changed this to an instance variable (can't be a "class constant" because one of the values if `self.venv`).

src/imitation/algorithms/dagger.py
qxcv: When you don't want to keep the process running while you wait for user input (e.g. because it could take several hours for the user to come back, and the run might last for a few days). So I guess checkpointing.

src/imitation/algorithms/dagger.py
```diff
@@ -75,65 +75,88 @@ def reconstruct_trainer(
...
         previously-saved one.
     """
-    checkpoint_path = os.path.join(scratch_dir, "checkpoint-latest.pt")
+    checkpoint_path = pathlib.Path(scratch_dir, "checkpoint-latest.pt")
     return th.load(checkpoint_path, map_location=utils.get_device(device))
 
 
-def _save_trajectory(
-    npz_path: str,
+def _save_dagger_demo(
```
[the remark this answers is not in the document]
qxcv: I probably wrote the original version of this. I don't know why I did it this way, and can't think of any obvious objection to changing it.

src/imitation/algorithms/dagger.py
qxcv: Context: I think my motivating application for this code was interactive IL with MTurk demonstrators. For that purpose it would be weird to need a background process running & communicate with IPC—instead you could just save some stuff to disk (or a DB or whatever) as requests come in, then periodically run a cron job to do incremental training.

src/imitation/algorithms/dagger.py
qxcv: That's right, it's safe to remove. Originally this class serialised itself using the "save all the kwargs and any mutable state" trick. After the Torch port it became possible to just pickle the whole thing, which is what the current code does.

src/imitation/algorithms/dagger.py
```diff
@@ -310,25 +358,39 @@ def _check_has_latest_demos(self):
...
+            #   data, or should we load all the data?
             transitions, num_demos = self._load_all_demos()
             logging.info(
                 f"Loaded {sum(num_demos)} new demos from {len(num_demos)} rounds"
             )
             data_loader = th_data.DataLoader(
                 transitions,
                 self.batch_size,
-                drop_last=True,
+                # If drop_last=True, then BC.train() fails on "no batches available"
```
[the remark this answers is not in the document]
qxcv: Yes, this is why I always set `drop_last=True`.

src/imitation/algorithms/dagger.py
qxcv: Yes, the intent is to load demos from all rounds.

src/imitation/scripts/config/train_dagger.py
```diff
@@ -0,0 +1,152 @@
...
+        venv = util.make_vec_env(env_name)  # Automatically initialized using `env_name`
+
+
+@train_dagger_ex.config
+def default_train_duration(bc_train_kwargs):
+    if (
+        bc_train_kwargs.get("n_epochs") is None
+        and bc_train_kwargs.get("n_batches") is None
+    ):
+        bc_train_kwargs["n_epochs"] = 4
```
qxcv, line 80: This default seems duplicated between here and `_default_bc_train_kwargs`.

src/imitation/scripts/train_dagger.py
```diff
@@ -0,0 +1,157 @@
...
+        expert_policy_path: Either a path to a policy directory containing model.zip
+            (and optionally, vec_normalize.pkl) if `expert_data_src_format == "ppo"`,
+            or None if `expert_data_src_format in ('zero', 'random'). This is used
+            as the `policy_path` argument to `imitation.policies.serialize.load_policy`.
+        expert_policy_type: Either 'ppo', 'zero', or 'random'. This is used as the
+            `policy_type` argument to `imitation.policies.serialize.load_policy`.
+        venv: The vectorized training environment matching the expert data and policy.
+        total_timesteps: The number of timesteps to train inside the environment. (In
+            practice this is a lower bound, as the number of timesteps is rounded up
+            to finish a DAgger training rounds.)
```
qxcv, line 57: Typo? ("finish a DAgger training rounds")

qxcv (commented):
A few small comments, otherwise looks good. Thanks for doing a big refactor in addition to adding the new convenience class!

src/imitation/algorithms/dagger.py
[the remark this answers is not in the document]
AdamGleave: To be fair, I think NPZ is usually more efficient than pickle for large NumPy arrays. Although the most efficient thing on disk would probably be to flatten the trajectories into transitions and then reconstruct them from dones, since the overhead of lots of small NumPy arrays is significant. This is probably premature optimization for now though.

src/imitation/algorithms/dagger.py
shwang: Thanks, I was confused there but that's right

src/imitation/algorithms/dagger.py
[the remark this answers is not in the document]
shwang: I don't have the bandwidth to refactor this right now I think, but I'll flag this as ~"possibly redundant with `types.save`, though saving to npz is more efficient than what we do there"

src/imitation/scripts/config/train_dagger.py
shwang: I think it's fine to have a duplicate setting in the Sacred config so that the user can see a value at a glance with `python -m imitation...train_dagger print_config` rather than None

src/imitation/algorithms/dagger.py
```diff
@@ -340,15 +402,18 @@ def extend_and_update(self, **train_kwargs) -> int:
...
         Arguments:
-            **train_kwargs: arguments to pass to `BC.train()`.
+            bc_train_kwargs: A dictionary of keyword arguments to pass to
+                `BC.train()`, or None to use `self._default_bc_train_kwargs()`.
 
         Returns:
             round_num: new round number after advancing the round counter.
         """
+        if bc_train_kwargs is None:
+            bc_train_kwargs = self._default_bc_train_kwargs()
```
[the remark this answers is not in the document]
shwang: Noted, changing this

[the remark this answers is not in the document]
shwang: Actually, I realize now that the reason I chose the default to be all-or-nothing is we can only accept one of `n_epochs=<not None>` and `n_batches=<not None>`. This way, the user can just specify one or the other without having to worry about setting the other argument to `None`.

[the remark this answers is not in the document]
shwang: OK, I've implemented a middle ground between these two :)

```
            bc_train_kwargs: Keyword arguments for calling `BC.train()`. If
                `log_rollout_venv` is not provided, then it is set to `self.venv`
                by default. If neither of `n_epochs` and `n_batches` is provided,
                then `n_epochs` is set to `self.DEFAULT_N_EPOCHS`.
```

src/imitation/algorithms/dagger.py
```diff
@@ -75,65 +75,88 @@ def reconstruct_trainer(
...
+    Every call to `.step(actions)` accepts and saves expert actions to `self.save_dir`,
+    but only forwards expert actions to the wrapped VecEnv with probability
+    `self.beta`. With probability `1 - self.beta`, a "robot" action (i.e
+    an action from the imitation policy) is forwarded instead.
+
+    Demonstrations are saved as `TrajectoryWithRew` to `self.save_dir` at the end
+    of every episode.
     """
 
+    # TODO(shwang): Consider adding a Callable param that determines whether we
```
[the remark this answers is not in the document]
shwang: I no longer agree with my comment. I think it was based on a misunderstanding earlier on, so I'm removing it.

src/imitation/algorithms/dagger.py
```diff
@@ -169,41 +199,49 @@ def step(self, user_action: np.ndarray) -> Tuple[np.ndarray, float, bool, dict]:
 
         # Replace the given action with a robot action 100*(1-beta)% of the time.
         if np.random.uniform(0, 1) > self.beta:
-            actual_act = self.get_robot_act(self._last_obs)
+            actual_acts = self.get_robot_acts(self._last_obs)
```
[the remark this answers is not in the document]
shwang: `actual_acts[mask] = self.get_robot_acts(self._last_obs[mask])` This line is spicy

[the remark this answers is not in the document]
shwang: OK, added both changes

src/imitation/algorithms/dagger.py
```diff
@@ -391,30 +450,182 @@ def save_trainer(self) -> Tuple[str, str]:
...
+            log_dir: Directory for storing Tensorboard logs, policies, checkpoints,
+                and demonstrations.
+            expert_policy: The expert policy used to generate synthetic demonstrations.
+            expert_trajs: Optional starting dataset that is inserted into the round 0
+                dataset.
+            batch_size: Number of samples in each batch during BC training.
+            bc_kwargs: additional arguments for constructing the `BC` that
+                will be used to train the underlying policy.
+        """
+        self.log_dir = pathlib.Path(log_dir)
```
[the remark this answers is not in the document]
shwang: I like using `log_dir` in all of the Sacred scripts for consistency. Maybe the right thing to do here is to rename `scratch_dir` in `DAggerTrainer` to `log_dir`?

I'm making this change for now -- let me know if I should rename to `log_dir` to `scratch_dir` instead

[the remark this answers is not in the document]
shwang: jk -- this is more complicated than what I described above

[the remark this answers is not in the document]
shwang: Switched everything to `scratch_dir` as per your suggestion (I assumed that BC used `log_dir` for some reason, but this is actually configured via `logger.configure()` inside the Sacred scripts.)

src/imitation/algorithms/dagger.py
shwang: fixed

src/imitation/algorithms/dagger.py
```diff
@@ -391,30 +450,182 @@ def save_trainer(self) -> Tuple[str, str]:
...
         Args:
             policy_path: path to save policy to.
         """
         self.bc_trainer.save_policy(policy_path)
+
+
+class SimpleDAggerTrainer(DAggerTrainer):
+    """Simpler subclass of DAggerTrainer for training with synthetic feedback."""
+
+    def __init__(
```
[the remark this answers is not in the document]
shwang: That's right, I've refactored accordingly

src/imitation/algorithms/dagger.py
```diff
@@ -169,41 +199,49 @@ def step(self, user_action: np.ndarray) -> Tuple[np.ndarray, float, bool, dict]:
...
-
-        return next_obs, reward, done, info
+        for traj in fresh_demos:
+            _save_dagger_demo(traj, self.save_dir)
+
+        return next_obs, rews, dones, infos
+
+    def flush_trajectories(self) -> None:
+        """Immediately save all partially completed trajectories to `self.save_dir`."""
+        # TODO(shwang): Since we are saving partial trajectories instead of
```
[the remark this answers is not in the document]
shwang: I'll leave this item as a todo

src/imitation/algorithms/dagger.py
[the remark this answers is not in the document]
shwang: It's annoying to fail due to not enough samples, because the error message you get is not at all obvious (something like "BC data loader yielded no batches unexpected"). Also, testing becomes a pain for both DAgger and the associated Sacred script because you have to set a high enough number of timesteps to ensure we don't run into this issue while still trying to shorten test duration.

I'm going to revert to `drop_last=True` as recommended, and then ensure that `train()` doesn't end a round without collecting `self.batch_size` samples, and also add a big `assert` or `RuntimeError` here if we still don't have enough samples for a single batch.

src/imitation/algorithms/bc.py
```diff
@@ -234,6 +234,11 @@ def __init__(
             self.device
         )  # pytype: disable=not-instantiable
         optimizer_kwargs = optimizer_kwargs or {}
+        # TODO(shwang): Raise error or warning if learning rate is included, as we
+        #   are setting learning rate via `policy_kwargs` and ConstantLRSchedule
+        #   instead? I also don't fully remember why we need to do this again. It would
+        #   be helpful to add a comment explaining the conflict between these two
+        #   learning rate parameters.
```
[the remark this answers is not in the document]
shwang: Man, I thought we already did this sometime in the past. Maybe that was on an ILR specific branch.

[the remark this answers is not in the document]
shwang: I'd like to punt this change off to a later PR, leaving this as a TODO. For interests of wrapping up this PR sooner rather than later.

src/imitation/algorithms/dagger.py
```diff
@@ -237,46 +275,55 @@ class DAggerTrainer:
...
     ):
-        """Trainer constructor.
+        """DaggerTrainer constructor.
 
         Args:
-            env: environment to train in.
-            scratch_dir: directory to use to store intermediate training
+            venv: Vectorized training environment. Note that when the robot
+                action is randomly injected (in accordance with `beta_schedule`
+                argument), every individual environment will get a robot action
```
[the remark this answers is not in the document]
shwang: made action source randomization independent across Envs, and updated the comment

src/imitation/algorithms/dagger.py
```diff
@@ -391,30 +450,182 @@ def save_trainer(self) -> Tuple[str, str]:
...
+        round_num = 0
+
+        while total_timestep_count < total_timesteps:
+            collector = self.get_trajectory_collector()
+            round_episode_count = 0
+            round_timestep_count = 0
+
+            obs = collector.reset()
+            ep_rewards = np.zeros([self.venv.num_envs])
+            # TODO(shwang): This while loop end condition causes rollout collection to
```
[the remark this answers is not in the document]
shwang: fixed. As part of this change, I modified the `rollout.make_sample_until` helper to accept `min_episodes` and `min_timesteps` simultaneously.

src/imitation/algorithms/dagger.py
```diff
@@ -391,30 +450,182 @@ def save_trainer(self) -> Tuple[str, str]:
...
+            # fails if the "round" also didn't finish any episodes (not enough
+            # timesteps).
+            collector.flush_trajectories()
+
+            logger.record("dagger/total_timesteps", total_timestep_count)
+            logger.record("dagger/round_num", round_num)
+            logger.record("dagger/round_episode_count", round_episode_count)
+            logger.record("dagger/round_timestep_count", round_timestep_count)
+
+            # TODO(shwang): It looks like BC might start looping Tensorboard
```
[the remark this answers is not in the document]
shwang: Yes, BC loops around Tensorboard x-axis right now

image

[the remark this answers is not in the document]
shwang: Could we leave this as an issue, or as a TODO comment for now? (It will require some refactoring and internal step counting for BC). I expect that tuning will be straightforward for the paper and I only need the final Sacred stats to generate the table.

shwang:
Thanks for the reviews guys. This PR is ready for another round of review.

Some of the code seems like it could still change after a round of review, so I plan on adding code coverage after this round.

qxcv (approved):
Checked the diff from last review & it looks good to me. Also happy with resolution to my previous comments.

src/imitation/algorithms/bc.py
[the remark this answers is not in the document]
AdamGleave: Since it's just a one-line change I've gone ahead and made it. NaN failed assertion, but huge LR worked OK.

AdamGleave (approved):
Thanks for the changes -- LGTM

shwang:
Looks like https://github.com/HumanCompatibleAI/imitation/pull/309/commits/e2fe9c1ad4602de945628f204bdafacf7f6c8109 isn't working right now because Adam optimizer doesn't initialize properly with it:  https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/913/workflows/884d3b80-f9bf-415d-8fc7-6ccc8e6c09dc/jobs/1903

AdamGleave:
Will take a look. Weird, tests passed on my machine.

AdamGleave:
After taking a look at the LR thing more, I think perhaps the learning rate should be being set on `self.optimizer` directly and not via `self.policy_kwargs` after all. Setting `lr_schedule` in `policy_kwargs` has the effect of setting `self.policy.optimizer`'s lr: https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/policies.py But we don't use `self.policy.optimizer` as far as I can see.

@qxcv @scottemmons how is this code meant to work? It's also odd that `ConstantLRSchedule` has a hard-coded learning-rate rather than being a configuration parameter. I feel like I'm probably missing something.

src/imitation/algorithms/dagger.py
```diff
@@ -391,30 +469,167 @@ def save_trainer(self) -> Tuple[str, str]:
...
+
+            round_episode_count += len(trajectories)
+
+            logger.record("dagger/total_timesteps", total_timestep_count)
+            logger.record("dagger/round_num", round_num)
+            logger.record("dagger/round_episode_count", round_episode_count)
+            logger.record("dagger/round_timestep_count", round_timestep_count)
+
+            # TODO(shwang): BC starts looping Tensorboard
+            #   back to x=0 with each new call to BC.train(). Consider adding a
```
shwang, line 629: I've fixed this Tensorboard reset problem locally because it was making DAgger runs annoying to analyze. Commit will come shortly

shwang:
This PR is ready for merge after adding coverage (which I don't think I have time to get to right now)

shwang:
Dagger results for hyperparams added by https://github.com/HumanCompatibleAI/imitation/pull/309/commits/d8b234425d54867a35093dd33ed35a1774f15ce1:

# seals CartPole
image

# seals MountainCar
image

# seals HalfCheetah
image

src/imitation/scripts/train_dagger.py
```diff
@@ -38,9 +38,13 @@ def train_dagger(
 
     Args:
         expert_data_src: Either a path to pickled `Sequence[Trajectory]` or
-            `Sequence[Trajectory]`.
+            `Sequence[Trajectory]` or None. If None, then ignore
+            `expert_data_src_format` and `n_expert_demos`, and don't load
+            initial demonstrations.
         expert_data_src_format: Either "path" if `expert_data_src` is a path, or
-            "trajectory" if `expert_data_src` if `Sequence[Trajectory]`.
+            "trajectory" if `expert_data_src` if `Sequence[Trajectory]`, or None.
```
AdamGleave, line 45:
```suggestion
            "trajectory" if `expert_data_src` is `Sequence[Trajectory]`, or None.
```

src/imitation/scripts/train_dagger.py
```diff
@@ -84,31 +88,34 @@ def train_dagger(
...
         expert_trajs = types.load(expert_data_src)
     elif expert_data_src_format == "trajectory":
         # Convenience option for launching experiment from Python script with
         # in-memory trajectories.
         expert_trajs = expert_data_src
     else:
         raise ValueError(
-            f"expert_data_src_format={expert_data_src_format} should be 'path'"
-            "or 'trajectory'."
+            f"expert_data_src_format={expert_data_src_format} should be 'path' "
```
AdamGleave, line 102:
```suggestion
            f"expert_data_src_format={expert_data_src_format} should be 'path', "
```

src/imitation/scripts/train_dagger.py
```diff
@@ -84,31 +88,34 @@ def train_dagger(
...
+            "or 'trajectory' or None."
```
AdamGleave, line 103:
```suggestion
            "'trajectory' or None."
```

AdamGleave (approved):
New changes LGTM -- added some minor suggestions.

PR #301, opened 2021-07-22
Torch JAX

qxcv:
This PR replaces JAX with Torch in the MCE IRL code. It also removes all mention of JAX from the rest of the repo (including dependencies).

Issue #246, opened 2020-10-24
Common reward model interface


AdamGleave:
Currently our `RewardNet` class is designed for `AIRL`. In particular, having separate `train` and `test` outputs is specific to AIRL, and does not make any sense for most others algorithms.

We currently have an alternative interface `tabular_irl.RewardModel`. This is intended for JAX -- but it'd be incompatible with `RewardNet` even after being ported to PyTorch. (And note that it's not actually specific to tabular methods -- although some of the concrete subclasses are.)

In the future we might add other algorithms like GCL, DRLHP, etc which also use rewards but do not have this train vs test distinction.

My proposal would be to make an abstract `RewardNet` or `RewardModel` class (either name seems fine) that is minimal and so should work across all algorithms. In particular, I imagine the `__init__` would take in an `observation_space` and `action_space`, there would be a single abstract method `reward` (v.s. `reward_train` and `reward_test` now) and a concrete implementation `predict` taking NumPy arrays. (You might also need some other things for PyTorch, e.g. `device`, that we have now.)

We can then have concrete instances for various different reward models. I would suggest having one that is an MLP reward model (with boolean flags indicating whether it uses observation, action, next observation and dones as input), one that is an MLP potential shaping only, a class that lets you take linear combinations of reward models, and then a shaped MLP can be defined in terms of these.

For AIRL you could then feed in an MLP reward model and an MLP potential shaping, it can internally sum them and have the train vs test distinction -- with no need for a separate class for this.

Note I've done something similar in https://github.com/HumanCompatibleAI/evaluating-rewards/blob/master/src/evaluating_rewards/rewards/base.py which has been working well (and I iterated on a couple of designs before getting to this stage). But that's all TF code, so cannot be used verbatim.

ejnnr:
#311 mostly fixes these, except that IRL still uses its own reward models. Should be very easy to change so that IRL uses the models in `rewards.reward_nets`. But I don't see a training script for IRL. Do people just use `mce_irl` directly or is there some other place we'd need to modify?

AdamGleave:
I don't think we have a training script for `tabular_irl`. Which is bad, we should, but I think that can probably be postponed -- it'd be a natural thing to do if/when we unify training different algorithms into a single script. So I think the only other place to update is `test_tabular`.

PR #311, opened 2021-08-18
Generalize RewardNet base class

ejnnr:
For the implementation of preference comparisons, we'll need implementations of reward models and it makes sense to share those with existing and other future algorithms. The current `RewardNet` ABC contains some things that don't make sense in general, such as the test/train split (see also #246). This PR introduces a very minimalistic `RewardNet` ABC and renames the old one to `AIRLRewardNet` (which is essentially a wrapper around the new `RewardNet`).

That's basically the minimal change I'd like to have for preference comparisons. But I think more refactoring would improve things:
- [x] Add the preprocessing logic to the new minimal `RewardNet`: its `.forward()` method would be abstract and take in pytorch tensors, we'd then have `RewardNet.predict()`, which takes in and outputs numpy arrays and does preprocessing
- [x] Create an abstract `ShapedRewardNet` that subclasses the new `RewardNet` and exposes a `.predict_without_potential(obs, acts, next_obs, dones)` method or similar (in addition to the usual `.predict(), which uses the potential)
- [x] Introduce a concrete implementation of `ShapedRewardNet` with MLPs (analogously to `BasicRewardMLP`)
- [x] Have a single AIRL-specific reward model class, `AIRLRewardNet`, which takes a `RewardNet` instance as an argument and would just be a very thin wrapper. I'd suggest just checking whether the `RewardNet` instance is an instance of `ShapedRewardNet` and using it without potentials for testing in that case. That's less general than the current ABC but I'd only change that if we ever actually need a more general version.
- [x] The argument to `AIRL` (in `algorithms`) would be the `RewardNet` type and kwargs that are passed to `AIRLRewardNet`

This would mean there is only one AIRL-specific class and everything else can be used for preference comparisons as well. We could even consider getting rid of the AIRL-specific class entirely but this would require a few more changes to the AIRL code I assume (haven't looked at that yet in detail).

Once #301 is merged, I assume it would also be pretty simple to use the new `RewardNet` interface for tabular IRL.

@AdamGleave could you please let me know what you think about these potential changes? And could you ping anyone else you think would be affected and should take a look? (I don't really know who is working on what)


shwang (commented):
Thanks for the refactor, Erik. Looking through your draft PR right now, I have a renaming suggestion for `AIRLRewardNet` subclasses

src/imitation/rewards/reward_nets.py
```diff
@@ -188,7 +210,7 @@ def device(self) -> th.device:
         return first_param.device
 
 
-class RewardNetShaped(RewardNet):
+class RewardNetShaped(AIRLRewardNet):
```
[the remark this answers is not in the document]
ejnnr: Agreed that the naming is very confusing after my change. If we decide to go with the additional changes I'm proposing, there would only be one AIRL-specific class left, which would be called `AIRLRewardNet` (or similar), otherwise I'll rename the existing classes like you suggest. I'll hold off on that though until we've decided whether to implement my suggestions

src/imitation/rewards/discrim_nets.py
```diff
@@ -182,19 +182,21 @@ def _eval_reward(
 
 
 class DiscrimNetAIRL(DiscrimNet):
-    r"""The AIRL discriminator for a given RewardNet.
+    r"""The AIRL discriminator for a given ARILRewardNet.
```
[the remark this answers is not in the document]
ejnnr: Oops, thanks for catching that!

AdamGleave:
Thanks for the PR! Will start looking at the diff now. From the textual summary: I agree with the high-level design. For `ShapedRewardNet`, perhaps it should have a composition relationship, so that the unshaped `RewardNet` can be accessed directly via `shaped_reward_net.base` so we don't need a separate `predict_without_potential`?

ejnnr:
I agree a single wrapper class makes more sense than an ABC for `ShapedRewardNet`, @AdamGleave, thanks for the suggestion.
I'll implement the remaining changes and then request re-review

ejnnr:
I've implemented everything outlined in my initial comment. There are some further things that could be improved IMO:
1. Combine reward nets with the IRL implementations (but that can only happen after #301)
2. Get rid of `AIRLRewardNet` entirely: the only thing it does is to check whether the wrapped `RewardNet` has a potential, and if so, disable that for testing. We could just move that logic to `DiscrimNetAIRL`. We would also need to modify `serialize.py` and either duplicate this logic there or remove the `"RewardNet_unshaped"` `reward_type`.
3. Maybe there should be a cleaner separation between the `rewards` package and GAIL/AIRL? If we get rid of `AIRLRewardNet`, then `reward_nets.py` would only have algorithm-independent code, but `discrim_nets.py` and `serialize.py` still contain a lot of GAIL/AIRL specific code.
4. If `AIRL` got its `RewardNet` instance directly (rather than as a class + kwargs), we could probably get rid of `BasicShapedRewardNet` (which I only added to make it easier to use AIRL via the scripts). Mentioning this because I already saw a FIXME comment with this suggestion by @qxcv: https://github.com/HumanCompatibleAI/imitation/blob/949662acce6d9f126d94fd1d177b301d452de315/src/imitation/algorithms/adversarial.py#L479

I could remove `AIRLRewardNet` (point 2. above) in this PR. I think 1. could be a small new PR once #301 is merged, before this one becomes too large? I'm not sure whether 3. and 4. are worth changing. In any case, I think they'd require working with code I'm not familiar with yet, so I'd rather not implement them in this PR and instead work on preference comparisons first. I or someone else could work on them in another PR later if we decide it makes sense.

ejnnr:
Converted back to draft because @AdamGleave and I found a potential shaping issue we want to fix in this PR. Should be a very small change though, you can still review everything else already

AdamGleave:
I like your proposal of doing 2) in this PR, and 1) in another PR.

I'm happy to take a stab at 3) & 4) in a separate PR.

Will review the changes now.

src/imitation/rewards/reward_nets.py
```diff
@@ -14,148 +14,51 @@
...
         self,
         observation_space: gym.Space,
         action_space: gym.Space,
-        *,
-        scale: bool = False,
-        use_state: bool = True,
-        use_action: bool = True,
-        use_next_state: bool = False,
-        use_done: bool = False,
+        normalize_images: bool = True,
```
AdamGleave, line 27: It's weird that this needs to be in the ABC but I don't see a good way around it, so seems fine to leave it in for now.

src/imitation/rewards/reward_nets.py
```diff
@@ -168,50 +71,63 @@ def _eval_reward(
             action=action,
             next_state=next_state,
             done=done,
-            device=self.device(),
-            scale=self.scale,
+            device=self.device,
+            scale=self.normalize_images,
```
AdamGleave, line 75: Consider renaming `scale` in `disc_rew_preprocess_inputs`?

src/imitation/rewards/reward_nets.py
```diff
@@ -14,148 +14,51 @@
...
 
         Args:
-            observation_space: The observation space.
-            action_space: The action space.
-            scale: Whether to scale the input.
-            use_state: Whether state is included in inputs to network.
-            use_action: Whether action is included in inputs to network.
-            use_next_state: Whether next state is included in inputs to network.
-            use_done: Whether episode termination is included in inputs to network.
+            observation_space (gym.Space): the observation space of the environment
```
AdamGleave, line 32: You don't need the types in parentheses when we have type annotations. Sphinx does (or at least can -- if not we should enable it) include type annotations in the generated docs.

src/imitation/rewards/reward_nets.py
```diff
@@ -168,50 +71,63 @@ def _eval_reward(
...
+        potential: Callable[[th.Tensor], th.Tensor],
+        discount_factor: float,
+        normalize_images: bool = True,
     ):
-        super().__init__(observation_space, action_space, **kwargs)
-        self._discount_factor = discount_factor
+        """Setup a ShapedRewardNet instance.
+
+        Args:
+            observation_space (gym.Space): observation space of the environment
```
AdamGleave, line 111: as before, types in docstring not necessary when they're annotated in the function

src/imitation/rewards/reward_nets.py
```diff
@@ -221,58 +137,141 @@ def __init__(
...
 
+
+class AIRLRewardNet(nn.Module):
+    """Wrapper around RewardNet with different forward passes for test and train.
+
+    This class is used for AIRL, where we want the potential shaping to be included
+    during training but not testing. If the wrapped RewardNet is not an instance
+    of ShapedRewardNet, then the test and train passes will be identical.
+
+    TODO(ejnnr): get rid of this class entirely, instead just implement the logic
```
AdamGleave, line 168: Yeah I agree this doesn't belong here. Not sure if it's better to just move the class to AIRL-specific code or delete it entirely.

src/imitation/rewards/reward_nets.py
```diff
@@ -339,124 +338,93 @@ def forward(self, state, action, next_state, done):
...
-    """A shaped reward network with simple, default settings.
+    This is just a very simple convenience class for instantiating a BasicRewardNet
+    and a BasicPotentialShaping and wrapping them inside a ShapedRewardNet.
+    Mainly exists for backwards compatibility after
+    https://github.com/HumanCompatibleAI/imitation/pull/311
+    to keep the scripts working.
 
-    With default parameters this RewardNet has two hidden layers [32, 32]
-    for the base reward network and shaping network.
+    TODO(ejnnr): if we ever modify AIRL so that it takes in a RewardNet instance
```
AdamGleave, line 350: :+1: but probably fine to leave for now

AdamGleave (approved):
LGTM, only some very small suggested changes. I think we're OK merging this PR (after those changes) and then doing potential shaping fix in separate PR, but also happy to re-review this PR with potential shaping fix.

src/imitation/algorithms/adversarial.py
```diff
@@ -466,7 +466,7 @@ def __init__(
 
 class AIRL(AdversarialTrainer):
 
-    reward_net: reward_nets.RewardNet
+    reward_net: reward_nets.AIRLRewardNet
```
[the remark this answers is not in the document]
ejnnr: I think you're right, removed it

src/imitation/rewards/reward_nets.py
ejnnr: Done, and also renamed it in the discriminator nets to make everything consistent

src/imitation/rewards/reward_nets.py
ejnnr: I've removed it entirely now, I think that makes the most sense. It's significantly shorter in terms of LOC and it also "flattens" things a bit because we don't have this wrapper everywhere.

ejnnr:
I've remove `AIRLRewardNet` now, I like the new version without it much more. But probably makes sense for you to take another look, @AdamGleave.

I've also renamed `scale` to `normalize_images` everywhere and modified one docstring, which made it sound as though this would scale *all* Box observations, rather than just images. Unfortunately, this might break people's scripts/configs because it also affects the argument passed to the discriminator nets. So happy to roll that change back if we're aiming to avoid breaking changes like that if they're avoidable.

I'm a bit confused by the fact that `normalize_images` appears so often anyway, maybe that could use a refactor as well (though that would mainly affect the discriminator nets and AIRL/GAIL, so it should be a separate PR I think).
I've also noticed that `DiscrimNetAIRL` doesn't have a `normalize_images` argument, so I think it will just always default to `False` (because of its parent class)? Not sure if that's intentional.

src/imitation/rewards/serialize.py
```diff
@@ -30,7 +30,14 @@ def loader(path: str, venv: VecEnv) -> common.RewardFn:
         """Load train (shaped) or test (not shaped) reward from path."""
         del venv  # Unused.
         net = th.load(str(path))
-        reward = net.predict_reward_train if shaped else net.predict_reward_test
+        if not shaped and hasattr(net, "base"):
```
AdamGleave, line 33: Hm -- it feels weird that this is even part of the serialization logic, but we can perhaps clean this up in a separate PR.

AdamGleave (approved):
LGTM.

> I'm a bit confused by the fact that normalize_images appears so often anyway, maybe that could use a refactor as well (though that would mainly affect the discriminator nets and AIRL/GAIL, so it should be a separate PR I think).
I've also noticed that DiscrimNetAIRL doesn't have a normalize_images argument, so I think it will just always default to False (because of its parent class)? Not sure if that's intentional.

In SB2 the parameter, then called `scale`, had more of an effect as it also renormalized finite-bounded Box-es: https://github.com/Stable-Baselines-Team/stable-baselines/blob/f877c85b6e45084da1d8ccf73e7e730dc2001c3f/stable_baselines/common/input.py#L6 I think this is why we have it in so many places.

I don't think DiscrimNetAIRL not having the parameter is intentional, just that we haven't needed it yet.

I agree cleaning this up is best addressed in another PR. Not sure how to fix this even really. We do need to configure this somewhere. It seems messy to have it in every constructor, but the alternatives seem worse: e.g. having it in the `preprocess` function would be error prone (could end up normalizing in some places and not others).

AdamGleave:
> Unfortunately, this might break people's scripts/configs because it also affects the argument passed to the discriminator nets. 

It's OK if we break external scripts -- we're still a 0.x release, so breaking API changes are to be expected.

Scripts in the repo should not break, but looks like you updated the ones in `train_adversarial.py`, and I'm not aware of it being used anywhere else.

PR #316, opened 2021-08-20
analyze: Generate TeX tables for paper

shwang:
Upgrade analyze script to generate TeX tables for paper of varying verbosity


src/imitation/scripts/analyze.py
```diff
@@ -13,13 +17,43 @@
...
+    )
+    sacred_dicts = []
+
+    for sacred_dir in sacred_dirs:
+        try:
+            sacred_dicts.append(sacred_util.SacredDicts.load_from_dir(sacred_dir))
+        except json.JSONDecodeError:
+            warnings.warn(f"Invalid JSON file in {sacred_dir}", RuntimeWarning)
+
+    if run_name is not None:
```
AdamGleave, line 37: Consider splitting this into a helper function along the lines of:

```
def _filter_dicts(dicts, key, expected_val):
    return filter(lambda sd: get(sd.run, key) == expected_val, sacred_dicts)
```

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +108,174 @@ def gather_tb_directories(
...
+
+def _get_algo_name(sd: sacred_util.SacredDicts) -> str:
+    exp_command = _get_exp_command(sd)
+
+    if exp_command == "train_adversarial":
+        algo = get(sd.config, "algorithm")
+        if algo is not None:
+            algo = algo.upper()
+        else:
+            print(sd.config)
```
AdamGleave, line 123: What's the purpose of this? Perhaps add a print statement before this to give some context?

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +108,174 @@ def gather_tb_directories(
...
+        (
+            "imit_return_summary",
+            lambda sd: _return_summaries(sd)["imit_return_summary"],
+        ),
+        ("imit_expert_ratio", lambda sd: _return_summaries(sd)["imit_expert_ratio"]),
+    ]
+)
+
+
+# If `verbosity` is at least the length of this tuple, then we use all table_entry_fns
```
AdamGleave, line 186:
```suggestion
# If `verbosity` is at least the length of this list, then we use all table_entry_fns
```

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +108,174 @@ def gather_tb_directories(
...
+# as columns of table.
+# Otherwise, use only the subset at index `verbosity`. The subset of columns is
+# still arranged in the same order as in the `table_entry_fns` OrderedDict.
+table_verbosity_mapping: List[Set[str]] = []
```
AdamGleave, line 190: This is a neat way of implementing verbosity :)

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +108,174 @@ def gather_tb_directories(
...
+# verbosity 2
+table_verbosity_mapping.append(
+    table_verbosity_mapping[-1]
+    | {"status", "imit_expert_ratio", "exp_command", "run_name"}
+)
+
+
+def _get_table_entry_fns_subset(table_verbosity: int) -> sd_to_table_entry_type:
+    assert table_verbosity >= 0
+    if table_verbosity >= len(table_entry_fns):
```
AdamGleave, line 214:
```suggestion
    if table_verbosity >= len(table_verbosity_mapping):
```
given it's used as an index into `table_verbosity_mapping` on line 217

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +108,174 @@ def gather_tb_directories(
...
-        associated with the `train_adversarial` Sacred script. Behavior is
-        undefined if there are Sacred subdirectories associated with other
-        scripts.
-      run_name: If provided, then only analyze results from Sacred directories
-        associated with this run name. `run_name` is compared against the
-        "experiment.name" key in `run.json`.
-      skip_failed_runs: If True, then filter out runs where the status is FAILED.
-      csv_output_path: If provided, then save a CSV output file to this path.
-      verbose: If True, then print the dataframe.
+        source_dir: A directory containing Sacred FileObserver subdirectories
```
AdamGleave, line 236: Argument does not exist

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +108,174 @@ def gather_tb_directories(
...
+            associated with the `train_adversarial` Sacred script. Behavior is
+            undefined if there are Sacred subdirectories associated with other
+            scripts. (Captured argument)
+        run_name: If provided, then only analyze results from Sacred directories
```
AdamGleave, line 240: Argument does not exist

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +108,174 @@ def gather_tb_directories(
...
+            associated with this run name. `run_name` is compared against the
+            "experiment.name" key in `run.json`. (Captured argument)
+        skip_failed_runs: If True, then filter out runs where the status is FAILED.
```
AdamGleave, line 243: Argument does not exist

AdamGleave (commented):
Thanks for adding these features -- overall looks pretty good, added a couple of comments. Ping me when it's out of draft and I'll take another look. (If you don't have time to finish it, then maybe write a comment outling what's left to do, and I or someone else can try to wrap it up?)

src/imitation/scripts/analyze.py
shwang: Yeah, this argument isn't part of the function signature, though it is required for this Sacred command to run correctly (It's captured by another command).

I'll add something to the docstring with the effect of "you also need to look at the captured arguments for ____".

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +108,176 @@ def gather_tb_directories(
...
+        for col_name, make_entry_fn in table_entry_fns_subset.items():
+            row[col_name] = make_entry_fn(sd)
 
     df = pd.DataFrame(rows)
+    if len(df) > 0:
+        assert set(df.columns) == set(table_verbosity_mapping[table_verbosity])
+        df.sort_values(by=["algo", "env_name"], inplace=True)
+
+    display_options = dict(index=False)
     if csv_output_path is not None:
```
shwang, line 271: `print_table` conditional clause should really come after `tex_output_path is not None`

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +108,176 @@ def gather_tb_directories(
...
-        df.to_csv(csv_output_path)
-    if verbose:
-        print(df.to_string())
+        df.to_csv(csv_output_path, **display_options)
+        print(f"Wrote CSV file to {csv_output_path}")
+    if print_table:
+        print(df.to_string(**display_options))
```
shwang, line 275: `print_table` conditional clause should really come after `tex_output_path is not None`

src/imitation/scripts/analyze.py
AdamGleave: :+1:

src/imitation/scripts/analyze.py
shwang: The helper pattern doesn't match `!= "FAILED"` on the `skip_failed_runs` conditional. (But otherwise I"d be happy with this change)

src/imitation/scripts/analyze.py
shwang: THis should probably be removed or otherwise editted. I think I added it here for debug purposes

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +123,170 @@ def gather_tb_directories(
...
+        The DataFrame generated from the Sacred logs.
     """
-    sacred_dicts = _get_sacred_dicts(source_dir, run_name, env_name, skip_failed_runs)
+    table_entry_fns_subset = _get_table_entry_fns_subset(table_verbosity)
 
     rows = []
-    for sd in sacred_dicts:
+    for sd in _gather_sacred_dicts():
         row = OrderedDict()
         rows.append(row)
```
AdamGleave, line 269: Nit: seems more natural to have this at the end of the `for` loop, once `row` has been fully set, although I know it doesn't actually matter since `row` is mutable and points at the same object throughout.

src/imitation/scripts/analyze.py
```diff
@@ -74,62 +123,170 @@ def gather_tb_directories(
...
-            row["imit_return_summary"] = _make_return_summary(imit_stats, "monitor_")
-            row["imit_expert_ratio"] = (
-                imit_stats["monitor_return_mean"] / expert_stats["return_mean"]
-            )
+        for col_name, make_entry_fn in table_entry_fns_subset.items():
+            row[col_name] = make_entry_fn(sd)
 
     df = pd.DataFrame(rows)
+    if len(df) > 0:
+        assert set(df.columns) == set(table_verbosity_mapping[table_verbosity])
```
AdamGleave, line 275: This fails if `table_verbosity >= len(table_verbosity_mapping)`, in which case all of `table_entry_fns` are used.

AdamGleave (approved):
LGTM apart from two very minor points.

PR #320, opened 2021-08-23
Preference comparisons

ejnnr:
A simple implementation of reward learning using preference comparisons. The plan is to only implement synthetic preference sources for now (i.e. no human feedback), but the design is such that extensions like that should be possible without changes later.

Requires #311. I based the `preference_comparisons` branch off the `rewardnet_refactor` branch, so you're seeing the `RewardNet` refactor changes here too until #311 is merged.

Still very much a WIP, created the PR as a place for discussion/early review.


AdamGleave:
I've changed the base so the diff only includes the changes introduced in this PR, not the more general refactor.

We should change the base back to master once https://github.com/HumanCompatibleAI/imitation/pull/311 is merged

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+        reward_model: RewardNet,
+        agent_timesteps: int,
+        fragmenter: Optional[Fragmenter] = None,
+        preference_gatherer: Optional[PreferenceGatherer] = None,
+        reward_trainer: Optional[RewardTrainer] = None,
+    ):
+        """Initialize the preference comparison trainer.
+
+        Args:
+            agent_trainer: generates trajectories while optionally training an RL agent
```
AdamGleave, line 245: If the training part is *optional*, and the primary purpose is to sample trajectories, then calling it `agent_trainer` seems like a bit of a misnomer.

However, `AgentTrainer` itself seems like it really is focused on agent training.

Maybe rename it to something like `trajectory_generator`? And rename `train` to something like `sample`? Granted, this will be confusing if `sample` also trains as a side-effect. I'm tempted to even just split the methods up into a separate `maybe_train()` and `sample()` method, but you could also have something like `sample_maybe_train()`.

This is just a rough sketch, there's probably something better you can do.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+                on the learned reward function (can also be a sampler from a static
+                dataset of trajectories though).
+            reward_model: a RewardNet instance to be used for learning the reward
+            agent_timesteps: number of environment steps to train the agent for between
```
AdamGleave, line 249: Is this the number of timesteps it'll return as a sample? Do they always have to be the same?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+                based on ground-truth rewards. Human preferences could be implemented
+                here in the future.
+            reward_trainer: trains the reward model based on pairs of fragments and
+                associated preferences. Default is to use the preference model
+                and loss function from DRLHP.
+        """
+        self.model = reward_model
+        if reward_trainer is None:
+            reward_trainer = CrossEntropyRewardTrainer(reward_model)
+        self.reward_trainer = reward_trainer
```
AdamGleave, line 266: So if user wants to do something that's not `CrossEntropyRewardTrainer`, they have to construct it and pass in `reward_model` themselves? That does seem like the most flexible approach, but I'm a bit nervous someone messes up and passes in a different model to this class and the RewardTrainer. Consider adding `assert self.reward_trainer.model == self.model`?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+        self.agent = agent_trainer
+        if fragmenter is None:
+            fragmenter = RandomFragmenter()
+        self.fragmenter = fragmenter
```
AdamGleave, line 270: Can delete last 2 lines and have:
```suggestion
        self.fragmenter = fragmenter or RandomFragmenter()
```

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+        if preference_gatherer is None:
+            preference_gatherer = SyntheticGatherer()
+        self.preference_gatherer = preference_gatherer
```
AdamGleave, line 273: Similar to above.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+
+        self.dataset = PreferenceDataset()
+
+    def train(self, steps: int):
+        """Train the reward model and the policy if applicable.
+
+        Args:
+            steps: number of iterations of the outer training loop
+        """
+        for _ in range(steps):
```
AdamGleave, line 284: Consider adding some logging here. You could do that in a separate PR, but I'd suggest doing that before trying to test whether your code works. Will be hard to debug without!

src/imitation/data/fragments.py
```diff
@@ -0,0 +1,87 @@
+import random
+import warnings
+from typing import Callable, List, Optional, Tuple
+
+from imitation.data.types import TrajectoryWithRew
+
+Fragmenter = Callable[
+    [List[TrajectoryWithRew]], List[Tuple[TrajectoryWithRew, TrajectoryWithRew]]
```
AdamGleave, line 8:
```suggestion
    [Sequence[TrajectoryWithRew]], Sequence[Tuple[TrajectoryWithRew, TrajectoryWithRew]]
```

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+import torch as th
+
+from imitation.data.fragments import Fragmenter, RandomFragmenter
+from imitation.data.rollout import flatten_trajectories_with_rew
+from imitation.data.types import TrajectoryWithRew, TransitionsWithRew
+from imitation.policies.trainer import AgentTrainer
+from imitation.rewards.reward_nets import RewardNet
+
+PreferenceGatherer = Callable[
+    [List[Tuple[TrajectoryWithRew, TrajectoryWithRew]]], np.ndarray
```
AdamGleave, line 21:
```suggestion
    [Sequence[Tuple[TrajectoryWithRew, TrajectoryWithRew]]], np.ndarray
```

src/imitation/data/fragments.py
```diff
@@ -0,0 +1,87 @@
...
+
+class RandomFragmenter:
+    """Sample fragments of trajectories uniformly at random with replacement.
+
+    Note that each fragment is part of a single episode and has a fixed
+    length. This leads to a bias: transitions at the beginning and at the
+    end of episodes are less likely to occur as part of fragments (this affects
+    the first and last fragment_length transitions).
+
+    TODO(ejnnr): should we correct for this bias and make all transitions equally
```
AdamGleave, line 21: Good question, how would you correct for the bias?

One approach would be to just partition the trajectories into `fragment_length` fragments, and then choose from those fragments randomly. Since no overlap, transitions at the start/end are as likely to occur as those in the middle. But you now have a new bias of never seeing trajectories that cross a certain boundary of timesteps. Also this won't work if episodes aren't a multiple of `fragment_length`.

In general I'm not convinced it's possible to get uniform probability over transitions without adding restrictions to how the trajectories are sliced.

src/imitation/data/fragments.py
```diff
@@ -0,0 +1,87 @@
...
+            fragment_length: the length of each sampled fragment
+            num_pairs: the number of fragment pairs to sample
+            seed: an optional seed for the internal RNG
+        """
+        self.fragment_length = fragment_length
+        self.num_pairs = num_pairs
+        self.rng = random.Random(seed)
+
+    def __call__(
+        self, trajectories: List[TrajectoryWithRew]
```
AdamGleave, line 43:
```suggestion
        self, trajectories: Sequence[TrajectoryWithRew]
```

src/imitation/data/fragments.py
```diff
@@ -0,0 +1,87 @@
...
+    ) -> List[Tuple[TrajectoryWithRew, TrajectoryWithRew]]:
```
AdamGleave, line 44:
```suggestion
    ) -> Sequence[Tuple[TrajectoryWithRew, TrajectoryWithRew]]:
```

src/imitation/data/fragments.py
```diff
@@ -0,0 +1,87 @@
...
+        fragments: List[TrajectoryWithRew] = []
+
+        # filter out all trajectories that are too short
+        trajectories = [
```
AdamGleave, line 48: This is also a bias, albeit an unavoidable one, but worth mentioning in docstring.

Also tempted to log how many trajectories are removed (just compute len before and after).

src/imitation/data/fragments.py
```diff
@@ -0,0 +1,87 @@
...
+        if len(trajectories) == 0:
+            raise ValueError(
+                "No trajectories are long enough for the desired fragment length."
+            )
+
+        weights = [len(traj) for traj in trajectories]
+
+        # TODO(ejnnr): since we're sampling with replacement, there could
+        # already be lots of duplicates below this threshold. Perhaps
+        # we should warn already if the number of transitions isn't larger
```
AdamGleave, line 60: Bumping warning threshold seems good to me.

Sampling without replacement is probably not worth the hassle. In most applications of DRLHP, human evaluation is expensive, collecting samples is cheap. So we'd expect `sum(weights)` to be much larger than `self.num_pairs`.

src/imitation/rewards/reward_nets.py
```diff
@@ -49,6 +50,25 @@ def forward(
     ):
         """Compute rewards for a batch of transitions and keep gradients."""
 
+    def preprocess(
```
AdamGleave, line 53: Mentioned this elsewhere, but am trying to wrap my head around why this should be a separate method, and not just rolled into the forward pass?

tests/test_preference_comparisons.py
```diff
@@ -0,0 +1,32 @@
...
+from imitation.util import util
+
+
+@pytest.fixture
+def expert_trajectories():
+    trajs = types.load("tests/data/expert_models/cartpole_0/rollouts/final.pkl")
+    return trajs
+
+
+def test_trainer_no_crash():
```
AdamGleave, line 20: aka a "smoke test": https://en.wikipedia.org/wiki/Smoke_testing_(software)

src/imitation/policies/trainer.py
```diff
@@ -0,0 +1,83 @@
...
+from stable_baselines3.common.vec_env import VecEnv
+
+from imitation.data import types
+from imitation.data.wrappers import BufferingWrapper
+from imitation.rewards.common import RewardFn
+from imitation.rewards.reward_nets import RewardNet
+from imitation.util.reward_wrapper import RewardVecEnvWrapper
+
+
+class AgentTrainer:
```
AdamGleave, line 14: Do you expect `AgentTrainer` to be used in other algorithms in `imitation`? I don't object it to being in its own file, just trying to understand why you split this up, and not e.g. the fragmenter, gatherer, etc.

src/imitation/policies/trainer.py
```diff
@@ -0,0 +1,83 @@
+"""Training policies with a specifiable reward function and collect trajectories."""
+from typing import List, Union
+
+from stable_baselines3.common.base_class import BaseAlgorithm
```
AdamGleave, line 4: Google Python style guide: import modules not classes. (mentioned previously in other files)

src/imitation/policies/trainer.py
```diff
@@ -0,0 +1,83 @@
...
+from imitation.data.wrappers import BufferingWrapper
+from imitation.rewards.common import RewardFn
+from imitation.rewards.reward_nets import RewardNet
+from imitation.util.reward_wrapper import RewardVecEnvWrapper
+
+
+class AgentTrainer:
+    """Wrapper for training an SB3 algorithm on an arbitrary reward function.
+
+    TODO(ejnnr): For preference comparisons, we might want to allow something more
```
AdamGleave, line 17: You can make them separate functions without being separate computation. The `train()` can just cache samples that can be returned by `sample()` (maybe with a fallback if there's no samples in the buffer when `sample()` is called?)

src/imitation/policies/trainer.py
```diff
@@ -0,0 +1,83 @@
...
+    but not when training), but on the other hand, sampling in addition to training
+    can be unnecessary overhead.
+
+    Args:
+        algorithm: the stable-baselines algorithm to use for training.
+            Its environment must be set.
+        reward_fn: either a RewardFn or a RewardNet instance that will supply
+            the rewards used for training the agent. In the latter case,
+            a simple wrapper is automatically applied to get a RewardFn.
+        device: an optional PyTorch device on which the RewardNet expects
```
AdamGleave, line 30: Not present in args.

src/imitation/policies/trainer.py
```diff
@@ -0,0 +1,83 @@
...
+        self,
+        algorithm: BaseAlgorithm,
+        reward_fn: Union[RewardFn, RewardNet],
+    ):
+        self.algorithm = algorithm
+        if isinstance(reward_fn, RewardNet):
+            reward_fn = reward_fn.predict
+        self.reward_fn = reward_fn
+
+        env = self.algorithm.get_env()
```
AdamGleave, line 44: Nit: `venv` is canonical for vector environments, `env` for single instance `gym.Env`s.

src/imitation/policies/trainer.py
```diff
@@ -0,0 +1,83 @@
...
+        return self.algorithm.policy
+
+    @property
+    def venv(self) -> RewardVecEnvWrapper:
+        venv = self.algorithm.get_env()
+        assert isinstance(venv, RewardVecEnvWrapper), "RewardVecEnvWrapper missing"
+        return venv
+
+    def _pop_trajectories(self) -> List[types.TrajectoryWithRew]:
+        # self.venv is the reward wrapper, so self.venv.venv is the BufferingWrapper
```
AdamGleave, line 82: This is fragile, why not just store it as `self.buffering_wrapper = BufferingWrapper(env)` on line 51?

AdamGleave (commented):
Thanks for the PR! Overall this looks pretty good. Most of my comments were stylistic. The `AgentTrainer` interface seems a bit clunky so discussed that a bit but I think you were aware this was an issue. Other main suggestion is to add logging -- it'll be very hard to figure out why it is/isn't working (or even if it's working) if you log nothing. You may be able to integrate it with SB logging, or write directly to TensorBoard files, or for now even just printing stuff is better than nothing.

src/imitation/policies/trainer.py
ejnnr: Yes, I thought `AgentTrainer` might be useful elsewhere (not even necessarily by other algorithms, but maybe by end users instead). I basically just think of it as a utility wrapper for training an SB3 algorithm on a reward function different than the one provided by the environment.

Admittedly, as long as we don't want to collect trajectories during training, it's probably simpler to just use the `RewardVecEnvWrapper` manually. And I might end up adding a bit more code to `AgentTrainer` that would rarely be useful outside preference comparisons. I'll leave this open for now and decide later whether the final version makes sense standing on its own

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+"""
+
+
+import abc
+from typing import Callable, List, Optional, Tuple
+
+import numpy as np
+import torch as th
+
+from imitation.data.fragments import Fragmenter, RandomFragmenter
```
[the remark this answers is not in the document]
ejnnr: Gotcha, I think I've fixed all occurrences of that now except for `imitation.data.types`. I've introduced type aliases `TrajectoryPair` and `TrajectoryWithRewPair` which make the type annotations much shorter, so adding the `types.` in front probably wouldn't be as bad anymore. IMO it's still more readable if we import `TrajectoryWithRewPair` etc. directly but I'm not sure how to weigh that against consistently following the Google style guide. Do you just want to make a decision on that, @AdamGleave? Either way is fine for me

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+    """Computes synthetic preferences using ground-truth environment rewards."""
+
+    def __init__(self, probabilistic: bool = False, noise_prob: float = 0.0):
+        """Initialize the synthetic preference gatherer.
+
+        Args:
+            probabilistic: if False (default), the preferences are either 1
+                (if fragment 1 has a higher sum of rewards), 0 (if fragment 2
+                has a higher sum of rewards) or 0.5 (if they are equal).
+                If True, a probability between 0 and 1 is returned
```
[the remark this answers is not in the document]
ejnnr: I think I also like your suggestion better, especially because it's closer to human feedback and slightly less trivial. In principle, I also like a temperature argument because I think it could replace both `probabilistic` and `noise_prob` (an infinite temperature would be equivalent to `noise_prob=1`). OTOH, the DRLHP paper as well as my own default implementation of the loss function assume the `noise_prob` version to model human error, rather than a temperature. It might be nice to have this be the same everywhere, and I don't really like having both `noise_prob` and `temperature`. Do you have a preference?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def __call__(
+        self, fragments: List[Tuple[TrajectoryWithRew, TrajectoryWithRew]]
+    ) -> np.ndarray:
+        rews1, rews2 = self._reward_sums(fragments)
+        if self.probabilistic:
+            # If probabilistic is True, we use the human model to compute probabilities.
+            # Instead of computing exp(rews1) / (exp(rews1) + exp(rews2)) directly,
+            # we divide enumerator and denominator by exp(rews1) to prevent overflows:
+            model_probabilities = 1 / (1 + np.exp(rews2 - rews1))
```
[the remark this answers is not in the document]
ejnnr: Good to know, but yeah, I think I'll leave it as it is

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -0,0 +1,111 @@
...
+        env_name,
+        num_vec,
+        seed=_seed,
+        parallel=parallel,
+        log_dir=log_dir,
+        max_episode_steps=max_episode_steps,
+    )
+
+    reward_net = reward_nets.BasicRewardNet(venv.observation_space, venv.action_space)
+    # HACK: verbose=1 prevents SB3 from configuring logger,
```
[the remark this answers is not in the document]
ejnnr: Created #325 to track this

ejnnr:
I'll add a few more tests tomorrow, then this should be ready to merge. But right now my code is incompatible with #326 (should be easy to change though). I'll adapt it to not rely on the global SB3 logger anymore and see if that breaks anything with current `master`. If so, we should probably wait for #326, otherwise order of merging the PRs doesn't matter.

ejnnr:
I switched to the local logger convention now, so this PR is now based off #328, which should be merged first! I left the base branch as `master` though to get code coverage reports.

ejnnr:
Given that I changed quite a few things to switch to the new logging convention, I think it'd be good if you did another (hopefully final) review, @AdamGleave. I also switched to a `temperature` parameter for the synthetic gatherer.

src/imitation/util/logger.py
```diff
@@ -125,6 +125,9 @@ def get_dir(self) -> str:
     def log(self, *args, **kwargs):
         self.default_logger.log(*args, **kwargs)
 
+    def warn(self, *args):
```
AdamGleave, line 128: Huh? Is this necessary? I think `warn` in the super class should call `self.log` which will then do the appropriate thing of forwarding to `self.default_logger`.

Note if we did want to add `warn`, we'd need to also add `info`, `error`, etc.

tests/test_agent_trainer.py
```diff
@@ -0,0 +1,45 @@
+# import gym  # noqa: F401
```
AdamGleave, line 1: delete?

tests/test_agent_trainer.py
```diff
@@ -0,0 +1,45 @@
...
+from imitation.rewards import common as rewards_common
+
+
+@pytest.fixture
+def agent():
+    return PPO("MlpPolicy", env="CartPole-v1")
+
+
+@pytest.fixture
+def reward_fn() -> rewards_common.RewardFn:
```
AdamGleave, line 15: Do you intend to change this in the future? If not, then I don't think it needs to be a fixture, you can just have:

```
def reward_fn(obs, acts, next_obs, dones):
    return dones.astype(float)
```

tests/test_agent_trainer.py
```diff
@@ -0,0 +1,45 @@
...
+    assert str(err.value) == "The environment for the agent algorithm must be set."
+
+
+def test_transitions_left_in_buffer(agent_trainer):
+    # Faster to just set the counter than to actually fill the buffer
+    # with transitions.
+    agent_trainer.buffering_wrapper.n_transitions = 2
+    with pytest.raises(RuntimeError) as err:
+        agent_trainer.train(steps=1)
+    assert str(err.value) == (
```
AdamGleave, line 42: You can use `pytest.raises(RuntimeError, match=msg)` instead.

tests/test_preference_comparisons.py
```diff
@@ -8,22 +10,100 @@
...
+
+def test_trainer_no_crash(agent_trainer, reward_net, fragmenter):
     main_trainer = preference_comparisons.PreferenceComparisons(
         agent_trainer, reward_net, sample_steps=10, fragmenter=fragmenter
     )
     main_trainer.train(2)
+
+
+def test_discount_rate_no_crash(agent_trainer, reward_net, fragmenter):
+    # also use a non-zero noise probability to ensure that doesn't cause errors
```
AdamGleave, line 53:
```suggestion
    # also use a non-zero noise probability to check that doesn't cause errors
```
I initially parsed this as saying we need non-zero noise probability otherwise it'll error, but I think this is saying that we're also testing that?

tests/test_preference_comparisons.py
```diff
@@ -8,22 +10,100 @@
...
+        # we'll ever reach
+        fragment_length=10000,
+        num_pairs=2,
+        seed=0,
+        warning_threshold=0,
+    )
+    with pytest.raises(ValueError) as err:
+        fragmenter(trajectories)
+
+    assert (
```
AdamGleave, line 89: Again can use the `match` kwarg

tests/test_preference_comparisons.py
```diff
@@ -8,22 +10,100 @@
...
+
+def test_preference_dataset_errors(agent_trainer, fragmenter):
+    dataset = preference_comparisons.PreferenceDataset()
+    trajectories = agent_trainer.sample(2)
+    fragments = fragmenter(trajectories)
+    # just create something with a different shape:
+    preferences = np.empty(len(fragments) + 1, dtype=np.float32)
+    with pytest.raises(ValueError) as err:
+        dataset.push(fragments, preferences)
+    assert "Unexpected preferences shape" in str(err.value)
```
AdamGleave, line 103: `match` again

tests/test_preference_comparisons.py
```diff
@@ -8,22 +10,100 @@
...
+
+    # Now test dtype
+    preferences = np.empty(len(fragments), dtype=np.float64)
+    with pytest.raises(ValueError) as err:
+        dataset.push(fragments, preferences)
+    assert str(err.value) == "preferences should have dtype float32"
```
AdamGleave, line 109: `match` again

AdamGleave (approved):
LGTM except for some minor changes. The only one I think is of significance is removing the `def warn` from `logger` which I think is unnecessary; the others are more cosmetic.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -410,13 +405,26 @@ def __init__(
...
+        self.logger = custom_logger or logger.configure()
         self.model = reward_model
-        self.reward_trainer = reward_trainer or CrossEntropyRewardTrainer(reward_model)
+        self.reward_trainer = reward_trainer or CrossEntropyRewardTrainer(
+            reward_model, custom_logger=self.logger
+        )
+        # If the reward trainer was created in the previous line, we've already passed
+        # the correct logger. But if the user created a RewardTrainer themselves and
+        # didn't manually set a logger, it would be annoying if a separate one was used.
+        if hasattr(self.reward_trainer, "logger"):
```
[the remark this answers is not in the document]
ejnnr: You're right, I think this is bad even in Python (for exactly the reason you describe). I went with 1. now because most subclasses seem to use loggers anyway. Had to turn `Fragmenter` etc., which where previously just type aliases into actual ABCs but maybe that would have made sense anyway at some point.

src/imitation/util/logger.py
ejnnr: Oh yeah, you're completely right

Issue #325, opened 2021-08-28
Upgrade stable-baselines3


ejnnr:
We are currently on `stable-baselines3~=0.10.0` while the newest version is `1.1.0`. There have been a lot of bug fixes since then (see https://github.com/DLR-RM/stable-baselines3/releases) and in particular, [issue #109](https://github.com/DLR-RM/stable-baselines3/issues/109) has been resolved, which would remove hacks in a number of places in `imitation` (just search for `#109` in the `imitation` code).

There are some breaking changes from `0.10.0` to `1.1.0`, we at least have to slightly modify logging in a few places because the global logger has been removed. Not sure if any of the other changes affect `imitation`.

PR #326, opened 2021-08-30
Remove dependency on global sb3 logger

ebuehrle:
Related to #325

I tried to keep the code changes minimal. Tests in `test_logger.py` pass, and `quickstart.py` as well as `quickstart.sh` seem to be functional.

### DIFF
--- a/src/imitation/algorithms/bc.py
+++ b/src/imitation/algorithms/bc.py
@@ -12,10 +12,12 @@ import numpy as np
 import torch as th
 import torch.utils.data as th_data
 import tqdm.autonotebook as tqdm
-from stable_baselines3.common import logger, policies, utils, vec_env
+from stable_baselines3.common import policies, utils, vec_env
 
 from imitation.data import rollout, types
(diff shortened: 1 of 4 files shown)


AdamGleave (approved):
LGTM. Long-term would be nice to move away from the global logger entirely, but this seems like a good stop-gap. I've made new issue https://github.com/HumanCompatibleAI/imitation/issues/327 to track that.

Will merge once tests pass (triggered)

AdamGleave:
There are some issues with the tests for the algorithms. Most proximally there needs to be a `record_mean` added. But more seriously there are a lot of cases (at least in tests) where we call `logger` without having called `logger.configure()`.

Since this looked like it'd be a fairly major change, I decided to just go ahead and switch to the local logger, which we wanted to do anyway, in https://github.com/HumanCompatibleAI/imitation/pull/328. Hope that's ok -- thanks for the PR!

ebuehrle:
Of course, no problem!

PR #331, opened 2021-09-01
Configure logger in training scripts

AdamGleave:
Hotfix for https://github.com/HumanCompatibleAI/imitation/pull/328

I forgot to update the scripts to use the new log configuration, which requires passing in the logger to the algorithm. Updates `train_{adversarial,bc,dagger}` to do this.

I also started seeing some unrelated unit test breakage in DAgger due to a mixture of trajectories that do and don't have `infos` (online demos have them, offline expert demos don't since they were stripped to save space). So I patched `rollouts` to handle this.

### DIFF
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -362,9 +362,13 @@ def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> Dict[str, 
     for t in trajectories:
         if t.infos is not None:
             ep_return = t.infos[-1].get("episode", {}).get("r")
-            if ep_return:
+            if ep_return is not None:
                 monitor_ep_returns.append(ep_return)
     if monitor_ep_returns:
+        # Note monitor_ep_returns[i] may be from a different episode than ep_return[i]
+        # since we skip episodes with None infos. This is OK as we only return summary
+        # statistics, but you cannot e.g. compute the correlation between ep_return and
+        # monitor_ep_returns.
         traj_descriptors["monitor_return"] = np.asarray(monitor_ep_returns)
         # monitor_return_len may be < n_traj when infos is sometimes missing
         out_stats["monitor_return_len"] = len(traj_descriptors["monitor_return"])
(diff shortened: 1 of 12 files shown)


src/imitation/data/rollout.py
```diff
@@ -358,10 +358,16 @@ def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> Dict[str,
...
     }
 
-    infos_peek = trajectories[0].infos
-    if infos_peek is not None and "episode" in infos_peek[-1]:
-        monitor_ep_returns = [t.infos[-1]["episode"]["r"] for t in trajectories]
+    monitor_ep_returns = []
+    for t in trajectories:
+        if t.infos is not None:
+            ep_return = t.infos[-1].get("episode", {}).get("r")
+            if ep_return:
```
ejnnr, line 365: If I understand the code correctly, this should be `if ep_return is not None`? Otherwise, we would skip episodes that happen to have exactly zero return

src/imitation/data/rollout.py
```diff
@@ -358,10 +358,16 @@ def rollout_stats(trajectories: Sequence[types.TrajectoryWithRew]) -> Dict[str,
...
+                monitor_ep_returns.append(ep_return)
+    if monitor_ep_returns:
         traj_descriptors["monitor_return"] = np.asarray(monitor_ep_returns)
+        # monitor_return_len may be < n_traj when infos is sometimes missing
+        out_stats["monitor_return_len"] = len(traj_descriptors["monitor_return"])
```
ejnnr, line 370: It feels potentially confusing that the i-th `monitor_return` doesn't necessarily correspond to the same episode as the i-th `return` or `len`. I'm not sure if there's a better way though; we could use masked numpy arrays but I'm not familiar enough with them to say whether that would just lead to more issues down the road. So probably just leave it as it is but add a sentence about it to the docstring?

ejnnr (approved):
LGTM except for my two small comments (I think at least the `if ep_return` statement should be fixed before merging, unless I'm missing something).

I saw that the `AdversarialTrainer` still has an `init_tensorboard` argument but AFAICT that's correct because it's logging some extra stuff. If the `custom_logger` also logs to tensorboard, it might be nice if `AdversarialTrainer` could use the same directory automatically. But not sure how easy that would be and in any case it should probably happen in a separate PR so we can merge this fix quickly.

(Btw, glad you caught the logging issue before I merged #320, I made the same mistake there :sweat_smile:)

AdamGleave:
Thanks for the review, both good points.

Issue #219, opened 2020-07-31
PyTorch complains when we pass in Transitions because arrays are nonwritable


shwang:
After #208, we get warnings like

```
tests/test_dagger.py::test_trainer_makes_progress
  /pytorch/torch/csrc/utils/tensor_numpy.cpp:141: UserWarning: The given NumPy array is not writeable, and PyTorch does not support non-writeable tensors. This means you can write to the underlying (supposedly non-writeable) NumPy array using the tensor. You may want to copy the array to protect its data or make it writeable before converting it to a tensor. This type of warning will be suppressed for the rest of this program.
```

According to what I read online, this makes PyTorch copy the array before Tensor conversion (which TF1 had to do anyways, I'm pretty sure) and incurs a small speed cost compared to using mutable array. Luckily PyTorch is polite enough to not make this warning appear 100 times.

Might be nice to suppress this warning somehow.

shwang:
Also I read that Pytorch can actually mutate the underlying numpy memory in some cases?

@qxcv I'm not very familiar with PyTorch -- I don't think this is the case, but do you know if there's any place in our code that would likely lead to numpy mutations?

As I understand it, one of the only ways you can mutate a numpy array after it is converted to a Tensor is by direct assignment to an element of that Tensor.

qxcv:
> do you know if there's any place in our code that would likely lead to numpy mutations?

I don't, no. Mutations are generally not recommended in Torch networks because they complicate backprop code. Scott also asked me about this & I recommended to just suppress the warning in the BC forward prop/backprop code. I guess if we really wanted to, we could also figure out what code is declaring arrays unwritable & fix that. My bet is one some environment code caching an observation that it doesn't want to be written to, or maybe some multiprocessing code that is keeping arrays in shared memory somewhere.

shwang:
I made the arrays in Transitions unwritable upon initialization.

qxcv:
Ahhhh okay. Yeah, in that case I guess our best bet is just to silence the errors (or add a redundant copy if running on CPU :shrug:)

PR #332, opened 2021-09-01
Copy to make arrays mutable for PyTorch

AdamGleave:
Fixes https://github.com/HumanCompatibleAI/imitation/issues/219

May introduce a slight performance overhead -- I've not tested. @shwang thought PyTorch was soft-copying anyway. I expect this to be insignificant for current workloads, we can do something smarter if/when profiling reveals this is an issue.

### DIFF
--- a/src/imitation/data/types.py
+++ b/src/imitation/data/types.py
@@ -109,7 +109,7 @@ def transitions_collate_fn(
     dicts.
     """
     batch_no_infos = [
-        {k: v for k, v in sample.items() if k != "infos"} for sample in batch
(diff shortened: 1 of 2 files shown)


AdamGleave:
@qxcv will you have time to review? If not I'll ask someone else to take a look.

qxcv:
Yep, can review!

qxcv (approved):
Looks good to me. I'm not too worried about perf impact. It it's a bottleneck (e.g. for image envs with huge observations) then it should be pretty obvious from profiling later on, since it's just one function call.

PR #334, opened 2021-09-03
Test refactor: move into subdirectories and other misc cleanup

AdamGleave:
The `tests/` directory was getting very unwieldy with lots of files as `imitation` has grown. I've refactored this to more or less follow the hierarchical structure of `src/`, with a few exceptions (e.g. `experiments/` and `scripts/` just have `test_experiments.py` and `test_scripts.py` and not their own subdirectory).


examples/quickstart.py
```diff
@@ -12,8 +12,8 @@
 from imitation.util import logger, util
 
 # Load pickled test demonstrations.
-with open("tests/data/expert_models/cartpole_0/rollouts/final.pkl", "rb") as f:
-    # This is a list of `imitation.data.types.Trajectory`, where
+with open("tests/testdata/expert_models/cartpole_0/rollouts/final.pkl", "rb") as f:
+    # This is a list of `imitation.testdata.types.Trajectory`, where
```
ejnnr, line 16: This is one replace to many, right? (should still be `imitation.data.types.Trajectory`)

ejnnr (approved):
LGTM except for the one typo and the fact that it looks like you still need to commit a `custom_logger` fixture to `conftest.py`? Happy to take another look after that but I don't think it's necessary once the tests pass.
I only looked at the diff though, didn't check whether you caught every instance of `data` => `testdata` that's needed.

ejnnr:
Just saw the `custom_logger` fixture exists in #333, looks good

examples/quickstart.py
AdamGleave: Yep, good catch.

AdamGleave:
Thanks for the review, sorry that the `conftest.py` change got missed, my bad for cherrypicking commits and not checking everything was included.

PR #335, opened 2021-09-03
Trajectory: add flag indicating if terminal

AdamGleave:
Adds a flag `terminal` to `Trajectory` indicating if the trajectory contains the terminal state of an episode. h/t @ejnnr for the idea.

This is useful in a couple of places:
  - For variable horizon checks introduced in https://github.com/HumanCompatibleAI/imitation/pull/333 this lets us discard incomplete trajectories from the check, without needing to mess around with separate functions to return full and partial trajectories.
  - When converting partial trajectories to transitions, we don't want `done=True` for the last (but non-terminal) state, especially when reward models depend on `done`.
  - We sometimes want to use `Trajectory` for trajectory fragments, e.g. in preference comparison. We could introduce a separate `TrajectoryFragment` type, but given they contain exactly the same fields, this is probably not necessary and `terminal` is simpler.

One thing to flag is that `Trajectory` can have `terminal` set to true even if it's a trajectory fragment, so long as the *last* state is terminal. We could introduce another field `initial` to catch this case, but I don't think we currently have any reason to care about this.


ejnnr (approved):
Thanks for implementing this! LGTM except for my comment on the `RandomFragmenter` and the fact that I don't think you meant to commit all those files in `tests/testdata/`? Once those things are addressed, should be fine to merge without a re-review

PR #336, opened 2021-09-06
Patches for preference comparison implementation

ejnnr:
This PR contains a few bug fixes for #320, as well as a more usable logger setup and a few more exposed hyperparameters. However, training still doesn't work with the current hyperparameters on `seals/CartPolev-v0`, so there are more issues and this is a WIP until those are fixed.


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -201,24 +207,42 @@ def __init__(
                 how good a fragment is. Default is to use undiscounted
                 sums of rewards (as in the DRLHP paper).
             seed: seed for the internal RNG (only used if temperature > 0)
+            threshold: preferences are sampled from a softmax of returns.
+                To avoid overflows, we clip differences in returns that are
+                above this threshold (after multiplying with temperature).
+                This threshold is therefore in logspace. The default value
+                of 50 means that probabilities below 2e-22 are rounded to 0.
```
AdamGleave, line 214: From looking at the code, I think it's the opposite? That probabilities are rounded up to 2e-22, since |returns_diff| is clamped to at most threshold. Not that this makes a practical difference.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -201,24 +207,42 @@ def __init__(
...
         # we divide enumerator and denominator by exp(rews1) to prevent overflows:
-        model_probs = 1 / (1 + np.exp(rews2 - rews1))
+        model_probs = 1 / (1 + np.exp(-returns_diff))
+        # Compute the mean binary entropy. This metric helps estimate
+        # how good we can expect the performance of the learned reward
+        # model to be at predicting preferences.
+        entropy = -(
+            special.xlogy(model_probs, model_probs)
+            + special.xlogy(1 - model_probs, 1 - model_probs)
+        ).mean()
```
AdamGleave, line 244: `sum()` not mean as `model_probs` are, well, probability weightings?

AdamGleave: I think just `scipy.stats.entropy(model_probs)` would do what you want.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -343,18 +370,27 @@ def __init__(
...
+            lr: the learning rate
             discount_factor: the model of preference generation uses a softmax
                 of returns as the probability that a fragment is preferred.
                 This is the discount factor used to calculate those returns.
                 Default is 1, i.e. undiscounted sums of rewards (which is what
                 the DRLHP paper uses).
+            threshold: the preference model used to compute the loss contains
+                a softmax of returns. To avoid overflows, we clip differences
+                in returns that are above this threshold.
+                This threshold is therefore in logspace. The default value
```
AdamGleave, line 383: Same issue as before.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -407,12 +446,13 @@ def train(self, dataset: PreferenceDataset):
             ),
             collate_fn=preference_collate_fn,
         )
-        for fragment_pairs, preferences in dataloader:
-            self.optim.zero_grad()
-            loss = self._loss(fragment_pairs, preferences)
-            loss.backward()
-            self.optim.step()
-            self.logger.record("reward/loss", loss.item())
+        for ep in range(self.episodes):
```
AdamGleave, line 449: What does this have to do with episodes? This seems closer to number of epochs, as we iterate over the whole `dataloader` (and thereby `dataset`), which AFAICT might contain data from multiple (RL) episodes?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -499,17 +541,21 @@ def train(self, iterations: int):
...
-            self.reward_trainer.train(self.dataset)
-            self.logger.log(f"Training agent for {self.agent_steps} steps")
-            self.trajectory_generator.train(steps=self.agent_steps)
+            with self.logger.accumulate_means("reward"):
+                self.logger.log("Training reward model")
+                self.reward_trainer.train(self.dataset)
+            with self.logger.accumulate_means("agent"):
+                self.logger.log(f"Training agent for {self.agent_steps} steps")
+                self.trajectory_generator.train(steps=self.agent_steps)
+            self.logger.dump(i)
```
AdamGleave, line 561: FYI I think this is meant to be globally unique, so weird things may happen with logging if you call `train` multiple times (but I think we don't do that?)

src/imitation/policies/trainer.py
```diff
@@ -45,6 +46,42 @@ def train(self, steps: int, **kwargs):
                 the training procedure.
         """
 
+    def set_logger(self, custom_logger: Optional[logger.HierarchicalLogger]):
```
AdamGleave, line 49: We used `@property` and `@property.setter` in `BaseImitationAlgorithm` so maybe stick to that convention (although, to be clear, both approaches seem fine in general)? Unless you want `logger` getter to not be exposed in public API?

src/imitation/policies/trainer.py
```diff
@@ -45,6 +46,42 @@ def train(self, steps: int, **kwargs):
...
+        """Set the internal logger.
+        For subclasses, this should also set the logger of any wrapped modules
```
AdamGleave, line 51:
```suggestion
        
        For subclasses, this should also set the logger of any wrapped modules
```

src/imitation/policies/trainer.py
```diff
@@ -45,6 +46,42 @@ def train(self, steps: int, **kwargs):
...
+        self._trajectories = types.load(path)
+        self.rng = random.Random(seed)
+
+    def sample(self, steps: int) -> Sequence[types.TrajectoryWithRew]:
+        available_steps = sum(len(traj) for traj in self._trajectories)
+        if available_steps < steps:
+            raise RuntimeError(
+                f"Asked for {steps} transitions but only {available_steps} available"
+            )
+        trajectories = list(self._trajectories)
```
AdamGleave, line 71: Similar logic to `AgentTrainer` -- factor out into helper method?

src/imitation/policies/trainer.py
```diff
@@ -45,6 +46,42 @@ def train(self, steps: int, **kwargs):
...
+        (such as SB3 algorithms).
+
+        Args:
+            logger: the HierarchicalLogger instance to use.
+        """
+        self._logger = custom_logger or logger.configure()
+
+
+class TrajectoryDataset(TrajectoryGenerator):
+    def __init__(self, path: types.AnyPath, seed: int = 0):
```
AdamGleave, line 61: You should call superclass `__init__`. If that does something you don't want (e.g. the logger), then consider moving that logic out of the superclass.

AdamGleave (commented):
Minor suggestions for improvement. Nothing jumps out in terms of potential bugs.

One high-level logging suggestion would be to record the accuracy: how often does the trajectory with higher predicted return correspond to the one that the "user" chose? I found this more interpretable than the loss. https://github.com/HumanCompatibleAI/evaluating-rewards/blob/master/src/evaluating_rewards/rewards/preferences.py#L228

You might also be able to log accuracy relative to the "true" (noiseless) comparison based on ground-truth reward, which should be able to reach 100% (even though the noise floor in "user" feedback lower bounds the loss), though this sounds tricky to do in a maintainable way so perhaps best limited to debugging.

AdamGleave:
Took a quick overall look at the preference comparison code to see if I could spot anything that might be causing the issues you see. Some thoughts:
  - **Big one**: I think there's a bug in `AgentTrainer` where trajectories captured via `sample()` that were not already in buffering wrapper are captured directly from `self.venv` which is *after* applying the reward wrapper. So effectively you're sampling trajectories labeled with whatever your reward model thinks is currently correct! Although I'd expect this to only slow down training rather than making loss actually go up. I think simplest option to fix this is to just *discard* the return value of `generate_trajectories`, and set `additional_trajectories = self._pop_trajectories()` instead. You could also call rollout directly on the original venv (before wrapping with buffer or reward wrapper).
  - You probably want to call `learn` in `AgentTrainer.train` with `reset_num_timesteps=False`. Otherwise SB3 will think you're training stuff from scratch. I think this mostly negatively affects logging rather than actually causing problems with training, though, so I doubt it explains the issues you're seeing.
  - I don't see you applying `VecNormalize` anywhere, which can really help with RL training in continuous control environments. I'd at least sanity check that you can learn a good policy without normalization on the ground-truth reward in the environments you're testing. You should also make sure your reward scale is reasonable -- if it's very large or very small, you'll need to decrease/increase learning rate for policy accordingly. Good to eventually add VecNormalize support, but not sure I'd recommend it immediately (it's tricky to get right, since it introduces a new non-stationarity, and you have to restore the normalization stats when restoring reward models).
  - IIUC, the preference dataset grows without bound, and is randomly sampled from. This effectively means that samples collected early on in training get trained on more than those collected later in training. I think this may just be how DRLHP works, but I found it a bit odd, and expect that importance sampling would help here. But not a priority.

Some general ideas for debugging the loss going up:
  - In the extreme case, if you set learning rate of RL algorithm to zero so the policy is static), does the reward model loss go down? Distribution is IID stationary then, but it's still a bit harder to train on than a fixed dataset which could just be memorized. If that works, start gradually increasing RL learning rate until the policy starts learning something -- does reward model loss still go down? You could also play around with number of RL policy agent training timesteps with similar effect.
  - If it's instability, try just setting a lower learning rate for the reward model!
  - We already discussed warm-starting the reward model at something good and seeing if it's preserved with subsequent training.

src/imitation/algorithms/preference_comparisons.py
ejnnr: Oh, you're completely right. Fixed

src/imitation/algorithms/preference_comparisons.py
ejnnr: I thin `sum` is correct and that `scipy.stats.entropy(model_probs)` does something different. What might be causing some confusion: each entry of `model_probs` is a complete description of a single Bernoulli distribution (namely for one fragment pair). I want to compute the entropy of each Bernoulli distribution and then take the mean of all of them. This is analogous to how the `binary_cross_entropy` loss I'm using in `CrossEntropyTrainer` works, so if I have both metrics, I can tell how well the model does (because the mean entropy I'm computing is the lower bound on the mean cross-entropy).

Using `sum` or `mean` is pretty arbitrary but `mean` makes it more interpretable because it's independent from the batch size (and it's what pytorch uses for `binary_cross_entropy` by default.

I could use `scipy.stats.entropy`, but I'd have to stack together `model_probs` and `1 - model_probs` because `scipy`.stats.entropy` works on arbitrary categorical distributions rather than Bernoulli probabilities.

src/imitation/algorithms/preference_comparisons.py
ejnnr: Also fixed

src/imitation/algorithms/preference_comparisons.py
ejnnr: Um, yeah, I meant epochs, it just seems that working on RL has had weird effects on my brain xD. Fixed now.

src/imitation/policies/trainer.py
ejnnr: Good point, I switched to properties now

src/imitation/policies/trainer.py
ejnnr: Good idea, done

src/imitation/policies/trainer.py
ejnnr: Done. We don't need the logger here for now but I think we decided to just make sure a logger always exists so we can set it in `PreferenceComparisons`

src/imitation/algorithms/preference_comparisons.py
AdamGleave: OK, I'm convinced by this.

src/imitation/algorithms/preference_comparisons.py
ejnnr: Our script only calls it once, but I think we should support calling it multiple times (as long as it's not much work), so I've added a global step counter

ejnnr:
I think I've addressed everything essential now. Some notable other changes:
- I've moved `AgentTrainer` etc. into `preference_comparisons.py` because it was getting pretty specific to that. This means that `preference_comparisons.py` is now almost 900 lines but if that's too much, I'd rather suggest we split it up into submodules inside the `algorithms` package (inside a `preference_comparisons` subpackage perhaps)
- There are some new configuration options, in particular for loading things from disk (e.g. warm starting the agent)
- Instead of specifying the number of iterations (i.e. complete cycles), `PreferenceComparisons.train()` now takes `total_timesteps` and `total_comparisons`, which of course required changing a few other arguments as well. I think this is more natural because it's easy to compare to other algorithms (and will probably be better suited to a unified training interface)

Things I'll probably want to implement at some point but not right now:
- Train the reward model for a while before starting agent training
- Decay the rate at which the reward model is trained and preferences gathered (because towards the end, the reward model is already very good and just needs to be trained a bit to prevent reward hacking)
- Discard old preferences by limiting the size of the preference dataset
- Tune default hyperparameters a bit (I'd guess they're currently pretty inefficient)
- Make sure we really only use about `total_timesteps` agent training timesteps (right now it's sometimes significantly more because of `PPO`s `n_steps` and parallel environments)

The first three of these are straight from the DRLHP paper and it seems reasonable that they could make things more efficient. Of course, there's also human feedback and an ensemble of reward models as potential extensions but those should definitely be separate PRs.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -452,9 +746,22 @@ def __init__(
                 Default (and currently the only option) is to use synthetic preferences
                 based on ground-truth rewards. Human preferences could be implemented
                 here in the future.
+            preferences_path: path to a pickled PreferenceDataset which will be
```
AdamGleave, line 749: This seems like a useful feature for debugging, but I don't like how it has so many effects throughout the code, being used as a flag to toggle whether to train the agent or gather data as well.

I'm inclined to remove this, and maybe keep the changes in some private branch you could use later.

An alternative I'd be happier with is to let the callee pass in a `dataset` directly (that could be loaded from disk), have a separate flag to disable adding to the dataset (or you could even just run that code anyway but have the fixed dataset ignore any additions -- though that's inefficient; or perhaps the dataset itself could have a frozen attribute that's checked?), and a separate flag to disable agent training (or you could maybe even just make an AgentTrainer wrapper with a `train` that just forwards to `sample`).

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -463,12 +770,20 @@ def __init__(
                 condition, and can seriously confound evaluation. Read
                 https://imitation.readthedocs.io/en/latest/guide/variable_horizon.html
                 before overriding this.
+            seed: seed to use for initializing subcomponents such as fragmenter.
+                Only used when default components are used; if you instantiate your
+                own fragmenter, preference gatherer, etc., you are responsible for
+                seeding them!
```
AdamGleave, line 776: Mention return value in docstring?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -481,35 +796,89 @@ def __init__(
...
+                num_steps = timesteps_per_iteration
+                # if the number of timesteps per iterations doesn't exactly divide
+                # the desired total number of timesteps, we train the agent a bit longer
+                # at the end of training (where the reward model is presumably best)
+                if i == iterations - 1:
+                    num_steps += extra_timesteps
+                with self.logger.accumulate_means("agent"):
+                    self.logger.log(f"Training agent for {num_steps} timesteps")
+                    self.trajectory_generator.train(steps=num_steps)
+            self.logger.dump(self._iteration)
```
AdamGleave, line 881:
```suggestion
            
            self.logger.dump(self._iteration)
```

tests/generate_test_data.sh
```diff
@@ -3,4 +3,18 @@
 set -e
 
 # Regenerate tests/testdata/expert_models (for various tests).
-experiments/train_experts.sh -r
+# experiments/train_experts.sh -r
```
AdamGleave, line 6: Don't think you want to commit this line.

tests/test_scripts.py
```diff
@@ -65,9 +65,32 @@ def test_main_console(script_mod):
...
+    },
+    {
+        "agent_path": (
+            "tests/testdata/expert_models/cartpole_0/policies/final/model.zip"
+        ),
+        # TODO(ejnnr): the policy we load was trained on 8 parallel environments
+        # and for some reason using it breaks if we use just 1 (like would be the
+        # default with the fast named_config)
+        "num_vec": 8,
+        # we're putting this here to test it at some point without having
```
AdamGleave, line 81: "this", "it" -- can you be a bit more precise? Even if just "the below" (e.g. right now I'm not sure if you're referring to the immediate next line or the next two lines).

AdamGleave (commented):
This is looking good. Left mostly minor comments. The one semi-major thing is I think `preferences_path` has complexified the code significantly, so I'd like to find some alternative way of implementing that (or just drop the feature). Please LMK when it's ready for me to take another look.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -3,48 +3,267 @@
...
+                self.venv,
+                sample_until=sample_until,
+            )
+            additional_trajectories = self._pop_trajectories()
+
+            trajectories = list(trajectories) + list(additional_trajectories)
+
+        return _get_trajectories(trajectories, steps)
+
+    def _pop_trajectories(self) -> Sequence[types.TrajectoryWithRew]:
```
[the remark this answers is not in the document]
ejnnr: Good point, removed

src/imitation/algorithms/preference_comparisons.py
ejnnr: I assume you mean for the `train()` method? Added it there

tests/generate_test_data.sh
ejnnr: Oops, thanks for catching that!

tests/test_scripts.py
ejnnr: I meant both of them, I've rewritten the comment

src/imitation/algorithms/preference_comparisons.py
ejnnr: Agreed that it makes things unnecessarily complex, I've removed it for now.

My motivation for leaving it in was mainly that I think it will be useful if we ever have an implementation of human feedback; in that case, you might want to quickly train a reward model on already collected preferences (though this is of course not equivalent to online training, so it would just be for quick tests).

But if we decide we do need it, I think the simplest option would just be to have a script that doesn't rely on `PreferenceComparisons` at all. If we have a fixed dataset, the code to train on that is literally just
```
reward_trainer = CrossEntropyRewardTrainer(reward_net, **kwargs)
dataset = PreferenceDataset.load(preference_path)
reward_trainer.train(dataset)
```
apart from logging, creating the `RewardNet` etc. So it would be very quick to create an extra script for that if needed.
(I wanted to integrate this into the existing training script at first, but even that isn't very clean)

ejnnr:
Thanks for the review! I think I've addressed everything (and I added a few more tests)

src/imitation/algorithms/preference_comparisons.py
ejnnr: (I did leave the preference saving feature in because that's not much overhead and if we ever have human feedback, I think it'll be good practice to save them)

src/imitation/algorithms/preference_comparisons.py
AdamGleave: yes, sorry for commenting in the wrong place, thanks.

src/imitation/algorithms/preference_comparisons.py
AdamGleave: Ah, good point about training from an offline dataset. That is a reasonable use case, although I think one might often want to warm-start from an offline dataset but then continue to do online training. Seems tricky to support all these possible use cases, so I think breaking it down into using the underlying components is probably more scalable as you suggest.

Being able to save datasets seems good.

AdamGleave (approved):
LGTM. I removed the testdata file you added for preferences which seems to no longer be used, and added some test coverage. Please check my changes and merge if happy.

ejnnr:
Thanks for the changes! One of the tests broke because my preference comparison scripts expects `agent_path` to be a path directly to the policy `.zip` file, rather than the folder containing it. The reason is that I'm not using the `NormalizePolicy` wrapper (since the policy and `RewardNet` should probably use the same normalization). So taking in the entire folder as an argument would be misleading.

On reflection, I think the correct thing to do in this case would be to also warm start the `VecNormalize` stats (but still train them, since we're still training the policy). If `normalize=False` and we warm start an agent with associated normalization stats, we should just raise an error because that's probably unintentional.

ejnnr:
I've implemented what I've described above now. There are some things that would probably be good to test (see e.g. the bug fix I added in 61ab6b3c4339d85498f4875f91175f778ec58467) but that seems hard to do because we'd have to "look inside" the script. Would probably require mocking the agent/VecNormalize/... to make sure everything is used correctly.

Regarding test coverage: looks like the main issue are the configs for environments we don't currently test. Would of course be easy to add those but I think the other algorithms don't do that, presumably to keep test runtimes more manageable?

AdamGleave (approved):
LGTM. The normalization thing is quite messy, but I think resolving this is going to require a significant re-design of how we handle this in `imitation` so let's address that in another PR (see my message on Slack in #imitation).

Coverage seems OK as-is, I agree we don't care about coverage of config files that much.

PR #340, opened 2021-09-07
rollout new type migration: update test data, fix conversion script, add test coverage

AdamGleave:
Fix-up a script and add test coverage


src/imitation/scripts/convert_trajs_in_place.py
```diff
@@ -1,8 +1,7 @@
 """
 This script takes as command-line input multiple paths to pickle files containing
-Sequence[imitation.types.TrajectoryWithRew] or Sequence[imitation.old_types.Trajectory].
-It overwrites each path with a new pickle where the demonstration data is saved as
-Sequence[imitation.types.TrajectoryWithRew].
+Sequence[imitation.types.TrajectoryWithRew]. It overwrites each path with a new pickle
```
ejnnr, line 3: May want to mention here that this adds `terminal=True`? Otherwise it's not clear what this script does (and in contrast to what this script did previously, adding `terminal=True` might not always be the right thing to do)

ejnnr (approved):
LGTM and thanks for adding a test case. Only comment I have is that adding `terminal=True` is not really right when a file contains partial trajectories. We can't do anything about that and it would happen anyway if someone loaded the old file but it should maybe be mentioned somewhere.

PR #341, opened 2021-09-07
Add `env_make_kwargs` to `imitation.util.util.make_vec_env`

kmdanielduan:
This PR adds an optional argument to `imitation.util.util.make_vec_env` to support passing `kwargs` into `gym.make` or `spec.make`. This is an innate feature in [gym](https://github.com/openai/gym/blob/a4c3f15e9b0c7525ec9006e92a3bf6bc8bd0a798/gym/envs/registration.py#L61).

This is needed to specify environment variants in reward-function-transfer.


src/imitation/util/util.py
```diff
@@ -41,6 +43,7 @@ def make_vec_env(
     log_dir: Optional[str] = None,
     max_episode_steps: Optional[int] = None,
     post_wrappers: Optional[Sequence[Callable[[gym.Env, int], gym.Env]]] = None,
+    env_make_kwargs: Optional[Mapping[str, Any]] = {},
```
AdamGleave, line 46: As written this isn't actually `Optional` -- the code will choke if you pass in a `None`.

However, it's also bad practice to use `{}` as a default argument, since it's actually a mutable object so the default can be *changed* if you were to mutate `env_make_kwargs` anywhere.

I'd suggest changing `{}` to `None`, and then adding inside the function:
```python
env_make_kwargs = env_make_kwargs or {}
```

AdamGleave (approved):
LGTM apart from one issue with the default value for `env_make_kwargs`

AdamGleave:
Tests won't automatically run on this branch -- when you make the changes, just ping me and I'll trigger them to run and can merge for you.

src/imitation/util/util.py
AdamGleave: I think I would prefer it have a default argument, don't want every caller to have to pass in `None` explicitly. It won't always be called from Sacred scripts.

kmdanielduan (commented):
I changed the `env_make_kwargs` default value accordingly. Sorry I didn't notice this is the `make_vec_env` function which doesn't have a Sacred config.

AdamGleave (approved):
LGTM

Issue #271, opened 2021-03-05
Generate trajectories from rule based agent (instead of RL algo)


SurferZergy:
Hi, 

Is there a way to generate the trajs from a rule based (or human) expert? As I imagine generating it from another RL is basically the same as training RL from scratch.

Looking through the project, my thinking is the only way to do this right now is to modify the ```get_action``` function in https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/data/rollout.py and remove the policy input to the ```generate_trajectories``` func, and implement the rule based agent inside that func as well.

Is there a "better way" to do this? like convert a rule base agent to a policy? 

Edit: I just read this:  
GAIL, and AIRL also accept as `expert_data` any Pytorch-style DataLoader that
iterates over dictionaries containing observations, actions, and next_observations.

Are there any examples of what an expert_data dict might look like?
like {[(np array)obs1]: ('(int)action', '(np array)[obs2]')}?

AdamGleave:
Closing due to inactivity -- feel free to reopen if this is still an issue.

PR #344, opened 2021-09-08
Make generate_trajectories more flexible

ejnnr:
`generate_trajectories` currently needs an SB3 policy or algorithm. With this PR, it also accepts a `Callable` that generates actions from observations, or `policy=None` (in which case random actions are used). I don't have any plans to use this inside `imitation` itself, only in the reward preprocessing project, so if you don't think it belongs here, feel free to close this and I'll just implement it there. But #271 already wanted support for `Callable`s so maybe others will use that as well in the future, which is why I created the PR here.

### DIFF
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -237,9 +237,12 @@ def make_sample_until(
 
     return sample_until
 
+# A PolicyCallable is a function that takes an array of observations
+# and returns an array of corresponding actions.
+PolicyCallable = Callable[[np.ndarray], np.ndarray]
 
 def generate_trajectories(
-    policy,
+    policy: Union[BaseAlgorithm, BasePolicy, PolicyCallable, None],
     venv: VecEnv,
     sample_until: GenTrajTerminationFn,
     *,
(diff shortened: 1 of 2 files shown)


AdamGleave (approved):
Thanks for the PR. This feature seems worth upstreaming into imitation even if we don't need it there yet -- it's a fairly natural thing to want to do.

I've refactored the policy to callable conversion into its own function -- I don't feel strongly about this (it's not used anywhere else), so feel free to revert that, but it seemed a bit easier to reason about and has a nice and simple type signature.

I also expanded the types of `generate_transitions` and `rollout_and_save` to allow for callables/None, too, since they just use `generate_trajectories` under the hood.

Please do check my changes before merging!

ejnnr:
Thanks for the review, and I like your changes :) Your reordering of the `elif` statements introduced a bug though -- luckily the test suite caught that because I wouldn't have (my original ordering wasn't deliberate). I've fixed it and added a comment (unfortunately also had to add back in the `pytype` ignore comment). I've also added some smoke tests for random and `Callable` policies.

I've requested another review in case you spot any other similar bugs, and so you can take a look at the new tests.

AdamGleave (approved):
> I've requested another review in case you spot any other similar bugs, and so you can take a look at the new tests.

LGTM.

> Your reordering of the `elif` statements introduced a bug though -- luckily the test suite caught that because I wouldn't have (my original ordering wasn't deliberate).

Ugh, sorry, should have checked the tests passed before committing. Thanks for fixing my mistake!

> Hm, not sure how to best deal with this line in the test suite that's never executed:

I just added a `# pragma: no cover` line. We demand 100% test coverage so we don't accidentally write test code that doesn't get executed. But things like this (input validation for a helper method in a test) are fine to not execute, so it's fine to override it.

PR #347, opened 2021-09-09
Adding Docker runner scripts to quickly setup Docker environments for development

kmdanielduan:
This PR aims to create two Docker runner scripts to quickly set up Docker environments for development. 

- `runners/build_push_image.sh base python-req` builds and pushes (optional) the docker images with tags `base` and `python-req`.
- `runners/launch_docker-dev.sh` launches the `python-req` image and provides a way of mounting local code and `mjkey.txt`.


ejnnr:
I wrote my own version of `launch_docker-dev.sh` just a few hours ago, so having a script like that in the repository might make sense :) Unless there's some other good way of developing `imitation` on a server that I'm missing?

runners/launch_docker-dev.sh
```diff
@@ -0,0 +1,25 @@
...
+
+# install imitation in developer mode
+CMD="pip install -e .[docs,parallel,test] gym[mujoco]"  # borrowed from ci/build_venv.sh
+
+# docker pull ${DOCKER_IMAGE}  # Uncomment this line to pull the image
+docker run -it --rm --init \
+       -v ${LOCAL_MNT}:/imitation \
+       -v ${MJKEY_MNT}:/root/.mujoco/mjkey.txt \
+       ${FLAGS} ${DOCKER_IMAGE} \
+       /bin/bash -c "${CMD} && bash"
```
ejnnr, line 25: I haven't tested it, but won't this give us an additional bash process so that we have to exit twice to get out of the Docker container? I'm just using exec bash in my script and I don't think that has any drawbacks

runners/launch_docker-dev.sh
```diff
@@ -0,0 +1,25 @@
...
+  LOCAL_MNT="${HOME}/imitation"
+fi
+
+# Pass your own mjkey.txt
+if [[ ${MJKEY_MNT} == "" ]]; then
+  MJKEY_MNT="${HOME}/mnt/mjkey.txt"
+fi
+
+# install imitation in developer mode
+CMD="pip install -e .[docs,parallel,test] gym[mujoco]"  # borrowed from ci/build_venv.sh
```
ejnnr, line 18: One thing I'm confused about: `build_venv.sh` creates a virtual environment `/venv`, so in my version of this script, I activated that. But apparently, `/venv/bin` is already part of `$PATH` for the Docker container, so this might be unnecessary?

runners/launch_docker-dev.sh
kmdanielduan: I tested and found the minimal version for me is `docker run -it --rm --init ${DOCKER_IMAGE} /bin/bash -c "${CMD} && bash"`. Deleting the first /bin/bash -c seems to results in XdummyDisplay errors. Deleting the second one works but will make the container stay in detached mode. I didn't find the case where I have to exit twice to get out of the container. Would be nice if you wanna test it.

ejnnr: I meant `/bin/bash -c "${CMD} && exec bash"` but if you don't have to exit twice, then nvm, I apparently just don't understand Docker and shells as well as I'd thought :)

runners/build_push_image.sh
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+for key in $KEYS; do
+    echo "----- Building and pushing humancompatibleai/imitation:${key} ..."
+    BUILD="docker build --target ${key} -t humancompatibleai/imitation:${key} ."
+    PUSH="docker push humancompatibleai/imitation:${key}"
+    echo "${BUILD}"
+    ${BUILD}
+    echo "${PUSH}"
+    # ${PUSH}  # Uncomment this line to push the image to Docker Hub
```
AdamGleave, line 22: Can we make this a flag? Having to edit source files feels awkward -- it's always annoying having dirty changes in your working directory as you have to make sure not to commit them accidentally.

runners/launch_docker-dev.sh
AdamGleave: @ejnnr I think you're right there's two shells, but when the innermost shell (launched by `&& bash`) exits, so will the outermost shell since it was run with `-c` which means "run this command then exit". `exec bash` has the effect of *replacing* the outermost shell with the innermost shell, so will save memory and I think is slightly better, although it's a negligible difference in practice.

runners/launch_docker-dev.sh
AdamGleave: Yeah I think it's not strictly necessary because the PATH is already set. activate does do some other things (like changing the prompt) but by default nothing else is important I think.

runners/launch_docker-dev.sh
kmdanielduan: Thanks for pointing that out. I missed out `exec bash` you mentioned. I also tested it and it works. I'll make the change accordingly.

kmdanielduan: It seems you're right. I missed out `exec bash`. It works, and I'll make changes accordingly

runners/build_push_image.sh
kmdanielduan: Sure, the latest update added usage and flag catcher for both files.

kmdanielduan:
Thanks for those comments! I made some changes to the files accordingly. Hope it's useful for people.

AdamGleave (approved):
LGTM

PR #352, opened 2021-09-15
Check docstrings with Darglint and fix high-priority errors

AdamGleave:
Adds https://github.com/terrencepreilly/darglint as a linter that runs on CI and in `ci/code_checks.sh` pre-commit hook. Checks for things like missing/superflous/misnamed arguments in docstrings relative to function signature. Also adds `pydocstyle`, which does some simpler checks (e.g. is a docstring missing? is it misformatted?)

Fixes high-priority errors that seemed to indicate incorrect docstrings.

Ignored most of the ones for missing docstrings. Ideally we'd add more docstrings later, but this PR is already big enough.

darglint ends up being a bit slow: around 4m to run on whole codebase (parallel across 8 cores). This is tolerable for CI purposes, but is way too slow for use a pre-commit hook as `ci/code_checks.sh` is intended. To fix this, I had to do a rather hacky thing of:
  1. Disable darglint by default in `setup.cfg` (it runs via a flake8 integration, so we can't just not run it).
  2. Enable it in CircleCI and `ci/code_checks.sh` at the relevant time by having a `.darglint` in CWD that re-enables it.
 I also do the hacky thing of running `flake8` twice in `ci/code_checks.sh`, since if we run `darglint` directly we lose some nice `flake8` features like being able to ignore darglint errors on a per-block basis (darglint requires the noqa comments are in docstring themself as discussed in https://github.com/terrencepreilly/darglint which pollutes the Sphinx HTML build).


src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -52,7 +52,7 @@ def train_preference_comparisons(
     agent_kwargs: Mapping[str, Any],
     gatherer_kwargs: Mapping[str, Any],
     allow_variable_horizon: bool,
-) -> Mapping[str, Any]:
+) -> Mapping[str, Mapping[str, float]]:
```
ejnnr, line 55: I don't think that's correct, the `reward_loss` and `reward_accuracy` keys currently live directly in the top-level dict (we could change that though)

ejnnr (approved):
Thanks for this PR! I think linting doc strings is a good idea and the hack to run `flake8` without `darglint` seems like a pretty good solution given there's no built-in support for that (I tried running `flake8` without `darglint` myself a while ago and didn't find a good way). Only had a few small/optional comments.

src/imitation/__init__.py
```diff
@@ -1 +1,3 @@
+"""imitation: implementations of imitation and reward learning algorithms."""
```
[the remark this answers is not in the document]
AdamGleave: good idea, done!

src/imitation/scripts/train_preference_comparisons.py
AdamGleave: Good catch, I'll revert this.

PR #353, opened 2021-09-17
Move adversarial algorithms into their own package

AdamGleave:
Moves `imitation.rewards.discrim_net` and `imitation.algorithms.adversarial` into a new package `imitation.algorithms.adversarial` with modules `common`, `airl` and `gail`.

Rationale for this is `DiscrimNet` is not used anywhere outside of AIRL/GAIL -- so why should it be in `rewards`?

Could have kept it all in `imitation.algorithms.adversarial` module (not splitting up into separate submodules) but that'd have been huge, and having one file per algorithm feels like a good rule of thumb.

Also took opportunity to do some other minor clean up, e.g. deleting `ActObsMLP` which is not needed as can just use `BasicRewardNet` instead. Does make me wonder if `RewardNet` is a bit misnamed (really just anything that maps transitions to a real number, which may or may not be a reward) but probably not worth changing just for this use case (and GAIL discriminator is really close to being a reward).


ejnnr (approved):
LGTM, I didn't spot any issues and I agree that organizing the files this way makes sense.

I'm wondering whether we need the discriminator nets at all, in some places they currently look like unnecessarily complicated wrappers around `RewardNet` (for example duplicating the input preprocessing). But getting rid of them would necessarily mean sharing a bit less code between AIRL and GAIL, I think, so not sure if that's worth it.

We could also consider making a few important classes, such as `GAIL`, `AIRL` and other algorithms available directly in the `imitation` namespace (i.e. make `from imitation import AIRL` possible like in SB3). I don't have a strong opinion on this but seems worth thinking about (unless you already decided against this a while ago). Probably depends on the target audience for the library: if we assume that people who use it spend some time working with it and getting familiar with the structure, then top-level exports make less sense. OTOH, someone who just wants to briefly use `AIRL` for something doesn't care where it's located. A potential middle ground would be to export all algorithms from the `imitation.algorithms` namespace.

AdamGleave:
Thanks for the review!

I've also been wondering about `DiscrimNet`'s -- I want to keep this PR small, but I'll see how much I can pare them down in https://github.com/HumanCompatibleAI/imitation/pull/354 and if there's not much left might remove them entirely.

Good point about the top-level imports, let's move that discussion to Slack.

PR #354, opened 2021-09-17
Eliminate DiscrimNet and have constructors take instantiated reward/policy objects

AdamGleave:
This PR makes three major changes:
  1. Changes our algorithms to have the convention of taking an instantiated reward net / RL algorithm / policy, rather than taking in the class and keyword arguments and constructing them inside the algorithm.
  2. Eliminates `DiscrimNet`, moving the relevant logic directly to the `AIRL` and `GAIL` classes. This lets us delete a lot of code that was previously duplicated with `RewardNet`.
  3. Deletes the MountainCar reward analysis code. We wrote this originally when none of our algorithms were working on MountainCar, but I don't think it's useful enough to keep around now that our algorithms do work. If we need it we can always resurrect it later. I decided to do it now as some of my changes broke this code, and didn't really want to fix it if going to delete it anyway.

It would have been better to split these up into separate PRs, but they were touching similar parts of the code, so tricky to disentangle now.

Switching to the instantiated rather than class convention had been a long-standing FIXME/TODO in the codebase. The class convention doesn't have much upside that I can see, and makes some some (admittedly rare) use cases awkward, like warm-starting a reward or policy or training with two different algorithms.

I stuck to the class/kwargs construction for optimizers, however. My rationale for this is: 1) the optimizer "belongs" to the algorithm in a way that a reward/policy network does not (you might want to use the reward or policy after the algorithm is finished training, but you'd throw away the optimizer); 2) the optimizer has to be initialized with the parameter set of the reward/policy network, which might be error prone if done outside the algorithm. I'm not sure this is the right call, though.

After this and previous refactoring, the `DiscrimNet`s didn't seem to have much basis for independent existence, being only used inside the `GAIL` and `AIRL` classes. So I just rolled in the relevant logic directly into those classes, and where possible leaned on existing implementations in `RewardNet`.


src/imitation/scripts/train_adversarial.py
```diff
@@ -53,8 +55,9 @@ def train_adversarial(
     checkpoint_interval: int,
     gen_batch_size: int,
     init_rl_kwargs: Mapping,
+    reward_net_cls: Optional[Type[reward_nets.RewardNet]],
```
ejnnr, line 58: These new arguments are missing from the docstring

tests/algorithms/adversarial/test_algorithms.py
```diff
@@ -124,18 +77,19 @@ def trainer(
...
-        n_envs=1,
-        parallel=_parallel,
-        log_dir=tmpdir,
-    )
-
-    _algorithm_cls = _algorithm_kwargs.pop("algorithm_cls")
-    gen_algo = util.init_rl(venv, verbose=1, **_algorithm_kwargs)
+    algorithm_kwargs = dict(algorithm_kwargs)
+    _algorithm_cls = algorithm_kwargs.pop("algorithm_cls")
+    venv = util.make_vec_env(env_name, n_envs=num_envs, parallel=parallel)
```
ejnnr, line 82: Did you remove `log_dir=tmpdir` deliberately here?

ejnnr (approved):
Thanks for all the refactoring you're doing! I think all three changes in this PR are a good idea and I especially like that input preprocessing is now more concentrated in one place. I think we could streamline things even more by inlining `disc_rew_preprocess_inputs` (see my comment in the code).

I don't have a clear opinion on the optimizer thing, so I'd say we just leave it as it is now, should be easy to change if we notice any issues with that solution later.

One small comment (because Github doesn't allow inline comments on unchanged lines):
https://github.com/HumanCompatibleAI/imitation/blob/fe6e9ee511c341b374630012d5dbda77ca336334/src/imitation/rewards/reward_nets.py#L308-L311
Not sure anymore if this makes sense; if not, we should remove the TODO. From an end-user perspective, keeping the class might make sense because it makes it very easy to just modify the size of the networks, even from the command line. Not sure how we can have an easy to use script interface if we get rid of this class. We could also leave the TODO for now and decide this when refactoring the training scripts.

src/imitation/data/types.py
```diff
@@ -209,7 +209,7 @@ def __post_init__(self):
                 f"{len(self.obs)} != {len(self.acts)}",
             )
 
-        if self.infos is not None and len(self.infos) != len(self.obs):
```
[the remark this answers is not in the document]
AdamGleave: I'm not sure this is the right decision tbh. Best justification I have is that a `Trajectory` is a single trajectory, so either the whole thing should have infos or they should all be omitted. Whereas `Transitions` can contain samples from multiple trajectories, some of which might have had `infos` recorded and others not. Given that code would have to handle this hybrid case anyway (representing missing infos with an empty info dict `{}`), may as well just make it mandatory.

However, it is a bit inefficient having all these empty dictionaries lying around. I actually tried to add `None` support briefly to this, but it broke a lot of things (e.g. collation, slicing). Could be fixed with some work, but it's a non-trivial change.

src/imitation/rewards/reward_nets.py
```diff
@@ -69,10 +78,10 @@ def preprocess(
         return rewards_common.disc_rew_preprocess_inputs(
```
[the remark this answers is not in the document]
AdamGleave: Good point, I've inlined it.

tests/algorithms/adversarial/test_algorithms.py
AdamGleave: I didn't, but I think it's actually fine to remove -- for `make_vec_env` it defaults to `None`, which will not log Monitor stats to any directory, which seems fine for testing.

src/imitation/scripts/train_adversarial.py
AdamGleave: Oh, good catch, I was relying on darglint to fix this but it looks like it's not running on CI -- I'll fix that in a separate PR. Added these arguments in now.

PR #359, opened 2021-09-23
Fix bugs in adversarial trainer

ejnnr:
1. Adversarial training could fail if `allow_variable_horizon=False` even on fixed-horizon environments. The issue was that we use `reset_num_timesteps=False` when calling SB3's `learn()`, which has the side effect of not resetting the environment. So it could happen that we had partial trajectories, which would be collected by `pop_trajectories`, but when training later resumed, it resumed from the middle of an episode. This then leads to trajectories that are terminal but still shorter than usual because they are not "initial". As a patch, I just reset the environment before calling `learn()`. A cleaner solution might be to just keep the partial trajectories in the buffer when calling `pop_trajectories`. Preference comparisons already reset the environment (like adversarial training now does); if we change one, we should change the other as well.
2. `normalize_obs` was called even when `normalize=False`, in which case the environment wasn't a `VecNormalize`, so the `normalize_obs` just resulted in an error. I wrapped the call in a check; I think that's better than always using a `VecNormalize` wrapper since it also works for `Discrete` state spaces.

### DIFF
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -388,6 +388,17 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
             learn_kwargs = {}
 
         with self.logger.accumulate_means("gen"):
+            # Because we use reset_num_timesteps=False (to get logging right),
+            # SB3 doesn't automatically reset the environment on .learn().
+            # This causes the following issue: if the previous learning cycle leaves
+            # unfinished trajectories, then we will collect the remainder of those
+            # trajectories during this cycle. They will therefore be incomplete episodes
+            # (because they don't start in the initial state) and in particular will
+            # be shorter, which raises the variable horizon check error.
+            # TODO(ejnnr): the same issue occurs in preference comparisons and I'm
+            # not sure that resetting is the "correct" solution. Better may be to just
+            # not collect partial trajectories when calling pop_trajectories?
+            self.venv.reset()
             self.gen_algo.learn(
                 total_timesteps=total_timesteps,
                 reset_num_timesteps=False,
@@ -512,8 +523,9 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
             [np.zeros(n_expert, dtype=int), np.ones(n_gen, dtype=int)],
         )
         # Policy and reward network were trained on normalized observations.
-        obs = self.venv_norm_obs.normalize_obs(obs)
-        next_obs = self.venv_norm_obs.normalize_obs(next_obs)
+        if isinstance(self.venv_norm_obs, vec_env.VecNormalize):
+            obs = self.venv_norm_obs.normalize_obs(obs)
+            next_obs = self.venv_norm_obs.normalize_obs(next_obs)
 
         # Calculate generator-policy log probabilities.
         with th.no_grad():


ejnnr:
Found and fixed another bug related to 2. (the adversarial training script tried to pickle `trainer.venv_norm_obs` under the assumption that it was a `VecNormalize` instance, which is not the case if `normalize=False`).
Maybe this suggests that `venv_norm_obs` being either a `VecNormalize` instance or not isn't a good idea but I don't know the adversarial trainer code well enough to suggest an alternative, so not sure.

AdamGleave (commented):
Thanks for spotting & fixing both of these bugs!

I've had a go refactoring this to: 
  - Have train_adversarial and preference_comparisons both only use finished trajectories.
  - Make `self.venv_norm_obs` None when observation normalization isn't taking place.

I've mixed feelings about the partial vs finished trajectory thing. Adversarial imitation learning and preference comparison should both be able to train fine on partial trajectories, so in one sense using partial trajectories is nicer as then the batch size should always be the same (though the replay buffer makes this matter relatively little for `AdversarialTrainer`). But having to reset environments each time is inefficient, and reset + using partial trajectories biases the data towards earlier on in episodes.

If it's mostly just the episode variable-horizon check causing the problem, I wonder if we should just make buffering wrapper keep track of episode length, then we can just check that directly? Was there any other reason you've needed to call `reset()`?

AdamGleave:
On reflection, I think checking episode length in `BufferingWrapper` is much better than this, and simplifies the code overall. In fact I'm not sure we even need the `terminal` flag in `Trajectory` now, but it seems like a useful thing to keep around, if we want to e.g. compute return on trajectory fragments need to handle temrinal state vs non-terminal final state differently.

I've currently gone for `AdversarialTrainer` including partial trajectories and `PreferenceComparison` only using finished trajectories. I'm not sure that's right. There's an argument for consistency. But preference comparison does compare trajectory fragments, and in some configurations these can be entire episodes (fragment = episode length), so it feels right for full trajectories to be the atomic element. Whereas GAIL/AIRL operate on a level of transitions, so feeding it lots of small trajectory fragments is fine, and just ensures the rate at which new samples is added is consistent across rounds.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -163,22 +163,10 @@ def train(self, steps: int, **kwargs) -> None:
...
-        # SB3 doesn't automatically reset the environment on .learn().
-        # This causes the following issue: if the previous learning cycle leaves
-        # unfinished trajectories, then we will collect the remainder of those
-        # trajectories during this cycle. They will therefore be incomplete episodes
-        # (because they don't start in the initial state) and in particular will
-        # be shorter, which raises the variable horizon check error.
-        # I think avoiding a reset() here would be tricky (assuming we even want to)
-        # because it would be hard to set a hypothetical Trajectory.initial attribute
-        # correctly like we do for Trajectory.terminal.
-        self.venv.reset()
```
AdamGleave, line 176: I think this is not needed if we always collect full trajectories from buffering wrapper. Bit worried we perhaps tried this and there was some edge case where it failed? But I expect we were doing something subtly different before.

ejnnr:
I can read over your changes later, for now just answering some questions:

> Was there any other reason you've needed to call reset()?
No, this was only necessary because of the variable horizon check. The only other thing I can think of is that resetting might be slightly more efficient for preference comparisons because if we start in the middle of an episode, there's a risk that the remaining episode will be shorter than the fragment length and needs to be discarded. But that's only applicable if we pop partial trajectories, shouldn't be an issue now.

> In fact I'm not sure we even need the terminal flag in Trajectory now
We still need it to set `done` correctly when calling `flatten_trajectories`, which is important for potential shaping or when `use_dones=True` for reward models.

> I've currently gone for AdversarialTrainer including partial trajectories and PreferenceComparison only using finished trajectories.
At least for PreferenceComparisons, only using full trajectories seems best to me -- if I'm understanding correctly we're not discarding partial ones but rather just finish them on the next iteration, right?
If we did it the same way for AdversarialTrainer, I think we could get rid of all the episode length code you've added to `BufferingWrapper`, the only issue I see with that would be that the number of transitions would vary slightly between iterations. Not sure whether that's worth it.

AdamGleave:
> At least for PreferenceComparisons, only using full trajectories seems best to me -- if I'm understanding correctly we're not discarding partial ones but rather just finish them on the next iteration, right?

Yep, that's my understanding.

> If we did it the same way for AdversarialTrainer, I think we could get rid of all the episode length code you've added to BufferingWrapper, the only issue I see with that would be that the number of transitions would vary slightly between iterations. Not sure whether that's worth it.

Using only full trajectories actually added a lot of complexity to `AdversarialTrainer` unfortunately. I did it that way first (you can go back in commit history on this PR to see), and then this meant that early in training sometimes the replay buffer was empty when discriminator training was meant to happen, so that had to be special cased. Overall I think the episode length tracking code is similarly complex to that.

src/imitation/data/wrappers.py
```diff
@@ -87,16 +98,42 @@ def _finish_partial_trajectories(self) -> Sequence[types.TrajectoryWithRew]:
...
+            consisting of data collected after the last call to `pop_trajectories`.
+            They may miss initial states (if `pop_trajectories` previously returned
+            a fragment for that episode), and terminal states (if the episode has
+            yet to complete). `ep_lens` is the total length of completed episodes.
+        """
+        if self.n_transitions == 0:
+            return [], []
+        partial_trajs = self._finish_partial_trajectories()
+        self._trajectories.extend(partial_trajs)
+        return self.pop_finished_trajectories()
```
kmdanielduan, line 136: If I understand it correctly, it's possible for both `pop_trajectories()` and `pop_finished_trajectories()` to have incomplete trajectories. The difference being that 
- All trajs from `pop_finished_trajectories()` are terminated. There can be one without initial state.
- `pop_trajectories()`: There can be one without initial state or non-terminated yet.

kmdanielduan:
Just flagging that I also encountered this problem earlier this day. Thanks to both of you for working on the fix. The current episode length tracking solution seems reasonable to me.
I didn't catch why "the number of transitions would vary slightly between iterations". I thought `gen_sample` was generated from the same `batch_size`. Did you mean the number of total transitions in the `ReplayBuffer` will differ across iterations?

AdamGleave:
> Just flagging that I also encountered this problem earlier this day. Thanks to both of you for working on the fix. The current episode length tracking solution seems reasonable to me.
> I didn't catch why "the number of transitions would vary slightly between iterations". I thought `gen_sample` was generated from the same `batch_size`. Did you mean the number of total transitions in the `ReplayBuffer` will differ across iterations?

The number of transitions added to the `ReplayBuffer` will vary. If replay buffer capacity is much greater than batch size, this has little effect. But by default we set capacity = batch size, so this would mean that in some iterations we discard some of the data (overflows from buffer), and in other cases the buffer still contains data from previous iterations. Which I expect could change the training dynamics.

src/imitation/data/wrappers.py
ejnnr: Yes, that sounds right. But I think the idea is that if you _always_ use `pop_finished_trajectories()`, then you'll only get complete trajectories.

ejnnr (commented):
I've looked over the changes now, LGTM (I just updated one docstring). Can't approve the PR though since I opened it. If you're happy with what we have, then feel free to merge, @AdamGleave

PR #361, opened 2021-09-29
ResettableEnv: make observation space dict combining states and observations

AdamGleave:
This PR modifies `ResettableEnv` (and subclass `TabularModelEnv`) to return observations consisting of a dict of the underlying state, and the POMDP-style observation.

This allows MCE IRL to fit in better with the Gym/SB3 interface. Specifically, MCE IRL policy is state-based (planned), while the MCE IRL reward is observation-based. This decomposition lets both depend on the relevant component.


src/imitation/algorithms/mce_irl.py
```diff
@@ -205,27 +204,31 @@ def predict(
         Returns:
             Tuple of the actions and new hidden states.
         """
-        if state is None:
-            state = np.zeros(len(observation), dtype=np.int)
+        timesteps = state  # rename to avoid confusion
+        del state
+
+        # obs_state = state from the MDP
```
qxcv, line 210: Where is the name `obs_state` used in the code?

src/imitation/envs/resettable_env.py
```diff
@@ -58,16 +59,31 @@ def state_space(self) -> gym.Space:
         """
         return self._state_space
 
+    @property
+    def raw_observation_space(self) -> gym.Space:
```
qxcv, line 63: "Raw" feels a bit nondescript to me—I can think of reasons why you could call either the state space (ints) or the observation space (vectors) the "raw" space. I'm struggling to think up a better name, though. Some ideas:

- `approx_observation_space`, `func_observation_space`, `func_approx_observation_space` or `fapprox_observation_space`
- `vector_observation_space`
- `semantic_observation_space`
- `qualitative_observation_space`

Personally `func_approx` is my favourite because it makes it totally explicit what purpose these observations are for. Curious to hear your thoughts, though—do you think this is actually an improvement? Do you have other ideas?

qxcv (approved):
Two comments about naming, but otherwise looks okay. Thank you for thinking up and implementing a solution to this—I like the idea you converged on.

src/imitation/envs/resettable_env.py
AdamGleave: Agree raw is rather vague, would like something better.

I was envisaging `ResettableEnv` as being a generic interface for environments with a resettable simulator. So I've been trying to avoid it being too tightly coupled to MCE IRL. This might be premature abstractification, but I have found other uses for resettable environment interfaces in the past (albeit not in `imitation`), and been frustrated there isn't a standard one.

`func_approx` and the ilk seem pretty coupled to how MCE IRL is using them, as opposed to it being a more standard POMDP.

Perhaps `underlying_observation_space` or `inner_observation_space`?

PR #371, opened 2021-10-22
Small improvements to preference comparisons

ejnnr:
These are some changes I made for the reward preprocessing project but hadn't merged so far:
- One small bug fix
- Allow sampling some comparisons at the beginning for pertaining the reward model before training the agent
- Slightly different default configs
- Allow using random actions for some fraction of the sampled trajectories

The last point adds a bit of complexity to the code, so we could also remove that until we actually need it (I used it during debugging and it seems plausible that it could sometimes be helpful, but I don't actually have an example environment where it's needed).

### DIFF
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -394,12 +394,22 @@ class RandomFragmenter(Fragmenter):
 class PreferenceGatherer(abc.ABC):
     """Base class for gathering preference comparisons between trajectory fragments."""
 
-    def __init__(self, custom_logger: Optional[imit_logger.HierarchicalLogger] = None):
+    def __init__(
+        self,
+        seed: Optional[int] = None,
+        custom_logger: Optional[imit_logger.HierarchicalLogger] = None,
+    ):
         """Initializes the preference gatherer.
 
         Args:
+            seed: seed for the internal RNG, if applicable
             custom_logger: Where to log to; if None (default), creates a new logger.
         """
+        # The random seed isn't used here, but it's useful to have this
+        # as an argument nevertheless because that means we can always
+        # pass in a seed in training scripts (without worrying about whether
+        # the PreferenceGatherer we use needs one).
+        del seed
         self.logger = custom_logger or imit_logger.configure()
 
     @abc.abstractmethod
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -728,6 +728,7 @@ class PreferenceComparisons(base.BaseImitationAlgorithm):
         comparisons_per_iteration: int = 50,
         fragment_length: int = 50,
         transition_oversampling: float = 10,
+        initial_comparison_fraction: float = 0.1,
         custom_logger: Optional[imit_logger.HierarchicalLogger] = None,
         allow_variable_horizon: bool = False,
         seed: Optional[int] = None,
@@ -763,6 +764,10 @@ class PreferenceComparisons(base.BaseImitationAlgorithm):
                 creating fragments. Since fragments are sampled with replacement,
                 this is usually chosen > 1 to avoid having the same transition
                 in too many fragments.
+            initial_comparison_fraction: fraction of the total_comparisons argument
+                to train() that will be sampled before the rest of training begins
+                (using the randomly initialized agent). This can be used to pretrain
+                the reward model before the agent is trained on the learned reward.
             custom_logger: Where to log to; if None (default), creates a new logger.
             allow_variable_horizon: If False (default), algorithm will raise an
                 exception if it detects trajectories of different length during
@@ -812,6 +817,7 @@ class PreferenceComparisons(base.BaseImitationAlgorithm):
         self.comparisons_per_iteration = comparisons_per_iteration
         self.fragment_length = fragment_length
         self.transition_oversampling = transition_oversampling
+        self.initial_comparison_fraction = initial_comparison_fraction
 
         self.dataset = PreferenceDataset()
 
@@ -829,6 +835,8 @@ class PreferenceComparisons(base.BaseImitationAlgorithm):
         Raises:
             ValueError: `total_comparisons < self.comparisons_per_iteration`.
         """
+        initial_comparisons = int(total_comparisons * self.initial_comparison_fraction)
+        total_comparisons -= initial_comparisons
         iterations, extra_comparisons = divmod(
             total_comparisons,
             self.comparisons_per_iteration,
@@ -852,7 +860,7 @@ class PreferenceComparisons(base.BaseImitationAlgorithm):
             # the desired total number of comparisons, we collect the remainder
             # right at the beginning to pretrain the reward model slightly
             if i == 0:
-                num_pairs += extra_comparisons
+                num_pairs += extra_comparisons + initial_comparisons
             num_steps = math.ceil(
                 self.transition_oversampling * 2 * num_pairs * self.fragment_length,
             )
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -594,11 +594,13 @@ class RewardTrainer(abc.ABC):
         self.logger = custom_logger or imit_logger.configure()
 
     @abc.abstractmethod
-    def train(self, dataset: PreferenceDataset):
+    def train(self, dataset: PreferenceDataset, epoch_multiplier: float = 1.0):
         """Train the reward model on a batch of fragment pairs and preferences.
 
         Args:
             dataset: the dataset of preference comparisons to train on.
+            epoch_multiplier: how much longer to train for than usual
+                (measured relatively).
         """
 
 
@@ -624,7 +626,9 @@ class CrossEntropyRewardTrainer(RewardTrainer):
                 is uniformly random (used for the model of preference generation
                 that is used for the loss)
             batch_size: number of fragment pairs per batch
-            epochs: number of epochs on each training iteration
+            epochs: number of epochs on each training iteration (can be adjusted
+                on the fly by specifying an `epoch_multiplier` in `self.train()`
+                if longer training is desired in specific cases).
             lr: the learning rate
             discount_factor: the model of preference generation uses a softmax
                 of returns as the probability that a fragment is preferred.
@@ -721,8 +725,8 @@ class CrossEntropyRewardTrainer(RewardTrainer):
         model_probability = 1 / (1 + returns_diff.exp())
         return self.noise_prob * 0.5 + (1 - self.noise_prob) * model_probability
 
-    def train(self, dataset: PreferenceDataset):
-        """Trains for `self.epochs` epochs over `dataset`."""
+    def train(self, dataset: PreferenceDataset, epoch_multiplier: float = 1.0):
+        """Trains for `epoch_multiplier * self.epochs` epochs over `dataset`."""
         # TODO(ejnnr): This isn't specific to the loss function or probability model.
         # In general, it might be best to split the probability model, the loss and
         # the optimization procedure a bit more cleanly so that different versions
@@ -734,7 +738,8 @@ class CrossEntropyRewardTrainer(RewardTrainer):
             shuffle=True,
             collate_fn=preference_collate_fn,
         )
-        for _ in range(self.epochs):
+        epochs = round(self.epochs * epoch_multiplier)
+        for _ in range(epochs):
             for fragment_pairs, preferences in dataloader:
                 self.optim.zero_grad()
                 loss = self._loss(fragment_pairs, preferences)
@@ -909,9 +914,16 @@ class PreferenceComparisons(base.BaseImitationAlgorithm):
             ##########################
             # Train the reward model #
             ##########################
+            # Usually, num_pairs is the same as comparisons_per_iteration,
+            # but on the first iteration, we might have additional pairs
+            # (i.e. num_pairs > comparisons_per_iteration). In that case,
+            # we want to train the reward model for correspondingly more
+            # epochs.
+            epoch_multiplier = num_pairs / self.comparisons_per_iteration
+
             with self.logger.accumulate_means("reward"):
                 self.logger.log("Training reward model")
-                self.reward_trainer.train(self.dataset)
+                self.reward_trainer.train(self.dataset, epoch_multiplier)
             reward_loss = self.logger.name_to_value["mean/reward/loss"]
             reward_accuracy = self.logger.name_to_value["mean/reward/accuracy"]
 
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -114,6 +114,7 @@ class AgentTrainer(TrajectoryGenerator):
         exploration_frac: float = 0.0,
         stay_prob: float = 0.5,
         random_prob: float = 0.5,
+        seed: Optional[int] = None,
         custom_logger: Optional[imit_logger.HierarchicalLogger] = None,
     ):
         """Initialize the agent trainer.
@@ -129,6 +130,7 @@ class AgentTrainer(TrajectoryGenerator):
                 step for the exploratory samples.
             random_prob: the probability of picking the random policy when switching
                 during exploration.
+            seed: random seed for exploratory trajectories.
             custom_logger: Where to log to; if None (default), creates a new logger.
 
         Raises:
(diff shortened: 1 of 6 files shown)


src/imitation/scripts/config/train_preference_comparisons.py
```diff
@@ -19,11 +19,15 @@
...
     fragment_length = 100  # timesteps per fragment used for comparisons
     total_timesteps = int(1e6)  # total number of environment timesteps
-    total_comparisons = 1000  # total number of comparisons to elicit
+    total_comparisons = 5000  # total number of comparisons to elicit
     # comparisons to gather before switching back to agent training
-    comparisons_per_iteration = 50
+    comparisons_per_iteration = 100
     # factor by which to oversample transitions before creating fragments
-    transition_oversampling = 10
+    transition_oversampling = 1
```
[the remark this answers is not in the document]
AdamGleave: OK with removing the warning

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         self,
         algorithm: base_class.BaseAlgorithm,
         reward_fn: Union[rewards_common.RewardFn, reward_nets.RewardNet],
+        random_frac: float = 0.0,
```
AdamGleave, line 113: This seems like a useful feature, but does it belong in the algorithm? One alternative way of implementing this is to have a wrapper for `BaseAlgorithm` that randomizes between a random agent and the agent being trained. WDYT?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -760,6 +792,10 @@ def __init__(
                 creating fragments. Since fragments are sampled with replacement,
                 this is usually chosen > 1 to avoid having the same transition
                 in too many fragments.
+            initial_comparison_frac: fraction of the total_comparisons argument
+                to train() that will be sampled before the rest of training begins
+                (using the randomly initialized agent). This can be used to pretrain
+                the reward model before the agent is trained on the learned reward.
```
AdamGleave, line 798:
```suggestion
                the reward model before the agent is trained on the learned reward,
                to help avoid irreversibly learning a bad policy from an untrained reward.
```
Paraphrasing the justification given in appendix A of https://arxiv.org/pdf/1706.03741.pdf

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -849,7 +888,7 @@ def train(self, total_timesteps: int, total_comparisons: int) -> Mapping[str, An
             # the desired total number of comparisons, we collect the remainder
             # right at the beginning to pretrain the reward model slightly
             if i == 0:
-                num_pairs += extra_comparisons
+                num_pairs += extra_comparisons + initial_comparisons
```
AdamGleave, line 891: Is this enough? Gathers a bunch of data, but only does a single reward model update AFAICT. Might need an `initial_updates` or make that proportional to the standard number of `self.comparisons_per_iteration`.

AdamGleave (commented):
Thanks for the PR and sorry for the slow review! Looks like some solid improvements. I have some questions about whether this is best way of implementing `random_frac` (although seems like a useful feature), and if `initial_frac` is doing enough.

Happy to discuss on this PR or in our next meeting.

src/imitation/algorithms/preference_comparisons.py
ejnnr: Yeah, agreed. I've just made the number of reward training epochs proportional to the number of new comparisons now, wanted to avoid yet another hyperparameter. If we decide we need something like `initial_updates` after all, then maybe we should just implement a schedule that regulates the amount of reward vs policy training (IIRC, the DRLHP paper decays the amount of reward training over time).

My implementation with `epoch_multiplier` seems a bit complicated, maybe we should just pass `epochs` to `reward_trainer.train()`, rather than making it a constructor argument. OTOH, the entire point of having abstract base classes was that the main preference comparison trainer shouldn't make many assumptions about the individual components, and an epoch multiplier is a weaker assumption than an absolute number of epochs.

We could consider just getting rid of the ABC for the reward trainer, or at least making a few more assumptions (such as having epochs over a fixed training set). I think that would simplify the code, but if we'll predictably want to have very different types of reward trainers at some point, I would leave things as they are.

src/imitation/scripts/config/train_preference_comparisons.py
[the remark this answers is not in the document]
ejnnr: I noticed I'd already built in the ability to suppress the warning with a parameter to `RandomFragmenter`, so I went with that instead of removing the code entirely. Assuming we implement human feedback at some point, we can then have it turned on by default when using that

src/imitation/algorithms/preference_comparisons.py
ejnnr: That sounds like a good idea because it would make it easy to switch policies within an episode like we discussed. I think we could have two parameters: probability of sticking with the current policy vs randomly picking a new one at each step, and probability of picking each policy (expert or random). That covers purely random trajectories as a special case, so we can replace the current implementation with that.

Unfortunately, I don't think this is really going to simplify the code in `AgentTrainer`. In general, we still need to collect additional expert trajectories, and then collect trajectories with some randomness. Of course we could just take the expert trajectories we already collected during training and fill those up with partially random trajectories, but then the user can't put an upper bound on the proportion of partially random trajectories (we might have collected very few trajectories during training). If you see a way around collecting additional trajectories twice, LMK

AdamGleave: >  I think we could have two parameters: probability of sticking with the current policy vs randomly picking a new one at each step, and probability of picking each policy (expert or random). That covers purely random trajectories as a special case, so we can replace the current implementation with that.

Seems easy to implement and strictly more general, so I'm in favour.

> Unfortunately, I don't think this is really going to simplify the code in AgentTrainer

You're right. I was thinking for some reason that we were doing the sampling step twice: once from the RL policy, once random. But of course the RL policy is "sampled" as a side-effect of the RL training process (so is effectively free). I think will need to keep current implementation then. I might rename it from `random_frac` to something like `extra_frac` or `exploration_frac` if the trajectories collected by this are not totally random.

src/imitation/algorithms/preference_comparisons.py
AdamGleave: I think I misunderstood what the code did previously: I'd assumed we just treated the PreferenceDataset as one big batch and did a single update, but I see now that we do batch it (which is reasonable). I'd have actually thought a priori that a fixed number of epochs for the initial comparison should be fine. Maybe it should be a different number to for non-initial comparisons (on the basis that it's a different size dataset), though.

It's an interesting question how the # of epochs should scale with dataset size. Having it be proportional to size of dataset seems not quite right: it'd mean the number of training updates is quadratic in the number of initial comparisons, as we train for more batches and more epochs the more data we have. I actually usually see the opposite -- when training on very big datasets like with unsupervised natural language models, people will often train for 1 epoch, or less (i.e. only train on random subset of data).

Was adding `epoch_multilpier` necessary to get good performance? If not, inclined to scrap it (sorry for the miscommunication). If so, then I think we should set it to some fixed value for initial vs remaining training (a new hyperparameter, which is annoying, but is OK if it has a sensible default), or implement a schedule as you suggest.

ejnnr: Since it's been a while, I'm unfortunately not sure whether `epoch_multiplier` was necessary. For now, I've reverted to the original version (i.e. same number of epochs no matter the dataset size). However, I think it makes sense to implement a schedule since this is what the DRLHP paper uses (and we should probably see if replicating that helps with our issues). So I've left the `epoch_multiplier` argument in place in the reward trainer, I'm just no longer using it for now. If you'd rather I remove it for now and re-introduce it when implementing a schedule, I can do that too

AdamGleave: How are you planning to use `epoch_multiplier` in the future? I see the DRLHP paper uses a schedule for the rate of labeling described in appendix A:

> In the Atari domain we also pretrain the reward predictor for 200 epochs before beginning RL training, to reduce the likelihood of irreversibly learning a bad policy based on an untrained predictor. For the rest of training, labels are fed in at a rate decaying inversely with the number of timesteps; after twice as many timesteps have elapsed, we answer about
half as many queries per unit time.

But this seems to be about the labels added to the dataset, not the number of training epochs, although I could imagine them having a similar effect.

They also use a replay buffer which I think we don't use?

> We maintain a buffer of only the last 3,000 labels and loop over this buffer continuously; this is to ensure that the predictor gives enough weight to new labels (which can represent a shift in distribution) when the total number of labels becomes large

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -394,12 +394,22 @@ def __call__(
...
-    def __init__(self, custom_logger: Optional[imit_logger.HierarchicalLogger] = None):
+    def __init__(
+        self,
+        seed: Optional[int] = None,
+        custom_logger: Optional[imit_logger.HierarchicalLogger] = None,
+    ):
         """Initializes the preference gatherer.
 
         Args:
+            seed: seed for the internal RNG, if applicable
```
AdamGleave, line 405: nice addition, always good to make things easy to replicate

AdamGleave (approved):
LGTM. I don't understand why we'll need `epoch_multiplier` but if you think it'll be useful after reading my comment then feel free to keep it -- easy enough to remove later anyway.

AdamGleave:
Might be worth adding some code coverage in tests for `random_steps > 0` branch which is currently being missed: https://app.codecov.io/gh/HumanCompatibleAI/imitation/compare/371/tree/src/imitation/algorithms/preference_comparisons.py

The other reductions in coverage seem innocuous from a quick skim, so if adding that test doesn't resolve the coverage requirement I'm happy to override that check.

src/imitation/algorithms/preference_comparisons.py
ejnnr: You're right, I misremembered what they use a schedule for. I don't have a good intuition for which aspects will be important to make preference comparisons works, so it might be best to stay very close to the DRLHP paper for now. That means I've added a new parameter that specifies the `epoch_multiplier` for pretaining (i.e. we can use that to train for 200 epochs in the first round but then train less during all remaining iterations). Can easily generalize this to a schedule if we ever need to.

I'll limit the size of our dataset and introduce the preference collection schedule in another PR.

src/imitation/algorithms/preference_comparisons.py
ejnnr: Implemented the more general version now and renamed to `exploration_frac`

ejnnr:
I think I've finished everything now, but requested re-review since there are quite a few changes (mainly the new `ExplorationWrapper` to implement the more general partially random policies we decided on).

Re `epoch_multiplier`: you're right, the DRLHP paper doesn't have a scheduler for epochs. I still left it in because I'm now using it to implement a longer first round of training (e.g. their 200 epochs). No plans to implement a scheduler for epochs for now, though it would be easy if we ever want one

src/imitation/policies/exploration_wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import Optional
+
+import numpy as np
+from stable_baselines3.common import vec_env
+
+from imitation.data import rollout
+
+
+class ExplorationWrapper:
+    """Wraps a PolicyCallable to create a partially randomized version.
```
AdamGleave, line 12: I think we could make this a more general `MixtureWrapper` without much implementation complexity: take a sequence of `policies` and `probs` (defining a categorical probability distribution), and `stay_prob`.

`ExplorationWrapper` is then just special-case where `policies = [policy, self._random_policy]` and `probs = [random_prob, 1 - random_prob]`.

I don't have a concrete use case in mind for this, so think it's fine to leave it as-is, but feels like a feature that's nice to have around. E.g. what if we want a random policy with some different distribution over actions?

src/imitation/policies/exploration_wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+
+    This wrapper randomly switches between two policies: the wrapped policy,
+    and a uniformly random one. After each action, the current policy is kept
```
AdamGleave, line 15: Not necessarily uniform, whatever the action spaces `sample()` method is (e.g. for continuous unbounded distributions it's Gaussian).

src/imitation/policies/exploration_wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+        return np.stack(acts, axis=0)
+
+    def _switch(self):
+        """Pick a new policy at random."""
+        if self.rng.rand() < self.random_prob:
+            self.current_policy = self._random_policy
+        else:
+            self.current_policy = self.wrapped_policy
+
+    def __call__(self, obs: np.ndarray):
```
AdamGleave, line 60:
```suggestion
    def __call__(self, obs: np.ndarray) -> np.ndarray:
```

src/imitation/policies/exploration_wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+
+        self.current_policy = policy
+        # Choose the initial policy at random
+        self._switch()
+
+    def _random_policy(self, states):
+        acts = [self.venv.action_space.sample() for _ in range(len(states))]
+        return np.stack(acts, axis=0)
+
+    def _switch(self):
```
AdamGleave, line 53:
```suggestion
    def _switch(self) -> None:
```

src/imitation/policies/exploration_wrapper.py
```diff
@@ -0,0 +1,64 @@
...
+        self.venv = venv
+
+        self.rng = np.random.RandomState(seed)
+        self.venv.action_space.seed(seed)
+
+        self.current_policy = policy
+        # Choose the initial policy at random
+        self._switch()
+
+    def _random_policy(self, states):
```
AdamGleave, line 49:
```suggestion
    def _random_policy(self, states: np.ndarray) -> np.ndarray:
```
Is it states or obs?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -245,3 +245,21 @@ def test_store_and_load_preference_dataset(agent_trainer, fragmenter, tmp_path):
 
         assert preference == loaded_preference
         _check_trajs_equal(fragments, loaded_fragments)
+
```
AdamGleave, line 248: Weak preference but I'd consider adding one or two more tests of `ExplorationWrapper`, mainly because it actually seems amenable to unit testing. Ideas:
  1. If `random_prob == 0`, is `self.current_policy` always `wrapped_policy`, and if `== 1`, always `random_policy`? If `0.5`, is it (roughly) equal after ~100 calls to switch? Can make this deterministic by setting seed.
  2. Check `__call__` always returns things inside `venv.action_space`

AdamGleave (commented):
Thanks for the changes! This looks basically ready. I have some suggestions for changes, but I don't think you need to incorporate all of them (e.g. generalizing `ExplorationWrapper`) if you want to just move on.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,11 +932,16 @@ def train(
             # Gather new preferences #
             ##########################
             num_pairs = self.comparisons_per_iteration
-            # if the number of comparisons per iterations doesn't exactly divide
+            epoch_multiplier = 1.0
+            # If the number of comparisons per iterations doesn't exactly divide
             # the desired total number of comparisons, we collect the remainder
-            # right at the beginning to pretrain the reward model slightly
+            # right at the beginning to pretrain the reward model slightly.
```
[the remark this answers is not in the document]
ejnnr: Good point, I've added a comment and a note in the doctoring for `comparisons_per_iteration`. With the hyperparameters I've been using so far, this shouldn't be a huge problem (e.g. for 5000 total comparisons, 100 comparisons per iteration, this is at most 2% of the total comparisons, whereas `initial_comparison_frac` is typically more like 10% or 20%). But could be an issue if `initial_comparison_frac` is 0.

If I implement a scheduler, I'll have to rewrite this logic anyway, I'll see then whether it's easy to do it correctly right away.

src/imitation/policies/exploration_wrapper.py
ejnnr: I had the same thought but I think this should be easy to change later if we ever need it, so I'll stay with the simple  special case for now

src/imitation/policies/exploration_wrapper.py
ejnnr: I went with `obs` now, I think `imitation` uses that (almost?) everywhere else

tests/algorithms/test_preference_comparisons.py
ejnnr: Done

PR #373, opened 2021-11-01
Make type of PreferenceGatherer configurable

ejnnr:
This makes it easy to re-use the `train_preference_comparisons` script with only a very light wrapper while using externally defined `PreferenceGatherer`s (needed for the reward preprocessing project). If we ever implement human feedback, we'd need to make this configurable anyway.

We could similarly make the `RewardTrainer` and `Fragmenter` configurable, happy to do that if you think it makes sense. Held off for now because I don't have specific plans to use custom implementations for those and wasn't sure if we wanted to add the (slight) complexity without a concrete need.

### DIFF
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -394,12 +394,21 @@ class RandomFragmenter(Fragmenter):
 class PreferenceGatherer(abc.ABC):
     """Base class for gathering preference comparisons between trajectory fragments."""
 
-    def __init__(self, custom_logger: Optional[imit_logger.HierarchicalLogger] = None):
+    def __init__(
+        self,
+        seed: Optional[int] = None,
+        custom_logger: Optional[imit_logger.HierarchicalLogger] = None,
+    ):
         """Initializes the preference gatherer.
 
         Args:
+            seed: seed for the internal RNG, if applicable
             custom_logger: Where to log to; if None (default), creates a new logger.
         """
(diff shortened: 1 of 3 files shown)


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -394,12 +394,21 @@ def __call__(
...
         """Initializes the preference gatherer.
 
         Args:
+            seed: seed for the internal RNG, if applicable
             custom_logger: Where to log to; if None (default), creates a new logger.
         """
+        # The random seed isn't used here, but it's useful to have this
+        # as an argument nevertheless because that means we can always
+        # pass in a seed in training scripts (without worrying about whether
+        # the PreferenceGatherer we use needs one).
```
AdamGleave, line 411: I suggest adding `del seed` at the start of this method -- good way of documenting "yes, I really do mean this argument to be unused" in a way that will also then be statically checked that it is in fact unused. (You can keep the comment too.)

AdamGleave (approved):
LGTM; one minor comment. Don't think we need to make this change for the other classes -- it's an easy change to make, so let's do that as/when use cases arise.

AdamGleave:
@ejnnr I think this is ready to merge (apart from the one minor comment), just wanted to check if you're waiting on anything from me?

ejnnr:
No, all good, I think I just wanted to merge #371 first

ejnnr:
Tests failed because of mujoco dependency issues, but they were fine before then, so I merged this to have everything in one PR (#371)

PR #375, opened 2021-11-09
Saving checkpoints periodically for preference comparisons

kmdanielduan:
This PR aims to
- Allow additional self-defined `callback` to happen in `main_trainer.train()`.
- Allow periodical saving of reward model and policy weights at the end of `main_trainer.train()`.

### DIFF
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -139,12 +139,12 @@ class AgentTrainer(TrajectoryGenerator):
         # them after training. This should come first (before the wrapper that
         # changes the reward function), so that we return the original environment
         # rewards.
-        self.buffering_wrapper = wrappers.BufferingWrapper(venv)
+        self.buffering_wrapped_venv = wrappers.BufferingWrapper(venv)
         self.venv = reward_wrapper.RewardVecEnvWrapper(
-            self.buffering_wrapper,
-            reward_fn,
+            self.buffering_wrapped_venv,
+            self.reward_fn,
         )
-        self.algorithm.set_env(self.venv)
+        self.log_callback = self.venv.make_log_callback()
 
     def train(self, steps: int, **kwargs) -> None:
         """Train the agent using the reward function specified during instantiation.
@@ -154,10 +154,10 @@ class AgentTrainer(TrajectoryGenerator):
             **kwargs: other keyword arguments to pass to BaseAlgorithm.train()
 
         Raises:
-            RuntimeError: Transitions left in `self.buffering_wrapper`; call
+            RuntimeError: Transitions left in `self.buffering_wrapped_venv`; call
                 `self.sample` first to clear them.
         """
-        n_transitions = self.buffering_wrapper.n_transitions
+        n_transitions = self.buffering_wrapped_venv.n_transitions
         if n_transitions:
             raise RuntimeError(
                 f"There are {n_transitions} transitions left in the buffer. "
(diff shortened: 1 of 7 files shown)


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -812,7 +816,14 @@ def __init__(
 
         self.dataset = PreferenceDataset()
 
-    def train(self, total_timesteps: int, total_comparisons: int) -> Mapping[str, Any]:
+        self.vec_normalize = vec_normalize
```
AdamGleave, line 819: This attribute doesn't seem to actually be used anywhere?

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -24,6 +24,20 @@
...
 
+def save(trainer, save_path):
+    """Save reward model and policy."""
+    os.makedirs(save_path, exist_ok=True)
+    th.save(trainer.reward_trainer.model, os.path.join(save_path, "reward_net.pt"))
+    if hasattr(trainer.trajectory_generator, "algorithm"):
+        serialize.save_stable_model(
+            os.path.join(save_path, "policy"),
+            trainer.trajectory_generator.algorithm,
+            trainer.vec_normalize,
```
AdamGleave, line 35: Why not have `save` take `vec_normalize` directly? It's created inside the script, and not used by the trainer class at all.

src/imitation/algorithms/preference_comparisons.py
kmdanielduan: You are right about it!

src/imitation/scripts/train_preference_comparisons.py
kmdanielduan: Good point. I missed out that previously.

src/imitation/scripts/config/train_preference_comparisons.py
```diff
@@ -38,6 +38,8 @@ def train_defaults():
     normalize = True  # Use VecNormalize
     normalize_kwargs = {"norm_reward": False}  # kwargs for `VecNormalize`
 
+    checkpoint_interval = 0  # Num epochs between checkpoints (<0 disables)
```
AdamGleave, line 41: `=0` disables all but final one right (which I'm not sure I'd call a "checkpoint" anyway)?

```suggestion
    checkpoint_interval = 0  # Num epochs between checkpoints (<0 disables, =0 final only)
```

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -24,6 +24,38 @@
...
+    os.makedirs(save_path, exist_ok=True)
+    th.save(trainer.reward_trainer.model, os.path.join(save_path, "reward_net.pt"))
+    if allow_save_policy:
+        # Note: We should only save the model as model.pkl if `trajectory_generator`
+        # contains one. Specifically we check if the `trajectory_generator` contains an
+        # `algorithm` attribute.
+        assert hasattr(trainer.trajectory_generator, "algorithm")
+        save_model(trainer.trajectory_generator, vec_normalize, save_path)
+    else:
+        print("trainer.trajectory_generator doesn't contain a policy to save.")
```
AdamGleave, line 56: Avoid print statements; use `custom_logger.warn` instead (it'll then get logged to files not just stdout)

AdamGleave (approved):
LGTM apart from one minor comment.

src/imitation/scripts/train_preference_comparisons.py
kmdanielduan: Good point! I would log the warning messages to trainer.logger.

PR #379, opened 2021-12-06
Make `n_rounds` integer

ebuehrle:
Fixes an issue with the default config of `imitation.scripts.train_adversarial`, which passes 1e6 for `total_timesteps`, causing `n_rounds` to become a float.


AdamGleave:
`total_timesteps` should be int, it's a bug that it was a float which seems to be fixed in `master`, so I'm closing this for now. But feel free to re-open a PR fixing any cases where `total_timesteps` are float and/or validating this in scripts if this still is a problem for you.

PR #382, opened 2022-01-13
Improvements-weight_decay_reward_logging

kmdanielduan:
This PR aims to 
- add `weight_decay` to reward training for preference comparisons.
- add logging for learnt reward.


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -676,6 +683,8 @@ def __init__(
                 in returns that are above this threshold. This threshold
                 is therefore in logspace. The default value of 50 means
                 that probabilities below 2e-22 are rounded up to 2e-22.
+            weight_decay: the weight decay factor for the reward model's weights,
+                equivalent to L2 regularization.
```
ejnnr, line 687: To be pedantic, weight decay and L2 regularization are _not_ equivalent for Adam (see https://arxiv.org/abs/1711.05101). `torch.optim.Adam` uses L2, whereas `torch.optim.AdamW` uses "real" weight decay. I don't think we need to discuss this in detail, but maybe write something like "weight decay factor (implemented as L2 regularization rather than AdamW, see https://arxiv.org/abs/1711.05101)" to avoid confusion?

src/imitation/scripts/config/train_preference_comparisons.py
```diff
@@ -32,6 +32,7 @@ def train_defaults():
 
     reward_trainer_kwargs = {
         "epochs": 3,
+        "weight_decay": 0.0,
```
ejnnr, line 35: Out of curiosity: Have you already done experiments with non-zero weight decay, and has that helped? (Though I agree with setting this to zero by default unless we're pretty sure that some non-zero value consistently works better)

ejnnr (approved):
Thanks for implementing this, I think both additions are very useful! LGTM except for one very minor comment on weight decay.

(I had already implemented something similar for my own experiments a while ago, sorry for not telling you about that! To avoid further duplication of efforts: I'm planning to implement a FIFO buffer for comparison labels as well as a scheduler for collecting reward labels pretty soon)

src/imitation/scripts/config/train_preference_comparisons.py
AdamGleave: Can omit this line, as it's already got a default value in `__init__`.

AdamGleave (approved):

ejnnr:
I just noticed that the original environment rewards (`agent/rollout/ep_rew_mean`) are missing from the terminal output in this PR (replaced by the reward model outputs, `agent/rollout/ep_rew_wrapped_mean`). This is especially confusing because the name is cut off (at least for me), I only see `agent/rollout/ep_...` in the terminal. Both are logged as expected in tensorboard. Does anyone know what's going on here? I would have expected both rewards to also be shown in the terminal

src/imitation/algorithms/preference_comparisons.py
AdamGleave: Should we just move to `AdamW` by default? From the linked paper it seems preferable, as it decouples the choice of this hyperparameter from the learning rate.

AdamGleave:
> I just noticed that the original environment rewards (`agent/rollout/ep_rew_mean`) are missing from the terminal output in this PR (replaced by the reward model outputs, `agent/rollout/ep_rew_wrapped_mean`). This is especially confusing because the name is cut off (at least for me), I only see `agent/rollout/ep_...` in the terminal. Both are logged as expected in tensorboard. Does anyone know what's going on here? I would have expected both rewards to also be shown in the terminal

Thanks for flagging this! I've uncovered two issues:
  1. When the RL algorithm is creating a new logger. `AgentTrainer`'s setter method for `logger` sets this, but by default that setter method is not called, as the `custom_logger` keyword argument to `__init__` is passed directly into superclass `TrajectoryGenerator`, and that set the logger using `self._logger`.
  2. The `HumanOutputFormat` truncates the keys as you've noticed, and then stores the truncated keys in a dict. If multiple keys have the same truncation (because they start with the same prefix), you'll only see one of them...

#2 seems like an upstream bug, so I'll open an issue in SB3 to track this. As a workaround we could bump `max_length` in `_truncate`, or abbreviate our names (but the max length seems quite short IMO, currently `23`).

AdamGleave:
@ejnnr I've opened https://github.com/DLR-RM/stable-baselines3/issues/750 to track the logger issue.

Once I hear back from you about `Adam` vs `AdamW`, I intend to make that change and then merge the PR, since it's been open for a while and I'd like to avoid any gnarly merge conflicts later down the line.

src/imitation/algorithms/preference_comparisons.py
ejnnr: I don't have any practical experience with `Adam` vs `AdamW` (haven't even read the entire paper). Don't think it matters much, but I'd lean slightly towards switching to `AdamW` (I don't think it has overtaken `Adam` as the "standard" optimizer, but I do think it has seen substantial adoption, so I'd guess they decoupling advantages are real?)

ejnnr:
Ah, thanks for debugging this, that explanation makes sense! Merging sounds good (either after or without switching to `AdamW`)

Issue #383, opened 2022-01-19
Expand quickstart tutorials


AdamGleave:
We currently have a brief (<70 LOC) [quickstart](https://github.com/HumanCompatibleAI/imitation/blob/master/examples/quickstart.py) Python file, illustrating some simple uses of the API. This is helpful, but it (a) doesn't walk the user through very much; (b) misses a lot of important features of the API, e.g. serializing and deserializing reward models and policies.

It would be nice to have more examples in the docs for common use cases, and a series of Jupyter notebook tutorials similar to the [SB3 RL tutorial](https://github.com/araffin/rl-tutorial-jnrr19/tree/sb3). Although as an initial step, I think it's best to err on the side of a fairly small tutorial, and we can then expand as we get feedback from users on what's confusing.

ernestum:
- quickstart.py should be actually quick. Therefore a simple< 20 loc example to demonstrate that imitation works in the most basic setting. Render a trajectory before training and one after to show that it actually works. This is a very light example with big "entertainment factor".
- train_using_{algo}.py for algo in [BC, DAgger, density, MCE, AIRL, GAIL, preferences]. To introduce all the algorithms with their individual quirks. Can be longer and with more explanations.
- warm_start.py Examples showing training an algorithm warm-started from a previously trained reward model and/or policy using AIRL.
- pause_and_resume.py demonstrating serialization of reward models and policies.
- data_analysis.py demonstrating how to visualize data e.g. rollout.rollout_stats.

PR #397, opened 2022-02-01
Overhaul of the examples folder

ernestum:
To resolve issue #383 this MR introduces a couple of new example files.


examples/utils.py
```diff
@@ -0,0 +1,12 @@
+def render_a_trajectory_and_print_reward(env, policy):
```
ernestum, line 1: My gut feeling says, that this function should already be in some gym utils, but I could not find it. Does anybody have a hint for me?

AdamGleave: You could do this using `InteractiveRender` from https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/scripts/eval_policy.py#L22 (move it from script to util), `rollout.generate_trajectories` and `rollout.rollout_stats`. Would need to wrap the env into a `VecEnv` first. Not obviously cleaner given the number of utilities needed, but introduces the reader to some relevant code at least.

ernestum: I went with `stable_baselines3.common.evaluation.evaluate_policy` for now. There is a `render` flag, that we can just set to `True`.

src/imitation/data/rollout.py
```diff
@@ -596,8 +594,54 @@ def rollout_and_save(
     if verbose:
         stats = rollout_stats(trajs)
         logging.info(f"Rollout stats: {stats}")
+    return trajs
+
+
+def rollout_and_save(
```
AdamGleave, line 600: I think `rollout_and_save` is only used in `train_rl`, so I'm tempted to just remove this function given it's basically a one-liner anyway

AdamGleave (commented):
Thanks, this is shaping up nicely!

I've left some minor comments. A few higher-level notes:
  - We really want `examples/quickstart.py` to actually be runnable, so having a dependency on `pytest_cache` is not ideal. 
  - I'm worried the notebooks will get broken over time and we won't realize. Can we add some smoke tests to make sure they run? I think `pytest-notebook` configured to [ignore the output](https://pytest-notebook.readthedocs.io/en/latest/user_guide/tutorial_config.html) would do this. (Or we could try to make the output actually deterministic by seeding, but that sounds tricky.)

examples/4_train_airl.ipynb
```diff
@@ -0,0 +1,185 @@
...
+   "source": [
+    "When we look at the histograms of rewards before and after learning, we can see that the learner is not perfect yet, but it made some progress at least.\n",
+    "If not, just re-run the above cell."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [
```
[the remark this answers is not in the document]
ernestum: No this was just an accident. I removed it.

src/imitation/data/rollout.py
ernestum: Good point. I removed it.

examples/1_train_bc.ipynb
```diff
@@ -0,0 +1,195 @@
...
+    "        )\n",
+    "expert.learn(100000)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's quickly check if the expert is any good.\n",
+    "A mean reward of 500 should be fine."
```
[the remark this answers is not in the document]
ernestum: Good point! I adapted the text.

ernestum:
I am a bit stuck with the MCE example. I figured out how to make it execute in some sensible way but the results are garbage. Maybe I need a different RL algorithm for the learner? I just took PPO with the same hyper-parameters as for CartPole which is probably not the smartest solution. Do you have any hints for me?

examples/quickstart.py
```diff
@@ -1,78 +1,21 @@
-"""Trains BC, GAIL and AIRL models on saved CartPole-v1 demonstrations."""
```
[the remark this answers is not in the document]
ernestum: For me the purpose is to have a minimal example, that fits (more or less, if you ignore the imports) on a single screen.
It also allows you to quickly launch something without having to spin up a jupyter notebook (in case that adds friction).
However, I see your point that the difference to the `train_bc.ipynb` is minimal right now. The only thing I really like is that in the python script we can set render=True to actually show what is going on.

ernestum:
> * I'm worried the notebooks will get broken over time and we won't realize. Can we add some smoke tests to make sure they run? I think `pytest-notebook` configured to [ignore the output](https://pytest-notebook.readthedocs.io/en/latest/user_guide/tutorial_config.html) would do this. (Or we could try to make the output actually deterministic by seeding, but that sounds tricky.)

There are actually already a `test_run_example_notebooks` test case in `test_examples.py`. 
It just runs the notebooks to verify that there are no exceptions. 
I think this is already quite a good smoke test.
However, I disabled the test for now using `pytest.mark.skip` since some of the notebooks take quite long.
I think there are two options here:
- We tweak the parameters in the notebooks so they finish quickly with garbage results but add the known good parameters in a comment. Something like:
  ```python
   expert.train(100)  # Note: change to 100000 to train a working expert
   ```
   I like that this approach nudges the user to play around with the code and contribute something of their own to make it "work".
- We mark the notebook tests (and potentially some others too) as `pytest.mark.slow` and set up a separate slow tests job. This job would be [triggered manually](https://circleci.com/docs/2.0/workflows/#holding-a-workflow-for-a-manual-approval) e.g. just before merging a PR but not on every push.

examples/quickstart.py
```diff
@@ -1,78 +1,21 @@
...
-    trajectories = pickle.load(f)
-
-# Convert List[types.Trajectory] to an instance of `imitation.data.types.Transitions`.
-# This is a more general dataclass containing unordered
-# (observation, actions, next_observation) transitions.
-transitions = rollout.flatten_trajectories(trajectories)
 
-venv = util.make_vec_env("seals/CartPole-v0", n_envs=2)
+env = gym.make("seals/CartPole-v0")
+with open(".pytest_cache/d/experts/CartPole-v1/rollout.pkl", "rb") as f:
```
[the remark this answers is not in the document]
AdamGleave: OK!

AdamGleave:
> I am a bit stuck with the MCE example. I figured out how to make it execute in some sensible way but the results are garbage. Maybe I need a different RL algorithm for the learner? I just took PPO with the same hyper-parameters as for CartPole which is probably not the smartest solution. Do you have any hints for me?

The example as given seems to generate trajectories from a random policy, I assume that's because you couldn't get a good RL expert to train?

I think your best option will be to use `mce_partition_fh` from `mce_irl` and then turn it into a `BasePolicy` using `TabularPolicy` to get rollouts. This does soft value iteration to produce an entropy-regularized optimal policy.

You could also do hard value iteration, but (a) we don't have that implemented; (b) it'll cause MCE IRL to do some strange things (basically reward scale will keep growing each iteration to try and explain the deterministic policy under a softmax model of demonstrations).

examples/quickstart.py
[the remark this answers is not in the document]
AdamGleave: Thanks for the explanation, yeah I can see the value of keeping this.

examples/quickstart.py
```diff
@@ -1,78 +1,64 @@
...
-gail_trainer = gail.GAIL(
-    venv=venv,
-    demonstrations=transitions,
-    demo_batch_size=32,
-    gen_algo=sb3.PPO("MlpPolicy", venv, verbose=1, n_steps=1024),
-    reward_net=gail_reward_net,
-    custom_logger=gail_logger,
-)
-gail_trainer.train(total_timesteps=2048)
+reward, _ = evaluate_policy(bc_trainer.policy, env, 3, render=True)
```
AdamGleave, line 57: Consider `n_eval_episodes=3` to make it explicit (`3` seems quite magical otherwise); same for next usage on line 63

ernestum: Good point. I fixed it.

ernestum:
> The example as given seems to generate trajectories from a random policy, I assume that's because you couldn't get a good RL expert to train?
> 
> I think your best option will be to use `mce_partition_fh` from `mce_irl` and then turn it into a `BasePolicy` using `TabularPolicy` to get rollouts. This does soft value iteration to produce an entropy-regularized optimal policy.

I just looked at the tests, where random trajectories are sampled. I assumed that MCE wants those random trajectories :-)

Thanks for the hint, I will try to go that way.

ernestum:
> Thanks for the hint, I will try to go that way.

I added your suggestion. Still no meaningful outcome unfortunately. I guess it is a matter of the hyper-parameters. Unless you have a hint on what to tweak, I would just run a hp-sweep to find good ones.
I would however do that in another PR so we can finally close this one.

AdamGleave (approved):
LGTM.

Note I've modified MCE IRL and density baselines notebook; please review those before merging, and tag me for re-review if you end up making any significant modifications.

The MCE IRL notebook is working now but might benefit from some clean up. E.g. the pretty histogram you had didn't integrate directly with the new rollout code I used so I just ripped it out, but it wouldn't be hard to plumb it in.

Density baselines also working though some rough edges, see discussion in #412.

MCE IRL notebook depends on changes in #414, tests may not pass in this PR until that is merged.

ernestum (commented):
LGTM and thanks for cleaning up after me!

PR #400, opened 2022-02-02
Upgrade package versions and fix CI errors

AdamGleave:
- Force recent version of SB3 (pip was backtracking to older version due to Gym dependency issue)
  - Blacken notebooks (now supported by black)  
  - Adds Ray to test requirements as it's necessary to test scripts parallel
  - Regenerate policy so that it can be loaded in more recent versions (of SB3/cloudpickle)


ernestum (approved):
lgtm

PR #404, opened 2022-02-07
Rename key of original environment reward in `RewardVecEnvWrapper` info dict

yulonglin:
In the logs, `ep_rew_wrapped_mean` refers to the reward model rewards, while `ep_rew_mean` reports original environment rewards from `Monitor`, which is applied to the environment immediately after creation.

In `RewardVecEnvWrapper`, we return an info dict with the original environment rewards under the key `wrapped_env_rew`. This is inconsistent with the logs where "wrapped" refers to reward model rewards.

As such, this PR changes the `wrapped_env_rew` key of the info dict to `original_env_rew`.


AdamGleave (approved):
LGTM, will merge once tests pass

PR #405, opened 2022-02-09
Train experts on demand

ernestum:
In the long run we want to avoid checked in binary blows with trained expert policies and their trajectory rollouts.

This PR removes the use those binary blobs from the test. Only the scripts still rely on them and therefore also the tests for the scripts.

Instead of loading the expert policies from the checked-in binary blobs they are trained and computed on-demand.
To avoid retraining the experts for each test run, they are cached in the pytest cache (located at `.pytest_cache`).
The pytest cache is cached using circleci caches.


.circleci/config.yml
```diff
@@ -140,7 +140,10 @@ jobs:
               sleep 1
             done
           background: true
-
```
AdamGleave, line 143: Nit: Rest of the file has whitespace between list entries. Think it makes it a bit more readable, but I don't have a strong preference on this, but if we remove whitespace here best remove it elsewhere in file for consistency.

.circleci/config.yml
```diff
@@ -151,7 +154,10 @@ jobs:
             mv .coverage .coverage.imitation
             coverage combine  # rewrite paths from virtualenv to src/
       - codecov/upload
-
+      - save_cache:
```
AdamGleave, line 157: It appears that CircleCI cache is immutable: https://circleci.com/docs/2.0/caching/#writing-to-the-cache-in-workflows So IIUC as written this will only save the cache the first time this runs, and it will never be invalidated again, until we bump `v1` to `v2` etc. Is this intended?

One option would be to checksum on `setup.py` like we do for dependencies. This will result in invalidating it sometimes when it's not necessary, though, probably only worthwhile if it's quick to regenerate.

tests/conftest.py
```diff
@@ -1,9 +1,172 @@
...
+def load_or_train_ppo(
+    cache_path: str,
+    training_function: Callable[[gym.Env], PPO],
+    venv,
+) -> PPO:
+    try:
+        return PPO.load(cache_path, venv)
+    except OSError:
+        pass
+    except AssertionError:
```
AdamGleave, line 31: When do we expect to see this? Internal checks of SB3 that could fail across versions?

AdamGleave: Also, it might be nice to log the traceback/exception information so that we can tell why policy was regenerated, even though it's no longer a hard-fail.

tests/conftest.py
```diff
@@ -1,9 +1,172 @@
...
+from imitation.data.types import TrajectoryWithRew
+from imitation.data.wrappers import RolloutInfoWrapper
+from imitation.policies.base import FeedForward32Policy, NormalizeFeaturesExtractor
 from imitation.util import logger
+from imitation.util.networks import RunningNorm
+
+CARTPOLE_ENV_NAME = "CartPole-v1"
+
+
+def load_or_train_ppo(
```
AdamGleave, line 22: How does this interact with `pytest-xdist`? Are we using a lock file somewhere, or did you work out a better solution?

tests/conftest.py
```diff
@@ -1,9 +1,172 @@
...
+        cache_path,
+        policy,
+        venv,
+        rollout.make_sample_until(min_timesteps=2000, min_episodes=57),
+    )
+    with open(cache_path, "rb") as f:
+        return pickle.load(f)  # TODO: not re-loading the trajectory would be nicer here
+
+
+@pytest.fixture(params=[1, 4])
```
AdamGleave, line 56: I think previously we only used different num envs for some of our tests (notably those in DAgger). Seems probably good to have this difference tested elsewhere, but can you quickly check what affect this has on test suite runtime? If it's a significant increase we might want to be more selective. I think we can use indirect parametrization to achieve this: https://docs.pytest.org/en/6.2.x/example/parametrize.html#indirect-parametrization

tests/conftest.py
```diff
@@ -1,9 +1,172 @@
...
+        policy_kwargs=policy_kwargs,
+        env=VecNormalize(cartpole_venv, norm_obs=False),
+        seed=0,
+        batch_size=64,
+        ent_coef=0.0,
+        learning_rate=0.0003,
+        n_epochs=10,
+        n_steps=64 // cartpole_venv.num_envs,
+    )
+    policy.learn(100000)
```
AdamGleave, line 83: Can we add some check that it meets a minimum level of performance, e.g. average return close to 500?

Our tests assume the expert is actually an expert, so I'd be expect a bunch of spurious failures (and maybe even spurious successes) if expert policy ends up being a dud.

AdamGleave: Also, does this get run twice because cartpole_venv is parametrized? That seems undesirable if so.

tests/conftest.py
```diff
@@ -1,9 +1,172 @@
...
+
+PENDULUM_ENV_NAME = "Pendulum-v1"
+
+
+@pytest.fixture
+def pendulum_venv() -> gym.Env:
+    return DummyVecEnv([lambda: RolloutInfoWrapper(gym.make(PENDULUM_ENV_NAME))] * 8)
+
+
+def train_pendulum_expert(pendulum_venv) -> PPO:
```
AdamGleave, line 123: Code for this + below is very similar to CartPole, I wonder if it's worth trying to be more DRY. Although most ways I can think of to do that are sufficiently magical that I'm not sure it's worth it, at least so long as we only have two environments.

AdamGleave (commented):
Thanks for this, looks like a significant improvement!

Mostly minor comments, main points are:
  - Does cache ever get invalidated on CircleCI? Think we should either make it so that it does invalidate automatically, or if manual invalidation is required then we should change it to *fail* if policy can't be loaded so that we notice tests failing and invalidate manually (rather than just wondering why our test suite got slower).
  - Interaction with parallel execution via `pytest-xdist`: do we need locking?
  - Some questions about impact on speed of test suite from parametrizing `num_envs` equally everywhere (although generally love that we're now using consistent fixtures, simplifies test code considerably).

.circleci/config.yml
ernestum: Do you mean the newlines between the steps of the job? Otherwise, that list entries are you talking about?

.circleci/config.yml
ernestum: That is true. I deliberately decided to just add the version number to keep it as simple as possible. However, I did not consider the problem, that we might silently retrain the experts for each test run, which would be wasteful.

I added the `setup.py` just as you recommended and also added a more generic key to the `restore_cache` option. This ensures, that whenever we change `setup.py`, we either have a cache miss (never trained anything previously) or just fetch the previous pytest cache. Either this cached experts are still properly loadable (then they are just copied to the new circleci cache) or they are retrained (and then placed in the new circleci cache). This is a nice way to prevent retraining when the `setup.py` changes are not relevant for the policy.

Now we still need to hope that changes outside of `setup.py` will cause or policies to not be loadable again.

tests/conftest.py
ernestum: Yes, I had AssertionErrors when loading outdated model files. This is why I added it there. Initially I just tried to catch `Exception` but PEP8 does not approve of that.
I now added a warning when the policy is retrained, rerolled and a comment to explain the AssertionErrors.

tests/conftest.py
ernestum: Yes, here I decided, that the LOC and duplication we save by generalizing are not worth the added complexity and readability of the code.

tests/conftest.py
ernestum: The `cartpole_expert_policy` fixture is executed twice. The first time it will train and store a policy. The second time it will just load the policy. So we do not train twice. I tried to ease the confusion be renaming the env parameter in the `train_cartpole_expert` such that there is no name-clash with the `cartpole_venv` fixture.

As for the policy evaluation: should I do some evaluation trajectories or should I check the return during training? What would be the appropriate value for the pendulum environment?

tests/conftest.py
ernestum: Good catch! I forgot about it, now added a lock file so the workers are waiting for each other to train the model.

.circleci/config.yml
AdamGleave: Yes, that's what I meant, sorry for ambiguity.

tests/conftest.py
```diff
@@ -1,9 +1,185 @@
...
+CARTPOLE_ENV_NAME = "CartPole-v1"
+
+
+def load_or_train_ppo(
+    cache_path: str,
+    training_function: Callable[[gym.Env], PPO],
+    venv,
+) -> PPO:
+    os.makedirs(os.path.dirname(cache_path), exist_ok=True)
+    with FileLock(cache_path + ".lock"):
```
AdamGleave, line 32: Just to note this lock is a bit trigger-happy as it'll stop multiple processes from *reading* even if it already exists. But that's probably fine, loading should be quick relative to the rest of the tests, so I don't think we need to change it but just something to watch out for if we do ever see test slowdowns.

tests/conftest.py
AdamGleave: Checking return during training is sufficient I think.

[Zoo](https://github.com/DLR-RM/rl-baselines3-zoo/blob/master/logs/benchmark/benchmark.md) reports -172 for `Pendulum-v0` with `ppo`, the best I see is -150 and the worst is around -210. So I'd guesstimate -185 should be enough to catch egregious errors but not be too flaky. Might be worth running a few seeds of PPO on Pendulum to check the mean and variance, then pick something about 2 SD below mean.

AdamGleave (commented):
This looks very nearly ready, the only thing I'd insist on before merging is some checks on return.

I would like to keep unit test suite fast, but I think that could be addressed in a separate lower-priority PR if we want to move on.

AdamGleave:
Oh, you're also missing some code coverage in the *tests*, meanings some lines of tests never get executed. It's probably something innocuous like an exceptional condition that doesn't occur, or the agent retraining not getting hit. Worth looking through CodeCov, and if there's anything that's being missed that you think is right to not be executed, just add `# pragma: no cover` to that block.

.circleci/config.yml
ernestum: OK fixed it.

.circleci/config.yml
ernestum: I updated the comments and also linked the docs just in case.

tests/conftest.py
ernestum: You mean we should wrap the FileLock only around training and not loading? That would unfortunately make things worse in comparison to no locks at all. Assume the following scenario with workers W1 and W2.
- W1 hits `load_or_train_ppo` first, loading fails, it acquires the lock and then starts training.
- W2 hits `load_or_train_ppo`, loading fails, it tries to acquire the lock but has to wait for W1
- W1 finishes training, releases the lock
- W2 gets the lock and starts training

So you see then we still have double training but now it's not even parallel any more. We need to wrap the FileLock around loading *and* training.

Parallel computing with shared state is a mess ...

AdamGleave: Yeah, I agree that locking just the training wouldn't work. 

I think there is something more sophisticated one could do, but do not think it is currently worth the added complexity given FileLock does not support this. But, the rough idea I had in mind was:
   1. Obtain a read-only lock. This can be obtained if there are other read-only holders, but *not* if there is a write holder.
   2. Attempt to load the policy.
   3. If loading succeeds, return and release lock.
   4. If loading fails, release read-only lock, then attempt to obtain write lock. This will block until all read-only lock holders relinquish the lock.
   5. Attempt to load policy. (This is needed in case we got the write lock after another process completed training.) Otherwise, train policy, and save it. 
   6. Return and release write lock.

tests/conftest.py
ernestum: Turns out that our pendulum expert is far worse than that. I trained it for a couple different seeds and this are the result from  a couple of trial runs. I evaluated using `stable_baselines3.common.evaluation.evaluate_policy`:

Mean Reward | Std over evaluation episodes
-------------------|---------------------
-232.0 | 315.3
-320.5 | 450.7
-621.3 | 314.6
-902.8 | 342.3
-421.4 | 339.2
-423.5 | 340.6
-528.8 | 334.0
-906.5 | 417.0
-500.9 | 382.0
-611.7 | 372.9

Mean over means: -547, stdev over means: 212

Turns out we do not use exactly the hyperparameters from Zoo. I am currently running a test with the parameters from there.

tests/conftest.py
ernestum: I see. That seems to be beyond the scope of the filelock library :-)

tests/conftest.py
ernestum: Results with the Zoo hyperparameters:

Mean Reward | Std over evaluation episodes
-------------------|---------------------
-368.2 | 389.6
-221.2 | 98.1
-132.5 | 102.7
-161.3 | 82.7
-183.3 | 142.6
-198.8 | 111.0
-152.9 | 71.5
-122.0 | 71.6
-342.6 | 142.1
-185.6 | 64.5

Mean over means: -207, stdev over means: 80

This does not quite match the results from the Zoo. However, note that the zoo uses v0 and we use v1 of the environment.

I will go with the median of -184.5 your guesstimate of -185 was not too bad! I will use that threshold now.

AdamGleave: Thanks for running these tests, and spotting the hyperparam difference.

If we fail if return is below median, this will fail 50% of the time? Do we want to automatically re-run it a certain number of times? Alternatively it's possible that boosting the number of `total_timesteps` moderately would improve consistency.

AdamGleave (approved):
LGTM.

I still have a mild concern about the increase in test suite runtime, but this is something we can address in another PR if it proves sufficiently annoying.

PR #409, opened 2022-02-17
Bits and pieces

ernestum:
While browsing the code I found lots of little typos and other details I would like to change, that are not big enough for their own PR.


src/imitation/algorithms/bc.py
```diff
@@ -260,7 +260,7 @@ def __init__(
         self.l2_weight = l2_weight
 
     @property
-    def policy(self) -> policies.BasePolicy:
+    def policy(self) -> policies.ActorCriticPolicy:
```
AdamGleave, line 263: Ah, you're right we need it to be an `ActorCriticPolicy` as we call `evaluate_actions`. Can you update other type annotations? `__init__` takes a `policy: Optional[policies.BasePolicy] = None` for example, we might still use `BasePolicy` elsewhere as well.

AdamGleave (approved):
LGTM; please update `BasePolicy` elsewhere in BC but no need to re-review.

ernestum:
Ah sorry, forgot to mark the PR as a draft. I added some more commits by now.

src/imitation/algorithms/bc.py
```diff
@@ -222,6 +223,7 @@ def __init__(
             ValueError: If `weight_decay` is specified in `optimizer_kwargs` (use the
                 parameter `l2_weight` instead.)
         """
+        self._demo_data_loader: Iterable[TransitionMapping] = None
```
AdamGleave, line 226: `None` is not an `Iterable`, should this be `Optional[...]` again?

src/imitation/util/logger.py
```diff
@@ -14,7 +14,7 @@
 def _build_output_formats(
     folder: str,
     format_strs: Sequence[str],
-) -> Sequence[sb_logger.KVWriter]:
+) -> List[sb_logger.KVWriter]:
```
AdamGleave, line 17: We generally try to go for the more generic type to keep flexibility of changing return type in future, e.g. making it a tuple actually seems reasaonable. Do we use the fact it's a list not a sequence anywhere (e.g. do we rely on mutability, `MutableSequence`)?

AdamGleave (commented):
A few minor points.

I think the `tests/testdata` files should be deleted, I'm guessing a bad rebase?

ernestum:
> I think the `tests/testdata` files should be deleted, I'm guessing a bad rebase?

No. I remember adding them explicitly in [5fb5e64](https://github.com/HumanCompatibleAI/imitation/pull/409/commits/5fb5e64c1895f24edb0a3ea1fac88007a0be8591). I think fixing the dependencies changed `setup.py` which invalidated the venv cache, which caused new versions of pickle (or something related) to be installed, which caused the old expert files to not be loadable. Therefore I just retrained them.

src/imitation/util/logger.py
ernestum: The output of `_build_output_formats` is always passed to a `stable_baselines3.common.logger.Logger`, which requires the type `List[KVWriter]` and not `Sequence[KVWriter]` (even though a sequence would be enough).
You are right though, just because we messed up the type annotations in SB3, we should not start messing them up in `imitation` as well.
I reverted the type annotation and added an explicit cast to `list` to please the type checker.

AdamGleave (approved):
LGTM.

Issue #410, opened 2022-02-23
Inconsistent naming of arguments in `ExplorationWrapper` of policies


kmdanielduan:
- In [`ExplorationWrapper`](https://github.com/HumanCompatibleAI/imitation/blob/e6e531180b26d94aec802aed5bb2ae96fb67fc31/src/imitation/policies/exploration_wrapper.py#L11), there are two arguments for `__init__()` to produce \epsilon-greedy behaviors of any wrapped policy:
    - `random_prob`: The probability of picking the random policy when switching.
    - `stay_prob`: The probability of staying with the current policy.
- `random_prob` seems fine, but `stay_prob` seems to produce the opposite behavior under ` if self.rng.rand() < self.stay_prob: self._switch()` when it's called. (i.e. If `stay_prob==1.0`, then always switch policy)

Proposed solution: refactor `stay_prob` to `switch_prob`.

See the source code below:
```
class ExplorationWrapper:
    """Wraps a PolicyCallable to create a partially randomized version.

    This wrapper randomly switches between two policies: the wrapped policy,
    and a random one. After each action, the current policy is kept
    with a certain probability. Otherwise, one of these two policies is chosen
    at random (without any dependence on what the current policy is).

    The random policy uses the action_space.sample() method.
    """

    def __init__(
        self,
        policy: rollout.PolicyCallable,
        venv: vec_env.VecEnv,
        random_prob: float,
        stay_prob: float,
        seed: Optional[int] = None,
    ):
        """Initializes the ExplorationWrapper.

        Args:
            policy: The policy to randomize.
            venv: The environment to use (needed for sampling random actions).
            random_prob: The probability of picking the random policy when switching.
            stay_prob: The probability of staying with the current policy.
            seed: The random seed to use.
        """
        self.wrapped_policy = policy
        self.random_prob = random_prob
        self.stay_prob = stay_prob
        self.venv = venv

        self.rng = np.random.RandomState(seed)
        self.venv.action_space.seed(seed)
        # breakpoint()

        self.current_policy = policy
        # Choose the initial policy at random
        self._switch()

    def _random_policy(self, obs: np.ndarray) -> np.ndarray:
        acts = [self.venv.action_space.sample() for _ in range(len(obs))]
        return np.stack(acts, axis=0)

    def _switch(self) -> None:
        """Pick a new policy at random."""
        if self.rng.rand() < self.random_prob:
            self.current_policy = self._random_policy
        else:
[...]

PR #411, opened 2022-02-23
Refactor `stay_prob` to `switch_prob` in `ExplorationWrapper`

kmdanielduan:
This PR aims to address #410. It refactors `stay_prob` to `switch_prob` in `ExplorationWrapper`.

### DIFF
--- a/src/imitation/policies/exploration_wrapper.py
+++ b/src/imitation/policies/exploration_wrapper.py
@@ -33,7 +33,7 @@ class ExplorationWrapper:
             policy: The policy to randomize.
             venv: The environment to use (needed for sampling random actions).
             random_prob: The probability of picking the random policy when switching.
-            switch_prob: The probability of switch away from the current policy.
+            switch_prob: The probability of switching away from the current policy.
             seed: The random seed to use.
         """
         self.wrapped_policy = policy
--- a/src/imitation/policies/exploration_wrapper.py
+++ b/src/imitation/policies/exploration_wrapper.py
@@ -24,7 +24,7 @@ class ExplorationWrapper:
         policy: rollout.PolicyCallable,
         venv: vec_env.VecEnv,
         random_prob: float,
-        stay_prob: float,
+        switch_prob: float,
         seed: Optional[int] = None,
     ):
         """Initializes the ExplorationWrapper.
@@ -33,12 +33,12 @@ class ExplorationWrapper:
             policy: The policy to randomize.
             venv: The environment to use (needed for sampling random actions).
             random_prob: The probability of picking the random policy when switching.
-            stay_prob: The probability of staying with the current policy.
+            switch_prob: The probability of switch away from the current policy.
             seed: The random seed to use.
         """
         self.wrapped_policy = policy
         self.random_prob = random_prob
-        self.stay_prob = stay_prob
+        self.switch_prob = switch_prob
         self.venv = venv
 
         self.rng = np.random.RandomState(seed)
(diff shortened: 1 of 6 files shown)


tests/policies/test_exploration_wrapper.py
```diff
@@ -57,11 +57,46 @@ def test_switch():
     assert num_constant > 450
 
 
+def test_switch_prob():
+    # Ensure that we test both the random and the wrapped policy
```
AdamGleave, line 61: Meaning of comment unclear. Code block immediately following it checks that the policy is never switched (when `switch_prob=0.0`) which does not seem to connect to this.

AdamGleave, line 61: Meaning of comment unclear. Code block immediately following it checks that the policy is never switched (when `switch_prob=0.0`) which does not seem to connect to this.

tests/policies/test_exploration_wrapper.py
```diff
@@ -57,11 +57,46 @@ def test_switch():
...
+    wrapper, venv = make_wrapper(random_prob=0.5, switch_prob=0.0)
+    policy = wrapper.current_policy
+    np.random.seed(0)
+    obs = np.random.rand(100, 2)
+    for action in wrapper(obs):
+        assert venv.action_space.contains(action)
+        assert wrapper.current_policy == policy
+
+    for random_prob in [0.0, 0.5]:
+        wrapper, venv = make_wrapper(random_prob=random_prob, switch_prob=1.0)
```
AdamGleave, line 72: I suggest factoring this code block out into a helper function, then calling it via something like:

```
num_random, num_constant = _wrapper_sample(random_prob=0.5, num_timesteps=1000, seed=0)
assert num_random > 450
assert num_constant > 450
num_random, num_constant = _wrapper_sample(random_prob=0.0, num_timesteps=1000, seed=0)
assert num_random == 0
assert num_constant == 1000
```

For loop seems a bit weird when we then have an if-else statement for every value the for loop can take.

src/imitation/policies/exploration_wrapper.py
```diff
@@ -33,12 +33,12 @@ def __init__(
             policy: The policy to randomize.
             venv: The environment to use (needed for sampling random actions).
             random_prob: The probability of picking the random policy when switching.
-            stay_prob: The probability of staying with the current policy.
+            switch_prob: The probability of switch away from the current policy.
```
AdamGleave, line 36:
```suggestion
            switch_prob: The probability of switching away from the current policy.
```

AdamGleave (commented):
Thanks for noticing and fixing this issue! Mostly looks good, some suggestion to clean up the tests, please request re-review once that's addressed.

tests/policies/test_exploration_wrapper.py
kmdanielduan: Yeah, I agree. Changed both tests into more detailed descriptions.

tests/policies/test_exploration_wrapper.py
kmdanielduan: Sounds good! I made corresponding changes to this.

kmdanielduan:
Sounds good! I have made changes to clean up the tests.

tests/policies/test_exploration_wrapper.py
```diff
@@ -68,33 +89,35 @@ def test_switch_prob():
...
-                else:  # pragma: no cover
-                    raise ValueError("Unknown policy")
-        if random_prob == 0.5:
-            # Holds with very high probability (and seeding means it's deterministic)
-            assert num_random > 450
-            assert num_constant > 450
-        else:
-            assert num_random == 0
-            assert num_constant == 1000
+            _ = wrapper(obs)
```
AdamGleave, line 99:
```suggestion
            wrapper(obs)
```
It's OK to just discard return values in Python.

AdamGleave (approved):
LGTM

PR #414, opened 2022-02-28
MCE IRL: fix sample-based occupancy measure computation

AdamGleave:
`MCEIRL` either takes the occupancy measure directly, or estimates it from a finite number of demonstrations. `mce_occupancy_measure` computes occupancy measure by summing discounted occupancies over an episode. For sample-based approximations, we were normalizing everything to sum to 1, when it should instead sum to `sum([gamma**t for t in range(horizon)])`.

This PR adjusts the sample-based approximation to match that of `mce_occupancy_measures`.

### DIFF
--- a/src/imitation/algorithms/mce_irl.py
+++ b/src/imitation/algorithms/mce_irl.py
@@ -356,7 +356,7 @@ class MCEIRL(base.DemonstrationAlgorithm[types.TransitionsMinimal]):
                         # TODO(adam): vectorize?
                         cum_discount = 1.0
                         # TODO(adam): do we want to omit last state?
-                        for obs in traj.obs[:-1]:
+                        for obs in traj.obs:
                             self.demo_state_om[obs] += cum_discount
                             cum_discount *= self.discount
                         num_demos += 1
--- a/src/imitation/algorithms/mce_irl.py
+++ b/src/imitation/algorithms/mce_irl.py
@@ -5,6 +5,7 @@ Follows the description in chapters 9 and 10 of Brian Ziebart's `PhD thesis`_.
 .. _PhD thesis:
     http://www.cs.cmu.edu/~bziebart/publications/thesis-bziebart.pdf
 """
+import collections
 import warnings
 from typing import Any, Iterable, Mapping, Optional, Tuple, Type, Union
 
@@ -94,10 +95,10 @@ def mce_occupancy_measures(
         discount: rate to discount the cumulative occupancy measure D.
 
     Returns:
-        Tuple of Dt (ndarray) and D (ndarray). Dt is a :math:`T*|S|`-dimensional vector
-        recording the probability of being in a given state at a given timestep. D is an
-        :math:`|S|`-dimensional vector recording the expected discounted number of times
-        each state is visited.
+        Tuple of ``D`` (ndarray) and ``Dcum`` (ndarray). ``D`` is of shape
+        ``(env.horizon, env.n_states)`` and records the probability of being in a
+        given state at a given timestep. ``Dcum`` is of shape ``(env.n_states,)``
+        and records the expected discounted number of times each state is visited.
     """
     # shorthand
     horizon = env.horizon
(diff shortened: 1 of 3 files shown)


src/imitation/algorithms/mce_irl.py
```diff
@@ -321,49 +321,92 @@ def __init__(
...
+                    batch.get("next_obs"),
                 )
+        else:
+            raise TypeError(
+                f"Unsupported demonstration type {type(demonstrations)}",
+            )
 
-            self.demo_state_om /= self.demo_state_om.sum()  # normalize
+        # Normalize occupancy measure estimates
+        self.demo_state_om *= (self.env.horizon + 1) / self.demo_state_om.sum()
```
qxcv, line 409: IMO this belongs in `_update_demo_from_obs`; it's hard to see why this line is correct when the OM update logic is in another function. You could still accommodate the `isinstance(demonstrations, Iterable)` branch by chaining together the batches in `demonstrations` and making just one `_update_demo_from_obs` call.

qxcv: (also if you change this then you may as well change `_update_demo_from_obs` and `_update_demo_from_trajectories` to zero out `demo_state_om` at the beginning of each call)

src/imitation/algorithms/mce_irl.py
```diff
@@ -321,49 +321,92 @@ def __init__(
...
+            self.demo_state_om[obs] += 1.0
+
+        # We assume the transitions were flattened from some trajectories,
+        # then possibly shuffled. So add next observations for terminal states,
+        # as they will not appear anywhere else; but ignore next observations
+        # for all other states as they occur elsewhere in dataset.
+        if next_obses is not None:
+            for done, obs in zip(dones, next_obses):
+                if isinstance(done, th.Tensor):
+                    done = done.numpy()
```
qxcv, line 352: Nit/defensive programming:

```suggestion
                    done = done.item()  # must be scalar
```

src/imitation/algorithms/mce_irl.py
```diff
@@ -321,49 +321,92 @@ def __init__(
...
+
+    def _update_demo_from_obs(
+        self,
+        obses: np.ndarray,
+        dones: Optional[np.ndarray],
+        next_obses: Optional[np.ndarray],
+    ) -> None:
         for obs in obses:
             if isinstance(obs, th.Tensor):
                 obs = obs.numpy()
```
qxcv, line 342: Is `obs` meant to be a scalar or an n-dimensional array?

qxcv (changes requested):
Some questions about docs and a few style nitpicks.

src/imitation/algorithms/mce_irl.py
AdamGleave: Changed. The collation code does make things a bit more complicated, but it's probably overall still easier to understand.

src/imitation/algorithms/mce_irl.py
AdamGleave: Ah you're right it's actually scalar in this case.

src/imitation/algorithms/mce_irl.py
```diff
@@ -109,9 +110,9 @@ def mce_occupancy_measures(
...
         _, _, pi = mce_partition_fh(env, reward=reward)
 
-    D = np.zeros((horizon, n_states))
+    D = np.zeros((horizon + 1, n_states))
     D[0, :] = env.initial_state_dist
-    for t in range(1, horizon):
+    for t in range(1, horizon + 1):
         for a in range(n_actions):
             E = D[t - 1] * pi[t - 1, :, a]
             D[t, :] += E @ T[:, a, :]
```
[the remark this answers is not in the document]
AdamGleave: Good suggestion, updated.

AdamGleave:
I believe I've addressed all of Sam's comments and since this now has a LGTM from Max I'm merging, but Sam do LMK if there are any outstanding issues and I'm happy to address in another PR.

PR #415, opened 2022-03-01
Changes to support dict_obs for BC

TaoHuang13:
This PR aims to address data preprocessing of dictionary observation (like in gym fetch envs) for BC.

The function `transitions_collate_fn` in `types.py` fails to collate dictionary observation. This issue comes from: the `batch_no_infos` in `transitions_collate_fn` will turn obs (as dictionaries) into np.array(dict). This makes the type of `batch_no_infos["obs"]` as object, which is not supported by the `default_collate` function of the `DataLoader` of torch.

This incurs a further problem: when feed `batch["obs"]` into the `CombinedFeatureExtractor`, the supported type is `TensorDict` (Dictionary of tenser) defined in `sb3.common.type_aliases`. The reason we'd like to use such type is that we need `preprocess_obs` in `sb3`, with which TensorDict type obs is compatible.

I make some modifications to address the two points above when running BC. But I think there could be a more concise and general way to face up dict_obs.

### DIFF
--- a/src/imitation/algorithms/bc.py
+++ b/src/imitation/algorithms/bc.py
@@ -289,7 +289,11 @@ class BC(algo_base.DemonstrationAlgorithm):
             stats_dict: Statistics about the learning process to be logged.
 
         """
-        obs = th.as_tensor(obs, device=self.device).detach()
+        if isinstance(obs, dict):
+            for key in obs:
+                obs[key] = th.as_tensor(obs[key], device=self.device).detach()
+        else:
+            obs = th.as_tensor(obs, device=self.device).detach()
         acts = th.as_tensor(acts, device=self.device).detach()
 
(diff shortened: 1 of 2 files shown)


AdamGleave (commented):
Thanks for opening this PR! Dictionary support has been on our wishlist for a while, glad to see progress happen here.

The high-level approach seems reasonable. Does BC just work with dict observations after this? It seems we'd need to flatten it at some point to feed into e.g. an MLP, but perhaps this happens as part of the preprocessing already.

To merge we'll need the existing tests + lint to pass (in particular `TensorDict` seems not defined anywhere), and should also add a test case for dict observations for data loader and showing it working with BC.

One other minor point: can you check for isinstance of `Mapping` instead of `dict` to be a bit more generic and match the type annotations?

PR #419, opened 2022-03-10
Fix typo in Dockerfile

AdamGleave:


AdamGleave:
Trivial change so I'm going to merge without review to keep things moving.

PR #422, opened 2022-03-17
Refactoring of Behavior Cloning

ernestum:
The current BC implementation mixes training, batch/epoch chunking and the progress-bar logic heavily. 
This hides the training code and thus makes it harder to understand the BC algorithm.

This PR is an attempt to separate those concerns from each other, while mostly keeping the original API intact.

TODO:
- [x] separate batch/epoch chunking from logging
- [x] add back the progress bar


src/imitation/algorithms/bc.py
```diff
@@ -4,181 +4,176 @@
...
-    def __call__(self, _):
-        """Returns the constant learning rate."""
-        return self.lr
+        ent_loss = -self.ent_weight * entropy
+        neglogp = -log_prob
+        l2_loss = self.l2_weight * l2_norm
+        loss = neglogp + ent_loss + l2_loss
 
+        return BehaviorCloningLoss(
+            neglogp,
```
AdamGleave, line 80: Maybe use kwargs? Would be easy to mix up positional order.

src/imitation/algorithms/bc.py
```diff
@@ -4,181 +4,176 @@
...
 import torch as th
-import tqdm.autonotebook as tqdm
-from stable_baselines3.common import policies, utils, vec_env
-
-from imitation.algorithms import base as algo_base
-from imitation.algorithms.base import TransitionMapping
+from stable_baselines3.common import utils, vec_env
+from stable_baselines3.common.policies import ActorCriticPolicy
+
+from imitation.algorithms.base import (
```
AdamGleave, line 17: Per https://github.com/HumanCompatibleAI/imitation/blob/a5b311ff32ea14859eb0f64fe3deef378a73c543/CONTRIBUTING.md we're following the Google Python style guide except where it conflicts with PEP8. In particular, https://google.github.io/styleguide/pyguide.html#22-imports "Use import statements for packages and modules only, not for individual classes or functions."

I don't feel that strongly about this (although I do prefer that style as it makes it more explicit where things are coming from), but if we do change it we should probably do that code-base wide in a separate PR.

AdamGleave (commented):
Only had time to look through quickly but seems to be moving in right direction, left a couple of minor comments.

High-level I like that we're moving stats and in particular logging code separate to training. I found the iterator approach fairly easy to follow, but I also have a bias towards using iterators a lot, may be worth thinking if there's a direct approach that is easier to understand (e.g. for loop over an iterator, call a training function, call a stats function) but I've not thought hard enough to establish if that would be better or not.

src/imitation/algorithms/bc.py
ernestum: Thx fixed.

src/imitation/algorithms/bc.py
ernestum: Ah this rule is new to me. I don't think its bad. I fixed it for now. I think the rule forces us to really think about our module names. E.g. I think we should reconsider the (repeated) usage of `base`, `common` etc. and maybe stick to more specific names.

AdamGleave: I agree we should try to improve the module names.

ernestum:
I would like to close this PR for now to keep it reasonably small.
Further changes would require modifications of the API/interface of the BC class and those I would like to do together with an overhaul of the tests in a separate PR.

src/imitation/algorithms/bc.py
```diff
@@ -4,178 +4,261 @@
...
+        self,
+        policy: policies.ActorCriticPolicy,
+        obs: Union[th.Tensor, np.ndarray],
+        acts: Union[th.Tensor, np.ndarray],
+    ) -> BehaviorCloningLoss:
+        """Calculate the supervised learning loss used to train the behavioral clone.
 
         Args:
-            lr: the constant learning rate that calls to this object will return.
+            policy: The actor-critic policy of which to compute the loss.
```
AdamGleave, line 108:
```suggestion
            policy: The actor-critic policy whose loss is being computed.
```

src/imitation/algorithms/bc.py
```diff
@@ -4,178 +4,261 @@
...
+                    )
+                if self.on_epoch_end is not None:
+                    self.on_epoch_end(epoch_num)
+
+        # Note: the islice here ensures we do not exceed self.n_batches
+        return itertools.islice(batch_iterator(), self.n_batches)
+
+
+@dataclasses.dataclass(frozen=True)
+class BehaviorCloningLoss:
```
AdamGleave, line 80: +1 on having this be a dataclass not just some unstructured dict. But is it really the loss? In line 127 we have `loss = neglogp + ent_loss + l2_loss`, so the other values are either intermediate values (`entropy`, `l2_norm` unweighted versions of `ent_loss` and `l2_loss`) or are purely information (`prob_true_act` I don't see used anywhere in computation of final `loss`).

Perhaps rename it to `BehaviorCloningStats` or similar?

src/imitation/algorithms/bc.py
```diff
@@ -4,178 +4,261 @@
...
+@dataclasses.dataclass(frozen=True)
+class BehaviorCloningTrainer:
+    """Functor to fit a policy to expert demonstration data."""
 
-    def close(self):
-        pass
+    loss: BehaviorCloningLossCalculator
+    optimizer: th.optim.Optimizer
+    policy: policies.ActorCriticPolicy
+    device: th.device  # TODO(max): not sure whether the device belongs in this class
```
AdamGleave, line 147: I think we can use `self.policy.device` instead; the only reason we need it for this class AFAICT is to send the inputs to `self.device` before evaluating the loss which is all in terms of `self.policy`.

src/imitation/algorithms/bc.py
```diff
@@ -4,178 +4,261 @@
...
 
-    def __iter__(
+class BCLogger:
+    """Utility class to help logging information relevant to Behavior Cloning."""
+
+    def __init__(self, logger: imit_logger.HierarchicalLogger):
+        """Create new BC logger.
+
+        Args:
+            logger: The logger to which to feed all the information.
```
AdamGleave, line 208:
```suggestion
            logger: The logger to feed all the information to.
```

src/imitation/algorithms/bc.py
```diff
@@ -4,178 +4,261 @@
...
-                batch_suffix = f"/{self.n_batches}"
-        else:
-            display = _NoopTqdm()
+        batch_num: int,
+        batch_size: int,
+        num_samples_so_far: int,
+        loss: BehaviorCloningLoss,
+        rollout_stats: Mapping[str, float],
+    ):
+        self._tensorboard_step += 1
```
AdamGleave, line 227: Feels a bit more natural to me to increment at end than at beginning (so we're 0-indexed), but doesn't really matter.

src/imitation/algorithms/bc.py
```diff
@@ -357,48 +397,57 @@ def train(
...
+
+        compute_rollout_stats = RolloutStatsComputer(
+            log_rollouts_venv,
+            log_rollouts_n_episodes,
         )
 
-        if reset_tensorboard:
-            self.tensorboard_step = 0
+        def _on_epoch_end(epoch_number: int):
+            if isinstance(batches_with_stats, tqdm.tqdm):
```
AdamGleave, line 409: The need for an `isinstance` is a little sad but probably harmless, does mean we can't wrap `batches_with_stats` any more without it causing this check to fail.

One alternative would be to have `batches_with_progress_bar` initialized to `None` at same time batches_with_stats` is set, and then `if progress_bar:` update both `batches_with_stats` and `batches_with_progress_bar` to tqdm wrapped version. Then we can drop line 411 that does a type cast, and replace the `isinstance` with just `if batches_with_progress_bar is not None`. WDYT?

src/imitation/algorithms/bc.py
```diff
@@ -357,48 +397,57 @@ def train(
...
+                total_num_epochs_str = f"of {n_epochs}" if n_epochs is not None else ""
+                batches_with_progress_bar: tqdm.tqdm = batches_with_stats
+                batches_with_progress_bar.display(
+                    f"Epoch {epoch_number} {total_num_epochs_str}",
+                    pos=1,
+                )
+            self._bc_logger.log_epoch(epoch_number)
```
AdamGleave, line 416: So the key `bc/epoch` only gets logged at the end of each epoch? Seems more intuitive to me to log it at each iteration (with epoch number only incrementing periodically).

src/imitation/algorithms/bc.py
```diff
@@ -357,48 +397,57 @@ def train(
...
+            batches_with_stats = tqdm.tqdm(
+                batches_with_stats,
+                unit="batch",
+                total=n_batches,
+            )
 
-            self.optimizer.zero_grad()
-            loss.backward()
-            self.optimizer.step()
+        for ((batch_num, batch_size, num_samples_so_far), batch) in batches_with_stats:
```
AdamGleave, line 435:
```suggestion
        for (batch_num, batch_size, num_samples_so_far), batch in batches_with_stats:
```

AdamGleave (commented):
Thanks, this looks like a significant improvement!

A few suggestions but nothing major, and nothing I'd insist on seeing changed.

Tag me when you've made any updates you want in response to my comments and I'll re-review.

src/imitation/algorithms/bc.py
```diff
@@ -4,178 +4,261 @@
...
-class ConstantLRSchedule:
-    """A callable that returns a constant learning rate."""
+    def __post_init__(self):
+        epochs_and_batches_specified = (
+            self.n_epochs is not None and self.n_batches is not None
+        )
+        neither_epochs_nor_batches_specified = (
+            self.n_epochs is None and self.n_batches is None
+        )
+        if epochs_and_batches_specified or neither_epochs_nor_batches_specified:
```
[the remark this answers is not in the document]
ernestum: Haha I can feel the urge with you to write it like this :-)
It is short and elegant but in my opinion ease of understanding trumps brevity and elegance here.

src/imitation/algorithms/bc.py
ernestum: You are right. This is not really the loss and I am also not happy with the name. `BehaviorCloningStats` is not wrong but too generic IMO. What do you think about `BCPerformanceStats`, `BCLossStats` (but then again its more than just loss)? Maybe you can come up with a better word than `Performance` here?

src/imitation/algorithms/bc.py
ernestum: Thanks for the hint! Implemented it!

src/imitation/algorithms/bc.py
ernestum: True. No idea how it ended up there. I moved it down.

src/imitation/algorithms/bc.py
ernestum: Its more or less the same LOC and equally confusing but I like that it is robust to further wrapping of the iterators. I implemented it!

src/imitation/algorithms/bc.py
ernestum: True, there were even other issues with epoch logging. I fixed it.

src/imitation/algorithms/bc.py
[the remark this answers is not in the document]
AdamGleave: Fair enough, resolving.

src/imitation/algorithms/bc.py
AdamGleave: `BCPerformanceStats` seems better than `BehaviorCloningLoss`. Another option is `BCTrainingMetrics`?

src/imitation/algorithms/bc.py
```diff
@@ -424,15 +424,17 @@ def _on_epoch_end(epoch_number: int):
             _on_epoch_end,
         )
         batches_with_stats = enumerate_batches(demonstration_batches)
+        tqdm_progress_bar: tqdm.tqdm = None
```
AdamGleave, line 427: Unless I'm misunderstanding the `tqdm.tqdm` type this shouldn't type check.

```suggestion
        tqdm_progress_bar: Optional[tqdm.tqdm] = None
```

AdamGleave (approved):
LGTM. Only things that need attending to before merge:
  - Decide if you want to rename `BehaviorCloningLoss`. I'm OK with any of the suggestions in the thread, trust your judgement here.
  - Correct type annotation for `tqdm_progress_bar` (change suggested).

I find the epoch logging still a bit weird; if `n_batches` is less than number of batches in an epoch, it'll just always log epoch 0, since `on_epoch_end` never gets called. But this is a bit of an edge case, the alternative of always incrementing the epoch when `n_batches` is hit also seems broken, at least if epoch is stuck at 0 it'll be obvious from the logs that something odd is going on. Still, maybe add a warning in the docs about this?

src/imitation/algorithms/bc.py
ernestum: I like `BCTrainingMetrics` I will go with that!

ernestum:
> I find the epoch logging still a bit weird

I agree. I plan on reducing the interface quite a lot and this should allow to simplify the implementation as well.
I can't imagine, that pytorch does not already ship with a solution for producing batches/epochs. There should be no need to re-implement that. I tried to stay as close to the original implementation as I could for now.

Issue #420, opened 2022-03-15
Randomness control for different `exploration_frac` in preference comparisons


yawen-d:
## Background
- Previously, [`exploration_frac`](https://github.com/HumanCompatibleAI/imitation/blob/3d7a76b8c587a25e380aeb09f65b764d7693aeea/src/imitation/algorithms/preference_comparisons.py#L212) was implemented to add exploratory trajectories for preference comparisons. Its aim was to add diversity to the dataset in order to escape from local minimum if any.
- In particular, `int(exploration_frac * steps)` transitions will be generated by [`\epsilon`-greedily](https://github.com/HumanCompatibleAI/imitation/blob/3d7a76b8c587a25e380aeb09f65b764d7693aeea/src/imitation/policies/exploration_wrapper.py#L62) apply the current policy after generating the normal transitions.

## Problems
1. Transition distribution shifts between normal vs. exploratory trajectories due to default `deterministic_policy` to be true.
2. Uncontrolled randomness among equivalent settings due to rounding, parallelisation and shuffling procedures.

### 1) Transition distribution shifts between normal vs. exploratory trajectories due to default `deterministic_policy` to be true for exploratory trajectories.
- For normal trajectories, policy actions are [generated stochastically](https://github.com/HumanCompatibleAI/imitation/blob/3d7a76b8c587a25e380aeb09f65b764d7693aeea/src/imitation/data/rollout.py#L305) (default to be non-deterministic policy).
- For exploratory trajectories, policy actions are [generated deterministically](https://github.com/HumanCompatibleAI/imitation/blob/3d7a76b8c587a25e380aeb09f65b764d7693aeea/src/imitation/algorithms/preference_comparisons.py#L167) from the [mode action](https://stable-baselines3.readthedocs.io/en/master/_modules/stable_baselines3/common/distributions.html#Distribution.get_actions) of the current policy distribution.
- Hence, the transition distribution shifts comes from (A) shifting between current policy and random policy, and (B) choosing actions deterministically with the highest probability. We indeed want (A) to increase exploration, but not (B) because (B) might strongly increase exploitation especially with a low `random_prob`.
- In extreme cases, for equivalent settings below, the training curves could be much different.
    a. `exploration_frac==0.0`
    b. `exploration_frac==0.2 & trajectory_generator_kwargs.random_prob=0.0`
[...]

PR #423, opened 2022-03-21
Set deterministic_policy of `exploration_wrapper` to False

kmdanielduan:
- Set deterministic_policy of `exploration_wrapper` to False.
- `exploration_wrapper.predict(...)` will
    - sample non-deterministic actions for stochastic algorithms.
    - generate deterministic actions for [deterministic algorithms](https://stable-baselines3.readthedocs.io/en/master/_modules/stable_baselines3/td3/policies.html) (e.g. TD3) because the flag of `deterministic_policy=False` is ignored when `.predict(obs)` is called.

Mentioned in #420 .


AdamGleave (commented):
Might be worth adding `deterministic_policy=False` to the `generate_trajectories` calls? It defaults to `False` there, but might be worth reiterating that both to (a) reassure the casual reader; (b) be defensive in case the default changes.

Might also be worth adding a comment briefly explaining this choice, e.g. that the `buffering_wrapper` collects rollouts from a non-deterministic policy so we do that here as well for consistency.

nohup.out
```diff
@@ -0,0 +1,18 @@
+
```
AdamGleave, line 1: I don't think you meant to commit this file?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -161,13 +161,15 @@ def __init__(
         self.log_callback = self.venv.make_log_callback()
 
         self.algorithm.set_env(self.venv)
-        policy = rollout._policy_to_callable(
+        policy_callable = rollout._policy_to_callable(
             self.algorithm,
             self.venv,
-            deterministic_policy=True,
+            # we keep the default policy scheme in determinism by setting
```
AdamGleave, line 167: Default where?

src/imitation/data/rollout.py
```diff
@@ -282,6 +282,14 @@ def get_actions(states):
             return acts
 
     elif isinstance(policy, Callable):
+        # When a policy callable is passed, by default we will use it directly.
+        # We are not able to change the determinism of the policy when it is a
+        # callable that only takes in the states.
+        if deterministic_policy is True:
```
AdamGleave, line 288:
```suggestion
        if deterministic_policy:
```

src/imitation/data/rollout.py
```diff
@@ -282,6 +282,14 @@ def get_actions(states):
...
+            raise ValueError(
```
AdamGleave, line 289: Good idea adding this check!

src/imitation/policies/exploration_wrapper.py
```diff
@@ -30,21 +30,22 @@ def __init__(
         """Initializes the ExplorationWrapper.
 
         Args:
-            policy: The policy to randomize.
+            policy_callable: The policy callable to randomize. Preferably a policy
+                callable to match the same type of self._random_policy(obs) call.
```
AdamGleave, line 34: What does "match the same type" mean? It's type is a `rollout.PolicyCallable` (according to the annotation), what other type should be aligned?

AdamGleave (commented):
Thanks, looks almost ready, main points:
  - nohup.out should be deleted
  - A few of the comments/docstring I found a bit hard to understand, can you edit where I had problems?

kmdanielduan:
Sorry that I requested re-review by mistake. I'm on it now.

src/imitation/data/rollout.py
kmdanielduan: Yeah!

src/imitation/policies/exploration_wrapper.py
kmdanielduan: `def _random_policy(self, obs: np.ndarray) -> np.ndarray:` seems to be the same type as `PolicyCallable = Callable[[np.ndarray], np.ndarray]`. Perhaps no need to include it.

AdamGleave (approved):
LGTM. (Thanks for adding test coverage.)

Issue #395, opened 2022-01-25
Reward Normalization


AdamGleave:
It is often desirable to have the reward used for RL training be normalized. In particular, scaling up the reward has similar effects to scaling up the learning rate. Normalizing ensures a consistent reward scale across environments, avoiding the need to tweak learning rate per environment.

There are two instances where the ground truth reward is used:
  - `train_rl`, for training expert policies. This currently uses `VecNormalize` with `norm_obs=False`, and I think it's probably fine to continue with this approach.
  - `PreferenceComparisons` for `SyntheticGatherer`. This currently does not normalize rewards, but it probably should, as currently the scale of reward affects the stochasticity of synthetic comparisons.

Note if `VecNormalize` is used, then the normalization statistics will need to be stored (and restored) from checkpoints to restart training. But currently the code is not set up to continue training from checkpoints anyway (although inference from checkpoints is possible).

Additionally, all reward learning algorithms that train both a reward and policy jointly (`AdversarialTrainer`, the parent of `AIRL` and `GAIL`, and `PreferenceComparisons`) uses the output of a reward network. Currently `PreferenceComparisons` does not use a normalized reward at all; `AdversarialTrainer` uses `VecNormalize`

Rather than using `VecNormalize` I suspect it is better to either (a) regularize the output of the reward network (similar to https://github.com/HumanCompatibleAI/evaluating-rewards/blob/master/src/evaluating_rewards/rewards/preferences.py#L155) which will tend to keep the scale small and relatively consistent between runs; (b) explicitly normalize the output by adding a `RunningNorm` (or `BatchNorm`?) layer to the end.

kmdanielduan:
I completed a simple implementation of that in #425.

AdamGleave:
This issue has now been addressed, closing.

PR #425, opened 2022-03-28
Normalize Wrapped Reward for PreferenceComparisons

kmdanielduan:
This PR aims to address #395 by
- Adding a `RunningNorm` layer to explicitly normalize the MLP output.


AdamGleave:
Just to flag this probably isn't how we want to implement things long-term, I think better to normalize the reward output directly coming out of the reward network. Some discussion of this in https://github.com/HumanCompatibleAI/imitation/issues/395

But your implementation looks correct (provided you don't serialize/deserialize anything, as the VecNormalize stats won't get restored), and is the right way to start to see if this even helps...

tests/rewards/test_reward_nets.py
```diff
@@ -18,7 +18,11 @@
 HARDCODED_TYPES = ["zero"]
 
 REWARD_NETS = [reward_nets.BasicRewardNet, reward_nets.BasicShapedRewardNet]
-REWARD_NET_KWARGS = [{}, {"normalize_input_layer": networks.RunningNorm}]
+REWARD_NET_KWARGS = [
+    {},
+    {"normalize_input_layer": networks.RunningNorm},
+    {"normalize_output_layer": networks.RunningNorm},
```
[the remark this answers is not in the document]
AdamGleave: (I've added this, feel free to revert if you disagree)

src/imitation/util/networks.py
```diff
@@ -107,7 +107,11 @@ def update_stats(self, batch: th.Tensor) -> None:
     def forward(self, x: th.Tensor) -> th.Tensor:
         """Updates statistics if in training mode. Returns normalized `x`."""
         if self.training:
-            self.update_stats(x)
+            # Buffered tensors' gradients cannot be cleared by `optimizer.zero_grad()`.
```
AdamGleave: I don't think this comment reflects our current understanding after our call, so I updated it; please edit my change as needed.

src/imitation/util/networks.py
```diff
@@ -107,7 +107,11 @@ def update_stats(self, batch: th.Tensor) -> None:
...
+            # We need to avoid computing gradient for running_mean, running_var, and
+            # count, otherwise gradients get computed twice in consecutive iterations.
+            with th.no_grad():
```
AdamGleave, line 114: I looked into this more and I'm confident it's doing what we want now, thanks for catching this bug :)

AdamGleave (commented):
Looks almost ready, some minor comments

src/imitation/util/networks.py
kmdanielduan: Oh, thank you for looking into this!

src/imitation/scripts/common/reward.py
```diff
@@ -43,6 +53,8 @@ def config_hook(config, command_name, logger):
         res["net_cls"] = default_net
     if "normalize_input_layer" not in config["reward"]["net_kwargs"]:
         res["net_kwargs"] = {"normalize_input_layer": networks.RunningNorm}
+    if "normalize_output_layer" not in config["reward"]["net_kwargs"]:
+        res["net_kwargs"] = {"normalize_output_layer": None}
```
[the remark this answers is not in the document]
kmdanielduan: `reward` ingredient is used in both `adversarial` and `PreferenceComparisons`. Enabling it by default might mean we need to change the code in `adversarial`. Do we want that in `GAIL` and `AIRL`?

tests/rewards/test_reward_nets.py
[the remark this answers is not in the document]
kmdanielduan: Sure!

tests/util/test_networks.py
```diff
@@ -91,3 +92,37 @@ def test_running_norm_matches_dist(batch_size: int) -> None:
...
+    [
+        {},
+        {
+            "normalize_input_layer": networks.RunningNorm,
+            "normalize_output_layer": networks.RunningNorm,
+        },
+    ],
+)
+def test_build_mlp_norm(init_kwargs):
+    """Tests that MLP can be built."""
```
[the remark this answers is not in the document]
kmdanielduan: Sure. I hope this explains it well.

tests/test_scripts.py
```diff
@@ -122,6 +122,32 @@ def test_train_preference_comparisons_main(tmpdir, config):
...
 
+@pytest.mark.parametrize(
+    "named_configs",
+    (
+        [],
+        ["reward.normalize_output_running"],
+        ["reward.normalize_output_disable"],
+    ),
+)
+def test_train_preference_comparisons_reward_norm(tmpdir, named_configs):
```
kmdanielduan, line 133: This is where the integrated tests happen.

src/imitation/scripts/common/reward.py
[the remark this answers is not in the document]
kmdanielduan: To change it to enabled by default, we might want to also delete the lines in [`imitation.algorithms.adversarials.common`](https://github.com/HumanCompatibleAI/imitation/blob/ed45793dfdd897d3ac1f3a863a8816b56d436887/src/imitation/algorithms/adversarial/common.py#L220).

For `AIRL`, it's fine to do this because the [`reward_train`](https://github.com/HumanCompatibleAI/imitation/blob/ed45793dfdd897d3ac1f3a863a8816b56d436887/src/imitation/algorithms/adversarial/airl.py#L90) of the `AdversarialTrainer` outputs the normalized output directly. Hence, it's equivalent.

For `GAIL`, the `AdversarialTrainer` would [first normalize the reward in `self.base.forward()`](https://github.com/HumanCompatibleAI/imitation/blob/ed45793dfdd897d3ac1f3a863a8816b56d436887/src/imitation/algorithms/adversarial/gail.py#L35), then take the negative log sigmoid of the logits. This order is different from the previous case of taking `-F.logsigmoid(logits)` first then normalize.

src/imitation/util/networks.py
kmdanielduan: Thanks!

src/imitation/scripts/common/reward.py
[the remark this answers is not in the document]
kmdanielduan: I basically ignored it for now by checking asserting it's running `train_preference_comparisons`, and raise `NotImplementedError` when it's running `train_adversarial`.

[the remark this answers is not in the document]
AdamGleave: Oh good point. We can have different defaults in the config hook depending on algorithm. But it sounds like we do want it normalized even in AIRL/GAIL, just right now in GAIL the code is taking logits of whatever the reward network outputs, and we want normalization after that?

This is a bit tricky to do cleanly. We could make `LogSigmoidRewardNet` take `normalize_output_layer` as an option. But making the config handle this would still be awkward. We could have separate config options for GAIL normalization, but that's error prone. Alternative would probably mean branching on whether or not its GAIL, both in the sacred config and in `train_adversarial`.

src/imitation/scripts/common/reward.py
```diff
@@ -43,6 +53,17 @@ def config_hook(config, command_name, logger):
         res["net_cls"] = default_net
     if "normalize_input_layer" not in config["reward"]["net_kwargs"]:
         res["net_kwargs"] = {"normalize_input_layer": networks.RunningNorm}
+    if "normalize_output_layer" not in config["reward"]["net_kwargs"]:
+        if command_name == "train_preference_comparisons":
+            res["net_kwargs"] = {"normalize_output_layer": networks.RunningNorm}
+        else:
+            res["net_kwargs"] = {"normalize_output_layer": None}
+    else:
+        if command_name != "train_preference_comparisons":
```
AdamGleave, line 62: Pleasingly I think the idea I had of overriding `predict` rather than normalizing at the `forward()` level would make this work with GAIL too, in which case I think we can enable it by default everywhere and remove this check.

tests/util/test_networks.py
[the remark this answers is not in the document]
AdamGleave: The docstring seems unchanged?

scottemmons:
> I'd appreciate other eyes on this @qxcv @scottemmons as this is proving more subtle than I expected.

If I understand correctly, we want the normalized RewardNet to output normalized rewards when it is being used by the RL algorithm, but still to output unnormalized rewards for its own training.

A few possible ways to do this:

1. Normalize the output of `.predict()` but not of `.forward()`. This is Adam's suggestion.
Pro: this will seamlessly fit into the current codebase with no need to change any code.
Con: now `.predict()` and `.forward()` have different scales. I think this could be confusing if someone tries to call `.predict()` to actually evaluate the reward model using the environment scale of reward.

2. Normalize the output when in `.eval()` mode but not when in `.train()` mode. Adam mentioned this but is leaning against it.
Pro: `.predict()` and `.forward()` will always be on the same scale
Con: there's no "evaluation" mode for the reward training itself (which should be unnormalized). additional code needs to be added to current workflows, and someone might forget to set the right mode

3. Define a new normalization mode, e.g. that is toggled by `.normalize()` and `.unnormalize()`, which sets the wrapper to (un)normalize the rewards in both `.predict()` and `.forward()`. This is a new idea I'm brainstorming.
Pro: it's now clear when normalization is (and isn't) being applied. `.predict()` and `.forward()` will always be on the same scale
Con:  additional code needs to be added to current workflows, and someone might forget to set the right mode

Overall, I like (1) the best because it fits seamlessly into existing code. My main worry is that it might be confusing that `.predict()` and `.forward()` behave differently by default. (The former normalizes rewards while the latter doesn't.) I propose addressing this by adding a `normalize_reward=True` parameter to `.predict()` and `normalize_reward=False` parameter to `.forward()`. This would let both methods normalize the rewards or not, but their default behavior is different.

AdamGleave:
> Overall, I like (1) the best because it fits seamlessly into existing code. My main worry is that it might be confusing that `.predict()` and `.forward()` behave differently by default. (The former normalizes rewards while the latter doesn't.) I propose addressing this by adding a `normalize_reward=True` parameter to `.predict()` and `normalize_reward=False` parameter to `.forward()`. This would let both methods normalize the rewards or not, but their default behavior is different.

I like the idea of adding extra arguments but this will break when the wrapper isn't applied. Same problem with `normalize()` and `unnormalize()`. I guess we could always wrap rewards with this in our codebase, and just toggle whether normalization is applied by default, or even make this a base feature of `RewardNet`, but both seem a bit nasty.

An alternative approach would be to break the abstraction layer a little and have `RewardVecEnvWrapper` check whether the function being passed in belongs to a normalization wrapper, and if so call it with `normalize_reward=True`, otherwise call it without. In this case `predict()` would still default to unnormalized matching `forward()`, but it'd be normalized when it matters. However, this would be a bit fragile (wrap the class again and this would break).

scottemmons:
It seems like there's tension between (i) letting the user control when the normalization is applied to `.predict()` and `.forward()`, and (ii) not wanting these controls to break when the wrapper is absent.

To resolve this tension, maybe it would work to have two separate wrappers. Both wrappers keep track of the normalization statistics on both `.predict()` and `.forward()`. But one wrapper applies normalization just to the output of `.predict()`, while the other wrapper applies normalization just to the output of `.forward()`.

AdamGleave:
> It seems like there's tension between (i) letting the user control when the normalization is applied to `.predict()` and `.forward()`, and (ii) not wanting these controls to break when the wrapper is absent.

Yes, this is a good way of putting it.

> To resolve this tension, maybe it would work to have two separate wrappers. Both wrappers keep track of the normalization statistics on both .predict() and .forward().

Innovative solution. I think adding a wrapper we don't use would be a bit confusing, but if the normalization logic is similar between the two of them, perhaps we could just have one wrapper and control which functions it normalizes by `__init__` parameters?

Another option is to add a `predict_unnormalized()` method that is pass-through to the raw `predict()`. This of course will also break when wrapper is absent, but I think it's at least more obvious that it will break, whereas keyword arguments feel a bit more insidious.

qxcv:
My 2c: if you really want things to break noisily, you could also ditch `.forward()` entirely and instead have a class with only `predict_normalized()` and `predict_unnormalized()` methods.

Otherwise: I agree with what is written above. From the perspective of writing idiomatic PyTorch, I feel that the cleanest approach would be to have a separate wrapper class that _always_ normalises reward on the forward pass. The policy training code would be given a wrapped model, and the reward learning code would be given an unwrapped model; the entire wrapped model (which contains the unwrapped model as a submodule) would be saved in checkpoints. This is analogous to the way that you might construct a classifier that outputs only logits so that you can apply `nn.LogSoftmax()` on top at train time and apply `nn.Softmax()` on top at test time. On the other hand, that would lead to the same bugs that happen with vecenv wrappers, so some variation on Adam's solution seems superior.

qxcv:
I think Scott's third solution, of having a `.normalized_mode()` method and an `.unnormalized_mode()` method, is also idiomatic PyTorch. It also has the advantage that it works well even when composing other modules on top of your reward model, since `.apply(enable_normalized_mode)` can be used to recursively set all reward models in a module hierarchy into normalized mode. On the other hand, it does not fail loudly when misused, as Scott pointed out.

AdamGleave:
> Otherwise: I agree with what is written above. From the perspective of writing idiomatic PyTorch, I feel that the cleanest approach would be to have a separate wrapper class that _always_ normalises reward on the forward pass. The policy training code would be given a wrapped model, and the reward learning code would be given an unwrapped model; the entire wrapped model (which contains the unwrapped model as a submodule) would be saved in checkpoints.

I like this in principle but it seems a bit tricky to actually implement. We could have `reward.make_reward_net` return an unnormalized and maybe-normalized version, and then have algorithms choose which one to use in appropriate places. 

I've got a new proposal: define a new abstract base class `RewardNetNormalized` that adds an abstract `predict_normalized` method. We then always wrap the `RewardNet` with either one that actually normalizes, or with a no-op that just passes through to `predict`. We pass `predict_normalized` into `RewardVecEnvWrapper`, and otherwise use the (always unnormalized) `predict`/`forward`.

It complicates the inheritance hierarchy a bit, but would require minimal changes to the algorithms (just change `predict` to `predict_normalized`) and seems fairly fool-proof: can't fail to serialize normalization statistics, or accidentally normalize things.

Main downside I see is it doesn't compose. But this seems like something you always want as the final layer? And if messed up it should fail at runtime and can probably be caught by static type analysis too.

qxcv:
I like that proposal! I agree that the lack of compositionality isn't a big deal. It would only matter if we wanted to do something really weird with the rewards later (like normalise and _then_ clip or something like that).

kmdanielduan:
> 1. Normalize the output of `.predict()` but not of `.forward()`. This is Adam's suggestion.
> 2. Normalize the output when in `.eval()` mode but not when in `.train()` mode. Adam mentioned this but is leaning against it.
> 3. Define a new normalization mode, e.g. that is toggled by `.normalize()` and `.unnormalize()`, which sets the wrapper to (un)normalize the rewards in both `.predict()` and `.forward()`.
> 4. Define a new abstract base class `RewardNetNormalized` that adds an abstract `predict_normalized` method. We then always wrap the `RewardNet` with either one that actually normalizes, or with a no-op that just passes through to predict. We pass `predict_normalized` into `RewardVecEnvWrapper`, and otherwise use the (always unnormalized) predict/forward.

I'm not a big fan of toggling `.normalize()` modes. I prefer the direction of (4) over others due to its readability and simplicity.

Some clarifying notes for where normalization is used: I searched through the codebase, and found there are three places where normalization of learnt rewards is needed.
1. For PreferenceComparisons, the status quo is to add an additional layer to reward_net, and use `reward_net.predict()` in [`RewardVecEnvWrapper`](https://github.com/HumanCompatibleAI/imitation/blob/ed45793dfdd897d3ac1f3a863a8816b56d436887/src/imitation/algorithms/preference_comparisons.py#L157).
2. For Adversarial, our current solution is to wrap `venv_train ` with [`VecNormalize`](https://github.com/HumanCompatibleAI/imitation/blob/ed45793dfdd897d3ac1f3a863a8816b56d436887/src/imitation/algorithms/adversarial/common.py#L221). `venv_train` is used to train generator and uses `predict()` to get reward signal. 
`forward()` is used in [`train_disc()`](https://github.com/HumanCompatibleAI/imitation/blob/ed45793dfdd897d3ac1f3a863a8816b56d436887/src/imitation/algorithms/adversarial/common.py#L327), and any reward postprocessing is applied within `logits_gen_is_high`. We only want to normalize the output of `predict()` but not `forward()`.
    - For AIRL, [`reward_train()`](https://github.com/HumanCompatibleAI/imitation/blob/ed45793dfdd897d3ac1f3a863a8816b56d436887/src/imitation/algorithms/adversarial/airl.py#L90) directly returns a `RewardNet`.
    - GAIL is the trickiest one. [`reward_train()`](https://github.com/HumanCompatibleAI/imitation/blob/ed45793dfdd897d3ac1f3a863a8816b56d436887/src/imitation/algorithms/adversarial/gail.py#L103) returns a [`LogSigmoidRewardNet`](https://github.com/HumanCompatibleAI/imitation/blob/ed45793dfdd897d3ac1f3a863a8816b56d436887/src/imitation/algorithms/adversarial/gail.py#L14) that modifies its own `forward()`, but its `predict()` remains the same.

It seems to me (4) is good in all cases, but it was pretty confusing to me at the beginning. Some additional comments might be helpful to clarify things.

src/imitation/scripts/common/reward.py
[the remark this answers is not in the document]
kmdanielduan: Sorry I was wrong about it. In GAIL, `train_gen()` and `train_disc()` used `reward_net.predict()` and `reward_net.forward()` respectively, and the normalization only applied to `reward_net.predict()`, so we are good.

src/imitation/scripts/common/reward.py
kmdanielduan: Thanks for pointing that out. I recently find out that you are right.

kmdanielduan:
After discussing with Adam, I made an attempt to restructure the inheritance hierarchy as the following.

- `RewardNet` ABC, with `predict_processed` concrete implementation that just calls `predict`.
- `BasicRewardNet` concrete implementation.
- `RewardNetWrapper` is ABC underneath `RewardNet`, taking base: `RewardNet` constructor argument and exposing base attribute
    - `ShapedRewardNet` is ABC, for adding shaping in general.
        - `BasicShapedRewardNet` concrete implementation using MLP for potential shaping.
    - `NormalizedRewardNet` is a concrete implementation, to implement `forward` as well as `predict_processed` that uses normalization layer given by normalize_class, like `NormalizeFeaturesExtractor`.
        - `BasicNormalizedRewardNet` is a concrete implementation using MLP as base net.

Would be great to receive some feedback on the high-level structures and implementations. If this is okay, I could further construct some test cases.

src/imitation/algorithms/adversarial/common.py
```diff
@@ -206,22 +204,23 @@ def __init__(
 
         venv = self.venv_buffering = wrappers.BufferingWrapper(self.venv)
 
+        if isinstance(self.reward_train, reward_nets.RewardNet):
```
AdamGleave, line 207: Why is this necessary? I think we can always use `predict_processed` as it's present in base `RewardNet`, just forwards to `predict` by default

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -142,7 +142,10 @@ def __init__(
         # will set self.logger, which also sets the logger for the algorithm
         super().__init__(custom_logger)
         if isinstance(reward_fn, reward_nets.RewardNet):
-            reward_fn = reward_fn.predict
```
AdamGleave, line 145: As with `adversarial/common.py` I do not see why this is necessary, think we can just use `predict_processed` always

src/imitation/rewards/reward_nets.py
```diff
@@ -142,6 +142,28 @@ def predict(
...
+        Its default behavior in RewardNet is to return the raw rewards from predict().
+
+        Args:
+            state: Current states of shape `(batch_size,) + state_shape`.
+            action: Actions of shape `(batch_size,) + action_shape`.
+            next_state: Successor states of shape `(batch_size,) + state_shape`.
+            done: End-of-episode (terminal state) indicator of shape `(batch_size,)`.
+
+        Returns:
+            Computed normalized rewards of shape `(batch_size,`).
```
AdamGleave, line 163:
```suggestion
            Computed processed rewards of shape `(batch_size,`).
```

src/imitation/rewards/reward_nets.py
```diff
@@ -142,6 +142,28 @@ def predict(
...
+        self,
+        state: np.ndarray,
+        action: np.ndarray,
+        next_state: np.ndarray,
+        done: np.ndarray,
+    ) -> np.ndarray:
+        """Compute the processed rewards for a batch of transitions without gradients.
+
+        Its default behavior in RewardNet is to return the raw rewards from predict().
+
```
AdamGleave, line 155:
```suggestion
Subclasses can override this to normalize or otherwise modify the rewards in
ways that may help RL training or other applications of the reward function.

```

src/imitation/rewards/reward_nets.py
```diff
@@ -163,78 +185,37 @@ def dtype(self) -> th.dtype:
             return th.get_default_dtype()
 
 
-class ShapedRewardNet(RewardNet):
-    """A RewardNet consisting of a base net and a potential shaping."""
+class RewardNetWrapper(RewardNet, abc.ABC):
```
AdamGleave, line 188: No need to inherit from `abc.ABC` as `RewardNet` already inherits from this. See https://stackoverflow.com/questions/48814211/python3-how-does-one-define-an-abstract-subclass-from-an-existing-abstract-cla

src/imitation/rewards/reward_nets.py
```diff
@@ -163,78 +185,37 @@ def dtype(self) -> th.dtype:
...
+    """A RewardNet wrapper with a base net."""
```
AdamGleave, line 189:
```suggestion
    """A RewardNet wrapping a base network."""
```

src/imitation/rewards/reward_nets.py
```diff
@@ -163,78 +185,37 @@ def dtype(self) -> th.dtype:
...
-            base: the base reward net to which the potential shaping
-                will be added.
-            potential: A callable which takes
-                a batch of states (as a PyTorch tensor) and returns a batch of
-                potentials for these states. If this is a PyTorch Module, it becomes
-                a submodule of the ShapedRewardNet instance.
-            discount_factor: discount factor to use for the potential shaping
+            observation_space: the observation space of the environment
+            action_space: the action space of the environment
+            base: a base RewardNet
```
AdamGleave, line 205:
```suggestion
            base: the base RewardNet to wrap.
```

src/imitation/rewards/reward_nets.py
```diff
@@ -265,7 +246,10 @@ def __init__(
             use_done: should the "done" flag be included as an input to the MLP?
             kwargs: passed straight through to `build_mlp`.
         """
-        super().__init__(observation_space, action_space)
+        super().__init__(
```
AdamGleave, line 249: why the reformatting?

src/imitation/rewards/reward_nets.py
```diff
@@ -318,6 +302,210 @@ def forward(self, state, action, next_state, done):
         return outputs
 
 
+class NormalizedRewardNet(RewardNetWrapper):
+    """A reward net that normalizes the output of its base net.
+
+    Only requires the implementation of a forward pass (calculating rewards given
```
AdamGleave, line 308: Huh? I think a forward pass is implemented in this class...

src/imitation/rewards/reward_nets.py
```diff
@@ -318,6 +302,210 @@ def forward(self, state, action, next_state, done):
...
+        return rew
+
+    def forward(
+        self,
+        state: th.Tensor,
+        action: th.Tensor,
+        next_state: th.Tensor,
+        done: th.Tensor,
+    ):
+        return self.base(state, action, next_state, done)
```
AdamGleave, line 374: Shouldn't this be?
```suggestion
        return self.base.forward(state, action, next_state, done)
```
(Maybe it's the same due to PyTorch magic...?)

src/imitation/rewards/reward_nets.py
```diff
@@ -163,78 +185,37 @@ def dtype(self) -> th.dtype:
...
         action_space: gym.Space,
         base: RewardNet,
-        potential: Callable[[th.Tensor], th.Tensor],
-        discount_factor: float,
         normalize_images: bool = True,
     ):
-        """Setup a ShapedRewardNet instance.
+        """A minimal abstract reward network wrapper with a base net.
+
+        A concrete implementation of forward() is needed.
```
AdamGleave, line 200: This belongs in class not constructor docstring.

src/imitation/rewards/reward_nets.py
AdamGleave: Can you also document what is the responsibility of subclasses to implement?

`forward()` definitely. But note by default `predict`, `preprocess`, `predict_processed`, `device` and all the Torch `nn.Module` methods will be inherited from `RewardNet` and *not* passed through to the `base`.

I think this actually works out OK most the time. E.g. `device` looks at parameters, but we get parameters of base network as Torch treats it like a submodule. `predict` calls `forward`, which subclass already needs to implement, etc.

But it's a bit fragile. If the base network had overrode `predict` for some reason, that will get lost as we don't forward to `base.predict`. So important users know this behaviour.

src/imitation/rewards/reward_nets.py
```diff
@@ -318,6 +302,210 @@ def forward(self, state, action, next_state, done):
...
+        """
+        super().__init__(
+            observation_space=observation_space,
+            action_space=action_space,
+            base=base,
+            normalize_images=normalize_images,
+        )
+        # assuming reward is always a scalar
+        self.rew_normalize_layer = (
+            rew_normalize_class(1) if rew_normalize_class else None
```
AdamGleave, line 340: I think `rew_normalize_class` should always be supplied (based on type annotation, and why wrap with a normalization layer if you don't want to normalize?) so can simplify this expression

src/imitation/rewards/reward_nets.py
```diff
@@ -318,6 +302,210 @@ def forward(self, state, action, next_state, done):
...
+        )
+
+    def predict_processed(
```
AdamGleave, line 343: This implementation looks like it should work but I think we can maybe come up with something a bit cleaner.

It's a shame you have to convert `rew` to Tensor `rew_th`, then go back again to NumPy. `predict` already has the tensor input, but it converts it to `NumPy` before returning.

Since we can make changes to base `RewardNet` class, one option is to just move most of `predict` into something like `predict_tensor` and then `predict` just calls `predict_tensor` and does the final conversion to NumPy: https://github.com/HumanCompatibleAI/imitation/blob/53d8ab17e4ef116695f73fcae673d1ddb79242ac/src/imitation/rewards/reward_nets.py#L141 This might actually be useful for other users anyway. WDYT?

src/imitation/rewards/reward_nets.py
```diff
@@ -318,6 +302,210 @@ def forward(self, state, action, next_state, done):
...
+            action: Actions of shape `(batch_size,) + action_shape`.
+            next_state: Successor states of shape `(batch_size,) + state_shape`.
+            done: End-of-episode (terminal state) indicator of shape `(batch_size,)`.
+
+        Returns:
+            Computed normalized rewards of shape `(batch_size,`).
+        """
+        rew = self.base.predict(state, action, next_state, done)
+        rew_th = th.as_tensor(rew, device=self.device)
+        rew = self.rew_normalize_layer(rew_th).detach().cpu().numpy().flatten()
```
AdamGleave, line 363: We want to be in eval mode when calling another layer, so can you use the `with networks.evaluating(self):` context manager here?

src/imitation/rewards/reward_nets.py
```diff
@@ -318,6 +302,210 @@ def forward(self, state, action, next_state, done):
...
+        self,
+        state: th.Tensor,
+        action: th.Tensor,
+        next_state: th.Tensor,
+        done: th.Tensor,
+    ):
+        return self.base(state, action, next_state, done)
+
+
+class BasicNormalizedRewardNet(NormalizedRewardNet):
```
AdamGleave, line 377: Hm, do we really need this class?

I was imagining we'd just have a flag for whether or not we want reward output to be normalized, and then the reward constructor in the Sacred config file applies the wrapper as appropriate.

Making a normalized version of every reward net will be a lot of bloat. (Admittedly `BasicShapedRewardNet` is already this anti-pattern...)

src/imitation/rewards/reward_nets.py
```diff
@@ -318,6 +302,210 @@ def forward(self, state, action, next_state, done):
...
+        super().__init__(
+            observation_space,
+            action_space,
+            base=base_reward_net,
+            rew_normalize_class=rew_normalize_class,
+        )
+
+
+class ShapedRewardNet(RewardNetWrapper):
+    """A RewardNet consisting of a base net and a potential shaping."""
```
AdamGleave, line 436:
```suggestion
    """A RewardNet consisting of a base net and potential shaping."""
```

src/imitation/rewards/reward_nets.py
```diff
@@ -367,6 +555,14 @@ def __init__(
             discount_factor: discount factor for the potential shaping.
             kwargs: passed straight through to `BasicRewardNet` and `BasicPotentialMLP`.
         """
+        # FIXME(yawen): why could the reward net and potential net use the same kwargs
```
AdamGleave, line 558: We do specify the hidden sizes of the reward & potential net separately. Most of the other parameters don't make sense to change from default at all anyway, but the ones I do see as being reasonable to configure (e.g. `activation`, `normalize_input_layer`) you probably do usually want shared between the networks.

So I think this is reasonable, although if you want to make it so you can specify different kwargs for each I'd be fine with that too so long as there's a way to specify kwargs shared for both.

src/imitation/scripts/common/reward.py
```diff
@@ -31,18 +32,30 @@ def normalize_input_running():
...
-        default_net = reward_nets.BasicRewardNet
+        default_net = reward_nets.BasicNormalizedRewardNet
         if command_name == "airl":
             default_net = reward_nets.BasicShapedRewardNet
         res["net_cls"] = default_net
     if "normalize_input_layer" not in config["reward"]["net_kwargs"]:
         res["net_kwargs"] = {"normalize_input_layer": networks.RunningNorm}
+    if "rew_normalize_class" not in config["reward"]["net_kwargs"]:
+        res["net_kwargs"] = {"rew_normalize_class": networks.RunningNorm}
     return res
```
AdamGleave, line 61: To elaborate a bit on what I mentioned in an earlier comment, I was imagining on adding something like an `output_normalize_class` config parameter (defaulting to None), modifying the below `make_reward_net` function to take that as a parameter and then if non-None wrapping the constructed reward network with the `NormalizedRewardNet` class before returning.

This seems cleaner, and right now I think you can't have a normalized shaped reward network for example, which is kind of unfortunate.

tests/rewards/test_reward_nets.py
```diff
@@ -30,6 +36,27 @@ def test_init_no_crash(env_name, reward_net_cls, reward_net_kwargs):
         reward_net_cls(env.observation_space, env.action_space, **reward_net_kwargs)
 
 
+@pytest.mark.parametrize("env_name", ENVS)
+@pytest.mark.parametrize(
+    "reward_net_kwargs",
+    [
+        {"rew_normalize_class": networks.RunningNorm},
```
AdamGleave, line 43: I think could write this as:
```python
{k: dict(rew_normalize_class=networks.RunningNorm, **v) for k, v in REWARD_NET_KWARGS.items()}
```
Better if you do want to inherit changes from `REWARD_NET_KWARGS`

tests/rewards/test_reward_nets.py
```diff
@@ -30,6 +36,27 @@ def test_init_no_crash(env_name, reward_net_cls, reward_net_kwargs):
...
+        {
+            "normalize_input_layer": networks.RunningNorm,
+            "rew_normalize_class": networks.RunningNorm,
+        },
+    ],
+)
+def test_init_normalized_reward_net_no_crash(env_name, reward_net_kwargs):
+    env = gym.make(env_name)
+    for i in range(3):
```
AdamGleave, line 52: I'm skeptical doing this three times really tests anything more (although I see we do this in an existing test too)

tests/test_scripts.py
```diff
@@ -139,11 +140,15 @@ def test_train_preference_comparisons_reward_norm(tmpdir, named_configs):
...
     elif "reward.normalize_output_disable" in named_configs:
-        assert run.config["reward"]["net_kwargs"]["normalize_output_layer"] is None
+        assert run.config["reward"]["net_kwargs"]["rew_normalize_class"] is nn.Identity
     else:
-        assert run.config["reward"]["net_kwargs"]["normalize_output_layer"] is not None
+        assert (
+            run.config["reward"]["net_kwargs"]["rew_normalize_class"] is not nn.Identity
+        )
     assert run.status == "COMPLETED"
     assert isinstance(run.result, dict)
```
AdamGleave, line 153: You could consider checking the scale of the RL algorithm return reported in `run.result` to see if reward normalization is/isn't applied. Might not be consistent enough, but would be more of an integration test. (Right now that config could be completely ignored and this would still pass..)

src/imitation/algorithms/adversarial/airl.py
```diff
@@ -92,7 +92,7 @@ def reward_train(self) -> reward_nets.RewardNet:
 
     @property
     def reward_test(self) -> reward_nets.RewardNet:
-        if isinstance(self._reward_net, reward_nets.ShapedRewardNet):
-            return self._reward_net.base
-        else:
-            return self._reward_net
+        reward_net = self._reward_net
+        while isinstance(reward_net, reward_nets.RewardNetWrapper):
```
AdamGleave, line 96: This is now sufficiently complex (& error-prone) I think it deserves a one-line comment explaining it

AdamGleave (commented):
Thanks for the implementation! High-level structure looks good to me, there are some areas where I've suggested more nuts-and-bolts changes.

Further test cases would be good but can wait until you've settled on implementation details (unless you feel like doing TDD).

src/imitation/rewards/reward_nets.py
kmdanielduan: Okay. A class only becomes concrete when there are no more `abstractmethod` and `abstractproperty` objects left in the implementation.

src/imitation/scripts/common/reward.py
kmdanielduan: Yeah, I think this makes sense.

src/imitation/rewards/reward_nets.py
kmdanielduan: Yeah. It is [PyTorch magic](https://github.com/pytorch/pytorch/blob/ad88816c864c04b478009eec3964d9210291a7a7/torch/nn/modules/module.py#L1175). It was [suggested](https://stackoverflow.com/questions/58508190/in-pytorch-what-is-the-difference-between-forward-and-an-ordinary-method) not to call `forward()`, but to call the layer itself.

src/imitation/rewards/reward_nets.py
kmdanielduan: deleted!

src/imitation/rewards/reward_nets.py
kmdanielduan: Makes sense.

src/imitation/rewards/reward_nets.py
kmdanielduan: If we decide to wrap RewardNet in reward.make_reward_net(), then it's not necessary to have this class.

src/imitation/rewards/reward_nets.py
kmdanielduan: make sense

src/imitation/rewards/reward_nets.py
kmdanielduan: make sense.

src/imitation/rewards/reward_nets.py
kmdanielduan: Seems reasonable. I was mostly worried about `kwargs` fed to `.build_mlp_kwargs` might contain unwanted arguments like `normalize_output_layer`. But now the problem doesn't exist because we moved the initialization of normalize_output_layer to `reward.make_reward_net`.

src/imitation/algorithms/adversarial/airl.py
kmdanielduan: `Recursively return the base network of the wrapped reward net`. Does this seem clear enough?

tests/rewards/test_reward_nets.py
kmdanielduan: Now I could directly add it to `REWARD_NET_KWARGS`.

src/imitation/algorithms/preference_comparisons.py
kmdanielduan: Agreed. It would be good to just simply this one.

src/imitation/algorithms/adversarial/airl.py
AdamGleave: It's clear what it does, but why do we do that? The point of reward_test is to get the unshaped version of the reward so I think there should be some mention of that.

kmdanielduan: would be better to include it in the docstring

tests/test_scripts.py
kmdanielduan: I think it would be better to test normalization somewhere else. `run.result` only records `reward_loss`, `reward_accuracy`, and non-wrapped `rollout` stats. This test aims to check the type of normalization applied is right under named_configs in `reward` ingredient.

kmdanielduan:
Hey, I just finished wrapping up the latest changes. Following the previous high-level changes, the current code structure is like this:
- `RewardNet` ABC, with `predict_processed` concrete implementation that just calls `predict`.
    - `BasicRewardNet` concrete implementation.
- `RewardNetWrapper` is ABC underneath `RewardNet`, taking base: `RewardNet` constructor argument and exposing base attribute
    - `ShapedRewardNet` is ABC, for adding shaping in general.
        - `BasicShapedRewardNet` concrete implementation using MLP for potential shaping.
- `NormalizedRewardNet` is a concrete implementation, to implement `forward` as well as `predict_processed` that uses normalization layer given by normalize_class, like `NormalizeFeaturesExtractor`.

The current issue is that we want the normalization layer in `NormalizedRewardNet` to only update the normalization stats during RL training, but not in `generate_trajectories` for evaluation (e.g. in `train.eval_policy` or `rollout.generate_trajectories`).
The current implementation introduced a `venv_train` and `venv_eval` in train_preference_comparisons to distinguish different use cases. They both wrap the `venv_buffering` using `RewardVecEnvWrapper` but with different reward_fn, one enables update_stat, and the other does not.
An alternative implementation is to introduce a context manager to explicitly turn on or off the evaluating mode for the normalization layer, but we thought it's not necessary to have such complexity for this case.

src/imitation/algorithms/adversarial/common.py
AdamGleave: I think we want to just delete this `isinstance` check? `self.reward_train` type annotation says it's always a `RewardNet` so it should always pass. And the current code does not handle the case where this check does not pass (leaves `reward_fn` undefined).

In fact I think we can probably remove the following line defining `reward_fn` and just inline it into the only place it's used, on line 217.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -142,8 +143,19 @@ def __init__(
...
+            if isinstance(reward_fn, reward_nets.NormalizedRewardNet):
```
AdamGleave, line 146: Why do we need to check this? If it's not a `NormalizedRewardNet`, it's harmless to have the partial application with `update_norm_stats=False`, it'll just be a no-op.

AdamGleave: Ah I see you only added `update_norm_stats` to `NormalizedRewardNet.predict_processed` not the base version. That makes sense, it is as written quite specific to normalization. But maybe we can just make the keyword argument more generic, e.g. `update_stats` or `is_rl`?

It's generally a bit error prone to have code depend on the types. If we add some new wrapper that implements its own `predict_processed`, it won't be obvious that we need to update preference comparisons to handle it.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -154,16 +166,26 @@ def __init__(
...
+            self.buffering_wrapper,
+            reward_fn=self.reward_fn_eval,
+        )
+        self.log_callback = self.venv_train.make_log_callback()
+
+        # venv_train should be used in self.algorithm.learn()
+        self.algorithm.set_env(self.venv_train)
 
-        self.algorithm.set_env(self.venv)
+        # TODO(yawen) Among all the attributes of venv, only venv.action_space is
```
AdamGleave, line 182: Huh? Surely the observation spaces need to match too (policy mapping from observation to a distribution over actions)?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -154,16 +166,26 @@ def __init__(
...
         # rewards.
         self.buffering_wrapper = wrappers.BufferingWrapper(venv)
-        self.venv = reward_wrapper.RewardVecEnvWrapper(
+        self.venv_train = reward_wrapper.RewardVecEnvWrapper(
             self.buffering_wrapper,
-            self.reward_fn,
+            reward_fn=self.reward_fn,
         )
-        self.log_callback = self.venv.make_log_callback()
+        self.venv_eval = reward_wrapper.RewardVecEnvWrapper(
```
AdamGleave, line 173: Looking at this more closely, do we actually need `venv_eval`? The only time we interact with the environment is https://github.com/HumanCompatibleAI/imitation/blob/6e3919b259ab75fe8aecc7906f9b238a5341d546/src/imitation/algorithms/preference_comparisons.py#L259 and https://github.com/HumanCompatibleAI/imitation/blob/6e3919b259ab75fe8aecc7906f9b238a5341d546/src/imitation/algorithms/preference_comparisons.py#L281 Critically, in both cases we don't make use of the reward in any way. So I think we could get away with just using the underlying `self.buffering_wrapper`. If so this would technically eliminate the need for even having `update_norm_stats` flag, although we might want to retain that feature anyway. What do you think?

src/imitation/rewards/reward_nets.py
```diff
@@ -138,9 +137,55 @@ def predict(
...
+    def predict_processed(
+        self,
+        state: np.ndarray,
+        action: np.ndarray,
+        next_state: np.ndarray,
+        done: np.ndarray,
+    ) -> np.ndarray:
+        """Compute the processed rewards for a batch of transitions without gradients.
+
+        Default to call `predict`. Subclasses can override this to normalize or
```
AdamGleave, line 175:
```suggestion
        Defaults to calling `predict`. Subclasses can override this to normalize or
```

src/imitation/rewards/reward_nets.py
```diff
@@ -163,78 +208,35 @@ def dtype(self) -> th.dtype:
             return th.get_default_dtype()
 
 
-class ShapedRewardNet(RewardNet):
-    """A RewardNet consisting of a base net and a potential shaping."""
+class RewardNetWrapper(RewardNet):
+    """An abstract RewardNet wrapping a base network.
+
+    A concrete implementation of `forward` method is needed.
```
AdamGleave, line 214:
```suggestion
    A concrete implementation of the `forward` method is needed.
```

src/imitation/rewards/reward_nets.py
```diff
@@ -163,78 +208,35 @@ def dtype(self) -> th.dtype:
...
+    Note: by default, `predict`, `predict_th`, `preprocess`, `predict_processed`,
+    `device` and all the PyTorch `nn.Module` methods will be inherited from `RewardNet`
+    and not passed through to the base network. If any of these methods is overridden
+    in the base `RewardNet`, it will be not inherited from `RewardNetWrapper`.
```
AdamGleave, line 218: Inherited has a technical meaning here that isn't appropriate (as it's a composition relationship)
```suggestion
    in the base `RewardNet`, this will not affect `RewardNetWrapper`.
```

src/imitation/rewards/reward_nets.py
```diff
@@ -293,6 +295,10 @@ def __init__(
                 # we do not want these overridden
                 "in_size": combined_size,
                 "out_size": 1,
+                # FIXME(yawen): by default, the reward output is squeezed to produce
```
AdamGleave, line 298: Duplicate of below

src/imitation/rewards/reward_nets.py
```diff
@@ -318,6 +324,134 @@ def forward(self, state, action, next_state, done):
...
+    ):
+        """Initialize the NormalizedRewardNet.
+
+        Args:
+            base: a base RewardNet
+            normalize_output_layer: The class to use to normalize rewards. This
+                can be any nn.Module that preserves the shape; e.g. `nn.Identity`,
+                `nn.LayerNorm`, or `networks.RunningNorm`.
+        """
+        # FIXME(yawen): by default, the reward output is squeezed to produce
```
AdamGleave, line 343: That's annyoing but I'm skeptical about BatchNorm for output normalization, so does not seem like a priority to fix. In particular RL training with every batch of mean zero reward is odd. I think it might still work: if the batch contains both a good and a bad action, gradients will point in the right direction. But any batch-to-batch difference is destroyed by this.

src/imitation/rewards/reward_nets.py
```diff
@@ -318,6 +324,134 @@ def forward(self, state, action, next_state, done):
...
+        update_norm_stats: bool = True,
+    ) -> np.ndarray:
+        """Compute normalized rewards for a batch of transitions without gradients.
+
+        Args:
+            state: Current states of shape `(batch_size,) + state_shape`.
+            action: Actions of shape `(batch_size,) + action_shape`.
+            next_state: Successor states of shape `(batch_size,) + state_shape`.
+            done: End-of-episode (terminal state) indicator of shape `(batch_size,)`.
+            update_norm_stats: Whether to update the running stats of the norm layer.
```
AdamGleave, line 366: Do we actually need this? Originally I thought we did as we had a use-case in preference comparison where, during training a reward learning algorithm, we wanted to call `predict_processed` but not update the stats (when gathering extra samples). But I think that's not needed: since we don't use the learned reward then (*only* for RL training), we can just use the `venv` *before* wrapping.

So now the only use case I can see where we want to call `predict_processed` but not update the stats is for actually evaluating the learned policy. But then it seems natural to put the network into evaluation mode, in which case the stats won't be updated.

So, I now lean towards getting rid of this flag, unless there's another use case in mind for it.

tests/rewards/test_reward_nets.py
```diff
@@ -156,3 +180,62 @@ def forward(self):
...
+@pytest.mark.parametrize("reward_net_cls", [reward_nets.BasicRewardNet])
+@pytest.mark.parametrize("normalize_input_layer", [networks.RunningNorm])
+@pytest.mark.parametrize("normalize_output_layer", [networks.RunningNorm])
+def test_training_regression(
+    reward_net_cls,
+    normalize_input_layer,
+    normalize_output_layer,
+):
+    """Test reward_net normalization by training a regression model."""
+    # TODO(yawen):
```
AdamGleave, line 194: ?

tests/rewards/test_reward_nets.py
```diff
@@ -156,3 +180,62 @@ def forward(self):
...
+            normalize_output_layer,
+        )
+
+    # Construct a loss function and an Optimizer.
+    criterion = th.nn.MSELoss(reduction="sum")
+    optimizer = th.optim.SGD(reward_net.parameters(), lr=1e-6)
+
+    for _ in range(20):
+        # Forward pass: Compute predicted y by passing x to the model
+        random = base.RandomPolicy(venv.observation_space, venv.action_space)
```
AdamGleave, line 213: Can move this out of the loop

tests/rewards/test_reward_nets.py
```diff
@@ -156,3 +180,62 @@ def forward(self):
...
+        loss = criterion(
+            th.as_tensor(transitions.rews, device=reward_net.device),
+            rews_th,
+        )
+
+        rews_predict = reward_net.predict(*trans_args)
+        rews_processed = reward_net.predict_processed(*trans_args)
+
+        assert (rews == rews_predict).all()
+        assert not (rews_processed == rews_predict).all()
```
AdamGleave, line 236: Isn't this only true when we have output normalization turned on?

tests/rewards/test_reward_nets.py
```diff
@@ -156,3 +180,62 @@ def forward(self):
...
+        # Compute and print loss
+        loss = criterion(
+            th.as_tensor(transitions.rews, device=reward_net.device),
+            rews_th,
+        )
+
+        rews_predict = reward_net.predict(*trans_args)
+        rews_processed = reward_net.predict_processed(*trans_args)
+
+        assert (rews == rews_predict).all()
```
AdamGleave, line 235: Is this actually true? The input normalization stats might have changed.

AdamGleave: Oh, nvm, this does work because we update stats *before* normalizing and only do that for forward not predict.

src/imitation/rewards/reward_nets.py
```diff
@@ -105,17 +105,16 @@ def preprocess(
...
         return state_th, action_th, next_state_th, done_th
 
-    def predict(
+    def predict_th(
         self,
         state: np.ndarray,
         action: np.ndarray,
         next_state: np.ndarray,
         done: np.ndarray,
     ) -> np.ndarray:
```
yulonglin, line 114: `) -> th.Tensor:`

AdamGleave: Yeah good point. Im sad `pytype` didn't complain about this ;(

src/imitation/algorithms/adversarial/common.py
kmdanielduan: Good point!

src/imitation/algorithms/preference_comparisons.py
kmdanielduan: Good point! I wasn't even aware of this. The `TrajectoryGenerator` directly get trajectories from `self.buffering_wrapper`, there is no need to wrap it with a `RewardVecEnvWrapper` at all!

I think generally it's useful to keep that feature. By default people can still do `rollout.generate_trajectories(policy, rew_wrapped_env)` without noticing the effects on the norm stats! I can also imagine people want to use `rollout.generate_trajectories(policy, rew_wrapped_env)` to generate trajectories with learned rewards, for analysis purpose.

I think it's fine to just use `buffering_wrapper` to better signal the sampled trajectories contain the original rewards. I also wish to refactor `self.buffering_wrapper` as `self.buffer_wrapped_venv` or `self.buffering_venv`.

tests/rewards/test_reward_nets.py
```diff
@@ -182,60 +182,54 @@ def forward(self):
...
     criterion = th.nn.MSELoss(reduction="sum")
-    optimizer = th.optim.SGD(reward_net.parameters(), lr=1e-6)
+    optimizer = th.optim.SGD(norm_rew_net.parameters(), lr=1e-6)
 
-    for _ in range(20):
-        # Forward pass: Compute predicted y by passing x to the model
-        random = base.RandomPolicy(venv.observation_space, venv.action_space)
+    # Getting transitions from a random policy
+    random = base.RandomPolicy(venv.observation_space, venv.action_space)
+    for i in range(2):
```
AdamGleave, line 205:
```suggestion
    for _ in range(2):
```
i is not used AFAICT

AdamGleave (approved):
LGTM (one very small suggestion)

kmdanielduan:
> LGTM (one very small suggestion). Thanks for seeing this through, bit of a marathon PR in the end!

I really appreciate everyone's ideas and replies @AdamGleave @scottemmons @qxcv @yulonglin, especially @AdamGleave's code review in great detail and patience! I also learned a lot!

PR #426, opened 2022-03-29
Remove dead reward input normalization code

AdamGleave:
Some code no longer needed after https://github.com/HumanCompatibleAI/imitation/pull/393 that we forgot to remove.

I'm tempted to delete `common.py` entirely as it's now just a single type alias. But it is used in several different modules in `imitation.rewards.*` so it doesn't make sense to relocate it to any of those in particular. We could move it into `imitation/rewards/__init__.py` but I tend to dislike sticking things in `__init__` (can make it a bit hard to follow where it's actually defined)

### DIFF
--- a/src/imitation/rewards/common.py
+++ b/src/imitation/rewards/common.py
@@ -3,52 +3,5 @@
 from typing import Callable
 
 import numpy as np
-from stable_baselines3.common import vec_env
 
 RewardFn = Callable[[np.ndarray, np.ndarray, np.ndarray, np.ndarray], np.ndarray]
-
-
-def build_norm_reward_fn(
-    *,
-    reward_fn: RewardFn,
-    vec_normalize: vec_env.VecNormalize,
-    norm_reward: bool = True,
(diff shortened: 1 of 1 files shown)


ernestum:
LGTM!

I am all for deleting `common.py`. But maybe that does not have to be part of this simple PR?

I also looked a bit into the rewards corner of the lib and I would propose the following refactoring (which would also get rid of `common.py`):
1. Turn `RewardFn` into an abstract class. Then it does not just define the signature of a reward function but also the argument names and a docstring explaining the semantics of the function call.
2. Rename `common.py` -> `reward_fn.py`
3. Make all the reward networks inherit from `RewardFn`. Turn their `.predict()` into `__call__()` (or make the `__call__()` execute predict

ernestum (approved):

AdamGleave:
I like those suggestions, I've opened https://github.com/HumanCompatibleAI/imitation/issues/427 to track it

PR #429, opened 2022-04-18
Remove mean_return helper and replace by evaluate_policy in SB3

ernestum:
Just a small PR to gradually reduce the number of helpers and utils that we have to maintain ourselves.

### DIFF
--- a/tests/algorithms/test_bc.py
+++ b/tests/algorithms/test_bc.py
@@ -5,7 +5,7 @@ import os
 
 import pytest
 import torch as th
-from stable_baselines3.common import vec_env
+from stable_baselines3.common import evaluation, vec_env
 from torch.utils import data as th_data
 
 from imitation.algorithms import bc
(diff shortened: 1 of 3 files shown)


tests/algorithms/test_bc.py
```diff
@@ -94,15 +94,14 @@ def test_train_end_cond_error(trainer: bc.BC):
...
-    sample_until = rollout.make_min_episodes(15)
-    novice_ret_mean = rollout.mean_return(trainer.policy, cartpole_venv, sample_until)
+    novice_ret_mean, _ = evaluation.evaluate_policy(trainer.policy, cartpole_venv, 15)
+
     trainer.train(
         n_epochs=1,
         on_epoch_end=lambda: print("epoch end"),
         on_batch_end=lambda: print("batch end"),
     )
-    trainer.train(n_batches=10)
```
AdamGleave, line 104: why is this line being deleted?

AdamGleave (approved):
LGTM apart from deleting one line (premature removal of n_batches from test?)

tests/algorithms/test_bc.py
ernestum: That was an accident. I will add it back in.

AdamGleave (approved):
LGTM

PR #430, opened 2022-04-18
Use permutations tests when comparing episode rewards

ernestum:
Right now we are comparing just the mean episode rewards between episodes from untrained and from trained policies to ensure that training is successful.
This is not a proper statistical test to ensure, that there was a certain training effect.
This PR introduces a proper permutation test to compute and ensure the significance of the change induced by training.

Note: right now this branch starts off the `remove_mean_return_helper`. It will be rebased on master as soon as that PR is merged.

### DIFF
--- a/src/imitation/testing/reward_improvement.py
+++ b/src/imitation/testing/reward_improvement.py
@@ -5,7 +5,7 @@ import numpy as np
 from scipy import stats
 
 
-def rewards_improved(
+def is_significant_reward_improvement(
     old_rewards: Iterable[float],
     new_rewards: Iterable[float],
     p_value: float = 0.05,
@@ -25,13 +25,13 @@ def rewards_improved(
         For this, the probability, that the old rewards are just as good as the new
         rewards must be below `p_value`.
 
-    >>> rewards_improved((5, 6, 7, 4, 4), (7, 5, 9, 9, 12))
+    >>> is_significant_reward_improvement((5, 6, 7, 4, 4), (7, 5, 9, 9, 12))
     True
 
-    >>> rewards_improved((5, 6, 7, 4, 4), (7, 5, 9, 7, 4))
+    >>> is_significant_reward_improvement((5, 6, 7, 4, 4), (7, 5, 9, 7, 4))
     False
 
-    >>> rewards_improved((5, 6, 7, 4, 4), (7, 5, 9, 7, 4), p_value=0.3)
+    >>> is_significant_reward_improvement((5, 6, 7, 4, 4), (7, 5, 9, 7, 4), p_value=0.3)
     True
     """
     permutation_test_result = stats.permutation_test(
--- a/src/imitation/testing/reward_improvement.py
+++ b/src/imitation/testing/reward_improvement.py
@@ -1,4 +1,4 @@
-"""Utility functions used to write tests."""
+"""Utility functions used to check if rewards improved wrt to previous rewards."""
 from typing import Iterable
 
 import numpy as np
--- a/src/imitation/testing/reward_improvement.py
+++ b/src/imitation/testing/reward_improvement.py
@@ -42,3 +42,28 @@ def is_significant_reward_improvement(
     )
 
     return permutation_test_result.pvalue < p_value
+
+
+def mean_reward_improved_by(
+    old_rewards: Iterable[float],
+    new_rewards: Iterable[float],
+    min_improvement: float,
+):
+    """Checks if mean rewards improved wrt. to old rewards by a certain amount.
+
+    Args:
+        old_rewards: Iterable of "old" trajectory rewards (e.g. before training).
+        new_rewards: Iterable of "new" trajectory rewards (e.g. after training).
+        min_improvement: The minimum amount of improvement that we expect.
+
+    Returns:
+        True if the mean of the new rewards is larger than the mean of the old rewards
+        by min_improvement.
+
+    >>> mean_reward_improved_by([5, 8, 7], [8, 9, 10], 2)
+    True
+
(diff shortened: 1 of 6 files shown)


tests/utils.py
```diff
@@ -0,0 +1,44 @@
...
+import numpy as np
+from scipy import stats
+
+
+def rewards_improved(
+    old_rewards: Iterable[float],
+    new_rewards: Iterable[float],
+    p_value: float = 0.05,
+) -> bool:
+    """Checks if that the new rewards are really better than the old rewards.
```
AdamGleave, line 13:
```suggestion
    """Checks if the new rewards are really better than the old rewards.
```

tests/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Ensures, that this is not just due to lucky sampling by a permutation test.
```
AdamGleave, line 15:
```suggestion
    Ensures that this is not just due to lucky sampling by a permutation test.
```

tests/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Args:
+        old_rewards: Iterable of "old" trajectory rewards (e.g. before training).
+        new_rewards: Iterable of "new" trajectory rewards (e.g. after training).
+        p_value: The maximum probabliity, that the old rewards are just as good as the
```
AdamGleave, line 20:
```suggestion
        p_value: The maximum probability, that the old rewards are just as good as the
```

tests/algorithms/test_bc.py
```diff
@@ -5,7 +5,8 @@
 
 import pytest
 import torch as th
-from stable_baselines3.common import vec_env
+import utils
```
AdamGleave, line 8: I'm a bit worried about this top-level import -- is this officially supported by `pytest`, can we depend on `sys.pythonpath` being set to include `tests/`?

https://stackoverflow.com/questions/33508060/create-and-import-helper-functions-in-tests-without-creating-packages-in-test-di has some other ideas on how to share helper methods between tests. I think the helper class in `conftest` might be slightly better? But it is also quite hacky, really.

AdamGleave: It's also a bit odd this is in the "third-party package" section of imports, despite being local... should probably add it to `known_first_party` in `isort` if we do stick with this.

AdamGleave (commented):
Thanks for the PR, definite improvement over just comparing means.

I'm a little uneasy on having top-level `import utils` to get helper methods, lean slightly towards adding as fixture in `conftest`, but if you're confident this is supported by `pytest` OK with sticking with current approach.

Do we just want to check that return improved with p>0.95, or that the improvement was *greater* than some threshold with p>0.95? In particular I'm worried the DAgger test might now be weaker (was previously checking means differed by >300...)

tests/algorithms/test_bc.py
AdamGleave: I agree putting it in `src/imitation/testing/` would be a good solution, it is a helpful utility.

ernestum:
> Do we just want to check that return improved with p>0.95, or that the improvement was _greater_ than some threshold with p>0.95? In particular I'm worried the DAgger test might now be weaker (was previously checking means differed by >300...)

You are right, that we now only check for p>0.95 and not for an absolute amount of improvement any more.
I would push tracking changes in algorithm performance in the realm of benchmarking/testing that the benchmarks stay roughly the same among commits. So instead of checking that the improvement is larger than some magic number, I would prefer comparing the episode rewards from a previous benchmark with episode rewards from the current benchmark using a permutation test again. This should be far less flaky. What do you think? @AdamGleave

AdamGleave:
> You are right, that we now only check for p>0.95 and not for an absolute amount of improvement any more. I would push tracking changes in algorithm performance in the realm of benchmarking/testing that the benchmarks stay roughly the same among commits. So instead of checking that the improvement is larger than some magic number, I would prefer comparing the episode rewards from a previous benchmark with episode rewards from the current benchmark using a permutation test again. This should be far less flaky. What do you think? @AdamGleave

I agree this is better but also let's not have the best be the enemy of the good. It will take some time for us to develop the benchmarks. In the meantime something like doing a permutation test between novice reward vs (trained reward - threshold) would be a more stringent but still hopefully non-flaky test with e.g. `threshold=100`, and we can just remove that later on.

I do not feel particularly strongly about this however.

tests/algorithms/test_bc.py
ernestum: done!

ernestum:
> I agree this is better but also let's not have the best be the enemy of the good. It will take some time for us to develop the benchmarks. In the meantime something like doing a permutation test between novice reward vs (trained reward - threshold) would be a more stringent but still hopefully non-flaky test [...]

You are right. I added the comparison again.

Also after the rebase to master with the new python version the test run.

src/imitation/testing/reward_improvement.py
```diff
@@ -0,0 +1,69 @@
...
+) -> bool:
+    """Checks if the new rewards are really better than the old rewards.
+
+    Ensures that this is not just due to lucky sampling by a permutation test.
+
+    Args:
+        old_rewards: Iterable of "old" trajectory rewards (e.g. before training).
+        new_rewards: Iterable of "new" trajectory rewards (e.g. after training).
+        p_value: The maximum probability, that the old rewards are just as good as the
+            new reawards, that we tolerate.
```
AdamGleave, line 21:
```suggestion
            new rewards, that we tolerate.
```

AdamGleave (approved):
LGTM apart from one small typo (suggested change)

PR #434, opened 2022-05-06
Remove redundant component of Xdummy-entrypoint.py

AdamGleave:
`Xdummy-entrypoint.py` tried to launch Xdummy in two ways: via a shell script (which no longer exists, and fails silently as output redirected to /dev/null) and directly via `Xorg` (which succeeds). This PR removes the silently failing part of it.

### DIFF
--- a/ci/Xdummy-entrypoint.py
+++ b/ci/Xdummy-entrypoint.py
@@ -30,11 +30,6 @@ if __name__ == "__main__":
             ":0",
         ]
     )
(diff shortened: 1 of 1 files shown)


PavelCz (approved):
Script works fine without that component.

PavelCz (approved):

PavelCz:
image
Guess my approval is not enough, since I don't have write access

PR #439, opened 2022-05-25
Refactored serialize.py to allow for dezerialization of normalized reward functions

levmckinney:
Within `serialize.py`, I've broke down `_load_reward_net_as_fn` into a series of composable functions that should allow for more flexibility in loading reward functions in the future. I have maintained the public interface of `load_reward` to avoid a breaking change. Finally, I've added a new reward function type that can be de-serialized `RewarNet_normalized` and `RewardNet_unnormalized`.  The former allows the user to maintain the normalization learned during training while the latter strips off the normalization wrapper and returns the underlying unnormalized reward function. In both cases the reward function will stop updating its normalization statistics after it is de-serialized.

### DIFF
--- a/src/imitation/rewards/serialize.py
+++ b/src/imitation/rewards/serialize.py
@@ -1,12 +1,18 @@
 """Load serialized reward functions of different types."""
 
-from typing import Callable
+from typing import Callable, List, Type
 
 import numpy as np
 import torch as th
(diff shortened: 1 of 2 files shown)


AdamGleave:
Thanks for the PR! I've given you push access to this repo, can you move your branch over to this repo so that the tests run?

Issue #437, opened 2022-05-10
`train_rl.py` depending on `rewards.serialize` still uses unnormalized reward


yawen-d:
## Background

#425 applied changes to save a `NormalizedRewardNet` containing an additional normalized layer and its norm statistics. The current state of `imitation.rewards.reward_nets` is as follows:

- `RewardNet` ABC, with `predict_processed` concrete implementation that just calls `predict`.
    - `BasicRewardNet` concrete implementation.
- `RewardNetWrapper` is ABC underneath `RewardNet`, taking `base`: `RewardNet` constructor argument and exposing base attribute.
    - `ShapedRewardNet` is ABC, for adding shaping in general.
        - `BasicShapedRewardNet` concrete implementation using MLP for potential shaping.
- `NormalizedRewardNet` is a concrete implementation, to implement `forward` as well as `predict_processed` that uses normalization layer given by normalize_class, like `NormalizeFeaturesExtractor`.

In `NormalizedRewardNet`, there are three methods that can output reward.
1. [`forward()`](https://github.com/HumanCompatibleAI/imitation/blob/f332680115380ce654cfd061fbaaa7bc47918d2c/src/imitation/rewards/reward_nets.py#L378) returns unnormalized reward (torch.Tensor) with gradient.
2. [`predict()`](https://github.com/HumanCompatibleAI/imitation/blob/f332680115380ce654cfd061fbaaa7bc47918d2c/src/imitation/rewards/reward_nets.py#L143) returns unnormalized reward (numpy.array) without gradient.
3. [`predict_processed()`](https://github.com/HumanCompatibleAI/imitation/blob/f332680115380ce654cfd061fbaaa7bc47918d2c/src/imitation/rewards/reward_nets.py#L347) returns normalized reward (numpy.array). Note that `predict_processed()` contains a `update_stats` (default to be `True`) argument to determine whether to update the norm stat or not.

[...]

PR #440, opened 2022-05-25
Refactored serialize.py to allow for de-serialization of normalized reward functions

levmckinney:
Within `serialize.py`, I've broke down `_load_reward_net_as_fn` into a series of composable functions that should allow for more flexibility in loading reward functions in the future. I have maintained the public interface of `load_reward` to avoid a breaking change. Finally, I've added a new reward function type that can be de-serialized `RewarNet_normalized` and `RewardNet_unnormalized`.  The former allows the user to maintain the normalization learned during training while the latter strips off the normalization wrapper and returns the underlying unnormalized reward function. In both cases the reward function will stop updating its normalization statistics after it is de-serialized.

### DIFF
--- a/experiments/transfer_learn_benchmark.sh
+++ b/experiments/transfer_learn_benchmark.sh
@@ -87,7 +87,7 @@ parallel -j 25% --header : --results "${LOG_ROOT}/parallel/" --colsep , --progre
   with \
   '{env_config_name}' seed='{seed}' \
   common.log_dir="${LOG_ROOT}/${ALGORITHM}/{env_config_name}_{seed}/n_expert_demos_{n_expert_demos}" \
-  reward_type="RewardNet_shaped" \
+  reward_type="RewardNet_unshaped" \
   reward_path="${REWARD_MODELS_DIR}/${ALGORITHM}/{env_config_name}_0/n_expert_demos_{n_expert_demos}/checkpoints/final/reward_test.pt" \
   "${extra_configs[@]}" \
   :::: ${CONFIG_CSV} \
--- a/src/imitation/rewards/serialize.py
+++ b/src/imitation/rewards/serialize.py
@@ -1,12 +1,18 @@
 """Load serialized reward functions of different types."""
 
-from typing import Callable
+from typing import Callable, List, Type
 
 import numpy as np
 import torch as th
 from stable_baselines3.common.vec_env import VecEnv
 
 from imitation.rewards import common
+from imitation.rewards.reward_nets import (
+    NormalizedRewardNet,
+    RewardNet,
+    RewardNetWrapper,
+    ShapedRewardNet,
+)
 from imitation.util import registry, util
 
 # TODO(sam): I suspect this whole file can be replaced with th.load calls. Try
@@ -17,33 +23,57 @@ RewardFnLoaderFn = Callable[[str, VecEnv], common.RewardFn]
 reward_registry: registry.Registry[RewardFnLoaderFn] = registry.Registry()
 
 
-def _load_reward_net_as_fn(shaped: bool) -> RewardFnLoaderFn:
-    def loader(path: str, venv: VecEnv) -> common.RewardFn:
-        """Load train (shaped) or test (not shaped) reward from path."""
-        del venv  # Unused.
-        net = th.load(str(path))
-        if not shaped and hasattr(net, "base"):
-            # If the "base" attribute exists, we are dealing with a ShapedRewardNet
-            # and will disable the potential shaping (if shaped is False).
-            # If no "base" attribute exists, we seem to use an unshaped RewardNet
-            # anyway, so we just use its predict() method directly.
-            reward = net.base.predict
-        else:
-            reward = net.predict
+def _validate_reward(reward: common.RewardFn) -> common.RewardFn:
+    """Add sanity check to the reward function for dealing with VecEnvs."""
 
-        def rew_fn(
-            obs: np.ndarray,
-            act: np.ndarray,
-            next_obs: np.ndarray,
-            dones: np.ndarray,
-        ) -> np.ndarray:
-            rew = reward(obs, act, next_obs, dones)
-            assert rew.shape == (len(obs),)
-            return rew
+    def rew_fn(
+        obs: np.ndarray,
+        act: np.ndarray,
+        next_obs: np.ndarray,
+        dones: np.ndarray,
+    ) -> np.ndarray:
+        rew = reward(obs, act, next_obs, dones)
+        assert rew.shape == (len(obs),)
+        return rew
 
-        return rew_fn
+    return rew_fn
 
-    return loader
+
+def _strip_wrappers(
+    net: RewardNet,
+    wrappers: List[Type[RewardNetWrapper]],
+) -> RewardNet:
+    """Attempts to remove provided wrappers.
+
+    wrappers until either it reaches a non-warper
+    reward net or it has removed all the listed wrappers.
+
+    Args:
+        net: an instance of a reward network that may be wrapped
+        wrappers: list of wrapper types to remove
+
+    Returns:
+        The reward network with the listed wrappers removed
+    """
+    for wrapper_type in wrappers:
+        if not isinstance(net, RewardNetWrapper):
+            # Reached base reward net
+            break
+
+        assert hasattr(net, "base")
+
+        if isinstance(net, wrapper_type):
+            net = net.base
+
+    return net
+
+
+def _make_functional(
+    net: RewardNet,
+    attr: str = "predict",
+    kwargs=dict(),
+) -> common.RewardFn:
+    return lambda *args: getattr(net, attr)(*args, **kwargs)
 
 
 def load_zero(path: str, venv: VecEnv) -> common.RewardFn:
@@ -62,14 +92,37 @@ def load_zero(path: str, venv: VecEnv) -> common.RewardFn:
 
 
 # TODO(adam): I think we can get rid of this and have just one RewardNet.
+
+# TODO figure out if you need to reset the input normalizationan)a)_and_update
 reward_registry.register(
     key="RewardNet_shaped",
-    value=_load_reward_net_as_fn(shaped=True),
+    value=lambda path, _: _validate_reward(_make_functional(th.load(str(path)))),
 )
 reward_registry.register(
     key="RewardNet_unshaped",
-    value=_load_reward_net_as_fn(shaped=False),
+    value=lambda path, _: _validate_reward(
+        _make_functional(_strip_wrappers(th.load(str(path)), [ShapedRewardNet])),
+    ),
 )
+
+reward_registry.register(
+    key="RewardNet_normalized",
+    value=lambda path, _: _validate_reward(
+        _make_functional(
+            th.load(str(path)),
+            attr="predict_processed",
+            kwargs={"update_stats": False},
+        ),
+    ),
+)
+
+reward_registry.register(
+    key="RewardNet_unnormalized",
+    value=lambda path, _: _validate_reward(
+        _make_functional(_strip_wrappers(th.load(str(path)), [NormalizedRewardNet])),
+    ),
+)
+
 reward_registry.register(key="zero", value=load_zero)
 
 
--- a/src/imitation/rewards/serialize.py
+++ b/src/imitation/rewards/serialize.py
@@ -78,7 +78,7 @@ def _make_functional(
 
 def _validate_type(net: RewardNet, type_: Type[RewardNet]):
     if not isinstance(net, type_):
-        raise TypeError("cannot load unnomralized reward as normalized")
+        raise TypeError(f"expected {type_} but found {type(net)}")
     return net
 
 
--- a/src/imitation/rewards/serialize.py
+++ b/src/imitation/rewards/serialize.py
@@ -45,7 +45,7 @@ def _strip_wrappers(
 ) -> RewardNet:
     """Attempts to remove provided wrappers.
 
-    wrappers until either it reaches a non-warper
+    Strips wrappers until either it reaches a non-warper
     reward net or it has removed all the listed wrappers.
 
     Args:
--- a/src/imitation/rewards/serialize.py
+++ b/src/imitation/rewards/serialize.py
@@ -64,6 +64,8 @@ def _strip_wrappers(
 
         if isinstance(net, wrapper_type):
             net = net.base
+        else:
+            break
 
     return net
 
@@ -74,6 +76,12 @@ def _make_functional(
     kwargs=dict(),
 ) -> common.RewardFn:
     return lambda *args: getattr(net, attr)(*args, **kwargs)
+
+
+def _validate_type(net: RewardNet, type_: Type[RewardNet]):
+    if not isinstance(net, type_):
+        raise TypeError("cannot load unnomralized reward as normalized")
+    return net
 
 
 def load_zero(path: str, venv: VecEnv) -> common.RewardFn:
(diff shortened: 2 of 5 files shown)


levmckinney:
This should resolve #437 since now train_rl can be called with the normalize flag set to false and the reward type set to `RewardNet_normalized`.

src/imitation/rewards/serialize.py
```diff
@@ -62,14 +100,36 @@ def f(
 
 
 # TODO(adam): I think we can get rid of this and have just one RewardNet.
+
 reward_registry.register(
     key="RewardNet_shaped",
-    value=_load_reward_net_as_fn(shaped=True),
+    value=lambda path, _: _validate_reward(_make_functional(th.load(str(path)))),
 )
 reward_registry.register(
```
yawen-d, line 108: Seems to me `NormalizedRewardNet` and `ShapedRewardNet` are both subclasses of `RewardNetWrapper`. Could simplify things by registering a `"RewardNet_unwrapped"`.

src/imitation/rewards/serialize.py
```diff
@@ -62,14 +100,36 @@ def f(
...
+    value=lambda path, _: _validate_reward(
+        _make_functional(
+            _validate_type(th.load(str(path)), NormalizedRewardNet),
+            attr="predict_processed",
+            kwargs={"update_stats": False},
+        ),
+    ),
+)
+
+reward_registry.register(
```
yawen-d, line 126: Same comments as previous ones

src/imitation/rewards/serialize.py
```diff
@@ -1,12 +1,18 @@
...
 
-from typing import Callable
+from typing import Callable, List, Type
 
 import numpy as np
 import torch as th
 from stable_baselines3.common.vec_env import VecEnv
 
 from imitation.rewards import common
+from imitation.rewards.reward_nets import (
```
yawen-d, line 10: Could simply this by only importing `RewardNetWrapper`.

src/imitation/rewards/serialize.py
```diff
@@ -17,33 +23,65 @@
...
+
+
+def _strip_wrappers(
+    net: RewardNet,
+    wrappers: List[Type[RewardNetWrapper]],
+) -> RewardNet:
+    """Attempts to remove provided wrappers.
+
+    wrappers until either it reaches a non-warper
+    reward net or it has removed all the listed wrappers.
```
yawen-d, line 49: Strip wrappers until either it reaches a non-warpped reward net or it has removed all the listed wrappers.

src/imitation/rewards/serialize.py
```diff
@@ -17,33 +23,65 @@
...
+
+    Args:
+        net: an instance of a reward network that may be wrapped
+        wrappers: list of wrapper types to remove
+
+    Returns:
+        The reward network with the listed wrappers removed
+    """
+    for wrapper_type in wrappers:
```
yawen-d, line 58: I would suggest simplying this part using a while loop, similar to [here](https://github.com/HumanCompatibleAI/imitation/blob/f332680115380ce654cfd061fbaaa7bc47918d2c/src/imitation/algorithms/adversarial/airl.py#L98).
```python
# Recursively return the base network of the wrapped reward net
while isinstance(reward_net, reward_nets.RewardNetWrapper):
    reward_net = reward_net.base
```

Though it would be better to add more detailed comments here. WDYT?

src/imitation/rewards/serialize.py
```diff
@@ -17,33 +23,65 @@
...
+        if not isinstance(net, RewardNetWrapper):
+            # Reached base reward net
+            break
+
+        assert hasattr(net, "base")
```
yawen-d, line 63: `base` is a property of `RewardNetWrapper`. I think no need to check here.

src/imitation/rewards/serialize.py
```diff
@@ -17,33 +23,65 @@
...
-    return loader
+def _make_functional(
+    net: RewardNet,
+    attr: str = "predict",
+    kwargs=dict(),
+) -> common.RewardFn:
+    return lambda *args: getattr(net, attr)(*args, **kwargs)
+
+
+def _validate_type(net: RewardNet, type_: Type[RewardNet]):
```
yawen-d, line 81: Change the function name to sth like `_validate_norm_reward_type`? Because the check doesn't apply to `RewardNet_shaped`. Another alternative would be to also extend it to `RewardNet_shaped`?

src/imitation/scripts/train_rl.py
```diff
@@ -102,6 +105,10 @@ def train_rl(
         # so normalizing it makes training more stable. Note we do *not* normalize
         # observations here; use the `NormalizeFeaturesExtractor` instead.
         venv = VecNormalize(venv, norm_obs=False, **normalize_kwargs)
+        if reward_type == "RewardNet_normalized":
```
yawen-d, line 108: Maybe we should ban the behavior of double normalization. This could produce silent error if users missed the warning. I would set an Assertion error check here.

src/imitation/scripts/train_rl.py
```diff
@@ -53,7 +54,9 @@ def train_rl(
 
     Args:
         total_timesteps: Number of training timesteps in `model.learn()`.
-        normalize_reward: If True, then rescale and clip reward.
+        normalize_reward: Applies normalization and clipping to the reward function by
```
yawen-d, line 57: Good point! Thanks for making a note about the difference between `VecNormalize` and `RunningNorm`. Though the difference is minor, I would suggest making an issue to track this difference and considering wrapping the `VecEnv` using a `RewardVecEnvWrapper` that takes in a `reward_fn` that takes in the ground truth reward and pass through a normalization layer.

This would be also useful if we want to use EMWA as the normalization layer for `train_rl`.

tests/test_scripts.py
```diff
@@ -366,6 +369,31 @@ def test_transfer_learning(tmpdir: str) -> None:
     _check_rollout_stats(run.result)
 
 
+def test_rl_train_warning(tmpdir: str):
```
yawen-d, line 372: Depending on the changes in `train_rl.py`.

yawen-d:
Sorry for getting to this a bit late. Thanks for the PR. I left a few comments and suggestions:
- Some code simplification in `reward_nets.py`
- Might be good to make an issue to track difference between `RunningNorm` and `VecNormalize`.

src/imitation/rewards/serialize.py
levmckinney: I guess I was imagining the case where we have a reward network that is shaped and then normalized. Maybe the user just wants to remove the normalization but not the shaping.

src/imitation/rewards/serialize.py
levmckinney: Ya I could definitely have it also validate that the reward function you are loading is shaped. I was just a bit nervous about breaking the current interface for anyone already using the code.

src/imitation/scripts/train_rl.py
levmckinney: Sure I can make it an assertion error.

src/imitation/rewards/serialize.py
levmckinney: See https://github.com/HumanCompatibleAI/imitation/issues/437#issuecomment-1136575041

src/imitation/scripts/train_rl.py
levmckinney: We could do that for consistency, but it seems a bit beyond the scope of what we are doing in this pull request. I've opened another issue to discuss it #442 like you suggested.

src/imitation/rewards/serialize.py
levmckinney: How should "RewardNet_unwrapped" behave in the presence of multiple wrappers? See https://github.com/HumanCompatibleAI/imitation/issues/437#issuecomment-1136575041.

src/imitation/rewards/serialize.py
yawen-d: That makes sense. Thanks for reminding me of that.

tests/test_scripts.py
```diff
@@ -358,14 +361,39 @@ def test_transfer_learning(tmpdir: str) -> None:
         named_configs=["cartpole"] + ALGO_FAST_CONFIGS["rl"],
         config_updates=dict(
             common=dict(log_dir=log_dir_data),
-            reward_type="RewardNet_shaped",
+            reward_type="RewardNet_unshaped",
```
levmckinney, line 364: This started throwing errors once I added validation to ensure that reward functions where actually shaped when they were loaded as shaped. I think this is supposed to be unshaped since it's loading the test reward function. Is that correct? Or is this supposed to be a shaped reward?

src/imitation/rewards/serialize.py
levmckinney: It turns out adding validation did break some test cases. I think there is likely either some caused by some other bug in the code or these test cases where not correct. See: https://github.com/HumanCompatibleAI/imitation/pull/440#discussion_r885117806

experiments/transfer_learn_benchmark.sh
```diff
@@ -87,7 +87,7 @@ parallel -j 25% --header : --results "${LOG_ROOT}/parallel/" --colsep , --progre
   with \
   '{env_config_name}' seed='{seed}' \
   common.log_dir="${LOG_ROOT}/${ALGORITHM}/{env_config_name}_{seed}/n_expert_demos_{n_expert_demos}" \
-  reward_type="RewardNet_shaped" \
+  reward_type="RewardNet_unshaped" \
```
levmckinney, line 90: Same issue as the previous one comment.

tests/test_scripts.py
AdamGleave: I think this is right, good catch!

AdamGleave (changes requested):
Thanks for the PR! Overall looks like a good implementation, I've got a few nIt-picky changes requested.

src/imitation/scripts/train_rl.py
```diff
@@ -80,6 +82,10 @@ def train_rl(
 
     Returns:
         The return value of `rollout_stats()` using the final policy.
+
+    Raises:
```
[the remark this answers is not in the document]
yawen-d: In general, on the one hand, I think it's dubious to use VecNormalize. On the other hand, I think we should track the reward fed into the RL algorithm.

For not using `VecNormalize`:
- Let's say the transfer distribution for learned reward shifted a lot from a scale of mean 0 to 100. It might be better to shift the mean back to 0 rather than only rescaling its variance using VecNormalize. We might also want to do it for train_rl using ground truth reward, otherwise the reward scale will be naturally different between GT reward and learned reward (normalized from train norm stats).
- For a reward with arbitrary scale of mean, we are not sure how to set the `clip_reward` scale.

My current best guess are to
1. Keep the norm stats updated during retraining/transfer.
2. Or wrap the loaded reward model with ... another `RewardWrapper` where we can easily choose which norm we wanna use between RunningNorm and EWMA.

At the same time, I think it's okay to keep the VecNormalize option on until we decide whether to replace VecNormalize.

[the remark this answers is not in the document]
levmckinney: Yawen and I had a discussion about having this be an error or a warning https://github.com/HumanCompatibleAI/imitation/pull/440#discussion_r885062805. I can change it back to being a warning fairly easily.

src/imitation/rewards/serialize.py
```diff
@@ -17,33 +23,63 @@
...
+
+
+def _strip_wrappers(
+    net: RewardNet,
+    wrappers: List[Type[RewardNetWrapper]],
+) -> RewardNet:
+    """Attempts to remove provided wrappers.
+
+    Strips wrappers until either it reaches a non-warper
+    reward net or it has removed all the listed wrappers.
```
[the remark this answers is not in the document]
levmckinney: I updated description to make this more clear.

src/imitation/rewards/serialize.py
```diff
@@ -17,33 +23,63 @@
...
+
+    Args:
+        net: an instance of a reward network that may be wrapped
+        wrappers: list of wrapper types to remove
+
+    Returns:
+        The reward network with the listed wrappers removed
+    """
+    for wrapper_type in wrappers:
+        if not isinstance(net, RewardNetWrapper):
```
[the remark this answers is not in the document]
levmckinney: Changed this into an assert to make sure the input has type `RewardNetWrapper`.

tests/rewards/test_reward_nets.py
```diff
@@ -16,7 +16,13 @@
 from imitation.util import networks, util
 
 ENVS = ["FrozenLake-v1", "CartPole-v1", "Pendulum-v1"]
-HARDCODED_TYPES = ["zero"]
+HARDCODED_TYPES = [
```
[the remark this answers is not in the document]
levmckinney: Switched to a more descriptive name.

src/imitation/scripts/train_rl.py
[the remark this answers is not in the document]
yawen-d: Based on our last conversations, we might want to convert it back to a warning for the purpose of this repo.

tests/rewards/test_reward_nets.py
```diff
@@ -75,13 +72,16 @@ def _make_env_and_save_reward_net(env_name, reward_type, tmpdir):
         net = reward_nets.NormalizedRewardNet(net, networks.RunningNorm)
     elif reward_type == "RewardNet_shaped":
         net = reward_nets.ShapedRewardNet(net, _potential, discount_factor=0.99)
+    assert (
```
AdamGleave, line 75: Does this function do the right thing for all reward types in the reward registry? If we add a reward to the registry, will we know to update this function?

Given this is just a helper for a *test*, I don't want to complicate things too much, but I might lean a bit towards actually just checking a hard-coded whitelist this helper was designed to work with.

AdamGleave (approved):
LGTM -- one minor suggestion.

src/imitation/scripts/train_rl.py
```diff
@@ -102,6 +105,12 @@ def train_rl(
         # so normalizing it makes training more stable. Note we do *not* normalize
         # observations here; use the `NormalizeFeaturesExtractor` instead.
         venv = VecNormalize(venv, norm_obs=False, **normalize_kwargs)
+        if reward_type == "RewardNet_normalized":
+            warnings.warn(
+                "Applying normalization to already normalized reward function. \
+                Consider removing the --normalize_reward flag",
```
yawen-d, line 111: Seems like `normalize_reward` isn't necessarily a flag? It's an argument in the train_rl function. Especially for Sacred experiment we usually pass in `python -m imitation.scirpts.train_rl with normalize_reward=True`. 

Suggestion: `"Consider setting normalize_reward as False"` (especially the default for normalize_reward is True in imitaion.scripts.configs.train_rl)

yawen-d (approved):

src/imitation/scripts/train_rl.py
AdamGleave: Oh yes, good catch.

levmckinney: It is changed.

tests/rewards/test_reward_nets.py
levmckinney: I've switch over to a hard coded white list.

src/imitation/scripts/train_rl.py
```diff
@@ -53,7 +53,9 @@ def train_rl(
 
     Args:
         total_timesteps: Number of training timesteps in `model.learn()`.
-        normalize_reward: If True, then rescale and clip reward.
+        normalize_reward: Applies normalization and clipping to the reward function by
+            keeping a running average of training rewards. Note: this is not
+            recommended if using a learned reward that is already normalized.
```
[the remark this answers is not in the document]
levmckinney: I've softened the language as you suggested.

src/imitation/scripts/train_rl.py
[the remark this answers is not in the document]
levmckinney: I converted it back to a warning.

Issue #441, opened 2022-05-28
EMA Reward Normalization


levmckinney:
# Feature Request
An exponential moving average normalization layer. 
### Current Behavior
The current behavior of the running normalization layer `RunningNorm` is to compute a running sample mean and variance.
### Expected Behavior
There should be an exponential moving average Normalization layer that computes the mean and variance of the features based on an exponentially waited sample mean and variance.

### Motivation

Throughout the process learning a reward function and policy the reward distribution will change due to changes in the reward network and the policy. Thus, the samples output by the reward network are not actually i.i.d over the course of training.  Therefor, when using  a running mean and variance to normalize rewards the algorithm puts too much weight on samples from early in reward function learning when the distribution of rewards was very different. Using an exponentially weighted moving average to compute the normalization statistics could fix this problem by putting more weight on recent samples.

AdamGleave:
Fixed in #443

PR #443, opened 2022-05-31
Created EMA normalization layer

levmckinney:
Implements the improvements described in #441.


src/imitation/util/networks.py
```diff
@@ -117,6 +117,57 @@ def forward(self, x: th.Tensor) -> th.Tensor:
         return (x - self.running_mean) / th.sqrt(self.running_var + self.eps)
 
 
+class EMARunningNorm(RunningNorm):
+    """Similar to RunningNorm but uses an exponentially weighting."""
```
AdamGleave, line 121:
```suggestion
    """Similar to RunningNorm but uses an exponential weighting."""
```

src/imitation/util/networks.py
```diff
@@ -117,6 +117,57 @@ def forward(self, x: th.Tensor) -> th.Tensor:
         return (x - self.running_mean) / th.sqrt(self.running_var + self.eps)
 
 
+class EMARunningNorm(RunningNorm):
```
AdamGleave, line 120: I weakly lean towards making `RunningNorm` an abstract class whose children must override `update_stats`, and then have concrete implementations `EMARunningNorm` and `SimpleRunningNorm` (currently `RunningNorm`).

Main reason is that it feels wrong to say that `EMARunningNorm` is-a simple moving average (which is what `RunningNorm` implements), and we don't use any of `RunningNorm.update_stats` but override it fully.

src/imitation/util/networks.py
```diff
@@ -117,6 +117,57 @@ def forward(self, x: th.Tensor) -> th.Tensor:
...
+    """Similar to RunningNorm but uses an exponentially weighting."""
+
+    running_mean: th.Tensor
```
AdamGleave, line 123: Is this needed given it's defined in parent class?

src/imitation/util/networks.py
```diff
@@ -117,6 +117,57 @@ def forward(self, x: th.Tensor) -> th.Tensor:
...
+        """Builds EMARunningNorm.
+
+        Args:
+            num_features: Number of features; the length of the non-batch dim.
+            lamb: value between (0, 1) update rate for the mean
+            r: value between (0, 1) update rate for the variance
+            eps: small constant for for numerical stability.
+        """
+        super().__init__(num_features, eps=eps)
+        self.lamb = lamb
```
AdamGleave, line 142: Add input validation between (0,1) for this and r and raise ValueError if it fails?

(I guess we should check eps > 0 in parent class too, missed that...)

src/imitation/util/networks.py
```diff
@@ -117,6 +117,57 @@ def forward(self, x: th.Tensor) -> th.Tensor:
...
+        self.r = r
+
+    def update_stats(self, batch: th.Tensor) -> None:
+        """Update `self.running_mean` and `self.running_var`.
+
+        Uses Macgregor & Harris (1993), "The Exponentially Weighted Moving Variance".
```
AdamGleave, line 148: always nice to have a citation :)

src/imitation/util/networks.py
```diff
@@ -117,6 +117,57 @@ def forward(self, x: th.Tensor) -> th.Tensor:
...
+            lamb: value between (0, 1) update rate for the mean
+            r: value between (0, 1) update rate for the variance
+            eps: small constant for for numerical stability.
+        """
+        super().__init__(num_features, eps=eps)
+        self.lamb = lamb
+        self.r = r
+
+    def update_stats(self, batch: th.Tensor) -> None:
+        """Update `self.running_mean` and `self.running_var`.
```
AdamGleave, line 146: you update count too right? But it's perhaps not actually used for anything beyond an auxiliary statistic.

src/imitation/util/networks.py
```diff
@@ -117,6 +117,57 @@ def forward(self, x: th.Tensor) -> th.Tensor:
...
+        """
+        if len(batch.shape) == 1:  # Assume list is a batch of 1-d vectors
+            batch = batch.unsqueeze(-1)
+
+        b_size = batch.shape[0]
+        b_mean = th.mean(batch, dim=0)
+        self.running_mean *= 1 - self.lamb
+        self.running_mean += b_mean * self.lamb
+
+        b_var = (
```
AdamGleave, line 161: Consider adding some intermediate variable, e.g:

```python
centered = batch - self.running_mean.unsqueeze(0)
```

Won't take up more LOC (as black already forces you to put line 162 on its own line), and I think will make it more readable if step-by-step.

tests/util/test_networks.py
```diff
@@ -11,13 +12,20 @@
 assert_equal = functools.partial(th.testing.assert_close, rtol=0, atol=0)
 
 
-def test_running_norm_identity() -> None:
+NORMALIZATION_LAYERS = [networks.RunningNorm, networks.EMARunningNorm]
+
+
+@pytest.mark.parametrize("normalization_layer", NORMALIZATION_LAYERS)
+def test_running_norm_identity(normalization_layer) -> None:
```
AdamGleave, line 19:
```suggestion
def test_running_norm_identity(normalization_layer: Type[networks.RunningNorm]) -> None:
```

AdamGleave: Type annotations in test code matter a lot less but may as well stick it in

tests/util/test_networks.py
```diff
@@ -28,13 +36,15 @@ def test_running_norm_identity() -> None:
         assert_equal(running_norm.forward(normalized), normalized)
 
 
+@pytest.mark.parametrize("normalization_layer", NORMALIZATION_LAYERS)
 def test_running_norm_eval_fixed(
+    normalization_layer,
```
AdamGleave, line 41: type annotation?

tests/util/test_networks.py
```diff
@@ -94,11 +104,42 @@ def test_running_norm_matches_dist(batch_size: int) -> None:
...
 
+@pytest.mark.parametrize("batch_size", [1, 8])
+@pytest.mark.parametrize("normalization_layer", NORMALIZATION_LAYERS)
+def test_parameters_converge(
+    batch_size: int,
+    normalization_layer: Type[networks.RunningNorm],
+) -> None:
+    """Test running norm parameters approximately converge to true values."""
+    mean = th.Tensor([42])
+    var = th.Tensor([42])
```
AdamGleave, line 115: I'd suggest using a different value for mean and variance, to catch bugs of the form `self.running_mean = self.running_var` (unlikely to be something actually that dumb, admittedly, but I'd have a vague fear there can be more subtle failures like this)

tests/util/test_networks.py
```diff
@@ -94,11 +104,42 @@ def test_running_norm_matches_dist(batch_size: int) -> None:
...
+    sd = th.sqrt(var)
+
+    num_dims = len(mean)
```
AdamGleave, line 118: Why not have two dimensions, with different mean/var? I think your test will still work! And this will catch bugs that mix statistics across dimensions somehow.

tests/util/test_networks.py
```diff
@@ -94,11 +104,42 @@ def test_running_norm_matches_dist(batch_size: int) -> None:
     assert running_norm.count == num_samples
 
 
+@pytest.mark.parametrize("batch_size", [1, 8])
+@pytest.mark.parametrize("normalization_layer", NORMALIZATION_LAYERS)
+def test_parameters_converge(
```
AdamGleave, line 109: Thanks for adding this test!

tests/util/test_networks.py
```diff
@@ -94,11 +104,42 @@ def test_running_norm_matches_dist(batch_size: int) -> None:
...
+
+    assert running_norm.count == num_samples
+
+
 @pytest.mark.parametrize(
     "init_kwargs",
     [
         {},
-        {"normalize_input_layer": networks.RunningNorm},
+        *[{"normalize_input_layer": layer} for layer in NORMALIZATION_LAYERS],
```
AdamGleave, line 142: Clever, but a bit obscure. How about?
```python
[{}] + [{"normalize_input_layer": layer} for layer in NORMALIZATION_LAYERS]
```

AdamGleave (changes requested):
Thanks for the PR. Looks correct, and I appreciate the additional test checking for parameter convergence.

The only high-level feedback I have is to make RunningNorm abstract and have two concrete implementations.

Left some more detailed feedback.

Please request re-review when addressed.

We should also at some point see what the empirical effect of these changes is. @yawen-d is that something you're already running? If not then perhaps @levmckinney you could do a basic test on a few environments?

yawen-d:
> We should also at some point see what the empirical effect of these changes is. @yawen-d is that something you're already running? If not then perhaps @levmckinney you could do a basic test on a few environments?

My next batch of experiments will basically about empirical tests of EWMA. So I'm happy to run it as it's aligned closely with my progress.

src/imitation/util/networks.py
levmckinney: What do you think of the class structure being
```
class BaseNorm(nn.Module, ABC)
class RunningNorm(BaseNorm)
class EMANorm(BaseNorm)
```
This would not break the public API and does not seem too awkward.

AdamGleave: I'm not too worried about breaking changes (we're not v1.0 yet, and `RunningNorm` was only recently introduced). But `BaseNorm` seems like a reasonable name for it, I'm happy with this hierarchy if it seems more natural to you.

levmckinney: Awesome! It's pushed.

src/imitation/util/networks.py
levmckinney: I wanted to keep the interface the same for everything.

yawen-d (commented):
Made some comments on the implementations. Mainly discussing the rules for updating.

src/imitation/util/networks.py
```diff
@@ -117,6 +117,57 @@ def forward(self, x: th.Tensor) -> th.Tensor:
...
+            th.sum(th.square(self.running_mean.unsqueeze(0) - batch), dim=0) / b_size
+        )
```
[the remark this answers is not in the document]
levmckinney: I think I like the derivations in [Incremental calculation of weighted mean and variance](https://fanf2.user.srcf.net/hermes/doc/antiforgery/stats.pdf) better than “The Exponentially Weighted Moving Variance”. Let me see if I can quickly work out how to take into account the batching.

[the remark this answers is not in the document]
levmckinney: I am convinced that using the old running mean is probably more correct, so I'm going to switch the current implementation over to doing that. The derivations in [this paper](https://fanf2.user.srcf.net/hermes/doc/antiforgery/stats.pdf) are extremely difficult for me to follow in a few places, and thus I'm not 100% sure the way I'm generalizing to batches is correct. I really doubt this will affect the performance of the system at all, so I'm not going to spend any more time trying to derive the incremental update from first principles.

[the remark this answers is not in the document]
levmckinney: This seems relevant https://openreview.net/pdf?id=SkgGjRVKDS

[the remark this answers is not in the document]
levmckinney: I have switched over to an implementation that does take batches into account and simply and updates incrementally. Howerver, it is a bit slower,
```bash
# New implementation
$ python -m timeit -s 'import torch; from imitation.util import networks; norm = networks.EMANorm(256); a = torch.randn(64, 256)' 'norm(a)'
> 200 loops, best of 5: 1.19 msec per loop
# Old implimentation
$ python -m timeit -s 'import torch; from imitation.util import networks; norm = networks.EMANorm(256); a = torch.randn(64, 256)' 'norm(a)'    
5000 loops, best of 5: 66.3 usec per loop
# Running mean implementation in use now
$ python -m timeit -s 'import torch; from imitation.util import networks; norm = networks.RunningNorm(256); a = torch.randn(64, 256)' 'norm(a)'
2000 loops, best of 5: 182 usec per loop
```

tests/util/test_networks.py
```diff
@@ -11,30 +12,52 @@
...
-    for i in range(10):
+    for _ in range(10):
         assert_equal(running_norm.forward(x), x)
+
+
+@pytest.mark.parametrize("normalization_layer", [networks.RunningNorm])
+def test_running_norm_identity_train(normalization_layer: Type[networks.BaseNorm]):
+    """Test that the running norm will not change already normalized data.
+
+    Note that we do not test the `EMANorm` here since its parameters will
```
AdamGleave, line 38: Sad, can we not do a more relaxed version of this test (assert_close) for that)? Although not essential given we still have a test later checking for parameter convergence.

AdamGleave (approved):
LGTM -- couple minor comments

tests/util/test_networks.py
levmckinney: I was a bit worried about reducing the tolerance on the original test so that it would pass on the new code, but it was not as bad as I thought. So I've changed it over to testing both versions now.

PR #444, opened 2022-06-06
Add Soft Actor Critic to imitation experiments

yawen-d:
This PR aims to 
- Add the option of choosing SAC as RL algorithm in the `imitation.scripts.common.rl` and `imitation.scripts.common.train` ingredient.
- Construct test cases for SAC runs
- Some minor changes (e.g. comments instead of docstring in sacred configs; upgrade HalfCheetah-v2 to HalfCheetah-v3; modify paths in launch_docker-dev.sh)

Add SAC support for #447

### DIFF
--- a/runners/launch_docker-dev.sh
+++ b/runners/launch_docker-dev.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e # Exit immediately if a command exits with a non-zero status.
+set -x # echo on
 
 __usage="launch_docker-dev.sh - Launching humancompatibleai/imitation:python-req
 
@@ -51,6 +52,7 @@ if [[ $PULL == 1 ]]; then
   echo "Pulling ${DOCKER_IMAGE} from DockerHub"
   docker pull ${DOCKER_IMAGE}
 fi
+
 
 docker run -it --rm --init \
   -v "${IMIT_LOCAL_MNT}:/imitation" \
--- a/runners/launch_docker-dev.sh
+++ b/runners/launch_docker-dev.sh
@@ -10,8 +10,8 @@ Usage: launch_docker-dev.sh [options]
 options:
   -p, --pull                pull the image to DockerHub
 
-Note: You can specify IMIT_LOCAL_MNT and IMIT_MJKEY_MNT environment variables to mount
-  local repository and MuJoCo license key respectively.
+Note: You can specify IMIT_LOCAL_MNT environment variables to mount local
+  repository and MuJoCo license key respectively.
 "
 
 PULL=0
@@ -39,11 +39,6 @@ if [[ ${IMIT_LOCAL_MNT} == "" ]]; then
   IMIT_LOCAL_MNT="${HOME}/imitation"
 fi
 
-# Pass your own mjkey.txt
-if [[ ${IMIT_MJKEY_MNT} == "" ]]; then
-  IMIT_MJKEY_MNT="${HOME}/mnt/mjkey.txt"
-fi
-
 # install imitation in developer mode
 CMD="pip install -e .[docs,parallel,test] gym[mujoco]" # copied from ci/build_and_activate_venv.sh
 
@@ -56,6 +51,5 @@ fi
 
 docker run -it --rm --init \
   -v "${IMIT_LOCAL_MNT}:/imitation" \
-  -v "${IMIT_MJKEY_MNT}:/root/.mujoco/mjkey.txt" \
   ${DOCKER_IMAGE} \
   /bin/bash -c "${CMD} && exec bash"
--- a/runners/launch_docker-dev.sh
+++ b/runners/launch_docker-dev.sh
@@ -9,7 +9,7 @@ Usage: launch_docker-dev.sh [options]
 options:
   -p, --pull                pull the image to DockerHub
 
-Note: You can specify LOCAL_MNT and MJKEY_MNT environment variables to mount
+Note: You can specify IMIT_LOCAL_MNT and IMIT_MJKEY_MNT environment variables to mount
   local repository and MuJoCo license key respectively.
 "
 
@@ -33,18 +33,18 @@ while test $# -gt 0; do
 done
 
 DOCKER_IMAGE="humancompatibleai/imitation:python-req"
-# Specify LOCAL_MNT if you want to mount a local directory to the docker container
-if [[ ${LOCAL_MNT} == "" ]]; then
-  LOCAL_MNT="${HOME}"
+# Specify IMIT_LOCAL_MNT if you want to mount a local directory to the docker container
+if [[ ${IMIT_LOCAL_MNT} == "" ]]; then
+  IMIT_LOCAL_MNT="${HOME}/imitation"
 fi
 
 # Pass your own mjkey.txt
-if [[ ${MJKEY_MNT} == "" ]]; then
-  MJKEY_MNT="${HOME}/mnt/mjkey.txt"
+if [[ ${IMIT_MJKEY_MNT} == "" ]]; then
+  IMIT_MJKEY_MNT="${HOME}/mnt/mjkey.txt"
 fi
 
 # install imitation in developer mode
-CMD="pip install -e .[docs,parallel,test] gym[mujoco]" # borrowed from ci/build_and_activate_venv.sh
+CMD="pip install -e .[docs,parallel,test] gym[mujoco]" # copied from ci/build_and_activate_venv.sh
 
 # Pull image from DockerHub if prompted
 if [[ $PULL == 1 ]]; then
@@ -53,7 +53,7 @@ if [[ $PULL == 1 ]]; then
 fi
 
 docker run -it --rm --init \
-  -v "${LOCAL_MNT}/imitation:/imitation" \
-  -v "${MJKEY_MNT}:/root/.mujoco/mjkey.txt" \
+  -v "${IMIT_LOCAL_MNT}:/imitation" \
+  -v "${IMIT_MJKEY_MNT}:/root/.mujoco/mjkey.txt" \
   ${DOCKER_IMAGE} \
   /bin/bash -c "${CMD} && exec bash"
--- a/runners/launch_docker-dev.sh
+++ b/runners/launch_docker-dev.sh
@@ -35,7 +35,7 @@ done
 DOCKER_IMAGE="humancompatibleai/imitation:python-req"
 # Specify LOCAL_MNT if you want to mount a local directory to the docker container
 if [[ ${LOCAL_MNT} == "" ]]; then
-  LOCAL_MNT="${HOME}/imitation"
+  LOCAL_MNT="${HOME}"
 fi
 
 # Pass your own mjkey.txt
@@ -53,7 +53,7 @@ if [[ $PULL == 1 ]]; then
 fi
 
 docker run -it --rm --init \
-  -v "${LOCAL_MNT}:/imitation" \
+  -v "${LOCAL_MNT}/imitation:/imitation" \
   -v "${MJKEY_MNT}:/root/.mujoco/mjkey.txt" \
   ${DOCKER_IMAGE} \
   /bin/bash -c "${CMD} && exec bash"
--- a/src/imitation/algorithms/adversarial/airl.py
+++ b/src/imitation/algorithms/adversarial/airl.py
@@ -1,7 +1,8 @@
 """Adversarial Inverse Reinforcement Learning (AIRL)."""
 
 import torch as th
-from stable_baselines3.common import base_class, vec_env
+from stable_baselines3.common import base_class, policies, vec_env
+from stable_baselines3.sac import policies as sac_policies
 
 from imitation.algorithms import base
 from imitation.algorithms.adversarial import common
@@ -54,7 +55,13 @@ class AIRL(common.AdversarialTrainer):
             reward_net=reward_net,
             **kwargs,
         )
-        if not hasattr(self.gen_algo.policy, "evaluate_actions"):
+        # AIRL needs a stochastic policy to compute the discriminator output.
+        # sac_policies.SACPolicy and policies.ActorCriticPolicy both have an actor
+        # to generate stochastic actions.
+        if not isinstance(
+            self.gen_algo.policy,
+            (sac_policies.SACPolicy, policies.ActorCriticPolicy),
+        ):
             raise TypeError(
                 "AIRL needs a stochastic policy to compute the discriminator output.",
             )
--- a/src/imitation/algorithms/adversarial/airl.py
+++ b/src/imitation/algorithms/adversarial/airl.py
@@ -8,6 +8,8 @@ from imitation.algorithms import base
 from imitation.algorithms.adversarial import common
 from imitation.rewards import reward_nets
 
+STOCHASTIC_POLICIES = (sac_policies.SACPolicy, policies.ActorCriticPolicy)
+
 
 class AIRL(common.AdversarialTrainer):
     """Adversarial Inverse Reinforcement Learning (`AIRL`_).
@@ -55,13 +57,8 @@ class AIRL(common.AdversarialTrainer):
             reward_net=reward_net,
             **kwargs,
         )
-        # AIRL needs a stochastic policy to compute the discriminator output.
-        # sac_policies.SACPolicy and policies.ActorCriticPolicy both have an actor
-        # to generate stochastic actions.
-        if not isinstance(
-            self.gen_algo.policy,
-            (sac_policies.SACPolicy, policies.ActorCriticPolicy),
-        ):
+        # AIRL needs a policy from STOCHASTIC_POLICIES to compute discriminator output.
+        if not isinstance(self.gen_algo.policy, STOCHASTIC_POLICIES):
             raise TypeError(
                 "AIRL needs a stochastic policy to compute the discriminator output.",
             )
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -11,12 +11,15 @@ import torch as th
 import torch.utils.tensorboard as thboard
 import tqdm
 from stable_baselines3.common import base_class, policies, vec_env
+from stable_baselines3.sac import policies as sac_policies
 from torch.nn import functional as F
 
 from imitation.algorithms import base
 from imitation.data import buffer, rollout, types, wrappers
 from imitation.rewards import reward_nets, reward_wrapper
 from imitation.util import logger, networks, util
+
+LOG_PROB_CUTOFF = -20.0
 
 
 def compute_train_stats(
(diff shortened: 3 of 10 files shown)


src/imitation/scripts/common/rl.py
```diff
@@ -84,13 +117,19 @@ def make_rl_algo(
...
-        assert "batch_size" not in rl_kwargs, "set 'batch_size' at top-level"
+        if "batch_size" not in rl_kwargs or rl_kwargs["batch_size"] is not None:
+            raise ValueError("set 'batch_size' at top-level")
         rl_kwargs["batch_size"] = batch_size
     else:
         raise TypeError(f"Unsupported RL algorithm '{rl_cls}'")
     rl_algo = rl_cls(
         policy=train["policy_cls"],
-        policy_kwargs=train["policy_kwargs"],
+        # Note(yawen): Here we make a copy of policy_kwargs as a temporary workaround
```
AdamGleave, line 127: This is a good point. What's temporary about it -- do you have an alternative solution in mind? If SB3 is going to go around mutating our configs I think this is the right thing to do.

I'd say something like:
```
# Note(yawen): Copy `policy_kwargs` as SB3 may mutate the config we pass.
# In particular, policy_kwargs["use_sde"] may be changed in rl_cls.__init__() for
# certain algorithms, such as Soft Actor Critic. See:
# https://github.com/DLR-RM/stable-baselines3/blob/30772aa9f53a4cf61571ee90046cdc454c1b11d7/stable_baselines3/common/off_policy_algorithm.py#L145
```

src/imitation/scripts/common/rl.py
```diff
@@ -42,6 +53,28 @@ def fast():
...
 
+@rl_ingredient.named_config
+def ppo():
+    # For recommended PPO hyperparams in each environment, see:
+    # https://github.com/DLR-RM/rl-baselines3-zoo/blob/master/hyperparams/ppo.yml
+    rl_cls = stable_baselines3.PPO
+    locals()  # quieten flake8
+
+
+# TODO(yawen): The current implementation of Soft Actor Critic in SB3 only supports
```
AdamGleave, line 64: This isn't really a TODO for imitation, I'd just add a one-line comment warning that SAC is currently continuous only. (And maybe link to that issue.)

src/imitation/scripts/common/train.py
```diff
@@ -32,6 +33,13 @@ def fast():
     locals()  # quieten flake8
 
 
+@train_ingredient.named_config
+def sac():
+    policy_cls = SACPolicy
+
+    locals()  # quieten flake8
```
AdamGleave, line 40: for a one-liner, just adding `# noqa: F841` is cleaner

tests/test_scripts.py
```diff
@@ -286,6 +331,33 @@ def test_train_adversarial(tmpdir, named_configs):
     _check_train_ex_result(run.result)
 
 
+def test_train_adversarial_sac(tmpdir):
```
AdamGleave, line 334: It's a bit sad how similar this is to `test_train_adversarial`, but there are enough differences (pendulum vs cartpole, additional named configs, additional assertion) I think you made the right call just copy-pasting and editing.

src/imitation/scripts/common/rl.py
```diff
@@ -20,19 +20,30 @@
...
     locals()  # quieten flake8
 
 
+@rl_ingredient.config_hook
+def config_hook(config, command_name, logger):
+    """Sets defaults equivalent to stable_baselines3.PPO default hyperparameters."""
+    del command_name, logger
+    res = {}
+    if config["rl"]["rl_cls"] is None:
+        default_rl = stable_baselines3.PPO
```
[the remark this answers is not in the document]
yawen-d: Oops

src/imitation/scripts/common/rl.py
```diff
@@ -42,6 +53,28 @@ def fast():
     locals()  # quieten flake8
 
 
+@rl_ingredient.named_config
+def ppo():
+    # For recommended PPO hyperparams in each environment, see:
+    # https://github.com/DLR-RM/rl-baselines3-zoo/blob/master/hyperparams/ppo.yml
+    rl_cls = stable_baselines3.PPO
```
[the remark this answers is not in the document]
yawen-d: This is a good point. I would think it's better to put it into the hook.

src/imitation/scripts/common/train.py
yawen-d: Sure!

tests/test_scripts.py
yawen-d: Simplified something that has been tested above.

src/imitation/scripts/common/rl.py
yawen-d: I don't have an alternative solution. There is [another way](https://github.com/IDSIA/sacred/issues/502) around: which is to set `SETTINGS.CONFIG.READ_ONLY_CONFIG = False`, but this is even less desirable because it would have effects globally.
```python
from sacred import SETTINGS
SETTINGS.CONFIG.READ_ONLY_CONFIG = False
```

yawen-d: Perhaps we should also add a warning here?

src/imitation/scripts/common/rl.py
yawen-d: You're right.

yawen-d:
Finished a round of revision based on the comments. Currently investigating the test failure for `airl`.

runners/launch_docker-dev.sh
```diff
@@ -33,27 +34,28 @@ while test $# -gt 0; do
...
-  LOCAL_MNT="${HOME}/imitation"
+# Specify IMIT_LOCAL_MNT if you want to mount a local directory to the docker container
+if [[ ${IMIT_LOCAL_MNT} == "" ]]; then
+  IMIT_LOCAL_MNT="${HOME}/imitation"
 fi
 
 # Pass your own mjkey.txt
-if [[ ${MJKEY_MNT} == "" ]]; then
-  MJKEY_MNT="${HOME}/mnt/mjkey.txt"
+if [[ ${IMIT_MJKEY_MNT} == "" ]]; then
```
ernestum, line 43: We probably don't need this feature any more since the mjkey.txt is already baked into the docker image.

yawen-d: Cool, thanks for the reminder!

src/imitation/scripts/common/rl.py
AdamGleave: Isn't the sacred warning here correct, though? Unless I'm missing something we don't want config objects to be mutated. So turning off the warning seems worse than copying it (which actually stops it from being mutated).

src/imitation/algorithms/adversarial/common.py
```diff
@@ -427,6 +430,45 @@ def _torchify_array(self, ndarray: Optional[np.ndarray]) -> Optional[th.Tensor]:
...
+
+        Args:
+            obs_th: A batch of observations.
+            acts_th: A batch of actions.
+
+        Returns:
+            A batch of log policy action probabilities.
+        """
+        log_policy_act_prob, log_policy_act_prob_th = None, None
+        if isinstance(self.gen_algo.policy, policies.ActorCriticPolicy):
```
AdamGleave, line 448:
```suggestion
        policy = self.gen_algo.policy
        if isinstance(policy, policies.ActorCriticPolicy):
```
You use policy a bunch here, so may as well shorten it.

src/imitation/algorithms/adversarial/common.py
```diff
@@ -427,6 +430,45 @@ def _torchify_array(self, ndarray: Optional[np.ndarray]) -> Optional[th.Tensor]:
...
+            mean_actions, log_std, _ = gen_algo_actor.get_action_dist_params(obs_th)
+            distribution = gen_algo_actor.action_dist.proba_distribution(
+                mean_actions,
+                log_std,
+            )
+            log_policy_act_prob_th = distribution.log_prob(acts_th)
+            # getting log probability of actions from a continuous distribution
+            # might result in NaN values.
+            nan_log_prob_idx = th.where(th.isnan(log_policy_act_prob_th))
+            log_policy_act_prob_th[nan_log_prob_idx] = LOG_PROB_CUTOFF
```
AdamGleave, line 467: I think it's clearer if we just add an:
```python
else:
    return None
```
branch here, and then can assume that `log_policy_act_prob_th` is non-None (and remove the initialization line).

src/imitation/algorithms/adversarial/common.py
```diff
@@ -427,6 +430,45 @@ def _torchify_array(self, ndarray: Optional[np.ndarray]) -> Optional[th.Tensor]:
...
+        elif isinstance(self.gen_algo.policy, sac_policies.SACPolicy):
+            gen_algo_actor = self.gen_algo.policy.actor
+            # generate log_policy_act_prob from SAC actor.
+            mean_actions, log_std, _ = gen_algo_actor.get_action_dist_params(obs_th)
+            distribution = gen_algo_actor.action_dist.proba_distribution(
+                mean_actions,
+                log_std,
+            )
+            log_policy_act_prob_th = distribution.log_prob(acts_th)
+            # getting log probability of actions from a continuous distribution
```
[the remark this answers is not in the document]
AdamGleave: Under what situation have you run into this being a problem? PPO also has a continuous distribution and we've tested on continuous environments and AFAIK not seen any NaNs. `evaluate_actions` in https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/policies.py#L632 does not seem to do any clipping.

src/imitation/algorithms/adversarial/airl.py
```diff
@@ -54,7 +55,13 @@ def __init__(
             reward_net=reward_net,
             **kwargs,
         )
-        if not hasattr(self.gen_algo.policy, "evaluate_actions"):
+        # AIRL needs a stochastic policy to compute the discriminator output.
+        # sac_policies.SACPolicy and policies.ActorCriticPolicy both have an actor
+        # to generate stochastic actions.
+        if not isinstance(
+            self.gen_algo.policy,
+            (sac_policies.SACPolicy, policies.ActorCriticPolicy),
```
AdamGleave, line 63: Consider defining a constant e.g. `STOCHASTIC_POLICIES` that defines this tuple. Then it's self-documenting (and you can abbreviate the comment).

src/imitation/policies/base.py
```diff
@@ -63,14 +64,30 @@ class FeedForward32Policy(policies.ActorCriticPolicy):
 
     Note: This differs from stable_baselines3 ActorCriticPolicy in two ways: by
     having 32 rather than 64 units, and by having policy and value networks
-    share weights except at the final layer.
```
AdamGleave, line 66: do they share weights at the final layer?

tests/policies/test_policies.py
```diff
@@ -36,31 +34,37 @@ def test_actions_valid(env_name, policy_type):
...
+@pytest.mark.parametrize(
+    "policy_env_pair",
+    [
+        ("ppo", SIMPLE_DISCRETE_ENV),
+        ("sac", SIMPLE_CONTINUOUS_ENV),
+    ],
+)
+def test_save_stable_model_errors_and_warnings(tmpdir, policy_env_pair):
     """Check errors and warnings in `save_stable_model()`."""
+    policy, env = policy_env_pair
```
AdamGleave, line 46:
```suggestion
    policy, env_name = policy_env_pair
```
we usually use `env` to refer to a `gym.Env`

tests/test_scripts.py
```diff
@@ -309,47 +324,39 @@ def _check_train_ex_result(result: dict):
         ["train.normalize_disable", "reward.normalize_input_disable"],
     ),
 )
-def test_train_adversarial(tmpdir, named_configs):
+@pytest.mark.parametrize("command", ("airl", "gail"))
```
AdamGleave, line 327: :) yes we should test both

AdamGleave (commented):
Thanks for the additional features! Some mostly minor comments. 

Only correctness related point is that I want to understand why we're getting NaNs in SAC log-probs and not in e.g. PPO, there might be some other way to address this.

tests/policies/test_policies.py
yawen-d: accepted!

tests/test_scripts.py
yawen-d: I really didn't expect the problem from airl lol...

src/imitation/policies/base.py
yawen-d: - SB3 [ActorCriticPolicy](https://github.com/DLR-RM/stable-baselines3/blob/d68f0a2411766beb6da58ee0e989d1a6a72869bc/stable_baselines3/common/policies.py#L455) uses `[dict(pi=[64, 64], vf=[64, 64])]` as its default network.
- According to [Custom Policy Network in SB3](https://stable-baselines3.readthedocs.io/en/master/guide/custom_policy.html#on-policy-algorithms), they don't share any layer.

src/imitation/algorithms/adversarial/airl.py
yawen-d: Good idea

src/imitation/algorithms/adversarial/common.py
yawen-d: Good point

src/imitation/algorithms/adversarial/common.py
yawen-d: Good suggestion

src/imitation/policies/base.py
AdamGleave: So if I understand correctly:
  - SB3 does no weight sharing by default.
  - We do weight sharing (except at the final layer, where there's a different linear head).
This seems consistent with the original comment, but I can see that it's a bit ambiguous, but perhaps clarify rather than remove the "except at the final layer" caveat.

src/imitation/scripts/common/rl.py
yawen-d: Mutating config objects with `SETTINGS.CONFIG.READ_ONLY_CONFIG = True` (which is the default) actually raises an exception instead of a warning. I think we agree that copying is better than setting `SETTINGS.CONFIG.READ_ONLY_CONFIG = False`. We can go ahead with the current approach of copying the config if you agree.

src/imitation/algorithms/adversarial/common.py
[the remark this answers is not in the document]
yawen-d: Oh, you're correct. I misidentified the problem.

- `nan` doesn't come from producing `log_prob` for actions far away from the distribution mean.
- The problem comes from the squashing function used by SAC (mentioned in [SpinningUp](https://spinningup.openai.com/en/latest/algorithms/sac.html#id1), [original paper Appendix C](https://arxiv.org/pdf/1801.01290.pdf), and [SB3 implementation](https://github.com/DLR-RM/stable-baselines3/blob/d68f0a2411766beb6da58ee0e989d1a6a72869bc/stable_baselines3/common/distributions.py#L195)) Specifically `SquashedDiagGaussianDistribution` squashes the sampled actions between (-1, 1) from a DiagGaussianDistribution by passing through a tanh function, hence it's expected that the actions are within (-1, 1).
- Then when [calculating log_prob](https://github.com/DLR-RM/stable-baselines3/blob/d68f0a2411766beb6da58ee0e989d1a6a72869bc/stable_baselines3/common/distributions.py#L225), the potentially unbounded actions could pass through `th.log(1-action**2+epsilon)`, which results in `nan` when `abs(actions) > 1`. This does happen for [pendulum](https://github.com/openai/gym/blob/053ee809324861dc1bec19feee97b47ba1557a94/gym/envs/classic_control/pendulum.py#L38) with its `action_space` being `Box(-2, 2)`.
- To be more specific, `acts_th` directly came from [`self._gen_replay_buffer`](https://github.com/HumanCompatibleAI/imitation/blob/2db81fec1490ea4624881fc44b9b566626f0658a/src/imitation/algorithms/adversarial/common.py#L459), which collects `gen_samples` with unscaled actions.

Proposed solutions:
- SB3.SAC does scale the [randomly sampled actions](https://github.com/DLR-RM/stable-baselines3/blob/d68f0a2411766beb6da58ee0e989d1a6a72869bc/stable_baselines3/common/off_policy_algorithm.py#L407) using `self.policy.scale_action`. I believe this would also work for our case.

src/imitation/policies/base.py
yawen-d: Oh I see what you mean. You're correct that it would be better to specify the different linear heads.

src/imitation/algorithms/adversarial/common.py
```diff
@@ -444,29 +442,28 @@ def _get_log_policy_act_prob(
...
-            log_policy_act_prob_th[nan_log_prob_idx] = LOG_PROB_CUTOFF
-        if log_policy_act_prob_th is not None:
-            log_policy_act_prob = log_policy_act_prob_th.detach().cpu().numpy()
+            # SAC applies a squashing function to bound the actions to a finite range
+            # `acts_th` need to be scaled accordingly before computing log prob.
+            scaled_acts_th = self.policy.scale_action(acts_th)
+            log_policy_act_prob_th = distribution.log_prob(scaled_acts_th)
+        else:
+            return None
+        log_policy_act_prob = log_policy_act_prob_th.detach().cpu().numpy()
```
AdamGleave, line 466:
```suggestion
        return log_policy_act_prob_th.detach().cpu().numpy()
```
and delete next line

src/imitation/algorithms/adversarial/common.py
```diff
@@ -427,6 +428,44 @@ def _torchify_array(self, ndarray: Optional[np.ndarray]) -> Optional[th.Tensor]:
...
+            gen_algo_actor = self.policy.actor
+            # generate log_policy_act_prob from SAC actor.
+            mean_actions, log_std, _ = gen_algo_actor.get_action_dist_params(obs_th)
+            distribution = gen_algo_actor.action_dist.proba_distribution(
+                mean_actions,
+                log_std,
+            )
+            # SAC applies a squashing function to bound the actions to a finite range
+            # `acts_th` need to be scaled accordingly before computing log prob.
+            scaled_acts_th = self.policy.scale_action(acts_th)
```
AdamGleave, line 462: This solution seems fine. It will break in environments that don't have bounded action spaces. But that's unusual, and there's not much reasonable we can do in that case.

AdamGleave: To be defensive, I might add an `assert self.policy.squash_output` as we'll be doing the wrong thing otherwise. It's currently hardcoded to True in SAC so we should be OK, but this feels like the kind of implementation detail that might change in the future.

AdamGleave (approved):
LGTM -- couple of minor suggestions.

src/imitation/algorithms/adversarial/common.py
yawen-d: That's a good suggestion!

src/imitation/algorithms/adversarial/common.py
yawen-d: check!

PR #445, opened 2022-06-07
Load expert models for testing from huggingface hub 

ernestum:
This is a draft until https://github.com/DLR-RM/rl-baselines3-zoo/pull/257 is resolved.


setup.py
```diff
@@ -32,6 +32,7 @@
     "ray[debug,tune]~=0.8.5",
     "scipy>=1.8.0",
     "wandb",
+    "huggingface_sb3",
```
araffin, line 35: careful, need to wait for new release, last release still has the annoying `pickle5` dependency

ernestum:
I am a bit stuck right now and hope to get some input. Maybe @AdamGleave or @araffin can help?

The first issue is with the helpers in `imitation.policies.serialize`.
They impose a format in which the model is always stored in a file called `model.zip` which is not compatible with the format, that is imposed by huggingface.
I can either work around this in tests by moving files around and renaming them (ugly) or try to modify/remove the module `imitation.policies.serialize`. 
The thing is, that I don't know who is using `imitation.policies.serialize` outside of the `imitation` library.
Any preferences here @AdamGleave ?

Another problem I have in the preferences comparison implementation.
Its test fail because even though I provide a venv with 2 envs upon construction/loading,  for some reason it expects an 8 env venv during execution. I could not figure out why yet. It is really confusing and frustrating.

AdamGleave:
> The first issue is with the helpers in `imitation.policies.serialize`. They impose a format in which the model is always stored in a file called `model.zip` which is not compatible with the format, that is imposed by huggingface. I can either work around this in tests by moving files around and renaming them (ugly) or try to modify/remove the module `imitation.policies.serialize`. The thing is, that I don't know who is using `imitation.policies.serialize` outside of the `imitation` library. Any preferences here @AdamGleave ?

In general I'm not too concerned about making breaking changes as imitation is not yet at a v1 release, more important that we move towards a good API than keeping consistency.

How are you intending for the HuggingFace loading to work? `serialize.py` maintains a registry of loaders `serialize.policy_registry`, so one way would just be to register a new loader `hugging_face` (or perhaps `hugging_face_ppo`, `hugging_face_sac`, etc) and have the path be the ID of the HF model. In that case there's no need to even use `load_stable_baselines_model` I don't think -- it's not doing much, anyway.

To clarify we do want to at least maintain the option of loading things from local disk, so I wouldn't get rid of `imitation/.policies.serialize` in favor of hugging face, although if there's a way to simplify the serialize code that'd be great.

> Another problem I have in the preferences comparison implementation. Its test fail because even though I provide a venv with 2 envs upon construction/loading, for some reason it expects an 8 env venv during execution. I could not figure out why yet. It is really confusing and frustrating.

What command can I run to replicate this failure? In CircleCI logs I see some scripts failing, but nothing that seemed preference comparison specific.

I've seen policies and reward networks sometimes have the number of envs get baked into their expected observation/action shape in the past. Although I thought that was no longer an issue with SB3 & PyTorch imitation.

src/imitation/scripts/common/demonstrations.py
```diff
@@ -27,36 +27,17 @@ def fast():
...
-        updates["rollout_path"] = os.path.join(
-            guess_expert_dir(data_dir, env_name),
-            "rollouts",
-            "final.pkl",
-        )
-    return updates
-
-
 @demonstrations_ingredient.capture
-def load_expert_trajs(
```
AdamGleave, line 52: I'd like to maintain the option of loading rollouts from disk. Otherwise this feels like a regression in functionality, e.g. we'd have no way on training from demonstrations provided by a human being rather than a synthetic policy.

src/imitation/scripts/common/experts.py
```diff
@@ -0,0 +1,18 @@
+from huggingface_sb3 import load_from_hub
```
AdamGleave, line 1: Should `experts` be its own ingredient, or rolled into `demonstrations`?

ernestum:
> What command can I run to replicate this failure? In CircleCI logs I see some scripts failing, but nothing that seemed preference comparison specific.

Seems like the CI tests timed out. I can reproduce it locally using:

```bash
pytest tests/test_scripts.py::test_train_preference_comparisons_main 
```

ernestum:
I see it makes sense to integrate the huggingface loading logic into `serialize.py`, especially if it is ok to make breaking changes.
When loading from huggingface we pull the model zip file to disk and then use the normal `.load()` method of the algorithm anyways quite similar to how we load in `serialize.py` so with some refactoring I should be able to find a satisfying solution that has no duplicated code.

ernestum:
Concerning demonstrations: I agree we still want to load demonstrations from disk. 
Right now there is no canonical way to upload RL-demonstrations to huggingface hub but there is the notion of [datasets on the hub](https://huggingface.co/docs/hub/datasets) so I hope to integrate that feature into `huggingface_sb3` soon.
In the meantime I will keep the option to load demonstrations from disk, or have them created ad-hoc from a model that was pulled from the hub.

AdamGleave:
> Concerning demonstrations: I agree we still want to load demonstrations from disk. Right now there is no canonical way to upload RL-demonstrations to huggingface hub but there is the notion of [datasets on the hub](https://huggingface.co/docs/hub/datasets) so I hope to integrate that feature into `huggingface_sb3` soon. In the meantime I will keep the option to load demonstrations from disk, or have them created ad-hoc from a model that was pulled from the hub.

Being able to download demos from HuggingFace would be cool but even so I think good to keep ability to load from disk. I don't want to force users to use a third-party vendor, and there are some scenarios where data might be sensitive and people might not want to store it in the cloud.

ejnnr:
> I've seen policies and reward networks sometimes have the number of envs get baked into their expected observation/action shape in the past. Although I thought that was no longer an issue with SB3 & PyTorch imitation.

I encountered that issue with SB3+pytorch as well a while ago but never got around to debugging it. I just used 8 envs for the test as a workaround, see
https://github.com/HumanCompatibleAI/imitation/blob/301e7a5ca27ce069242fe7e4426feffa0089fa12/tests/test_scripts.py#L89-L92

@ernestum Looks like this PR removes that part of the test config, I'd guess that's what causes the error? Though of course it would be nice to fix the underlying bug instead

AdamGleave:
> > What command can I run to replicate this failure? In CircleCI logs I see some scripts failing, but nothing that seemed preference comparison specific.
> 
> Seems like the CI tests timed out. 

Worth looking into what's causing the timeouts as well. Hopefully not an actual slow down from downloading stuff from HF? (If so we could always cache it.)

> I can reproduce it locally using:
> 
> ```shell
> pytest tests/test_scripts.py::test_train_preference_comparisons_main 
> ```

OK, I can replicate this, and also with `python -m imitation.scripts.train_preference_comparisons with load_expert=True common.num_vec=2`.

The issue appears to be that `self.algorithm.rollout_buffer.n_envs` is 8 even though `self.algorithm.n_envs` and the actual number of environments is 2. I assume `PPO.load` is restoring the rollout buffer. This seems like an issue that needs to be fixed upstream.

Separately, I will note that using a pre-trained agent with preference comparisons is a bit of an edge case. Seems nice to have support for it for debugging, or for building things like [learning from demonstrations and human preferences](https://arxiv.org/abs/1811.06521) (IRL followed by preference comparison), but it's not the vanilla algorithm.

ernestum (commented):
Regarding: bfa8831
@AdamGleave I had to remove this section since it made the test for the `benachmark_and_table.sh` experiment fail.
Specifically it failed when running `imit_benchmark.sh` for the cheetah environment.

What was it initially supposed to do? It seems to abort the script when some log files don't include the term "Result". But what is the consequence? What does it mean if the check fails and what are the required actions?
Is it safe to remove this section?

ernestum:
Right now the remaining failing tests are `test_scripts.py::test_analyze_imitation` and `test_scripts.py::test_convert_trajs_in_place`.
Before I put time into understanding and debugging those I wanted to verify, that  they still used?

`test_convert_trajs_in_place` seems to fail since the analyze script returns no results, which is again due to the `_gather_sacred_dicts` helper function not returning anything. @AdamGleave do you have a hint for me where to look for the issue before I dive into this on my own?

AdamGleave:
> What was it initially supposed to do? It seems to abort the script when some log files don't include the term "Result". But what is the consequence? What does it mean if the check fails and what are the required actions?
Is it safe to remove this section?

When our training scripts complete successfully, they output a line in the logs like (from running `python -m imitation.scripts.train_adversarial gail with fast rl.fast train.fast common.fast`):
```
INFO - train_adversarial - Result: {'imit_stats': {'n_traj': 2, 'monitor_return_len': 2, 'return_min': 3.2304581999778748, 'return_mean': 3.4182530641555786, 'return_std': 0.18779486417770386, 'return_max': 3.6060479283332825, 'len_min': 5, 'len_mean': 5.0, 'len_std': 0.0, 'len_max': 5, 'monitor_return_min': 5.0, 'monitor_return_mean': 5.0, 'monitor_return_std': 0.0, 'monitor_return_max': 5.0}, 'expert_stats': {'n_traj': 40, 'return_min': 500.0, 'return_mean': 500.0, 'return_std': 0.0, 'return_max': 500.0, 'len_min': 500, 'len_mean': 500.0, 'len_std': 0.0, 'len_max': 500}}
```
This is output by Sacred automatically, from the return value of the command executed (in this case, the function `train_adversarial`).

So, the lines you removed in https://github.com/HumanCompatibleAI/imitation/commit/bfa8831aa56c85185bb8e5999b48f1d9906c23ed were going through all the log directories and searching for the results. It's not a check so much as about printing out relevant results -- although if it fails, that does indeed suggest something's off.

> Right now the remaining failing tests are `test_scripts.py::test_analyze_imitation` and `test_scripts.py::test_convert_trajs_in_place`. Before I put time into understanding and debugging those I wanted to verify, that they still used?

`convert_trajs_in_place` seems useful still, although we should update the documentation. We switched to an npz style format for trajectories in https://github.com/HumanCompatibleAI/imitation/pull/448 so having a script to convert from the old-style pickle format is handy.

`analyze` produces CSV/TeX tables of results from benchmarking. I've never been that happy with how it works, so I'm happy to see it replaced, but it's better than nothing so I wouldn't want to delete it without a plan to improve on it.

> `test_convert_trajs_in_place` seems to fail since the analyze script returns no results, which is again due to the `_gather_sacred_dicts` helper function not returning anything. @AdamGleave do you have a hint for me where to look for the issue before I dive into this on my own?

I tried to run `experiments/imit_benchmark.sh --fast` on your branch to see what happened. As far as I can tell the problem is just the scripts it's running fail? For example, `output/imit_benchmark/2022-07-07T16\:39\:16-07\:00/cartpole_0/n_expert_demos_1/sacred/run.json` (timestamp will be different on your machine but otherwise the logs should still exist) shows this traceback:

```
  "fail_trace": [
    "Traceback (most recent call last):\n",
    "  File \"/home/adam/dev/imitation/venv/lib/python3.8/site-packages/sacred/config/captured_function.py\", line 42, in captured_function\n    result = wrapped(*args, **kwargs)\n",
    "  File \"/home/adam/dev/imitation/src/imitation/scripts/train_adversarial.py\", line 155, in gail\n    return train_adversarial(algo_cls=gail_algo.GAIL)\n",
    "  File \"/home/adam/dev/imitation/venv/lib/python3.8/site-packages/sacred/config/captured_function.py\", line 42, in captured_function\n    result = wrapped(*args, **kwargs)\n",
    "  File \"/home/adam/dev/imitation/src/imitation/scripts/train_adversarial.py\", line 112, in train_adversarial\n    expert_policy = load_expert_policy(env_name)\n",
    "  File \"/home/adam/dev/imitation/src/imitation/scripts/common/experts.py\", line 15, in load_expert_policy\n    load_from_hub(\n",
    "  File \"/home/adam/dev/imitation/venv/lib/python3.8/site-packages/huggingface_sb3/load_from_hub.py\", line 16, in load_from_hub\n    downloaded_model_file = hf_hub_download(\n",
    "  File \"/home/adam/dev/imitation/venv/lib/python3.8/site-packages/huggingface_hub/file_download.py\", line 1020, in hf_hub_download\n    _raise_for_status(r)\n",
    "  File \"/home/adam/dev/imitation/venv/lib/python3.8/site-packages/huggingface_hub/utils/_errors.py\", line 78, in _raise_for_status\n    raise RepositoryNotFoundError(\n",
    "huggingface_hub.utils._errors.RepositoryNotFoundError: 401 Client Error: Repository Not Found for url: https://huggingface.co/ernestumorga/ppo-CartPole-v1/resolve/main/ppo-CartPole-v1.zip. If the repo is private, make sure you are authenticated. (Request ID: N4BPRkJkdBtp8hwuZOKtG)\n"
  ],
```

But perhaps this would work if I was logged in as you...? Still, that log directory seems like a good place to start.

ernestum:
I can easily make the `test_convert_trajs_in_place` case pass by adding back the demonstrations (`final.pkl`) in the testdata folder.
However, I figured out that the `final.pkl` in the current master branch already seems to be in the new format. 
At least the diff between that file and the one created in the temp directory during testing indicates so.
Therefore the test is pointless unless we get hold of such a trajectory in the old format.
Do you know where to get one @AdamGleave ?

AdamGleave:
> I can easily make the `test_convert_trajs_in_place` case pass by adding back the demonstrations (`final.pkl`) in the testdata folder. However, I figured out that the `final.pkl` in the current master branch already seems to be in the new format. At least the diff between that file and the one created in the temp directory during testing indicates so. Therefore the test is pointless unless we get hold of such a trajectory in the old format. Do you know where to get one @AdamGleave ?

Thanks for flagging this. I think I removed the backwards-compatibility logic in `load` in https://github.com/HumanCompatibleAI/imitation/pull/337 but forgot to remove the script. However, this script `convert_trajs_in_place.py` has a new purpose now that https://github.com/HumanCompatibleAI/imitation/pull/448 is merged. On `master` it does produce different output:

```bash
$ python -m imitation.scripts.convert_trajs_in_place tests/testdata/expert_models/cartpole_0/rollouts/final.pkl 
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   tests/testdata/expert_models/cartpole_0/rollouts/final.pkl
$ file tests/testdata/expert_models/cartpole_0/rollouts/final.pkl  # note it was pkl before
tests/testdata/expert_models/cartpole_0/rollouts/final.pkl: Zip archive data, at least v2.0 to extract, compression method=deflate
```

That said, the documentation in the script is out of date, and it should probably rename pkl->zip to avoid confusion. I'll make a PR addressing that.

Note this branch is quite a bit behind `master` at this point, you may want to merge it in so you can see these changes...

src/imitation/scripts/common/expert.py
```diff
@@ -0,0 +1,51 @@
...
+        return policy_path
+    else:
+        model_name = ModelName(policy_type, EnvironmentName(common["env_name"]))
+        return load_from_hub(get_huggingface_repo_id(), model_name.filename)
+
+
+@expert_ingredient.capture
+def get_expert_policy(policy_type) -> policies.BasePolicy:
+    env = make_venv()
+    if policy_type == "ppo":
```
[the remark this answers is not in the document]
AdamGleave: Looking further I see you've ripped out all the calls to `load_policy`. Unfortunately that leads to a fairly substantial regression in functionality. E.g. we no longer support SAC, although that's easy to fix. More significantly, there's no way for an end user to define new policy types, unlike with the policy registry.

I'm all for simplifying this, `imitation.policies.serialize` was a bit of a mess, but I fear this might be too simple. Ideally the logic for deserialization would also not be tightly coupled with the scripts, so that people can load policies programatically elsewhere.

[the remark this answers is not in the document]
AdamGleave: Curious to hear others thoughts e.g. @norabelrose @scottemmons @yawen-d in terms of features we'd like to have for loading policies.

tests/scripts/test_scripts.py
```diff
@@ -52,11 +52,9 @@
 
 TEST_DATA_PATH = pathlib.Path("tests/testdata")
 CARTPOLE_TEST_DATA_PATH = TEST_DATA_PATH / "expert_models/cartpole_0/"
-CARTPOLE_TEST_ROLLOUT_PATH = CARTPOLE_TEST_DATA_PATH / "rollouts/final.pkl"
 CARTPOLE_TEST_POLICY_PATH = CARTPOLE_TEST_DATA_PATH / "policies/final"
 
 PENDULUM_TEST_DATA_PATH = TEST_DATA_PATH / "expert_models/pendulum_0/"
```
AdamGleave, line 57: Is this used anywhere now?

tests/scripts/test_scripts.py
```diff
@@ -95,9 +93,6 @@ def test_main_console(script_mod):
 
 PREFERENCE_COMPARISON_CONFIGS = [
     {},
-    {
-        "trajectory_path": CARTPOLE_TEST_ROLLOUT_PATH,
```
AdamGleave, line 99: Are we no longer testing ability to load rollouts from a file in script? Might be good to retain some coverage of that (new key `demonstrations.rollout_path` I believe)

tests/scripts/test_scripts.py
```diff
@@ -232,93 +225,84 @@ def test_train_dagger_main(tmpdir):
     assert isinstance(run.result, dict)
 
 
-def test_train_dagger_warmstart(tmpdir):
+def test_train_bc_main(tmpdir):
```
AdamGleave, line 228: This seems like it's testing a lot less than before -- why? Are we no longer able to warmstart runs in DAgger?

tests/scripts/test_scripts.py
```diff
@@ -232,93 +225,84 @@ def test_train_dagger_main(tmpdir):
...
 
-def test_train_bc_main(tmpdir):
+def test_train_bc_main_with_expert_path(tmpdir):
     run = train_imitation.train_imitation_ex.run(
         command_name="bc",
-        named_configs=["cartpole"] + ALGO_FAST_CONFIGS["imitation"],
+        named_configs=["seals_cartpole"] + ALGO_FAST_CONFIGS["imitation"],
         config_updates=dict(
             common=dict(log_root=tmpdir),
-            demonstrations=dict(rollout_path=CARTPOLE_TEST_ROLLOUT_PATH),
```
AdamGleave, line 291: I'm OK with the test as-written but I do want to flag that loading rollouts directly is the *most realistic* use case as it's the only option for human demos (even if for quick experiments we more often use synthetic data). So, I don't want rollouts to be treated as a second-class citizen. Ideally we have tests for both (could parametrize fairly easily, or just mix-and-match).

tests/scripts/test_scripts.py
```diff
@@ -232,93 +225,84 @@ def test_train_dagger_main(tmpdir):
...
+            expert=dict(policy_path=CARTPOLE_TEST_POLICY_PATH / "model.zip"),
         ),
     )
     assert run.status == "COMPLETED"
     assert isinstance(run.result, dict)
 
 
-def test_train_bc_warmstart(tmpdir):
+def test_train_bc_main_with_huggingface_repo_id(tmpdir):
```
AdamGleave, line 263: Basically identical to above test, might be good use for `@pytest.mark.parametrize`?

tests/scripts/test_scripts.py
```diff
@@ -232,93 +225,84 @@ def test_train_dagger_main(tmpdir):
...
+            ),
         ),
     )
     assert run.status == "COMPLETED"
+    assert isinstance(run.result, dict)
+
 
-    policy_path = pathlib.Path(run.config["common"]["log_dir"]) / "final.th"
-    run_warmstart = train_imitation.train_imitation_ex.run(
+def test_train_bc_main_with_random_expert(tmpdir):
```
AdamGleave, line 278: Also very similar to above, parametrize?

tests/scripts/test_scripts.py
```diff
@@ -232,93 +225,84 @@ def test_train_dagger_main(tmpdir):
...
+            ),
         ),
     )
+    assert run.status == "COMPLETED"
+    assert isinstance(run.result, dict)
 
-    assert run_warmstart.status == "COMPLETED"
-    assert isinstance(run_warmstart.result, dict)
+
+def test_train_bc_main_with_zero_expert(tmpdir):
```
AdamGleave, line 293: Also very similar to above, parametrize?

tests/scripts/test_scripts.py
```diff
@@ -436,35 +421,6 @@ def test_train_adversarial(tmpdir, named_configs, command):
     _check_train_ex_result(run.result)
 
 
-@pytest.mark.parametrize("command", ("airl", "gail"))
-def test_train_adversarial_warmstart(tmpdir, command):
```
AdamGleave, line 440: Why are we removing this test? Do we no longer support warmstarting?

tests/scripts/test_scripts.py
```diff
@@ -563,60 +500,6 @@ def test_transfer_learning(tmpdir: str) -> None:
...
 
 
-@pytest.mark.parametrize(
-    "named_configs",
-    (
-        [],
-        ["reward.reward_ensemble"],
-    ),
-)
-def test_preference_comparisons_transfer_learning(
```
AdamGleave, line 573: Why are we deleting this test?

AdamGleave (changes requested):
Thanks for working on this, it's exciting that we can now directly use experts from HuggingFace! This will make expert management a lot easier, and help users get off the ground quicker.

Most of my comments are minor. There's two main areas I'm hesitant about.

# 1) Serialization
We seem to be de-facto replacing `imitation.policies.serialize` by `imitation.scripts.common.expert`. But it's a bit of a regression in functionality. It can only be used in Sacred scripts; it doesn't support SAC (just PPO); and there's no way for end users to add support for new policy types. 

The lack of SAC support could actually break experiments @yawen-d is running. Now, it's a two-line fix to add that support, but overall this change feels a bit hasty. I'd feel better about building this functionality on top of `imitation.policies.serialize`. I think all we need is to add a flag to the registry to indicate whether or not the policy type requires a path (PPO, SAC, etc) or is hardcoded (random, none, etc) so that we know whether or not to try and download something from the hub (or maybe that could even be implicit based on whether an agent path/HuggingFace ID is passed in).

Long-term I think `imitation.policies.serialize` would benefit from being cleaned up, and you'd be welcome to work on that next. But seems best done in a separate PR, and in the mean time I don't want to lose core features.

# 2) Test coverage: warmstarts and rollouts

You were able to simplify the tests with the new structure which is great, but in some cases I think may have gone too far.

In particular it'd be nice to have some tests that still involve loading demos from a saved file of rollouts on disk. This is the use case we'll have if working with actual human demos. So, I'd be inclined to keep one of the pickle files around (possibly updating it to the new NPZ format).

Separately, all the tests involving warmstarting (train a policy with algo X; then load that policy again with fresh instance of algo X) seem to have been deleted. Why? This is a pretty common experimental workflow for us, so it'd be an issue if it's actually not supported any longer.

But looking at the code I think it should be supported: can't we just pass the policy trained by algo X in as an "expert"? If there are some issues (`BasePolicy` vs `BaseAlgorithm` is always annoying) it should still be fairly easy to workaround.

Please request a re-review once comments addressed. Heads up I may be a bit slow to respond Mon-Wed as traveling. Feel free to ask @norabelrose or others for a review.

tests/scripts/test_scripts.py
ernestum: You are right. My intention was to improve readability of the test outputs, but it turns out there is a smarter way to do this via configuration id's.

tests/scripts/test_scripts.py
ernestum: Good catch! I added it back!

tests/scripts/test_scripts.py
ernestum: This was an accident. Thanks for pointing it out. Should be back now.

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -102,7 +103,7 @@ def seals_ant():
 @train_adversarial_ex.named_config
 def half_cheetah():
     locals().update(**MUJOCO_SHARED_LOCALS)
-    common = dict(env_name="HalfCheetah-v2")
+    common = dict(env_name="seals/HalfCheetah-v0")
```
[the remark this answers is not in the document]
ernestum: `seals/HalfCheetah-v0` is the environment ID. I guess we can't change the env id. Do you suggest using/introducing a named sacred config?

src/imitation/scripts/common/demonstrations.py
```diff
@@ -27,25 +31,45 @@ def fast():
...
+        The expert trajectories.
+
+    Raises:
+        ValueError: If n_expert_demos is None.
+    """
+    if n_expert_demos is None:
+        raise ValueError("n_expert_demos must be specified when rollout_path is None")
+
+    rollout_env = DummyVecEnv(
+        [lambda: RolloutInfoWrapper(gym.make(common["env_name"])) for _ in range(4)],
```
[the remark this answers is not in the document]
ernestum: I think the rationale was to keep all modules as self contained as possible. Since generating the vec-env is just a 1-liner here I decided not to use any of the utility functions. 
I will have a look at the two utils you mentioned. But tbh just the fact that we have to similar utils scattered in different modules makes me feel uneasy.

experiments/imit_benchmark.sh
```diff
@@ -107,18 +105,12 @@ parallel -j 25% --header : --results "${LOG_ROOT}/parallel/" --colsep , --progre
...
-  demonstrations.data_dir="${DATA_DIR}" \
   demonstrations.n_expert_demos='{n_expert_demos}' \
   checkpoint_interval=0 \
   "${extra_configs[@]}" \
   :::: $CONFIG_CSV \
   ::: seed "${SEEDS[@]}"
 
-# Directory path is really long. Enter the directory to shorten results output,
-# which includes directory of each stdout file.
-pushd "${LOG_ROOT}/parallel"
```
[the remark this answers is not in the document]
ernestum: Apparently I removed it since it made some tests fail and I could not understand what it does. Right now I had another go at it but still could not figure out what those lines are good for. The comment seems to be unrelated and does not help in understanding. Do you have a hint for me @AdamGleave ?

What I did find out:
1. It makes `benchmark_and_table.sh --fast` fail (and therefore the experiment tests).
2. `benchmark_and_table.sh` fails on a call to `imit_benchmark.sh`
3. Just running `imit_benchmark.sh --fast` does not fail.
4. The line seems to ensure that certain result files or stdout files have been stored. In the case of the failing `benchmark_and_table.sh`, there is just one result file while `imit_benchmark.sh` itself generates two.

src/imitation/scripts/common/common.py
```diff
@@ -5,6 +5,7 @@
 from typing import Any, Mapping, Sequence, Tuple, Union
 
 import sacred
+import seals  # noqa: F401
```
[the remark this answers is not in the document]
ernestum: You are right. I removed it.

experiments/imit_benchmark.sh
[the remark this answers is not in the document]
AdamGleave: In general please don't delete lines just because it makes the tests pass ;) Do feel free to ask me or anyone that shows up in git blame if the purpose of code is unclear. Sometimes it really should be deprecated, but often it's important.

I thought we discussed what these lines do back in https://github.com/HumanCompatibleAI/imitation/pull/445#issuecomment-1178374293 The TL;DR is that these lines print out a summary of the results from all the scripts being run. This is useful logging output. If the lines fail, it's because there's no results, which should never happen.

You can probably figure out what's going on by looking through the logs. Either the output format has changed, or some of the Python scripts are failing silently (if they returned non-zero exit code this should be passed through by parallel and fail earlier).

If you have trouble figuring this out let me know and I can take a look if you send me which tests are failing.

src/imitation/scripts/common/demonstrations.py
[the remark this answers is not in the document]
AdamGleave: My point is done right the environment creation *isn't* a one-liner. This doesn't seed the environments, so they could all end up producing identical rollouts. Better to use a utility method that handles subtleties like this.

`util.make_vec_env` is a programmatic API that can be used from any Python code. `common.make_venv` is a Sacred capture function designed to only be called from scripts that use that ingredient, and is a light wrapper around `util.make_vec_env`. IMO it seems reasonable to have both, though happy to consider refactors.

src/imitation/scripts/config/train_adversarial.py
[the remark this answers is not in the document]
AdamGleave: Adding a named config `seals_half_cheetah` SGTM.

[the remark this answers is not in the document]
ernestum: Fixed in [0303a85](https://github.com/HumanCompatibleAI/imitation/pull/445/commits/0303a8591fabdca8e04c3aeacc23aec35090a055)

tests/scripts/test_scripts.py
ernestum: I added back the test here: [8c3a7a9](https://github.com/HumanCompatibleAI/imitation/pull/445/commits/8c3a7a95c5c700e5a52badee148ec61e448bf167)

experiments/imit_benchmark.sh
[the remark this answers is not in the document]
ernestum: You are right, deleting lines to make tests pass is not a very smart move :grimacing: . I also deleted it since I simply could not figure out what it is there fore. So I dug some deeper into the following line:

```bash
find . -name stdout -print0 | sort -z | xargs -0 tail -n 15 | grep -E '==|Result'
```

My best guess for it's intended action is: for each `stdout` file in the `parallel` subdir, check if it contains the string "Result" in the last 15 lines.
What I don't get is 1. why `sort`? and 2. why match for the string `==` OR for `Result`.
The sorting is meaningless but matching for `==` actually brings us into trouble later.

The first issue with this is that the `Result` seems to occur in `stderr` instead of `stdout`. Ok that is an easy fix; lets also remove the `sort`:

```bash
find . -name stderr -print0 | xargs -0 tail -n 15 | grep -E '==|Result'
```
Now as I understand the documentation of `find -print0` and `xargs -0`, the command executed in a directory like this:
```
.
└── env_config_name
    ├── seals_cartpole
    │   └── n_expert_demos
    │       └── 40
    │           └── seed
    │               └── 0
    │                   ├── seq
    │                   ├── stderr
    │                   └── stdout
    └── seals_mountain_car
        └── n_expert_demos
            └── 40
                └── seed
                    └── 0
                        ├── seq
                        ├── stderr
                        └── stdout
```
**should** result in two independent invocations of `tail`. What actually happens (you can verify this by enabling interactive mode of `xargs` with the `-p` flag) is that tail is executed once with the two files as an argument:
```bash
tail -n 15 ./env_config_name/seals_mountain_car/n_expert_demos/40/seed/0/stderr ./env_config_name/seals_cartpole/n_expert_demos/40/seed/0/stderr
```
This will output (with `-n 5` for brevity):
```
==> ./env_config_name/seals_mountain_car/n_expert_demos/40/seed/0/stderr <==
INFO - imitation.scripts.train_adversarial - Using '<class 'imitation.algorithms.adversarial.airl.AIRL'>' algorithm
round: 100%|██████████| 5/5 [00:00<00:00, 20.37it/s]
INFO - root - Saved policy to output/imit_benchmark/2022-08-26T12:42:38+02:00/seals_mountain_car_0/n_expert_demos_40/checkpoints/final/gen_policy
INFO - test-2022-08-26T12:42:24+02:00 - Result: {'imit_stats': {'n_traj': 2, 'monitor_return_len': 2, 'return_min': -4.088658452033997, 'return_mean': -1.3765016160905361, 'return_std': 2.7121568359434605, 'return_max': 1.3356552198529243, 'len_min': 5, 'len_mean': 5.0, 'len_std': 0.0, 'len_max': 5, 'monitor_return_min': -5.0, 'monitor_return_mean': -5.0, 'monitor_return_std': 0.0, 'monitor_return_max': -5.0}, 'expert_stats': {'n_traj': 40, 'return_min': -200.0, 'return_mean': -200.0, 'return_std': 0.0, 'return_max': -200.0, 'len_min': 200, 'len_mean': 200.0, 'len_std': 0.0, 'len_max': 200}}
INFO - test-2022-08-26T12:42:24+02:00 - Completed after 0:00:02

==> ./env_config_name/seals_cartpole/n_expert_demos/40/seed/0/stderr <==
INFO - imitation.scripts.train_adversarial - Using '<class 'imitation.algorithms.adversarial.airl.AIRL'>' algorithm
round: 100%|██████████| 5/5 [00:00<00:00, 20.15it/s]
INFO - root - Saved policy to output/imit_benchmark/2022-08-26T12:42:38+02:00/seals_cartpole_0/n_expert_demos_40/checkpoints/final/gen_policy
INFO - test-2022-08-26T12:42:24+02:00 - Result: {'imit_stats': {'n_traj': 2, 'monitor_return_len': 2, 'return_min': 0.24264731258153915, 'return_mean': 5.492057155817747, 'return_std': 5.249409843236208, 'return_max': 10.741466999053955, 'len_min': 5, 'len_mean': 5.0, 'len_std': 0.0, 'len_max': 5, 'monitor_return_min': 5.0, 'monitor_return_mean': 5.0, 'monitor_return_std': 0.0, 'monitor_return_max': 5.0}, 'expert_stats': {'n_traj': 40, 'return_min': 500.0, 'return_mean': 500.0, 'return_std': 0.0, 'return_max': 500.0, 'len_min': 500, 'len_mean': 500.0, 'len_std': 0.0, 'len_max': 500}}
INFO - test-2022-08-26T12:42:24+02:00 - Completed after 0:00:04
```
Note the lines like `==> ./env_config_name/seals_cartpole/n_expert_demos/40/seed/0/stderr <==`! They contain `==`! This is just `tail`, printing the file name surrounded by `==> FILENAME <==` as a header as soon as you pass multiple files.

So the current version of the command produces false positives due to two reasons:
1. As soon as more than one `stderr` file is found by `find`, `tail` produces the `==> FILENAME <==` headers, which are matched by `grep -E '==|Result'` no matter the content of the `stderr` files.
2. Even if we fixed 1. by changing to `grep -E 'Result'`, there is only one execution of of `grep`. So if only **one single** of the `stderr` files contains a `'Result'`, the test is positive.

Maybe I missed the intention of the bash command. Maybe there are ways to make it work in a meaningful way. But I think that is not the right way to go.
If one of the runs fails, that should be reflected in return code of `python -m imitation.scripts.train_adversarial` and the return code will be propagated by `parallel`.
If we **must** (for whatever reason) ensure, that successful training is indicated by a "Result" string in the `stdout` file, then that is a case for a unit test and not a bash one-liner.

So I propose:
1. To still remove the lines but add a unit test to check if there is `Result` in `stderr` iff the return code is 0.
2. Avoid setting up experiments as bash scripts in the future and minimize the number of bash scripts in the repo. They are hard to understand, easy to mess up and hard to debug.

What do you think @AdamGleave ?

src/imitation/scripts/common/expert.py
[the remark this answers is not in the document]
ernestum: I tried to make the expert ingredient use the `serialize` module again but then we do really run into the issues with the degenerate cases (zero/random policies) and it becomes messy again.

In the end we want the functionality to "turn a bunch of strings (model type, path, hf-repo-id, hf-orga) into some policy". Right now this is scattered between the expert ingredient and `serialize`. We need to if we either want to place all of it in one or the other:

## Option 1: Everything goes to `serialize`:

In this case the expert ingredient becomes a really thin wrapper around `serialize`. It is just about plugging the right configuration values into the right loader function.
In addition to the `ppo` and `sac` type that we register in the registry right now, we would also add `ppo-huggingface` and `sac-huggingface` types. The path, that is passed for the factory function then becomes the huggingface repo id or the env id (we have to figure out what is most convenient).
**Pro:** Powerful model loading features, that can be used without sacred while still being extendable by the user.
**Con:** `serialize` becomes more complex than it already is.

## Option 2: Everything goes to the expert ingredient

In this case the `serialize` module would become so thin, that we might even be able to remove it entirely (no registry would be needed any more).
To let the user extend the types of loadable experts, they could add new sub-ingredients to the expert ingredient:

```python
@expert_ingredient.config
def config():
    type: str = None  # The type of expert to load

@expert_ingredient.capture
def load(type):
    for i in expert_ingredient.ingredients:
        if i.path == type:
            return i.load()

@ppo_expert_ingredient.config
def config():
    path: str = None # The path to load the PPO policy from

@ppo_expert_ingredient.config
def load(path):
    return PPO.load(path).policy

expert_ingrdient.ingrdients.append(ppo_expert_ingrdient)
```

**Pros:** 
- Nudge users to use sacred and reuse sacred ingredients. 
- New policy types can introduce their own configuration values without changing the code in `imitation`. 
- No more degenerate cases (e.g. random policy which has to ignore the path).
**Con:** 
- We enforce the usage of sacred
- Adding a new policy type requires writing a new sacred ingredient which is harder than just registering a callback.

Which solution would you prefer @AdamGleave , @norabelrose , @scottemmons , @yawen-d ?

I would be in favor of Option 2, since it seems less hacky. But maybe you have other options in mind?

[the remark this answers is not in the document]
AdamGleave: I think there's two key questions here:
  1. Do we want the serialization to work just from scripts, or also from ad-hoc Python code?
  2. Do we want to tie ourselves more deeply to Sacred specifically?

For 1) I lean slightly towards allowing ad-hoc Python code, for research frameworks flexibility is valuable as hard to anticipate use cases. But I don't feel strongly about this, ["There's Only One Way To Do it"](https://wiki.python.org/moin/TOOWTDI) has its benefits too for consistency across projects.

For 2) though I'd really like to avoid having Sacred too deeply integrated. The idea of Sacred was great. But at this point it's largely unmaintained. There are known serious issues with it that would require a significant redesign to fix and seem unlikely to occur. At some point I'd like to survey the other options and potentially migrate to something else, although there's nothing that's obviously better. Still, this pushes me towards trying to keep the Sacred side as thin as possible.

Would definitely appreciate other people's thoughts here!

I do like the design of option 2 allowing you to specify different configs for different policy types. Maybe we can keep that part of the design even with serialize.py; i.e. move from a type signature of `load_policy(type: str, path: str)` to `load_policy(type: str, **kwargs)`?

experiments/imit_benchmark.sh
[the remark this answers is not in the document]
AdamGleave: Thanks for the detailed analysis! This seems broadly correct. However, I think the point of these lines was not purely to check for successful execution (I agree we should check return code of parallel for that), but also to return to the user a convenient summary. That's the reason why `sort -z` was included (so that you get results in a consistent order, that's more user parseable), and why `==` was included in the `grep` (so you can see which file each one is coming from).

It sounds like Sacred changed where it prints Result from stdout to stderr, and we can fix this just by changing to `stderr`. I'd be inclined to do that, and add some comments documenting what this line is doing, since it was clearly easier to write than to read :roll_eyes:

I do agree we should eventually try to burn these bash files and replace them with Python, and that'll allow for much prettier presentation of results...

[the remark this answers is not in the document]
ernestum: Ok as a "printing the results" feature this totally makes sense. I added it back with the `stderr` fix and a comment, that hopefully makes it easier to understand in the future.

tests/scripts/test_scripts.py
ernestum: I added that case back again.

tests/scripts/test_scripts.py
ernestum: I added this case back.

tests/scripts/test_scripts.py
ernestum: Yes it is.

src/imitation/scripts/common/expert.py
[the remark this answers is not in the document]
ernestum: Turns out that my proposed option 2 was not so easy to implement after all. To avoid too much entanglement with `sacred` I implemented your proposal about modifying the interface of `load_policy`. I am happy with the solution, `expert.py` became rather short now. Am happy to have your eyes on this again @AdamGleave

ernestum:
The remaining failing test is for `benchmark_and_table.sh`. It fails since it calls `bc_benchmark.sh` with the `--paper` option. This will try to train a `seals_half_cheetah` policy which fails since mujoco has been removed in #515 . 
_No I am quite confused:_
Why did this not make the pipeline fail earlier?
Answer: After #515, the Docker image has not been updated even though the Dockerfile has been changed. Therefore mujoco was/is still present in the (now stale) Dockerfile.
But then: why is mujoco gone now? Since the test fails with the usual
```
Exception: 
You appear to be missing MuJoCo.  We expected to find the file here: /root/.mujoco/mujoco210

This package only provides python bindings, the library must be installed separately.

Please follow the instructions on the README to install MuJoCo

    https://github.com/openai/mujoco-py#install-mujoco

Which can be downloaded from the website

    https://www.roboti.us/index.html
```
message.

src/imitation/policies/serialize.py
```diff
@@ -72,14 +79,43 @@ def _load_stable_baselines(
...
         return getattr(model, "policy")
 
     return f
 
 
+def _load_stable_baselines_from_huggingface(
+    algo_name: str,
+    cls: Type[base_class.BaseAlgorithm],
+) -> PolicyLoaderFn:
+    """Higher-order function, returning a policy loading function.
```
AdamGleave, line 94: This docstring is almost identical to the one for `_load_stable_baselines_from_file`: add one or two sentences to the docstrings?

src/imitation/policies/serialize.py
```diff
@@ -90,42 +126,59 @@ def f(path: str, venv: vec_env.VecEnv) -> policies.BasePolicy:
...
+    **kwargs,
 ) -> policies.BasePolicy:
     """Load serialized policy.
 
+    Note on the kwargs:
+
+    - `zero` and `random` policy take no kwargs
+    - `ppo` and `sac` policies take a `path` argument with a path to a zip file or to a
+      folder containing a `model.zip` file.
+    - ppo-huggingface and sac-huggingface policies take an `env_name` and optional
```
AdamGleave, line 163:
```suggestion
    - `ppo-huggingface` and `sac-huggingface` policies take an `env_name` and optional
```

src/imitation/scripts/common/demonstrations.py
```diff
@@ -27,25 +31,45 @@ def fast():
...
+    )
+    return rollout.rollout(
+        expert.get_expert_policy(),
+        rollout_env,
+        rollout.make_sample_until(min_timesteps=2000, min_episodes=n_expert_demos),
```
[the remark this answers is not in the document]
AdamGleave: (Bump on this comment.)

src/imitation/scripts/common/demonstrations.py
```diff
@@ -62,11 +61,14 @@ def generate_expert_trajs(
     if n_expert_demos is None:
         raise ValueError("n_expert_demos must be specified when rollout_path is None")
 
-    rollout_env = DummyVecEnv(
-        [lambda: RolloutInfoWrapper(gym.make(common["env_name"])) for _ in range(4)],
+    rollout_env = vec_env.DummyVecEnv(
```
AdamGleave, line 64: I thought I commented about this before but can't see it now -- don't we need to seed the environments individually, to make it (a) replicable, (b) uncorrelated between environments?

Would suggest using an existing helper method from `util` for this rather than rolling our own.

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -102,6 +102,35 @@ def seals_ant():
 
 @train_adversarial_ex.named_config
 def half_cheetah():
+    locals().update(**MUJOCO_SHARED_LOCALS)
```
AdamGleave, line 105: A shame there's so much duplication between this and the seals HalfCheetah config below. Factor out common config parameters (I think everything except `common=dict(env_name=...)`) into a dictionary e.g. `HALFCHEETAH_SHARED_LOCALS` and use the `locals().update` trick?

AdamGleave:
> The remaining failing test is for `benchmark_and_table.sh`. It fails since it calls `bc_benchmark.sh` with the `--paper` option. 

It looks like it was skipped before, because it was in the `USES_FULL_ROLLOUTS` tuple in `test_experiments.py`. You removed that check as part of the port, so it's no longer skipped.

I've pushed a commit that resolves this by having `benchmark_and_table.sh` no longer pass the `--paper` flag to `bc_benchmark.sh` and `dagger_benchmark.sh` when `--fast` is specified. I've also taken the liberty of removing `mujoco-py` from `setup.py` test requirements as I think it's no longer needed given this change (and will break on CI anyway as you observed).

I've also updated the Docker image on DockerHub now -- thanks for catching that.

tests/scripts/test_scripts.py
```diff
@@ -132,11 +105,43 @@ def test_main_console(script_mod):
...
+            "gatherer_kwargs": {"sample": False},
+            "agent_path": CARTPOLE_TEST_POLICY_PATH,
+            # FIXME(yawen): the policy we load was trained on 8 parallel environments
+            #  and for some reason using it breaks if we use just 1 (like would be the
+            #  default with the fast named_config)
+            "common": dict(num_vec=8),
+        },
+        with_checkpoints={
+            "checkpoint_interval": 1,
+            # Test that we can save checkpoints
```
AdamGleave, line 134: move comment to line above this?

tests/scripts/test_scripts.py
```diff
@@ -132,11 +105,43 @@ def test_main_console(script_mod):
...
-def test_train_preference_comparisons_main(tmpdir, config):
+@pytest.fixture(
+    params=[
+        "plain",
+        "with_expert_trajectories",
+        "warmstart",
+        "with_checkpoints",
+    ],
+)
+def preference_comparison_config(request):
```
AdamGleave, line 116: nice idea to give the configs explanatory names!

tests/scripts/test_scripts.py
```diff
@@ -132,11 +105,43 @@ def test_main_console(script_mod):
...
+    )[request.param]
+
+
+def test_train_preference_comparisons_main(tmpdir, preference_comparison_config):
     config_updates = dict(common=dict(log_root=tmpdir))
-    sacred.utils.recursive_update(config_updates, config)
+    sacred.utils.recursive_update(config_updates, preference_comparison_config)
     run = train_preference_comparisons.train_preference_comparisons_ex.run(
+        # Note: we have to use the cartpole and not the seals_cartpole config because
+        #  the seals_cartpole config needs a fixed horizon.
```
AdamGleave, line 144:
```suggestion
        #  the seals_cartpole config needs rollouts with a fixed horizon,
        #  and the saved trajectory rollouts are variable horizon.
```

tests/scripts/test_scripts.py
```diff
@@ -301,65 +282,87 @@ def test_train_dagger_warmstart(tmpdir):
...
     run = train_imitation.train_imitation_ex.run(
         command_name="bc",
-        named_configs=["cartpole"] + ALGO_FAST_CONFIGS["imitation"],
+        named_configs=["seals_cartpole"] + ALGO_FAST_CONFIGS["imitation"],
         config_updates=dict(
             common=dict(log_root=tmpdir),
             demonstrations=dict(rollout_path=CARTPOLE_TEST_ROLLOUT_PATH),
+            expert=dict(
+                policy_type="ppo-huggingface",
+                loader_kwargs=dict(env_id="seals/CartPole-v0"),
```
AdamGleave, line 355: How does `env_id` differ from the `env_name` automatically set in `expert.get_expert_policy`?

AdamGleave (commented):
I like the new design! Thanks for the persistence, I think this PR is now basically ready. I've left a few comments but most of them are minor -- please request re-review once addressed.

setup.py
```diff
@@ -60,6 +60,7 @@
         "scipy~=1.9.0",
         "wandb==0.12.21",
         "setuptools_scm~=7.0.5",
+        "mujoco-py~=2.1.2.14",
```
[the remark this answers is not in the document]
ernestum: You are right. I only added it to debug the mujoco pipeline issue.

tests/scripts/test_scripts.py
ernestum: thx

tests/scripts/test_scripts.py
ernestum: Good catch. This is a relic from when the `env_name` was not automatically set in the expert ingredient.
That is one of the drawbacks of not modelling the different loaders as sacred ingredients: we don't catch unused configuration parameters. I fixed it.

tests/scripts/test_scripts.py
ernestum: done

src/imitation/scripts/common/demonstrations.py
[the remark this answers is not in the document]
ernestum: Thanks for the bump. I can't remember why I added this magic number in the first place.
I fixed the underlying issue in `make_sample_until`: it declared both parameters as optional in the type annotations and had code handling this but was missing the `=None` default assignments to make them actually optional.

src/imitation/policies/serialize.py
ernestum: Done

src/imitation/scripts/config/train_adversarial.py
ernestum: I implemented the variant with the `locals().update()` for now for the sake of not expanding the scope of this PR any further.

However I don't like it since it is depends on knowing the smart hack that sacred uses to implement configs as functions. I feel it would be more hones to just use

```python
train_adversarial_ex.add_named_config("seals_half_cheetah", **CHEETAH_CONFIG, common = dict(env_name="HalfCheetah-v2"))
```

instead.

src/imitation/policies/serialize.py
```diff
@@ -42,25 +46,28 @@ def load_stable_baselines_model(
...
     """
     logging.info(f"Loading Stable Baselines policy for '{cls}' from '{path}'")
-    policy_dir = pathlib.Path(path)
-    if not policy_dir.is_dir():
+    path = pathlib.Path(path)
+
+    if path.is_dir():
+        path = path / "model.zip"
+
+    if not path.parent.is_dir():
```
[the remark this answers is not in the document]
ernestum: You are right. This happened during the migration of loading from a folder to loading from a file. I replaced this code with a more sensible one. Feel free to check if it makes more sense to you now @AdamGleave.

setup.py
ernestum: I think this is fixed by now.

src/imitation/scripts/common/demonstrations.py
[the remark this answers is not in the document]
ernestum: Good point. I used `common.make_venv` now.

src/imitation/scripts/common/demonstrations.py
ernestum: Yes you commented before. I fixed it  by using `common.make_venv()` instead.

src/imitation/scripts/common/experts.py
```diff
@@ -0,0 +1,18 @@
...
+from stable_baselines3 import PPO
+from stable_baselines3.common import policies
+
+
+def load_expert_policy(env_name: str, venv = None):
```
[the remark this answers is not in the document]
ernestum: As discussed in another thread, most functionality has been pushed to `imitation.policies.serialize`.

ernestum:
Not sure why the coverage suddenly dropped in a totally unrelated file.
The naive solution which would be just running a config of preference comparisons with the `active_selection` flag set fails for some reason. Before I dig deeper into this I am waiting for your thoughs @AdamGleave .

src/imitation/scripts/config/train_adversarial.py
AdamGleave: Using `locals().update()` just sets the local variables the same as a direct variable definition. So in that sense I don't think it requires any knowledge of Sacred internals beyond "Sacred somehow extracts the local variables" -- which is already needed to understand the function definition style in the first place.

I do think there's a case to be made for using `add_named_config` over function decorator as the default if we're not doing any actual computation in the function, as I agree the function decorator approach is unnecessarily magic. But IMO  if we do this we should make the change everywhere we can, not just for `locals().update`. Probably best addressed in a separate.

src/imitation/algorithms/bc.py
```diff
@@ -59,14 +59,14 @@ def batch_iterator():
...
+                for batch in self.batch_loader:
                     yield batch
+                    some_batch_was_yielded = True
 
-                if num_batches_in_epoch == 0:
+                if not some_batch_was_yielded:
                     raise AssertionError(
-                        f"Data loader returned no data after "
-                        f"{num_batches_in_epoch} batches, during epoch "
+                        f"Data loader returned no data after during epoch "
```
AdamGleave, line 69:
```suggestion
                        f"Data loader returned no data during epoch "
```

src/imitation/policies/serialize.py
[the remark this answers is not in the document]
AdamGleave: This makes sense now, thanks for making the change!

src/imitation/policies/serialize.py
```diff
@@ -91,7 +91,7 @@ def _load_stable_baselines_from_huggingface(
     algo_name: str,
     cls: Type[base_class.BaseAlgorithm],
 ) -> PolicyLoaderFn:
-    """Higher-order function, returning a policy loading function.
+    """Creates a policy loading function to load from huggingface.
```
AdamGleave, line 94:
```suggestion
    """Creates a policy loading function to load from Hugging Face.
```

AdamGleave (approved):
LGTM -- just one minor suggestion.

Well done pushing this PR over the finish line. It was a bit of a marathon but I think definitely worth it in the end. I've already enjoyed being able to see stats, videos, etc of the experts on Hugging Face.

AdamGleave:
The coverage seems fine to me -- the only check that's failing is an optional coverage check on the patch. There seems to be a bunch of indirect coverage changes due to you changing the tests -- so some environment named configs that were previously covered no longer are. This seems harmless.

Issue #399, opened 2022-02-02
Upgrade Ray


AdamGleave:
Currently pinned to 0.8.x but there's a stable release 1.9.2, we're way behind. Expect will require only minor tweaks to the parallel script.

norabelrose:
Fixed by #446

PR #446, opened 2022-06-13
Change dependencies for Python 3.10 support

norabelrose:
Currently, `imitation` does not work properly on Python 3.10 due to its reliance on the PyPI version of Sacred (0.8.2). In particular, if you try to run `python -m imitation.scripts.train_rl`, you'll get errors like this:
```
Exception originated from within Sacred.
Traceback (most recent calls):
  File "/opt/homebrew/Caskroom/miniforge/base/lib/python3.10/site-packages/sacred/utils.py", line 345, in recursive_update
    if isinstance(v, collections.Mapping):
AttributeError: module 'collections' has no attribute 'Mapping'
```
The `Mapping` type was removed from the `collections` library in Python 3.10. The GitHub version of Sacred was actually updated for Python 3.10 back in March (it's now on 0.8.3), but unfortunately the maintainer in charge of pushing new releases to PyPI seems to be MIA (see https://github.com/IDSIA/sacred/issues/857), so it's not clear when the PyPI package will be updated. This PR fixes the issue by simply changing the Sacred dependency in `setup.py` to the GitHub version.

I also noticed that the tests currently require an out-of-date version of Ray Tune (~=0.8.5), but Ray just added support for Python 3.10 in its most recent release a few days ago ([1.13.0](https://github.com/ray-project/ray/releases/tag/ray-1.13.0)). So I changed that dependency from `~=0.8.5` to `>=0.8.5` to allow for downloading the most recent version for testing.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ TESTS_REQUIRE = [
     "pytest-notebook",
     "pytest-xdist",
     "pytype",
-    "ray[debug,tune]~=0.8.5",
+    "ray[debug,tune]>=0.8.5",
     "scipy>=1.8.0",
     "wandb",
 ]
@@ -76,7 +76,8 @@ setup(
             "@gym_fixes#egg=stable-baselines3"
         ),
         "stable-baselines3>=1.4.0",
-        "sacred~=0.8.1",
+        # TODO(nora) switch back to PyPi once following makes it to release:
+        "sacred@git+https://github.com/IDSIA/sacred.git",
         "tensorboard>=1.14",
     ],
     tests_require=TESTS_REQUIRE,
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ from setuptools import find_packages, setup
 
 import src.imitation  # pytype: disable=import-error
 
+PARALLEL_REQUIRE = ["ray[debug,tune]>=1.13.0"]
 TESTS_REQUIRE = [
     "seals",
     "black",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -76,7 +76,8 @@ def get_readme() -> str:
             "@gym_fixes#egg=stable-baselines3"
         ),
         "stable-baselines3>=1.4.0",
-        "sacred~=0.8.1",
+        # TODO(nora) switch back to PyPi once 0.8.3 makes it to release:
+        "sacred@git+https://github.com/IDSIA/sacred.git",
```
AdamGleave, line 80: do we want to use master v.s. the 0.8.3 release tag?

setup.py
```diff
@@ -29,7 +29,7 @@
     "pytest-notebook",
     "pytest-xdist",
     "pytype",
-    "ray[debug,tune]~=0.8.5",
+    "ray[debug,tune]>=0.8.5",
```
AdamGleave, line 32: Heh, upgrading Ray was one of the things I was going to ask you to look into. Good this just works, I thought `src/imitation/scripts/parallel.py` might need changes. If we're going to upgrade it let's specify >= the latest release (1.13.0), 0.84 is quite ancient now.

AdamGleave (commented):
Thanks for the PR! Couple of minor points about what versions we should specify but generally happy with the upgrade.

Note CI will still be running on 3.8 as Dockerfile not changed. But we probably want that so long as we want to maintain compatibility with 3.8, as Python is backwards compatible more likely to break on 3.8 and not 3.10 than vice-versa.

AdamGleave:
Fixes #399

setup.py
norabelrose: Just changed to specify 0.8.3

setup.py
norabelrose: Now specifies `>=1.13.0`

setup.py
```diff
@@ -29,7 +29,7 @@
     "pytest-notebook",
     "pytest-xdist",
     "pytype",
-    "ray[debug,tune]~=0.8.5",
+    "ray[debug,tune]>=1.13.0",
```
AdamGleave, line 32: let's delete this line and concatenate PARALLEL_REQUIRE to TEST_REQUIRE

AdamGleave (approved):
LGTM

PR #452, opened 2022-06-16
Preference comparisons example improvement

### LINKED ISSUE #450: Preference Comparisons Improvement
For the [preference comparisons notebook example](https://github.com/HumanCompatibleAI/imitation/blob/59d474988c744c27fe4b5e6494170eeba14207eb/examples/5_train_preference_comparisons.ipynb)
In the last cell, the agent's policy is used to evaluate but it should be using learner's policy instead.

I suggest changing
`reward, _ = evaluate_policy(agent.policy, venv, 10)` 
to
`reward, _ = evaluate_policy(learner.policy, venv, 10)`

SugarmanZhu:
Fix #450

### DIFF
--- a/examples/5_train_preference_comparisons.ipynb
+++ b/examples/5_train_preference_comparisons.ipynb
@@ -164,7 +164,7 @@
    "source": [
     "from stable_baselines3.common.evaluation import evaluate_policy\n",
     "\n",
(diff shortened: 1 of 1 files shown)


AdamGleave (approved):
LGTM

Issue #390, opened 2022-01-20
Scheduler for preference comparison gathering


ejnnr:
We currently gather one big chunk of preference comparison labels at the beginning (determined by `initial_comparison_frac`) and then gather the same number of labels in each iteration. In contrast, the [DRLHP paper](https://arxiv.org/pdf/1706.03741.pdf) uses a schedule where the rate of collecting new labels decreases over time, e.g. for Atari:

> We collect 500 comparisons from a randomly initialized policy network at the beginning of training, and our rate of labeling after T frames of training is decreased every 5 ∗ 10^6 frames, to be roughly proportional to 5 ∗ 10^6/(T + 5 ∗ 10^6).

It makes sense that this would help (after a while, we've learned a good reward model and just need to fine-tune as the policy changes). I'd guess it's less important than #389 but would still be good to have implemented.

The most flexible choice would be to pass a scheduler function to `PreferenceComparisons`, but it would be nice to make it easy to use either a constant schedule (like we currently have) or a hyperbolic schedule (like in the paper), in both cases combined with a configurable number of additional initial steps. (e.g. we could have a `make_hyperbolic_schedule(decrease_every: int, initial_comparison_frac: float) -> Callable[[int], float]` helper function).

The other design decision is how to specify the total number of comparisons we gather. From an end-user perspective, I think it makes much more sense to continue specifying this as a single number (`total_comparisons`) rather than implicitly via the scheduler. So I think we need the scheduler function to be normalized (i.e. its outputs should sum to one over all iterations). Might be best to just do this inside `PreferenceComparisons` by calling the scheduler for all iterations right at the beginning (since the scheduler itself won't know about the total number of iterations).

PR #453, opened 2022-06-16
Implement FIFO buffer for preference comparison labels; a few related fixes

### LINKED ISSUE #389: Implement FIFO buffer for preference comparison labels
We currently maintain a dataset of preference comparison labels that can grow arbitrarily large as we continue collecting preference comparisons during training. This has two issues:
1. Since we always train the reward model over the entire dataset in each iteration, the reward model is effectively trained more later on (which is perhaps even the opposite of what we want). This would be easy to fix though by fixing the number of steps, instead of epochs.
2. More importantly, there may be significant distributional shift because of the changing policy, and once the dataset has grown very large, new labels will have little effect---we're effectively training on an outdated transition distribution.

Because of this second point, the [DRLHP paper](https://arxiv.org/pdf/1706.03741.pdf) uses a buffer of 3,000 preference comparisons (at least for Atari). I think this might help significantly with performance in some cases, so we should implement it as well.

This means the `PreferenceDataset` class needs to be generalized so that it can take a `max_size` argument or similar, specifying the maximum allowed number of comparisons to store. If `max_size=None`, it should revert to the current behavior of growing without limit. Otherwise, it should behave as a FIFO buffer (i.e. discard the oldest comparisons when new ones are added via `PreferenceDataset.push()`).

Since the `PreferenceDataset` instance is created inside `PreferenceComparisons.__init__()`, the `max_size` argument needs to be added there as well and passed on (and it should be added to the training script).

The simplest implementation would just continue using lists to store everything (overhead from copying when we need to drop prefixes of the list is probably not too big). Pythons `collections.deque` has O(n) random access, so I'd guess it's less efficient (we need random access during reward model training).

norabelrose:
Fixes #389 and #452. In particular, there is now a `comparison_queue_size` flag on `PreferenceComparisons.__init__()` and `PreferenceDataset.__init__()` which defaults to `None`. If it is non-`None` then the oldest fragments will be evicted from the queue as needed to make room for new fragments appended by `push()`. Other changes:
- `5_train_preference_comparisons.ipynb` now runs the algorithm on Pendulum instead of Cartpole. It seems that Cartpole is not a very good environment for this algorithm, at least how it's currently implemented. Since reward is constant until the pole falls over, and with a good policy the pole may stay standing until the episode is truncated, fragments of episodes frequently have identical ground truth rewards, leading to very high entropy preferences. The reward model then either overfits to the noise in these preferences, leading to a very poorly shaped reward, or it simply fails to fit to the preferences at all (getting barely above 50% accuracy in predicting preferences). The algorithm seems to do somewhat better with Pendulum, although I am still seeing some issues with the preference entropy getting too high as the policy improves, so we might want to look at allowing`SyntheticGatherer` to produce ties to see if that helps. Adding a preference gathering schedule as mentioned in #390 would also probably help.
- Fixed the following warning that was being emitted from PyTorch inside of `RewardNet.preprocess`:
> UserWarning: The given NumPy array is not writable, and PyTorch does not support non-writable tensors. This means writing to this tensor will result in undefined behavior. You may want to copy the array to protect its data or make it writable before converting it to a tensor. This type of warning will be suppressed for the rest of this program. (Triggered internally at  ../torch/csrc/utils/tensor_numpy.cpp:178.)

I fixed this by adding a function `util.safe_to_tensor` which checks if the NumPy array is writeable and copies it before converting to a PyTorch tensor if so.
- I also added a `dropout_prob` flag to `build_mlp`. Since the DLFHP paper uses dropout in the reward model to prevent overfitting, I figured we should probably allow for it too.
- Also, when a GPU is available, `imitation.scripts.train_preference_comparisons` now trains the reward model on the first GPU. More specifically, it puts the reward model on the same device that the agent is on. This seems to significantly speed up training especially in the initial phase where we might be running the training loop for hundreds of epochs.
- `CrossEntropyRewardTrainer` now displays a `tqdm` progress bar when training the reward model and the number of epochs is greater than 1


setup.py
```diff
@@ -68,12 +68,6 @@ def get_readme() -> str:
         "torch>=1.4.0",
         "tqdm",
         "scikit-learn>=0.21.2",
-        # TODO(adam): switch back to PyPi once following makes it to release:
-        # https://github.com/DLR-RM/stable-baselines3/pull/734 is released
-        (
-            "stable-baselines3@git+https://github.com/carlosluis/stable-baselines3.git"
-            "@gym_fixes#egg=stable-baselines3"
-        ),
         "stable-baselines3>=1.4.0",
```
AdamGleave, line 71: maybe bump version to 1.5.0 to get the logging bugfix?

examples/5_train_preference_comparisons.ipynb
```diff
@@ -78,7 +74,7 @@
     "    initial_comparison_frac=0.1,\n",
     "    allow_variable_horizon=False,\n",
     "    seed=0,\n",
-    "    initial_epoch_multiplier=2,  # Note: set to 200 to achieve sensible results\n",
+    "    initial_epoch_multiplier=1,\n",
```
AdamGleave, line 77: Do we want to add this back in or is it not necessary? `# Note: set to 200 to achieve sensible results`

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -805,7 +822,13 @@ def _train(self, dataset: PreferenceDataset, epoch_multiplier: float = 1.0):
             collate_fn=preference_collate_fn,
         )
         epochs = round(self.epochs * epoch_multiplier)
-        for _ in range(epochs):
+        outer_loop = range(epochs)
+        if epochs > 1:  # Show a progress bar iff we have more than one epoch
```
AdamGleave, line 826: why not just always use tqdm?

AdamGleave (commented):
Will take a more detailed look tomorrow but some initial things that jump out

src/imitation/algorithms/preference_comparisons.py
norabelrose: We could do that I suppose, I think progress bars with a total of one unit look kind of tacky but it's up to you

src/imitation/util/util.py
```diff
@@ -160,6 +160,25 @@ def endless_iter(iterable: Iterable[T]) -> Iterator[T]:
     return itertools.chain.from_iterable(itertools.repeat(iterable))
 
 
+def safe_to_tensor(numpy_array: np.ndarray) -> th.Tensor:
```
AdamGleave, line 163: Consider adding `device` flag? From https://pytorch.org/docs/stable/generated/torch.as_tensor.html it seems that when `th.as_tensor` copies, it does so like `data.to(dtype=dtype, device=device)`. I'd expect `data.to(dtype=dtype).to(device=device)` to be slower in many cases (copying twice).

tests/algorithms/test_preference_comparisons.py
```diff
@@ -250,6 +250,20 @@ def test_preference_dataset_errors(agent_trainer, fragmenter):
         dataset.push(fragments, preferences)
 
 
+def test_preference_dataset_queue(agent_trainer, fragmenter):
+    dataset = preference_comparisons.PreferenceDataset(max_size=5)
+    trajectories = agent_trainer.sample(10)
+
+    gatherer = preference_comparisons.SyntheticGatherer()
+    for _ in range(6):
```
AdamGleave, line 258:
```suggestion
    for i in range(6):
```

tests/algorithms/test_preference_comparisons.py
```diff
@@ -250,6 +250,20 @@ def test_preference_dataset_errors(agent_trainer, fragmenter):
...
+        fragments = fragmenter(trajectories, fragment_length=2, num_pairs=1)
+        preferences = gatherer(fragments)
+        dataset.push(fragments, preferences)
```
AdamGleave, line 261:
```suggestion
        assert len(dataset) == min(i, 5)
        dataset.push(fragments, preferences)
        assert len(dataset) == min(i + 1, 5)
```
you can then remove the assertion below

AdamGleave (commented):
Thanks for the PR! Looks almost ready, I've left a few minor comments, the only substantial one is to add `device` flag to the new `util.safe_to_tensor` function.

Relatedly should probably update some of the other uses of `th.as_tensor` in the codebase, unless we've caught all the non-writable ones? But that probably belongs in a separate PR.

examples/5_train_preference_comparisons.ipynb
norabelrose: It's really not necessary, at least for a simple continuous control environment like Pendulum. I think the 200 figure came from the DLFHP paper where it says they trained the reward model for 200 epochs on a random policy _in Atari only_. For continuous control it sounds like they didn't train the reward model at all before starting to train the agent.

examples/5_train_preference_comparisons.ipynb
```diff
@@ -96,11 +92,32 @@
...
    "source": [
     "pref_comparisons.train(\n",
-    "    total_timesteps=1000,  # Note: set to 40000 to achieve sensible results\n",
-    "    total_comparisons=120,  # Note: set to 4000 to achieve sensible results\n",
+    "    total_timesteps=1000,\n",
+    "    total_comparisons=200,\n",
     ")"
    ]
   },
+  {
```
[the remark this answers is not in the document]
norabelrose: I don't think we need it. Deleting it now.

examples/6_train_mce.ipynb
```diff
@@ -32,7 +32,6 @@
     ")\n",
     "import gym\n",
     "import imitation.envs.examples.model_envs\n",
-    "from imitation.algorithms import base\n",
```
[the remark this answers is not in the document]
norabelrose: Done

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -601,6 +610,14 @@ def push(
         self.fragments2.extend(fragments2)
         self.preferences = np.concatenate((self.preferences, preferences))
 
+        # Evict old samples if the dataset is at max capacity
+        if self.max_size is not None:
+            extra = len(self.preferences) - self.max_size
+            if extra > 0:
+                self.fragments1 = self.fragments1[extra:]
```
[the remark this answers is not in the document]
norabelrose: Yeah I thought about using a `collections.deque`, but that doesn't have O(1) random access, which `DataLoader` relies on. `PreferenceDataset.__getitem__()` is getting called `batch_size` times per gradient step, whereas `PreferenceDataset.push()` only gets called once each time we gather new preferences, so it's pretty clear we should optimize for random access. Technically the ideal data structure for this would be a ring buffer, I think, which would get you both properties.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -980,12 +1006,18 @@ def train(
             #
             # In addition, we collect the comparisons specified via
             # initial_comparison_frac.
-            if i == 0:
-                num_pairs += extra_comparisons + initial_comparisons
+            num_pairs = (
```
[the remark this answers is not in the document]
norabelrose: Yeah that seems right

src/imitation/util/util.py
norabelrose: Done

examples/5_train_preference_comparisons.ipynb
```diff
@@ -26,23 +26,19 @@
...
     "    venv.observation_space, venv.action_space, normalize_input_layer=RunningNorm\n",
     ")\n",
     "\n",
     "fragmenter = preference_comparisons.RandomFragmenter(warning_threshold=0, seed=0)\n",
     "gatherer = preference_comparisons.SyntheticGatherer(seed=0)\n",
-    "reward_trainer = preference_comparisons.CrossEntropyRewardTrainer(\n",
-    "    model=reward_net,\n",
-    "    epochs=3,\n",
-    ")\n",
+    "reward_trainer = preference_comparisons.CrossEntropyRewardTrainer(model=reward_net)\n",
```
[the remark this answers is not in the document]
norabelrose: I think I initially changed that to get the running time consistently under a minute, but now that I run the notebook again it looks like I can bump it back up to 3 epochs and it's still fast enough, at least on RNN. So I'll revert that change and let's hope it doesn't time out in the CI. I also bumped up the number of timesteps to 10,000. Results seem reasonable-ish. Reward model accuracy is higher in the `pref-schedule` branch w/ hyperbolic query scheduling

examples/5_train_preference_comparisons.ipynb
```diff
@@ -96,11 +92,32 @@
    "outputs": [],
    "source": [
     "pref_comparisons.train(\n",
-    "    total_timesteps=1000,  # Note: set to 40000 to achieve sensible results\n",
-    "    total_comparisons=120,  # Note: set to 4000 to achieve sensible results\n",
+    "    total_timesteps=1000,\n",
```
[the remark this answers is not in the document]
norabelrose: I changed `total_timesteps` to 10_000 which yields a reward model accuracy of 95.1% with low preference entropy, so I think these parameters are actually pretty good as-is.

AdamGleave (approved):
LGTM, can merge once tests pass

PR #454, opened 2022-06-16
Expose name_to_{value,count,excluded} maps in HierarchicalLogger

AdamGleave:
SB3 Logger has several (publicly accessible) maps `name_to_value`, `name_to_count` and `name_to_excluded` that store intermediate values prior to being "dumped" (persisted to an output format). The `HierarchicalLogger` has these attributes (as it inherits from Logger), but is currently not updating them as all the work is happening in sub-loggers.

This PR exposes the `name_to_*` maps of the current logger. It does this in a kind of hacky way by rewriting the attributes to match those of the current logger. Unfortunately we can't use `@property` which would be a cleaner way to do this, as the `__init__` method tries to write to these. (We could have a `@property.setter` that is a no-op but I think this is even more hacky.)

This fixes a bug that `PreferenceComparisons.train` was always returning zero reward loss and accuracy; I've also added a regression test for that.

### DIFF
--- a/ci/build_and_activate_venv.sh
+++ b/ci/build_and_activate_venv.sh
@@ -8,7 +8,7 @@ if [[ ${venv} == "" ]]; then
 fi
 
 virtualenv -p python3.8 ${venv}
-# shellcheck disable=SC1090
+# shellcheck disable=SC1090,SC1091
 source ${venv}/bin/activate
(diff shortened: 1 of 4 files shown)


norabelrose (approved):
Looks good to me!

PR #455, opened 2022-06-17
Customizable schedules for querying preferences

### LINKED ISSUE #390: Scheduler for preference comparison gathering

norabelrose:
Fixes #390.

This PR makes a **BC-breaking change** to the `PreferenceComparisons` API and `train_preference_comparisons` script, in order to allow for customizable preference querying schedules. In particular, the `comparisons_per_iteration` parameter has been removed, since the number of comparisons gathered will generally be different at each iteration.

In its place, there are two new parameters: `num_iterations`, which specifies the number of times the algorithm will switch from training the reward model to training the agent against the reward model, and `query_schedule`, which is either a `str` or a `Callable[[float], float]` which takes in a number between 0 and 1 indicating the fraction of total timesteps elapsed in the training run, and outputs a (possibly unnormalized) probability density which is used to determine the fraction of comparisons that should be gathered at the beginning of that iteration. The possible string values of `query_schedule` are `constant`, `hyperbolic`, and `inverse_quadratic`, corresponding to a constant density, a density of `1 / (1 + t)`, and a density of `1 / (1 + t ** 2)` respectively. The default query schedule is `hyperbolic`, since this is the schedule used by the DLFHP paper.


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -946,46 +993,35 @@ def train(
...
-        )
-        if iterations == 0:
-            raise ValueError(
-                f"total_comparisons={total_comparisons} is less than "
-                f"comparisons_per_iteration={self.comparisons_per_iteration}",
-            )
-        timesteps_per_iteration, extra_timesteps = divmod(total_timesteps, iterations)
 
+        # Compute the number of comparisons to request at each iteration in advance.
+        vec_schedule = np.vectorize(self.query_schedule)
```
AdamGleave, line 1001: We could make query schedule ndarray->ndarray rather than float->float. No real performance difference as only call it once, and the float->float is more general. But is there a case where we couldn't easily write a vector implementation? (I think all the current ones would "just work" with ndarray)

src/imitation/scripts/config/train_preference_comparisons.py
```diff
@@ -48,6 +55,7 @@ def train_defaults():
     allow_variable_horizon = False
 
     checkpoint_interval = 0  # Num epochs between saving (<0 disables, =0 final only)
+    query_schedule = "hyperbolic"
```
AdamGleave, line 58: Reasonable default given it's what's used in paper, but worth noting this is changing the default behavior (previously we were effectively constant). I think that's fine but @yawen-d @levmckinney you should be aware of this -- you may want to override it in your experiments if you want to make direct comparisons.

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -74,17 +75,20 @@ def train_preference_comparisons(
     rl: Mapping[str, Any],
     allow_variable_horizon: bool,
     checkpoint_interval: int,
+    query_schedule: str,
```
AdamGleave, line 78: No need to restrict this to `str` -- we can define functions in Sacred configs. (Hacky admittedly, but may be necessary in some cases.)

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -134,7 +143,7 @@ def train_preference_comparisons(
 
     reward_net = reward.make_reward_net(venv)
     if agent_path is None:
-        agent = rl_common.make_rl_algo(venv)
+        agent = rl_common.make_rl_algo(venv, rl_kwargs=rl["rl_kwargs"])
```
AdamGleave, line 146: shouldn't it automatically get the `rl_kwargs` as `make_rl_algo` is a `sacred.capture` decorated function?

src/imitation/util/util.py
```diff
@@ -25,6 +25,34 @@
 from stable_baselines3.common.vec_env import DummyVecEnv, SubprocVecEnv, VecEnv
 
 
+def oric(x: np.ndarray) -> np.ndarray:
+    """Optimal rounding under integer constraints.
+
+    Given a vector of real numbers such that the sum is an integer, returns a
+    vector of rounded integers that preserves the sum and which minimizes the
+    Lp-norm of the difference between the rounded and original vectors. Algorithm
```
AdamGleave, line 33: p isn't a parameter of the function, so which Lp norm have we chosen to minimize? (I think the L1?)

src/imitation/scripts/config/train_preference_comparisons.py
yawen-d: Thanks for the heads up!

src/imitation/util/util.py
norabelrose: I think L1 is what we care about, but the method actually guarantees that _all_ Lp norms with p >= 1 are minimized by the resulting rounded values (see Proposition 3 in the paper).

src/imitation/scripts/train_preference_comparisons.py
norabelrose: You're probably right, I didn't realize `sacred.capture` behaves that way.

src/imitation/scripts/train_preference_comparisons.py
norabelrose: Interesting, good to know. `Union[str, Callable[[float], float]]` should work then.

src/imitation/algorithms/preference_comparisons.py
norabelrose: I thought about doing `np.ndarray -> np.ndarray`, and actually coded it up that way at first, but I decided to change it to `float -> float` for two main reasons: 1) `torch.optim.lr_scheduler.LambdaLR` uses `float -> float` and I wanted to err on the side of being similar to existing "schedule" APIs, and 2) the docstring I wrote up for the `np.ndarray` approach seemed wordier and harder to follow than the current one.

I can't think of a case where `np.ndarray -> np.ndarray` wouldn't also work.

AdamGleave: Keeping it the same as the existing schedule API is a good reason, we do this in `BetaSchedule` in DAgger as well and `stable_baselines3.Schedule`, so I think I'm overall convinced.

src/imitation/util/util.py
AdamGleave: thanks, clarified in docstring

AdamGleave (approved):
LGTM

Issue #456, opened 2022-06-21
Allow `train_rl.py` script to load a saved policy to continue training


yawen-d:
## Motivation

- `train_preference_comparisons` has a feature of loading a pretrained/warm-started `stable_baselines3.common.base_class.BaseAlgorithm` object to continue training. This could be useful for `train_rl` and potentially for `train_adversarial`.

## Proposed solution

- create a captured function `load_rl_algo_from_path` in `rl` ingredient to load existing models.

yawen-d:
Issue closed due to merged branch.

PR #457, opened 2022-06-21
Allow train_rl.py script to load a saved policy to continue training

yawen-d:
This PR aims to address #456.


src/imitation/scripts/common/rl.py
```diff
@@ -132,3 +133,25 @@ def make_rl_algo(
...
+    rl_cls: Type[base_class.BaseAlgorithm],
+    rl_kwargs: Mapping[str, Any],
+    _seed: int,
+) -> base_class.BaseAlgorithm:
+    agent = serialize.load_stable_baselines_model(
+        rl_cls,
+        agent_path,
+        venv,
+        seed=_seed,
+        **rl_kwargs,
```
AdamGleave, line 151: What do you want to happen when `rl_kwargs` in the script does not match the `rl_kwargs` that was used during training `agent_path`? Right now we're overriding them (by default they're copied from the saved pickle file). This is reasonable -- sometimes you might want to restart training with e.g. a different learning rate -- but will sometimes cause things to fail. I think most of the failures should be obvious (e.g. you try and instantiate with a different sized policy network and things blow up), but it's possible some failures could be silent.

tests/test_scripts.py
```diff
@@ -79,22 +79,26 @@ def test_main_console(script_mod):
         script_mod.main_console()
 
 
+_rl_agent_loading_configs = {
```
AdamGleave, line 82: Constant, so uppercase?

tests/test_scripts.py
```diff
@@ -79,22 +79,26 @@ def test_main_console(script_mod):
...
+    "agent_path": CARTPOLE_TEST_POLICY_PATH,
+    # FIXME(yawen): the policy we load was trained on 8 parallel environments
```
AdamGleave, line 84: @ernestum was running into a similar error restoring PPO policies for preference comparison, we had a discussion [here](https://github.com/HumanCompatibleAI/imitation/pull/445#issuecomment-1163496244)

AdamGleave (approved):
LGTM, some small changes in comments

tests/test_scripts.py
yawen-d: yep

src/imitation/scripts/common/rl.py
yawen-d: I checked the source, and I think that is what I want it to behave. I want kwargs passed into `load_stable_baselines_model` to override the saved kwargs.

Issue #458, opened 2022-07-01
Add reward relabelling to `train_preference_comparisons` and `train_adversarial` for off-policy algorithms


yawen-d:
## Suggested Feature
- Allow `AgentTrainer` to relabel the learned rewards of all transitions in the replay buffer after the reward model is updated for off-policy algorithms.
- This feature applies to both `train_adversarial.py` and `train_preference_comparisons.py` when an off-policy algorithm is used as their RL component.

## Motivation
- Reward relabelling was used in the paper
[PEBBLE: Feedback-Efficient Interactive Reinforcement Learning via Relabeling Experience and Unsupervised Pre-training](https://arxiv.org/pdf/2106.05091.pdf) to mitigate the effects of a non-stationary reward function.

yawen-d:
- Preference Comparisons + SAC (w/ Reward Relabeling) outperforms (w/o Reward Relabeling)
![Screenshot 2022-07-06 at 11 14 20](https://user-images.githubusercontent.com/35738259/177616118-9b95e9ad-0995-441c-86ae-4768dc1510bc.png)

- Preference Comparisons + SAC (w/ Reward Relabeling) significantly outperforms (w/o Reward Relabeling)
![Screenshot 2022-07-06 at 11 16 13](https://user-images.githubusercontent.com/35738259/177616370-35a61562-7409-4ed3-9187-4602c2f24055.png)

AdamGleave:
Experimental results definitely show the benefit of this feature, plus the results in prior work (PEBBLE), so I'm in favour of us having an implementation :)

yawen-d:
Close issue because #459 has been merged.

PR #459, opened 2022-07-01
Add reward relabelling to `train_preference_comparisons` and `train_adversarial` for off-policy algorithms

yawen-d:
Aim to address #458 .


AdamGleave (commented):
Sorry for being slow to take a look at this.

If I understand correctly, the way this PR works is:
  - We add a callback to preference comparison, that gets executed each preference comparison iteration on RL training start.
  - This callback evaluates the reward function for all the data in the buffer, and relabels it.

I'd like to better understand why you went for this approach rather than making your own `ReplayBuffer` that wraps other replay buffers and relabels the reward. I think we discussed this before so I assume there's some disadvantage you came across while implementing it. To be concrete, what I had in mind was something like:

```
class ReplayBufferRewardWrapper(ReplayBuffer):
  def __init__(replay_buffer_cls: Type[ReplayBuffer] = ReplayBuffer, reward_fn: RewardFn, **kwargs):
     self.replay_buffer = replay_buffer_cls(**kwargs)
     self.reward_fn = RewardFn

  def add(*args, **kwargs):
      self.replay_buffer.add(*args, **kwargs)

   def sample(*args, **kwargs):
       samples = self.replay_buffer.sample(*args, **kwargs)
       rewards = reward_fn(samples.observations, samples.actions, samples.next_observations, samples.dones)
      return ReplayBufferSamples(samples.observations, samples.actions, samples.next_observations, samples.dones, rewards)
```

Naively, I'd expect this to be simpler as no need for a callback, or directly accessing internals of the `ReplayBuffer`. Just pass a different `replay_buffer_cls` into `SAC` or other off-policy algorithm constructor and you're done.

yawen-d:
Thanks for the comments!

### Option A

> If I understand correctly, the way this PR works is:
- We add a callback to preference comparison, that gets executed each preference comparison iteration on RL training start.
- This callback evaluates the reward function for all the data in the buffer, and relabels it.

Yep, the current implementation (option A) does the above.

### Option B

- Making a new `ReplayBufferRewardWrapper` that relabels the rewards from `.sample()` only.
- Instantiating `ReplayBufferRewardWrapper` requires a `replay_buffer_cls` and a reward function `reward_fn`.

I pursued option A simply to follow the procedure in the PEBBLE paper to relabel the whole ReplayBuffer. No deep reasons or strong preference then.

---

After reflecting on this a bit, I agree Option A is more complicated than B, but there are two problems with B:
1. For B, `ReplayBufferRewardWrapper` needs to be initialized after `reward_fn` is given. However, in `AgentTrainer.__init__`, when we have the `reward_fn`, the agent and its replay buffer has already been initialized.
    - https://github.com/HumanCompatibleAI/imitation/blob/d9993ae2760963c76cc79bb6cb65e49ced86369c/src/imitation/algorithms/preference_comparisons.py#L156
    - We can do something hacky to [reinitialize a new replay buffer](https://github.com/DLR-RM/stable-baselines3/blob/ef10189d80dbb2efb3b5391cba4eb3c2ab5c7aae/stable_baselines3/common/off_policy_algorithm.py#L205) to replace the original by `self.algorithm.replay_buffer = <new_buffer>`, but when SB3 changes its way of initializing replay_buffer, we need to make changes.
    - Another way is to init the wrapper with a replay_buffer object and a reward function directly by `  def __init__(replay_buffer: ReplayBuffer, reward_fn: RewardFn): ...`, but this is pretty similar to the above approach.
2. Another potential problem is that when saving the SB3 model, I worry the `ReplayBufferRewWrapper` class would be saved (although I haven't checked). In this case, when loading this model again, we have to provide a reward function.
    - We can set reward_fn default to None as a workaround.

---

Tentative solution: Initialize the wrapper with a replay_buffer object and a RewardFn directly by `  def __init__(replay_buffer: ReplayBuffer, reward_fn: RewardFn): ...`

AdamGleave:
> 1. For B, `ReplayBufferRewardWrapper` needs to be initialized after `reward_fn` is given. However, in `AgentTrainer.__init__`, when we have the `reward_fn`, the agent and its replay buffer has already been initialized.
>    
>    * https://github.com/HumanCompatibleAI/imitation/blob/d9993ae2760963c76cc79bb6cb65e49ced86369c/src/imitation/algorithms/preference_comparisons.py#L156
>    * We can do something hacky to [reinitialize a new replay buffer](https://github.com/DLR-RM/stable-baselines3/blob/ef10189d80dbb2efb3b5391cba4eb3c2ab5c7aae/stable_baselines3/common/off_policy_algorithm.py#L205) to replace the original by `self.algorithm.replay_buffer = <new_buffer>`, but when SB3 changes its way of initializing replay_buffer, we need to make changes.
>    * Another way is to init the wrapper with a replay_buffer object and a reward function directly by `  def __init__(replay_buffer: ReplayBuffer, reward_fn: RewardFn): ...`, but this is pretty similar to the above approach.

We could just make it the callees responsibility to set up. For example, change [train_preference_comparisons.py](https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/scripts/train_preference_comparisons.py#L141) to be:

```python
    reward_net = reward.make_reward_net(venv)
    rl_kwargs = dict(venv=venv)
    if reward_relabel:
        rl_kwargs["replay_buffer_class"] = ReplayBufferWrapper
        rl_kwargs["replay_buffer_kwargs"] = dict(reward_fn=reward_net)   
    if agent_path is not None:
        rl_kwargs["agent_path"] = agent_path
    agent = rl_common.make_rl_algo(**rl_kwargs)
```

I don't love this solution. Ideally we'd keep the scripts simple and most of the logic in the algorithm so that we can easily use the algorithms programatically in other contexts (in a Jupyter notebook, as part of a more complicated pipeline, etc).

But given the script is responsible for creating the algorithm, and that `replay_buffer_class` and `replay_buffer_kwargs` are part of the SB3 API so safer to rely on, this might be the least-bad option.

Although it occurs to me that reward relabeling is probably desirable in things other than preference comparisons as well? In that case, the best option might be to implement the logic inside `rl_common.make_rl_algo` which would then be shared between scripts. For example, we could add an optional `reward_net` parameter. If that's specified and the algo is an `OffPolicyAlgorithm`, then use the replay buffer wrapper.

> 2. Another potential problem is that when saving the SB3 model, I worry the `ReplayBufferRewWrapper` class would be saved (although I haven't checked). In this case, when loading this model again, we have to provide a reward function.
>    
>    * We can set reward_fn default to None as a workaround.

I think you're right it'd save it; at least with the solution I proposed above using `replay_buffer_class` and `replay_buffer_kwargs`. 

Definitely some subtleties here. It might save the reward network too in which case it'd at least be deserializable. But would we want to continue training with that reward network, or override it to a new one? etc.

AdamGleave:
I just want to check if this PR is blocked on me or not? I was intending to review it fully once it's been re-architected, but happy to take a closer look at the current implementation if that'd be helpful.

yawen-d:
> I just want to check if this PR is blocked on me or not? I was intending to review it fully once it's been re-architected, but happy to take a closer look at the current implementation if that'd be helpful.
Thanks for checking! I'll re-request review after making changes.

yawen-d:
> I don't love this solution. Ideally we'd keep the scripts simple and most of the logic in the algorithm so that we can easily use the algorithms programatically in other contexts (in a Jupyter notebook, as part of a more complicated pipeline, etc).
> But given the script is responsible for creating the algorithm, and that replay_buffer_class and replay_buffer_kwargs are part of the SB3 API so safer to rely on, this might be the least-bad option.
> Although it occurs to me that reward relabeling is probably desirable in things other than preference comparisons as well? In that case, the best option might be to implement the logic inside rl_common.make_rl_algo which would then be shared between scripts. For example, we could add an optional reward_net parameter. If that's specified and the algo is an OffPolicyAlgorithm, then use the replay buffer wrapper.

I personally prefer using the algorithms programatically in other contexts. Putting the logic into rl_common.make_rl_algo is interesting.

> I think you're right it'd save it; at least with the solution I proposed above using replay_buffer_class and replay_buffer_kwargs.
Definitely some subtleties here. It might save the reward network too in which case it'd at least be deserializable. But would we want to continue training with that reward network, or override it to a new one? etc.

Well, I don't want the `reward_net` to be stored along with the `replay_buffer_cls`. Let's say we want to do retraining using `scripts.train_rl` with a `reward_net`. I would prefer passing some argument into rl_common.make_rl_algo to wrap the replay_buffer with the loaded reward_net.

yawen-d:
Thanks for explaining this in detail! I do agree that this is not the ideal solution and I changed it to the proposed one. Here are the summary:

- ReplayBufferRewardWrapper takes in a `replay_buffer_cls` and `reward_fn` to initialize.
```python
class ReplayBufferRewardWrapper(ReplayBuffer):
    """Relabel the rewards in transitions sampled from a ReplayBuffer."""

    def __init__(
        self,
        *args,
        replay_buffer_cls: Type[ReplayBuffer] = ReplayBuffer,
        reward_fn: Optional[RewardFn] = None,
        **kwargs,
    ):
        self.replay_buffer = replay_buffer_cls(*args, **kwargs)
        self.reward_fn = reward_fn
    
    def sample(self, *args, **kwargs):
        ...
```
- I incoporated the replay buffer into `rl.make_rl_algo` and `rl.load_rl_algo_from_path`. So that any off-policy algorithm will use `ReplayBufferRewardWrapper` **by default** if a reward_fn is passed in.
- This implementation doesn't support `DictReplayBuffer` mainly because the current signature of `RewardFn` is [[np.ndarray], [np.ndarray], [np.ndarray], [np.ndarray]] -> np.ndarray, which doesn't support dict-based observation yet. We could add it in future PRs if needed.

src/imitation/policies/replay_buffer_wrapper.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(
+        self,
+        *args,
+        replay_buffer_cls: Type[ReplayBuffer] = ReplayBuffer,
+        reward_fn: Optional[RewardFn] = None,
+        **kwargs,
+    ):
+        """Builds ReplayBufferRewardWrapper.
+
+        Note(yawen): we directly inherit ReplayBuffer in this case and leave out the
```
AdamGleave, line 39: Including your username is better than just leaving an anonymous note (gives us someone to follow up with), but I think better yet is to link to the GitHub issue/PR so people have surrounding context. You could link to the permalink https://github.com/HumanCompatibleAI/imitation/pull/459#issuecomment-1201997194?

I also think this should maybe be a comment not a docstring, as it's more an implementation detail not part of public API docs, but close call -- don't feel strongly about this.

src/imitation/policies/replay_buffer_wrapper.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class ReplayBufferRewardWrapper(ReplayBuffer):
+    """Relabel the rewards in transitions sampled from a ReplayBuffer."""
+
+    def __init__(
+        self,
+        *args,
+        replay_buffer_cls: Type[ReplayBuffer] = ReplayBuffer,
+        reward_fn: Optional[RewardFn] = None,
```
AdamGleave, line 34: When do we want this to be `None`? Can't we just not wrap the replay buffer in that case?

src/imitation/scripts/common/rl.py
```diff
@@ -142,7 +155,17 @@ def load_rl_algo_from_path(
...
     rl_kwargs: Mapping[str, Any],
     _seed: int,
+    relabel_reward_fn: Optional[RewardFn] = None,
 ) -> base_class.BaseAlgorithm:
+    rl_kwargs = dict(rl_kwargs)
+    if relabel_reward_fn and issubclass(
+        rl_cls,
+        off_policy_algorithm.OffPolicyAlgorithm,
+    ):
+        assert not isinstance(venv.observation_space, gym.spaces.Dict)
```
AdamGleave, line 165: A bit of code duplication here. Consider factoring out this into a helper method e.g. `_maybe_add_relabel_buffer`?

src/imitation/scripts/common/rl.py
```diff
@@ -117,6 +123,13 @@ def make_rl_algo(
         if rl_kwargs.get("batch_size") is not None:
             raise ValueError("set 'batch_size' at top-level")
         rl_kwargs["batch_size"] = batch_size
+        if relabel_reward_fn:
+            assert not isinstance(venv.observation_space, gym.spaces.Dict)
+            # Note(yawen): We don't support HerReplayBuffer or DictReplayBuffer yet,
+            # since current RewardFn only takes in NumPy array-based inputs.
+            rl_kwargs["replay_buffer_class"] = ReplayBufferRewardWrapper
```
AdamGleave, line 130: What if `rl_kwargs["replay_buffer_class"]` is already specified? Should we save this and pass it through as `replay_buffer_cls` in `replay_buffer_kwargs`?

src/imitation/policies/replay_buffer_wrapper.py
```diff
@@ -0,0 +1,83 @@
...
+    )
+
+
+class ReplayBufferRewardWrapper(ReplayBuffer):
+    """Relabel the rewards in transitions sampled from a ReplayBuffer."""
+
+    def __init__(
+        self,
+        *args,
+        replay_buffer_cls: Type[ReplayBuffer] = ReplayBuffer,
```
AdamGleave, line 33: SB3 calls this `replay_buffer_class` -- use that as well for consistency? https://stable-baselines3.readthedocs.io/en/master/_modules/stable_baselines3/common/off_policy_algorithm.html

AdamGleave (commented):
Short on time this week so I've only done a shallow review. @tomtseng would you mind taking a closer look?

@yawen-d: thanks for rearchitecting it, this looks a lot cleaner now, and this will be a really useful feature.

It might make sense to add relabeling in `train_adversarial` as well, I think it should only be a few lines, seems like you've done most of the work needed to support it already. But if it turns out to be more involved, then just open an issue to track it.

src/imitation/policies/replay_buffer_wrapper.py
yawen-d: Agreed! Thanks!

src/imitation/policies/replay_buffer_wrapper.py
yawen-d: +1

src/imitation/policies/replay_buffer_wrapper.py
yawen-d: This is actually pretty annoying. Because ReplayBuffer is created by https://github.com/DLR-RM/stable-baselines3/blob/c1f1c3d3d796054f68b3fa741c1a0be4ce2187b5/stable_baselines3/common/off_policy_algorithm.py#L212

```python
self.replay_buffer = self.replay_buffer_class(
    self.buffer_size,
    self.observation_space,
    self.action_space,
    device=self.device,
    n_envs=self.n_envs,
    optimize_memory_usage=self.optimize_memory_usage,
    **self.replay_buffer_kwargs,
)
```
so we need to set reward_fn as a kwarg rather than an arg, otherwise an error of `TypeError: got multiple values for argument` will occur. That's why I set default values to `reward_fn` and `replay_buffer_class`. For now I'm not aware of a better way of working around this, happy to take ideas.

src/imitation/scripts/common/rl.py
yawen-d: Yep, I think you're correct, although this could be silent.

src/imitation/policies/replay_buffer_wrapper.py
AdamGleave: I'm not sure I quite understand this. Keyword arguments don't need to have default values. You can force a keyword argument by including a `*`: `def f(x,*,y)` requires that `y` be specified as a kwarg.

src/imitation/scripts/common/rl.py
yawen-d: yep

src/imitation/policies/replay_buffer_wrapper.py
yawen-d: Sounds good! This is exactly what I'm asking about. Thanks

tests/policies/test_replay_buffer_wrapper.py
```diff
@@ -0,0 +1,94 @@
+"""Tests ExplorationWrapper."""
```
tomtseng, line 1: typo?

yawen-d:
> It might make sense to add relabeling in train_adversarial as well, I think it should only be a few lines, seems like you've done most of the work needed to support it already. But if it turns out to be more involved, then just open an issue to track it.

Just did that in the issue and also made changes to this PR. Also made some unit tests for the code.

tests/policies/test_replay_buffer_wrapper.py
```diff
@@ -0,0 +1,94 @@
...
+    assert total_timesteps == replay_buffer_wrapper.size() == replay_buffer.size()
+    assert (replay_buffer_wrapper.sample(total_timesteps).rewards == 0.0).all()
+    assert (replay_buffer.sample(total_timesteps).rewards != 0.0).all()  # seed=42
+
+    # reset()
+    replay_buffer_wrapper.reset()
+    assert 0 == replay_buffer_wrapper.size() == replay_buffer.size()
+
+    # to_torch()
+    replay_buffer_wrapper.to_torch(np.ones(42))
```
tomtseng, line 94: just checking: is this just testing that this doesn't raise an exception? or should there be an assertion?

tests/policies/test_replay_buffer_wrapper.py
```diff
@@ -0,0 +1,94 @@
...
+    state: np.ndarray,
+    action: np.ndarray,
+    next_state: np.ndarray,
+    done: np.ndarray,
+):
+    del action, next_state, done
+    return np.zeros(state.shape[0], dtype=np.float32)
+
+
+def make_wrapper_n_algo(
```
tomtseng, line 27: what does the `_n_` refer to here?

tomtseng (approved):
overall looks pretty nice to me, just left a few nits

src/imitation/policies/replay_buffer_wrapper.py
```diff
@@ -0,0 +1,93 @@
...
+        action_space: spaces.Space,
+        *,
+        replay_buffer_class: Type[ReplayBuffer],
+        reward_fn: RewardFn,
+        **kwargs,
+    ):
+        """Builds ReplayBufferRewardWrapper.
+
+        Args:
+            buffer_size: Max number of element in the buffer
```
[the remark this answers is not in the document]
yawen-d: Thanks! LOL, then stablebaselines3.common.buffers also have typos

src/imitation/scripts/common/rl.py
```diff
@@ -72,6 +75,26 @@ def sac():
     locals()  # quieten flake8
 
 
+def _maybe_add_relabel_buffer(
+    rl_kwargs: Dict[str, Any],
+    relabel_reward_fn: Optional[RewardFn] = None,
+) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
yawen-d: Cool!

src/imitation/scripts/common/rl.py
```diff
@@ -72,6 +75,26 @@ def sac():
...
+    _rl_kwargs = dict(rl_kwargs)
+    if relabel_reward_fn:
+        _buffer_kwargs = dict(reward_fn=relabel_reward_fn)
+        if "replay_buffer_class" in _rl_kwargs:
+            _buffer_kwargs["replay_buffer_class"] = _rl_kwargs["replay_buffer_class"]
+        else:
+            _buffer_kwargs["replay_buffer_class"] = buffers.ReplayBuffer
```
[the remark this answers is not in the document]
yawen-d: Thanks! Looks good!

tests/policies/test_replay_buffer_wrapper.py
yawen-d: Good catch!

tests/policies/test_replay_buffer_wrapper.py
yawen-d: Oh, I meant `and`. Maybe it's better to name it as `make_algo_with_wrapped_buffer`

src/imitation/policies/replay_buffer_wrapper.py
```diff
@@ -0,0 +1,90 @@
...
+        action=samples.actions.cpu().numpy(),
+        next_state=samples.next_observations.cpu().numpy(),
+        done=samples.dones.cpu().numpy(),
+    )
+
+
+class ReplayBufferRewardWrapper(ReplayBuffer):
+    """Relabel the rewards in transitions sampled from a ReplayBuffer."""
+
+    def __init__(
```
AdamGleave, line 31: `__init__` is failing to set several attributes that would usually be set by `BaseBuffer`: namely `buffer_size`, `observation_space`, `action_space`, ...

I get that we can't call `super().__init__` directly because `ReplayBuffer.__init__` sets some things we don't want (e.g. `self.observations`), but perhaps it's worth adding a call to `BaseBuffer.__init__` directly?

Bit hacky but seems better than not calling `__init__` at all.

Thoughts @tomtseng?

tests/policies/test_replay_buffer_wrapper.py
yawen-d: Could make an assertion.

src/imitation/scripts/common/rl.py
```diff
@@ -75,6 +75,26 @@ def sac():
     locals()  # quieten flake8
 
 
+def _maybe_add_relabel_buffer(
+    rl_kwargs: Dict[str, Any],
+    relabel_reward_fn: Optional[RewardFn] = None,
+) -> Dict[str, Any]:
+    """Use ReplayBufferRewardWrapper in rl_kwargs if relabel_reward_fn is not None."""
+    _rl_kwargs = dict(rl_kwargs)
```
AdamGleave, line 83: You don't use `rl_kwargs` anywhere in this function after this, so why create a 2nd variable?
```suggestion
    rl_kwargs = dict(rl_kwargs)
```

tests/policies/test_replay_buffer_wrapper.py
```diff
@@ -0,0 +1,96 @@
...
+from imitation.policies.replay_buffer_wrapper import ReplayBufferRewardWrapper
+from imitation.util import util
+
+
+def zero_reward_fn(
+    state: np.ndarray,
+    action: np.ndarray,
+    next_state: np.ndarray,
+    done: np.ndarray,
+):
```
AdamGleave, line 23:
```suggestion
) -> np.ndarray:
```

tests/policies/test_replay_buffer_wrapper.py
```diff
@@ -0,0 +1,96 @@
...
+    del action, next_state, done
+    return np.zeros(state.shape[0], dtype=np.float32)
+
+
+def make_algo_with_wrapped_buffer(
+    rl_cls: Type[off_policy_algorithm.OffPolicyAlgorithm],
+    policy_cls: Type[BasePolicy],
+    replay_buffer_class: Type[buffers.ReplayBuffer],
+):
```
AdamGleave, line 32:
```suggestion
) -> off_policy_algorithm.OffPolicyAlgorithm:
```

AdamGleave (approved):
Couple of minor suggestions, otherwise LGTM

tests/policies/test_replay_buffer_wrapper.py
yawen-d: nice catch!

src/imitation/policies/replay_buffer_wrapper.py
yawen-d: `BaseBuffer.__init__` seems better than `ReplayBuffer.__init__`. Foreseeable problems might be `self.pos` and `self.full` will not be incremented by `self.add()`. One way of workaround would be adding the following into `self.add()`:

```python
self.pos += 1
if self.pos == self.buffer_size:
    self.full = True
    self.pos = 0
```

But not sure whether this is desirable.

yawen-d: FWIW, `BaseBuffer` has an abstractmethod `_get_samples()` that's not used anywhere in the wrapper, so in practice we could do `def _get_samples(): NotImplementedError`

AdamGleave:
An alternative would be to add Python properties that return the value of
these and other attributes from the wrapped replay buffer, and skip the
init.

AdamGleave:
We'll still inherit the concrete implementation from ReplayBuffer so long
as we subclass it. But overriding it and raising an error is reasonable
given it should never get called.

src/imitation/policies/replay_buffer_wrapper.py
tomtseng: Not sure if this is appropriate, but another (verbose) workaround would be to use `@property` decorators like
```py
  @property
  def pos(self):
    return self.replay_buffer.pos
```
This is what `gym` does with their environment wrappers — the `Wrapper` class  is a subclass of `Env`, and instead of calling `super().__init__()`, `Wrapper` has a bunch of property decorators like
```py
    @property
    def action_space(self) -> spaces.Space[ActType]:
        """Returns the action space of the environment."""
        if self._action_space is None:
            return self.env.action_space
        return self._action_space
```
Then all environment wrappers inherit can subclass `Wrapper`, and calling `super().__init__(env)` behaves nicely

yawen-d: Gotcha, this helps!

yawen-d:
I finalized the modifications to add `super().__init__(buffer_size, observation_space, action_space, **_base_kwargs)` as well as @property to self.pos and self.full.

It also came to me that with self.pos and self.full we could get rid of some methods like size(), reset(), and to_torch() that only manipulate these two attributes.

src/imitation/policies/replay_buffer_wrapper.py
```diff
@@ -80,11 +97,8 @@ def sample(self, *args, **kwargs):
...
-        return self.replay_buffer.size()
-
-    def reset(self) -> None:
-        self.replay_buffer.reset()
-
-    def to_torch(self, array: np.ndarray, copy: bool = True) -> th.Tensor:
-        return self.replay_buffer.to_torch(array, copy)
+    def _get_samples(self):
+        raise NotImplementedError(
+            "_get_samples() is intentionally not implemented."
```
tomtseng, line 102: nit: needs a space after the period

tomtseng (approved):

AdamGleave:
I think this is the cleanest solution, but may require a lot of boilerplate
code, so I'm not sure whether it's worth it for a one-off. Will let Yawen
make final call here.

Issue #427, opened 2022-04-01
Refactor RewardFn


AdamGleave:
From ernestum's suggestions in https://github.com/HumanCompatibleAI/imitation/pull/426:

> I also looked a bit into the rewards corner of the lib and I would propose the following refactoring (which would also get rid of common.py):
> 
> - Turn RewardFn into an abstract class. Then it does not just define the signature of a reward function but also the argument names and a docstring explaining the semantics of the function call.
> - Rename common.py -> reward_fn.py
> - Make all the reward networks inherit from RewardFn. Turn their .predict() into __call__() (or make the __call__() execute predict

AdamGleave:
I agree with the suggestion to turn `RewardFn` into an ABC. But I think having `RewardNet` subclass it won't work as they already inherit an incompatible `__call__`  from nn.Module` (that runs hooks then calls `forward`). 

So I think we can just do the first two points but not the third, WDYT @ernestum?

AdamGleave:
Fixed in https://github.com/HumanCompatibleAI/imitation/pull/478

PR #460, opened 2022-07-05
Adding reward ensembles and conservative reward functions

levmckinney:
This pull request creates a base class for reward functions that keep track of their epistemic uncertainty, provides an ensemble based implementation for it and a conservative reward wrapper.

### TODO:
- [x] Implement seeding correctly for ensemble.
- [x] Add ensemble reward function to  `make_reward`.
- [x] Add conservative reward wrapper.
- [x] Add option for training with reward ensemble to `train_preference_comparison.py`.
- [x] Write tests for ensemble creation.
- [x] Write tests for evaluating the ensemble
- [x] Write tests for saving and loading ensemble
- [x] Write tests for conservative reward wrapper
- [x] Write reward ensemble trainer in `train_preference.py`
- [x] Write tests for reward ensemble
- [x] Add logging for reward standard deviation
- [x] Add option to make reward conservative to train_rl.py.


src/imitation/rewards/reward_nets.py
```diff
@@ -550,3 +574,114 @@ def __init__(
...
+            state: Current states of shape `(batch_size,) + state_shape`.
+            action: Actions of shape `(batch_size,) + action_shape`.
+            next_state: Successor states of shape `(batch_size,) + state_shape`.
+            done: End-of-episode (terminal state) indicator of shape `(batch_size,)`.
+
+        Returns:
+            Computed lower confidence bounds on rewards of shape `(batch_size,`).
+        """
+        reward_mean = self.base.predict_processed(state, action, next_state, done)
+        reward_std = self.base.predict_processed(state, action, next_state, done)
```
AdamGleave, line 683:
```suggestion
        reward_std = self.base.standard_deviation(state, action, next_state, done)
```
?

AdamGleave: Also a bit inefficient (doing forward pass twice, once for `reward_mean` and once for `reward_std`)

AdamGleave:
Finished initial review now. Seems like a reasonable design. Left some more detailed comments in line. Please request re-review when addressed.

src/imitation/rewards/reward_nets.py
```diff
@@ -550,3 +574,114 @@ def __init__(
...
+            base_networks: Constituent reward networks that will make up the
+                ensemble.
+            observation_space: the observation space of the environment
+            action_space: the action space of the environment
+            **kwargs: passed along to superclass
+        """
+        super().__init__(observation_space, action_space, **kwargs)
+        self.base_networks = base_networks
+
+    @abc.abstractmethod
```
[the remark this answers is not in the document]
levmckinney: Nope fixing this.

levmckinney:
@yawen-d It should be possible to train the reward ensemble now. If you have time can you test that this most basic version works? You can do this by just setting the reward class used by confg/train_preference_comparison.py to be a reward ensemble.

I have not quite figured out how to get the conservative wrapper applied during retraining with `train_irl.py`. It's an issue with reward deserialization. Currently, it returns a `RewardFn` so I'll have to think more carefully about how to introduce the reward wrapper. Once that is completed the entire pipeline for training ensembles of reward functions and making them conservative should be done.

src/imitation/rewards/reward_nets.py
```diff
@@ -550,3 +575,200 @@ def __init__(
...
+            done,
+        )
+
+        return reward_mean - self.alpha * np.sqrt(reward_var)
+
+    def forward(self, *args):
+        self.base.forward(*args)
+
+
+def make_reward_net(
```
yawen-d, line 742: Slight preference over putting this function before RewardEnsemble

src/imitation/rewards/reward_nets.py
```diff
@@ -550,3 +575,200 @@ def __init__(
...
+    members: nn.ModuleList
+
+    def __init__(
+        self,
+        observation_space: gym.Space,
+        action_space: gym.Space,
+        num_members: int = 5,
+        member_cls: Type[RewardNet] = BasicRewardNet,
+        member_kwargs: Mapping[str, Any] = {},
+        member_normalize_output_layer: Optional[Type[nn.Module]] = None,
```
yawen-d, line 592: For now, I think `member_normalize_output_layer` isn't used even if it's specified because `forward_all()` and `forward()` don't used `predict_processed()` at all. I think we could directly remove `member_normalize_output_layer` unless we expect we'll ever have `RewardNetWrapper` as member_cls.

yawen-d:
Thanks for the implementations!

> @yawen-d It should be possible to train the reward ensemble now. If you have time can you test that this most basic version works? You can do this by just setting the reward class used by the reward ingredient to be a reward ensemble.

Sure. I just started some light benchmarking environments on the current version.

yawen-d:
> Thanks for the implementations!
> 
> > @yawen-d It should be possible to train the reward ensemble now. If you have time can you test that this most basic version works? You can do this by just setting the reward class used by the reward ingredient to be a reward ensemble.
> 
> Sure. I just started some light benchmarking environments on the current version.
> 
> In addition, one another feature to consider is to log the reward variance over time.

src/imitation/rewards/reward_nets.py
```diff
@@ -390,13 +390,17 @@ def predict_processed(
...
             update_stats: Whether to update the running stats of the normalization
                 layer.
+            **kwargs: kwargs passed to base predict_processed call.
 
         Returns:
             Computed normalized rewards of shape `(batch_size,`).
         """
         with networks.evaluating(self):
             # switch to eval mode (affecting normalization, dropout, etc)
-            rew_th = self.base.predict_th(state, action, next_state, done)
```
levmckinney, line 370: I changed this to improve composability of the wrappers. For example, if we add a `NormalizedRewardNet` on top of an `AddSTDRewardWrapper`, we want both to have an effect on the output of `predict_processed`.

While this will cause a performance hit from moving tenors back and forth from the GPU, I think it is worth it. There is probably a better solution, but it would need a larger refactor #464.

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
```
AdamGleave, line 1: Historically we haven't committed IDE configs. I can see a case for it, but probably best discussed in a separate issue/PR?

src/imitation/rewards/reward_nets.py
```diff
@@ -169,6 +169,7 @@ def predict_processed(
         action: np.ndarray,
         next_state: np.ndarray,
         done: np.ndarray,
+        **kwargs,
```
AdamGleave, line 172: Consider adding `del kwargs` to function body to self-document that these keyword arguments are ignored by this function.

src/imitation/rewards/reward_nets.py
```diff
@@ -181,6 +182,8 @@ def predict_processed(
             action: Actions of shape `(batch_size,) + action_shape`.
             next_state: Successor states of shape `(batch_size,) + state_shape`.
             done: End-of-episode (terminal state) indicator of shape `(batch_size,)`.
+            kwargs: additional kwargs may be passed to change the functionality of
+                subclasses.
```
AdamGleave, line 186: To check I understand how this is meant to work -- I can just call `net.predict_processed(special_kwarg=42)` and have it either (a) used if `isinstance(net, MySpecialClass)`, and (b) silently ignored otherwise?

This has the downside that if I typo it to `spcial_kwarg=42)` then `MySpecialClass` will presumably also just ignore it? But it has the benefit I don't have to do extensive `isinstance` checks, and it's more ducktype-friendly as you can implement `EvenMoreSpecialClass` that also takes `special_kwarg`.

src/imitation/rewards/reward_nets.py
```diff
@@ -239,6 +242,31 @@ def base(self) -> RewardNet:
         return self._base
 
 
+class RewardNetWithVariance(RewardNet):
+    """A reward net that keeps track of its epistemic uncertainty through variance."""
+
+    @abc.abstractmethod
+    def reward_moments(
```
AdamGleave, line 249: Other functions that take numpy inputs are prefixed with `predict_`, do we want to stick to that convention?

src/imitation/rewards/reward_nets.py
AdamGleave: Good catch -- correctness/composability trumps efficiency.

This makes me wonder why we return things as `numpy` at all, though. AFAICT, the main reason seems to be "RewardFn demands that" / "`VecEnvRewardWrapper` expects it". We could just add a simple utility method that torchifies arrays, and apply that as the very last step at the callee when needed?

AdamGleave (commented):
Got about 1/3rd of the way through, will try to take a look at the rest tomorrow!

.vscode/settings.json
levmckinney: Didn't mean to add this I will remove it.

src/imitation/rewards/reward_nets.py
levmckinney: Yes that is the tradeoff.

.vscode/settings.json
levmckinney: Can I add .vscode the git ignore, so this does not happen again?

src/imitation/rewards/reward_nets.py
levmckinney: However, there is a little more depth to it. The expected behavior is that all wrappers will pass along their `kwargs` to their base network. So in your example even if `net.base` was an instance of `MySpecialClass` it's `predict_processed` method should be called with special_kwarg=42. For an example of this see `test_normalization_wrappers_passes_on_kwargs`.

levmckinney: Just realized that this was not the behavior of `ShapedRewardNet`. I've made the aforementioned test a bit more general, now it's called `test_wrappers_passes_on_kwargs`  and added a default implementation of `predict_processed` to `RewardNetWrapper`. This was all done in commit 23b1f4d0da4bff33efe8ae9e7eec53b6fd1873a4. So might want to look over the affected lines again.

Note: that I have not forced the add standard deviation wrapper to propagate `kwargs`. The reasons for this are: First, there are no current use cases for propagating `kwargs` through an ensemble. Second, it would require use to pass `kwargs`, intended for `predict_processed` to, `predict_reward_moments` which seems a bit odd. I could be convinced to change my mind on this, If you think there may be a valid use case for having the `kwargs` propagate though to `predict_reward_moments` or beyond to the individual ensemble members.

src/imitation/rewards/reward_nets.py
levmckinney: I think it's a reasonable convention. I've switched it over to `predict_reward_moments`.

src/imitation/rewards/reward_nets.py
levmckinney: I've added this in a couple of places see 0246fb2713f0d70b3eed81132ae711d0e9940cc1.

tests/rewards/test_reward_nets.py
```diff
@@ -272,6 +322,168 @@ def test_potential_net_2d_obs():
...
+        reward_nets.RewardEnsemble(
+            env_2d.action_space,
+            env_2d.observation_space,
+            num_members=0,
+        )
+
+
+def test_reward_ensemble_predict_reward_moments(
+    two_ensemble: reward_nets.RewardEnsemble,
+    numpy_transitions: tuple,
```
AdamGleave, line 416: `tuple` fairly undefined. Could give type alias `NumPyTransitions = Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]`?

tests/rewards/test_reward_nets.py
```diff
@@ -272,6 +322,168 @@ def test_potential_net_2d_obs():
...
+    make_wrapper: reward_nets.RewardNetWrapper,
+    env_2d: Env2D,
+    numpy_transitions: tuple,
+):
+    basic_reward_net = reward_nets.BasicRewardNet(
+        env_2d.observation_space,
+        env_2d.action_space,
+    )
+    basic_reward_net.predict_processed = mock.Mock(return_value=np.zeros((10,)))
+    normalized_reward_net = make_wrapper(
```
AdamGleave, line 474: This isn't necessarily normalized (can also be shaped from `MAKE_BASIC_REWARD_NET_WRAPPERS`)

AdamGleave (commented):
This is in pretty good shape, only minor suggested changes.

Tests are missing docstrings. I think that's OK as they're generally pretty self explanatory from the name, but you might want to consider adding them for some of the more confusing/complex tests

.vscode/settings.json
AdamGleave: sure

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -809,27 +784,192 @@ def _probability(self, rews1: th.Tensor, rews2: th.Tensor) -> th.Tensor:
...
+                self.optim.step()
+                self.logger.record("loss", loss.item())
+                for metric in output["metrics"]:
+                    self.logger.record(metric, output["metrics"][metric])
+
+
+class RewardEnsembleTrainer(BasicRewardTrainer):
+    """Train a reward ensemble."""
+
+    _model: reward_nets.RewardEnsemble
```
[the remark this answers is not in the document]
levmckinney: I'm just specifying that the model has to be of a specific type. That is not the case for the base class. Both should have type annotations though.

src/imitation/rewards/reward_nets.py
levmckinney: I agree this smell's. Might be an instance of https://refactoring.guru/smells/message-chains. It also is related to #464 and #427. When we refactor reward nets we will need to accommodate for subclass specific configuration.

src/imitation/rewards/reward_nets.py
```diff
@@ -238,6 +244,42 @@ def __init__(
...
 
+    def predict_processed(
+        self,
+        state: np.ndarray,
+        action: np.ndarray,
+        next_state: np.ndarray,
+        done: np.ndarray,
+        **kwargs,
+    ) -> np.ndarray:
+        super().predict_processed.__doc__
```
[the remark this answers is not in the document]
levmckinney: Read the docs does in fact pick this up as a doc string: https://imitation--460.org.readthedocs.build/en/460/modules/imitation.rewards.html?highlight=rewardnetwrapper#imitation.rewards.reward_nets.RewardNetWrapper

src/imitation/rewards/reward_nets.py
```diff
@@ -550,3 +597,228 @@ def __init__(
...
+            done: The done flags as a torch tensor
+
+        Returns:
+            The reward given by each ensemble member. This tenor has a shape of
+                `(batch_size, num_members)`.
+        """
+        rewards = []
+        for net in self.members:
+            rewards.append(net(state, action, next_state, done))
+        rewards = th.stack(rewards, dim=-1)
```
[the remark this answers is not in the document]
levmckinney: No stack will create a new dim. `dim=-1` means we are adding it to the end of the shape. I've added an assent to clarify this.

src/imitation/rewards/serialize.py
```diff
@@ -73,15 +74,83 @@ def _strip_wrappers(
 def _make_functional(
     net: RewardNet,
     attr: str = "predict",
-    kwargs=dict(),
+    default_kwargs=dict(),
```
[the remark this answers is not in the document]
levmckinney: The main reason I have this is that the`update_states` `kwarg` in NorilizedRewardNet.predict_processed defaults to `True`. Currently, when loading a saved reward we expect `update_states=False`. `default_kwargs` is supped to account for this.

Thanks for the heads-up about this. Maybe we should add some a lint to check for this.

src/imitation/rewards/serialize.py
```diff
@@ -73,15 +74,83 @@ def _strip_wrappers(
...
+    **kwargs,
```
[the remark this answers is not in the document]
levmckinney: The main reason I have this is that the`update_states` `kwarg` in NorilizedRewardNet.predict_processed defaults to `True`. Currently, when loading a saved reward we expect `update_states=False`. `default_kwargs` is supped to account for this.

src/imitation/rewards/serialize.py
```diff
@@ -103,51 +172,80 @@ def f(
 
 reward_registry.register(
     key="RewardNet_shaped",
-    value=lambda path, _: _validate_reward(
-        _make_functional(_validate_type(th.load(str(path)), ShapedRewardNet)),
+    value=lambda path, _, **kwargs: _validate_reward(
+        _make_functional(
+            _validate_wrapper_structure(th.load(str(path)), [[ShapedRewardNet]]),
```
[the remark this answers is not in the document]
levmckinney: Compromise uses a tuple on the inside and a set on the outside? So `{(ShapedRewardNet,)}`. It's unlikely to cause typos, it looks nice, and it conveys that the order of the prefixes does not matter. These should only be passed to `_validate_reward_structure` which will just treat the set like an iterable. Thus, It should be a type error to try to mutate the set of prefixes within `_validate_reward_structure` anyway.

src/imitation/rewards/serialize.py
```diff
@@ -103,51 +172,80 @@ def f(
...
+    **kwargs,
+) -> common.RewardFn:
     """Load serialized reward.
 
     Args:
         reward_type: A key in `reward_registry`. Valid types
             include {reward_types}.
         reward_path: A path specifying the reward.
         venv: An environment that the policy is to be used with.
+        **kwargs: kwargs to pass to reward fn
```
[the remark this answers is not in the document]
levmckinney: The primary use case is for adjusting how the `alpha` given to the `AddRewardSTDWrapper.predict_processed`. I've written a test case demonstrating this use case. See `test_load_reward_passes_along_alpha_to_add_std_wrappers_predict_processed_method`.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -809,27 +784,192 @@ def _probability(self, rews1: th.Tensor, rews2: th.Tensor) -> th.Tensor:
...
+    def _make_data_loader(self, dataset: PreferenceDataset):
+        """Make a dataloader."""
+        return th.utils.data.DataLoader(
             dataset,
             batch_size=self.batch_size,
             shuffle=True,
             collate_fn=preference_collate_fn,
         )
+
+    def _train(self, dataset: PreferenceDataset, epoch_multiplier: float = 1.0):
```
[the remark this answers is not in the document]
levmckinney: It returns None. Personal preference is to leave these blank. That seems to be the convention in `imitation` as well.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -642,71 +643,45 @@ def preference_collate_fn(
...
+            fragment_pairs: Batch consisting of pairs of trajectory fragments.
+            preferences: The probability that the first fragment is preferred
+                over the second. Typically 0, 1 or 0.5 (tie).
 
-        Args:
-            dataset: the dataset of preference comparisons to train on.
-            epoch_multiplier: how much longer to train for than usual
-                (measured relatively).
+        Returns: # noqa: DAR202
+            loss: the loss
```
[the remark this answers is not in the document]
levmckinney: Switched over to using a named tuple instead of reward loss.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -767,17 +735,23 @@ def _loss(
...
-        self.logger.record("accuracy", accuracy.item())
-        return th.nn.functional.binary_cross_entropy(probs, preferences_th)
+        return {
+            "loss": th.nn.functional.binary_cross_entropy(probs, preferences_th),
+            "metrics": {"accuracy": accuracy.detach().cpu()},
+        }
 
-    def _rewards(self, transitions: Transitions) -> th.Tensor:
-        preprocessed = self.model.preprocess(
+    def _rewards(
```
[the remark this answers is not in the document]
levmckinney: `predict_th` evaluates the reward network with `no_grad` set. We need gradients on the reward function since this is a loss. Thus `predict_th` cannot be used here.

taufeeque9:
In case we want the average of the normalized reward of each network in `RewardEnsemble` (as stated in section 2.2.3 of the DRLHP paper), 

> We fit an ensemble of predictors, each trained on |D| triples sampled from D with replacement. The estimate \hat{r} is defined by independently normalizing each of these predictors and then averaging the results.

we'll need to make each of the member networks `NormalizedRewardNet`. However, the current implementation of forward in `RewardEnsemble` calls the `forward` method for all the members. And since the `forward` method of `NormalizedRewardNet` doesn't normalize the rewards (it is only done in `predict_processed`), we wouldn't be able to average the normalized rewards of the member networks.

It is also mentioned that the different networks in the ensemble are trained by sampling points from D with replacement. However, in our current implementation, we train all the networks on the same set D.

levmckinney:
> 

This should be addressed with [commit](https://github.com/HumanCompatibleAI/imitation/pull/460/commits/b431311dae66a70cba2b8319a6966c5ee119dd19)

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -809,27 +784,192 @@ def _probability(self, rews1: th.Tensor, rews2: th.Tensor) -> th.Tensor:
...
+
+
+def make_reward_trainer(
+    reward_model: reward_nets.RewardNet,
+    loss: RewardLoss,
+    reward_trainer_kwargs: Mapping[str, Any] = {},
+) -> RewardTrainer:
+    """Construct the correct type of reward trainer for this reward function."""
+    base_model = reward_model
+    while hasattr(base_model, "base"):
```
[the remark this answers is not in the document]
levmckinney: Seems like a good idea though I'm still in favor of some kind of more radical refactor of reward networks.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -939,16 +1079,23 @@ def __init__(
...
+        else:
+            self.reward_trainer = reward_trainer
+
         # If the reward trainer was created in the previous line, we've already passed
         # the correct logger. But if the user created a RewardTrainer themselves and
         # didn't manually set a logger, it would be annoying if a separate one was used.
         self.reward_trainer.logger = self.logger
         # the reward_trainer's model should refer to the same object as our copy
-        assert self.reward_trainer.model is self.model
+        # the only exception to this is when we are using a wrapped reward ensemble
```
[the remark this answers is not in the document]
levmckinney: I've made the model private, so I'm just going to remove this asserts.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -809,27 +784,192 @@ def _probability(self, rews1: th.Tensor, rews2: th.Tensor) -> th.Tensor:
...
+    reward_trainer_kwargs: Mapping[str, Any] = {},
+) -> RewardTrainer:
+    """Construct the correct type of reward trainer for this reward function."""
+    base_model = reward_model
+    while hasattr(base_model, "base"):
+        base_model = base_model.base
+
+    if isinstance(base_model, reward_nets.RewardEnsemble):
+        return RewardEnsembleTrainer(
+            base_model,
```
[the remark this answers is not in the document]
levmckinney: It's intentional. The only case where this would not be good is if for some reason we wanted to apply shaping on top of an ensemble. The current trainer does not support this, and I do not see a use case for it.

tests/rewards/test_reward_nets.py
```diff
@@ -148,6 +159,45 @@ def test_strip_wrappers_complex():
...
+        [[WrapperB, RewardNetA], [[RewardNetA]]],
+    )
+
+    # This should raise a type error since none the prefix is in the incorrect order
+    with pytest.raises(
+        TypeError,
+        match=r"Wrapper structure should be match \(one of\) \[.*\] but found \[.*\]",
+    ):
+        serialize._validate_wrapper_structure(reward_net, [[RewardNetA, WrapperB]])
+
```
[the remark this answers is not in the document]
levmckinney: Done

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -809,27 +784,192 @@ def _probability(self, rews1: th.Tensor, rews2: th.Tensor) -> th.Tensor:
...
+                output = self.loss(self._model, fragment_pairs, preferences)
+                loss = output["loss"]
+                loss.backward()
+                self.optim.step()
+                self.logger.record("loss", loss.item())
+                for metric in output["metrics"]:
+                    self.logger.record(metric, output["metrics"][metric])
+
+
+class RewardEnsembleTrainer(BasicRewardTrainer):
```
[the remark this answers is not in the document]
levmckinney: Changed to `EnsembleTrainer`

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -790,6 +764,7 @@ def _probability(self, rews1: th.Tensor, rews2: th.Tensor) -> th.Tensor:
             The softmax of the difference between the (discounted) return of the
             first and second trajectory.
         """
+        # TODO(lev) this could be extracted into its own class.
```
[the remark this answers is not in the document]
levmckinney: It was the second one. Though, I don't think this really needs to be an inline comment. The reason for this to be a class is that others can now try out different loss functions without having to fork imitation.

src/imitation/rewards/reward_nets.py
```diff
@@ -550,3 +597,228 @@ def __init__(
...
+        member_normalize_output_layer: Optional[Type[nn.Module]] = None,
+        **kwargs,
+    ):
+        """Initialize the RewardEnsemble.
+
+        Args:
+            observation_space: the observation space of the environment
+            action_space: the action space of the environment
+            num_members: the number of members in the ensemble. Must be at least 1.
+            member_cls: class of the constituent reward networks
```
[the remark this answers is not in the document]
levmckinney: Mainly I did this to make configuration simpler. Pushing things out to make reward net would mean that we would need a new key, `reward_ensemble_members_config`, that would only ever be used by the reward ensemble and nothing else. It didn't feel particularly clean, so I decided to push ensemble creation into the constructor. However, I see your point about keeping this out of the core API. I'll change it back to taking a list of reward networks and move the functinality out.

src/imitation/rewards/reward_nets.py
```diff
@@ -550,3 +597,228 @@ def __init__(
...
+            done,
+        )
+
+        return reward_mean + alpha * np.sqrt(reward_var)
+
+    def forward(self, *args):
+        return self.base.forward(*args)
+
+
+def make_reward_net(
```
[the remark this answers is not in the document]
levmckinney: Let's consolidate our discussion of this in the above thread.

src/imitation/scripts/common/reward.py
```diff
@@ -73,19 +81,20 @@ def make_reward_net(
...
+        add_std_alpha: multiple of reward function standard deviation to add to the
+            reward in predict_processed. Must be None when using a reward function that
+            does not keep track of variance. Defaults to None.
 
     Returns:
-        None if `reward_net_cls` is None; otherwise, an instance of `reward_net_cls`.
+        A, possibly wrapped, instance of `net_cls`.
     """
-    reward_net = net_cls(
+    reward_net = reward_nets.make_reward_net(
```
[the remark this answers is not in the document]
levmckinney: Again consolidating discussion in another thread. I'm going to refactor this.

src/imitation/algorithms/preference_comparisons.py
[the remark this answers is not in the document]
AdamGleave: Ah, good point, I missed that this was restricting it to an ensemble. Adding annotation to both base and this class sounds good!

AdamGleave:
> It is also mentioned that the different networks in the ensemble are trained by sampling points from D with replacement. However, in our current implementation, we train all the networks on the same set D.

Good point, this method ("bagging") can work a lot better than just relying on seeds for diversity. I think might be best addressed in a separate PR though as this one is already getting long and it is not a core functionality. Can you open a new issue to track this (and consider working on it yourself once done with the active learning side)?

src/imitation/rewards/reward_nets.py
```diff
@@ -600,98 +628,85 @@ def forward(self, state: th.Tensor) -> th.Tensor:
...
 
         Raises:
             ValueError: if num_members is less than 1
         """
-        del kwargs
         super().__init__(observation_space, action_space)
-        if num_members < 1:
+
+        members = list(members)
+        if len(members) == 0:
```
AdamGleave, line 660: More Pythonic:
```suggestion
        if not members:
```

src/imitation/rewards/reward_nets.py
```diff
@@ -700,35 +715,77 @@ def predict_reward_moments(
             action: Actions of shape `(batch_size,) + action_shape`.
             next_state: Successor states of shape `(batch_size,) + state_shape`.
             done: End-of-episode (terminal state) indicator of shape `(batch_size,)`.
+            **kwargs: passed along to predict processed.
 
         Returns:
-            Computed reward std of shape `(batch_size,)`.
+            * Reward mean of shape `(batch_size,)`.
+            * Reward std of shape `(batch_size,)`.
```
AdamGleave, line 722: docstring mentions std but return value is variance

src/imitation/rewards/reward_nets.py
```diff
@@ -700,35 +715,77 @@ def predict_reward_moments(
...
+    def predict(
+        self,
+        state: np.ndarray,
+        action: np.ndarray,
+        next_state: np.ndarray,
+        done: np.ndarray,
+        **kwargs,
+    ):
+        """Return the mean of the ensemble members."""
+        return self.predict_processed(state, action, next_state, done, **kwargs)
```
AdamGleave, line 762: It seems marginally more natural to me to have predict_processed call predict than the other way around (as if they're equal then the return value is presumably "unprocessed").

src/imitation/rewards/reward_nets.py
```diff
@@ -700,35 +715,77 @@ def predict_reward_moments(
...
 
 class AddSTDRewardWrapper(RewardNetWrapper):
-    """That adds a multiple of the standard deviation to the reward function."""
+    """Adds a multiple of the estimated standard deviation to mean reward."""
 
     base: RewardNetWithVariance
 
     def __init__(self, base: RewardNetWithVariance, default_alpha: float = 0.0):
-        """Create a conservative reward network.
+        """Create a reward network that added a multiple of the standard deviation.
```
AdamGleave, line 771:
```suggestion
        """Create a reward network that adds a multiple of the standard deviation.
```

src/imitation/rewards/reward_nets.py
AdamGleave: Superclass mentions variance too in docstring.

src/imitation/rewards/reward_nets.py
```diff
@@ -700,35 +715,77 @@ def predict_reward_moments(
...
+                This is used to compute the standard deviation.
             default_alpha: multiple of standard deviation to add to the reward mean.
                 Defaults to 0.0.
+
+        Raises:
+            ValueError: if base is not an instance of RewardNetWithVariance
         """
         super().__init__(base)
+        if not isinstance(base, RewardNetWithVariance):
+            raise ValueError(
```
AdamGleave, line 784:
```suggestion
            raise TypeError(
```

src/imitation/rewards/serialize.py
```diff
@@ -88,21 +88,42 @@ def _strip_wrappers(
...
     default_kwargs.update(kwargs)
     return lambda *args: getattr(net, attr)(*args, **default_kwargs)
 
 
-WrapperPrefix = List[Type[RewardNet]]
+WrapperPrefix = Sequence[Type[RewardNet]]
+
+
+def _prefix_matches(wrappers: Sequence[Any], prefix: Sequence[Any]):
+    """Return true if prefix is a prefix of wrappers."""
```
AdamGleave, line 104:
```suggestion
    """Return true if `prefix` is a prefix of `wrappers`."""
```

src/imitation/rewards/serialize.py
```diff
@@ -88,21 +88,42 @@ def _strip_wrappers(
...
+    # Base cases
+    if len(prefix) == 0:
+        # If we run of of prefix before running of of wrappers
```
AdamGleave, line 107:
```suggestion
        # If we run out of prefix before running out of wrappers
```

src/imitation/rewards/serialize.py
```diff
@@ -88,21 +88,42 @@ def _strip_wrappers(
...
+        return True
+    elif len(wrappers) == 0:
+        # If we run out of wrappers before we run of to prefix
```
AdamGleave, line 110:
```suggestion
        # If we run out of wrappers before we run out of prefix
```

src/imitation/rewards/serialize.py
```diff
@@ -88,21 +88,42 @@ def _strip_wrappers(
...
+        default_kwargs = {}
     default_kwargs.update(kwargs)
     return lambda *args: getattr(net, attr)(*args, **default_kwargs)
 
 
-WrapperPrefix = List[Type[RewardNet]]
+WrapperPrefix = Sequence[Type[RewardNet]]
+
+
+def _prefix_matches(wrappers: Sequence[Any], prefix: Sequence[Any]):
```
AdamGleave, line 103:
```suggestion
def _prefix_matches(wrappers: Sequence[Type[Any]], prefix: Sequence[Type[Any]]):
```

src/imitation/scripts/common/reward.py
```diff
@@ -62,16 +72,43 @@ def config_hook(config, command_name, logger):
         res["net_cls"] = default_net
     if "normalize_input_layer" not in config["reward"]["net_kwargs"]:
         res["net_kwargs"] = {"normalize_input_layer": networks.RunningNorm}
+
+    if config["reward"]["net_cls"] is reward_nets.RewardEnsemble:
```
AdamGleave, line 76: This will only apply to `RewardEnsemble` itself, not subclasses -- is that desired?

src/imitation/scripts/common/reward.py
```diff
@@ -84,17 +121,42 @@ def make_reward_net(
...
+        ensemble_member_config: The configuration for individual ensemble
+            members. Note that ensemble_members.net_cls must not be
+            :class: `reward_nets.RewardEnsemble`. Must be set if using
+            `net_cls = ` :class: `reward_nets.RewardEnsemble`.
 
     Returns:
         A, possibly wrapped, instance of `net_cls`.
+
+    Raises:
+        ValueError: Using a reward ensemble but fail to provide configuration.
```
AdamGleave, line 135:
```suggestion
        ValueError: Using a reward ensemble but failed to provide configuration.
```

src/imitation/scripts/common/reward.py
```diff
@@ -84,17 +121,42 @@ def make_reward_net(
...
+
+        reward_net = net_cls(venv.observation_space, venv.observation_space, members)
+
+        if add_std_alpha is not None:
+            reward_net = reward_nets.AddSTDRewardWrapper(
+                reward_net,
+                default_alpha=add_std_alpha,
+            )
+
+        # TODO should we support normalization on top of the ensemble or just within
```
AdamGleave, line 158: Good question. Given it'd complexify the code I lean towards not supporting it for now and we can add it later if a research use case emerges for it.

tests/algorithms/test_preference_comparisons.py
```diff
@@ -183,8 +191,62 @@ def test_trainer_no_crash(
...
+    if isinstance(reward_net, reward_nets.RewardEnsemble):
+        assert isinstance(
+            main_trainer.reward_trainer,
+            preference_comparisons.EnsembleTrainer,
+        )
+    elif hasattr(reward_net, "base") and isinstance(
+        reward_net.base,
+        reward_nets.RewardEnsemble,
+    ):
+        assert isinstance(
```
AdamGleave, line 236: Same body as previous if clause. Could simplify by merging check, e.g.

```python
base_reward_net = reward_net.base if hasattr(reward_net, "base") or reward_net
if isinstance(base_reward_net, reward_nets.RewardEnsemble):
  # ...
```

tests/rewards/test_reward_nets.py
```diff
@@ -3,6 +3,7 @@
 import logging
 import numbers
 import os
+from tempfile import TemporaryDirectory
```
AdamGleave, line 6: Import modules not individual classes: https://google.github.io/styleguide/pyguide.html#22-imports

tests/rewards/test_reward_nets.py
```diff
@@ -168,21 +196,27 @@ class WrapperB(reward_nets.RewardNetWrapper):
...
     with pytest.raises(
         TypeError,
-        match=r"Wrapper structure should be match \(one of\) \[.*\] but found \[.*\]",
+        match=r"Wrapper structure should match \[.*\] but found \[.*\]",
     ):
-        serialize._validate_wrapper_structure(reward_net, [[RewardNetA]])
+        serialize._validate_wrapper_structure(reward_net, {(RewardNetA,)})
+
+    # Reward net is not wrapped at all this should raise a type error.
+    with pytest.raises(TypeError):
```
AdamGleave, line 215: I notice you drop the `match=` keyword argument here. If that's just for brevity, you could define earlier on:

```
raise_ctx_mgr = pytest.raises(TypeError, match=...)
```

and then do `with raise_ctx_mgr` each time to avoid duplication.

(I think this works; if not, you could definitely make this work using functools.partial.)

tests/rewards/test_reward_nets.py
```diff
@@ -323,20 +359,13 @@ def test_potential_net_2d_obs():
...
+    """A simple test of the RewardEnsemble constructor."""
     env = gym.make(env_name)
-    ensemble = reward_nets.RewardEnsemble(
-        env.action_space,
-        env.observation_space,
-        num_members,
-        net_cls,
-    )
+    ensemble = make_ensemble(env.observation_space, env.action_space, num_members)
     assert ensemble
```
AdamGleave, line 367: What is this testing? (That it's not None/False?)

AdamGleave (commented):
Thanks for the changes, seems almost ready! Mostly minor suggestions, but test cases still failing -- once changes addressed & tests passing let me know and I can do a final review.

tests/rewards/test_reward_nets.py
levmckinney: This test is extraneous now I've removed it.

src/imitation/algorithms/preference_comparisons.py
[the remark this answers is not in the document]
levmckinney: The issue is that wrappers will hide attributes list `ensemble.members` and `predict_processed_all`. Which we need access to train the ensemble.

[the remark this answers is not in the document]
AdamGleave: What's the use case where we want to have a wrapper around `RewardEnsemble` but ignore it during training? Is it just normalization, that's not used during training anyway and just for RL training?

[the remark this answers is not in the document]
levmckinney: We might want to add a `AddSTDRewardWrapper` during training to encourage optimism. In fact the current default configuration will always do this but just leave alpha=0.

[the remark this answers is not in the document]
levmckinney: If we don't want to support any wrapping at all I would advocate for removing AddSTDRewardWrapper entirely and moving the ability to add the standard deviation to the abstract class RewardNetWithVariance.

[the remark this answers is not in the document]
AdamGleave: OK, the `AddSTDRewardWrapper` use case makes sense for me (add it to RL training but not reward model training). My suggestion is that we have a whitelist of "compatible" wrappers, and strip off those for reward model training. Anything else we throw an error or issue a warning for.

Generally the heuristic here is it's worth minor inconvenience for users/adding new features (i.e. update the whitelist) to avoid subtle bugs (a wrapper getting stripped that people didn't expect).

This whole thing is a good example of a use case we should try to support more cleanly post-refactor.

AdamGleave (approved):
LGTM. 

*Before merging*, please:
  - Review my changes. I pushed some small changes to restructure one piece of the code, and fix some typos.
  - Wait for `windows-ci-improvements` to get merged (if not already reviewed, you can review it to accelerate that). We should then let this PR be rebased and merge into `master`.

AdamGleave:
I think this is good to merge once conflicts resolved.

levmckinney:
Seems like rerunning the tests fixed things. see  #502

PR #465, opened 2022-07-08
[tests] Support loading cached npz files & remove pip workaround

AdamGleave:
Some minor improvements to tests & CI:
  - Change `tests/conftest.py` to use `types.load_with_rewards` rather than `pickle.load` directly which is breaking with new `npz` format. Also change filenames to default to `.npz`. This should stop CircleCI from taking 45m to complete as it had to regenerate the rollouts each time :)
  - Remove a lengthly workaround for `pip` being slow to install as  https://github.com/pypa/pip/issues/2195 was closed after 8 years :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,7 +63,7 @@ commands:
 
       - run:
           name: install imitation
-          command: pip install .
+          command: pip install --upgrade --force-reinstall --no-deps .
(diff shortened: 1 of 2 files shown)


levmckinney (approved):
LGTM. Thanks for the speed-up!

PR #466, opened 2022-07-08
Update `convert_trajs` script to support new format

AdamGleave:
`convert_trajs` script loads trajectories (perhaps in an old format) and saves in latest format. This PR: 
  - Changes `convert_trajs` to save with `".npz"` file format rather than overwriting original `".pkl"` file, and related to this renames `convert_trajs_in_place` -> `convert_trajs` (it might still do it in place, but only if it's already in npz format, in which case should not be any actual change as idempotent).
  - Adds deprecation warning when loading rollouts in old format. It's not much compatibility code at all, but we'll probably still drop support for it eventually.
  - Make test check that the deserialized trajectories are the same after conversion -- not just for idempotence.


src/imitation/testing/types.py
```diff
@@ -0,0 +1,31 @@
...
+    for k, a_v in dict_a.items():
+        b_v = dict_b[k]
+        if k == "infos":
+            # Treat None equivalent to sequence of empty dicts
+            a_v = [{}] * len(traj_a) if a_v is None else a_v
+            b_v = [{}] * len(traj_b) if b_v is None else b_v
+        assert np.array_equal(a_v, b_v)
+
+
+def assert_traj_sequences_equal(
```
ernestum, line 24: Maybe we would not need this if `assert_traj_equal` would be implemented as `__eq__(self, other)` in the trajectory class?

tests/test_scripts.py
```diff
@@ -687,21 +689,29 @@ def test_analyze_gather_tb(tmpdir: str):
     assert run.result["n_tb_dirs"] == 2
 
 
-def test_convert_trajs_in_place(tmpdir: str):
+def test_convert_trajs(tmpdir: str):
+    """Tests that convert_trajs is idempotent and does not change the data."""
     shutil.copy(CARTPOLE_TEST_ROLLOUT_PATH, tmpdir)
     tmp_path = os.path.join(tmpdir, os.path.basename(CARTPOLE_TEST_ROLLOUT_PATH))
     exit_code = subprocess.call(
```
ernestum, line 696: Could we run this directly via sacred instead of subprocess? This way we could also check if the deprecation warning is raised, ensuring that the test really converts from the old format to the new format.

ernestum (commented):
Looks good. I left some comments.
I would also move the test trajectory to the `testdata` root and call it something like `demonstrations_old_format.pkl` so we don't accidentally replace it with demonstrations in the new format.

tests/test_scripts.py
AdamGleave: I've added a check that original input file is pickle format and output is in npz format.

Unfortunately `convert_trajs` is not written in Sacred. We could call the `main()` method directly and mock `sys.argv`, or call the `update_traj_file_in_place` method directly. Perhaps ideally we'd do both unit and integration test, but feels a bit overkill to me for such a simple piece of code.

src/imitation/testing/types.py
AdamGleave: Good idea, I've implemented `__eq__` now (plus added a test for it).

AdamGleave:
Thanks for the review, made the requested changes.

tests/data/test_types.py
```diff
@@ -146,6 +140,52 @@ def test_valid_trajectories(
         for traj in trajs:
             assert len(traj) == length
 
+    def test_equals(
```
ernestum, line 143: This test turned out beautiful!
In the spirit of only testing one thing in each test, we could split it up into multiple tests.
I would propose tests called: `test_that_traj_unequal_to_primitives`, `test_that_traj_equal_to_self`, `test_that_traj_equal_to_copy`, `test_that_traj_unequal_to_copy_with_reward`, `test_that_traj_unequal_to_perturbed`.
But I can do this later when I do a general sweep through all of the tests.

src/imitation/data/types.py
```diff
@@ -60,10 +60,35 @@ class Trajectory:
     contain the final state of an episode (even if missing the start of the episode).
     """
 
-    def __len__(self):
+    def __len__(self) -> int:
         """Returns number of transitions, equal to the number of actions."""
         return len(self.acts)
 
+    def __eq__(self, other) -> bool:
```
ernestum, line 67: Ahh, almost too beautiful for such a simple function :-)

tests/test_scripts.py
ernestum: I would prefer mocking `sys.argv` or even calling `update_traj_file_in_place` directly. I think it would make the test easier to read and mostly also easier to debug. I try to avoid subprocesses in tests at all costs since it makes debugging so much more painful.

ernestum (approved):
This turned out nicely! I have a few minor nitpicks left, which I would consider optional for merging this PR.

tests/data/test_types.py
AdamGleave: Thanks for the suggestion. I've split up the tests. I went for a slightly different decomposition to avoid having some tiny tests with lots of boilerplate.

AdamGleave:
Thanks, addressed outstanding comments -- merging now.

PR #472, opened 2022-07-14
Uniformize logits sign convention for adversarial algorithms

### LINKED ISSUE #275: In GAIL and AIRL networks, use logits_expert_is_high instead of logits_gen_is_high
For consistency with GAIL/AIRL paper implementation and equations.

We currently use discriminator logits that are high if they predict a particular sample is from the generator, rather than the expert (reason: I chose the wrong arbitrary sign when I first made this repo). By reversing the meaning of our logits we will better align with "the literature" and won't have to [run into confusion when editing the code](https://github.com/HumanCompatibleAI/imitation/pull/198#discussion_r444539134).

Rocamonde:
Fixes #275. This PR uniformizes the convention of the logits outputted by reward nets in adversarial networks (e.g. GAIL and AIRL) as reported on #275 to mean that high logits correspond to the discriminator considering the trajectory to come from the expert, and conversely, low logits mean that the discriminator believes the trajectory comes from the generator. 

Since the monotonicity of the reward net output is reversed, the reward passed to the generator for policy training is also adequately updated for GAIL and AIRL.

### DIFF
--- a/src/imitation/algorithms/adversarial/airl.py
+++ b/src/imitation/algorithms/adversarial/airl.py
@@ -71,25 +71,38 @@ class AIRL(common.AdversarialTrainer):
         done: th.Tensor,
         log_policy_act_prob: th.Tensor,
     ) -> th.Tensor:
-        """
-        Compute the discriminator's logits for each state-action sample.
+        r"""Compute the discriminator's logits for each state-action sample.
 
         In Fu's AIRL paper (https://arxiv.org/pdf/1710.11248.pdf), the
-        discriminator output was given as 
+        discriminator output was given as
         $$
-        D_{\theta}(s,a) = 
+        D_{\theta}(s,a) =
         \frac{ \exp{r_{\theta}(s,a)} } { \exp{r_{\theta}(s,a)} + \pi(a|s) }
         $$
-        with a high value corresponding to expert and a low value corresponding to generator.
+        with a high value corresponding to the expert and a low value corresponding to
+        the generator.
 
-        In other words, the discriminator output is the probability that the action is taken 
-        by the expert rather than the generator.
+        In other words, the discriminator output is the probability that the action is
+        taken by the expert rather than the generator.
 
         The logit of the above is given as
         $$
         \operatorname{logit}(D_{\theta}(s,a)) = r_{\theta}(s,a) - \log{ \pi(a|s) }
         $$
-        which is what is returned by this function.        
+        which is what is returned by this function.
+
+        Args:
+            state: ?
+            action: ?
+            next_state: ?
+            done: ?
+            log_policy_act_prob: ?
+
+        Returns:
+            The logits of the discriminator for each state-action sample.
+
+        Raises:
+            TypeError: If `log_policy_act_prob` is not a tensor.
         """
         if log_policy_act_prob is None:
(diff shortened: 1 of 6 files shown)


AdamGleave:
Your assert doesn't handle the case that `log_policy_act_prob` is none which is why tests are continuing to fail for off-policy algorithms.

AdamGleave:
Tests are failing, didn't have time to track down the root cause, ping me when fixed and I'll take another look!

AdamGleave:
Will try and look at this tomorrow but not sure I'll have time before I leave. Nora, can you take a look on Monday if I've not reviewed by then?

src/imitation/algorithms/adversarial/airl.py
```diff
@@ -63,32 +63,53 @@ def __init__(
...
+        taken by the expert rather than the generator.
+
+        The logit of the above is given as
+        $$
+        \operatorname{logit}(D_{\theta}(s,a)) = r_{\theta}(s,a) - \log{ \pi(a|s) }
+        $$
+        which is what is returned by this function.
+
+        Args:
+            state: ?
```
norabelrose, line 95: It looks like you were using `?` as a placeholder here and forgot to fill it in? In the analogous method for GAIL, we don't have an `Args` section in the docstring at all. For completeness it would probably be best to include full `Args` sections in both docstrings. At least it should be consistent for the two methods.

Rocamonde:
Hi @norabelrose , thanks for your comments. I left the docstrings with question marks as I was not sure what to say that would be useful. I have filled something in now, let me know your thoughts, and feel free to tweak it if you can think of more useful or accurate descriptions.

norabelrose (approved):
LGTM. Adam says he will do a quick run through as well and then we can merge.

Rocamonde:
Sorry, saw "approved" and merged from my phone thinking you had missed it, didn't notice Adam wanted to have another look. Only saw now on my Mac. Feel free to point out any issues and I'll revert the merge with any relevant fixes if necessary.

src/imitation/algorithms/adversarial/gail.py
```diff
@@ -3,16 +3,65 @@
...
 
 from imitation.algorithms import base
 from imitation.algorithms.adversarial import common
 from imitation.rewards import reward_nets
 
 
-class LogSigmoidRewardNet(reward_nets.RewardNet):
-    """Wrapper for reward network that takes log sigmoid of wrapped network."""
+class RewardNetFromDiscriminatorLogit(reward_nets.RewardNet):
+    r"""Converts the discriminator logits raw value to a reward signal.
```
AdamGleave, line 16: thanks for writing this explanation, the math checks out and is much clearer than before :)

AdamGleave (commented):
Mostly minor suggestions, but should be addressed -- can probably make a new PR to fix them rather than reverting this merge.

Higher-level question: have we checked the algorithms still get good results? We should be able to get expert (500) reward on seals/CartPole-v0 with GAIL and AIRL for example.

src/imitation/algorithms/adversarial/gail.py
Rocamonde: Glad to hear! :)

PR #473, opened 2022-07-14
Remove references to DiscrimNet

### LINKED ISSUE #468: DiscrimNet nomenclature referring to AdversarialTrainer class
In the following docstring:

https://github.com/HumanCompatibleAI/imitation/blob/2a383ebfeef1753a9538435cb35e7b1f7f0ddb80/src/imitation/algorithms/adversarial/common.py#L23-L35

Should line 32 not refer to the class `AdversarialTrainer` instead of `DiscrimNet`, which in my understanding is not to be found anywhere in the repo? (apart from line 79 in a docstring within `scripts/train_adversarial.py` file). That is where the `logits_gen_is_high` function is found. 

I am asking because the AdversarialTrainer class is not a torch module, which is what the `<>Net` notation suggests based on other classes.

Happy to change this myself if this was just an errata.

Rocamonde:
Fixes #468 - removes DiscrimNet from those few places in the docstrings where it was still referenced, and brings any other part of the docstring up to speed if required.

### DIFF
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -29,7 +29,7 @@ def compute_train_stats(
 
     Args:
         disc_logits_gen_is_high: discriminator logits produced by
-            `algorithms.adversarial.common.AdversarialTrainer.logits_gen_is_high`.
+            `AdversarialTrainer.logits_gen_is_high`.
         labels_gen_is_one: integer labels describing whether logit was for an
             expert (0) or generator (1) sample.
         disc_loss: final discriminator loss.
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -29,7 +29,7 @@ def compute_train_stats(
 
     Args:
         disc_logits_gen_is_high: discriminator logits produced by
-            `DiscrimNet.logits_gen_is_high`.
+            `algorithms.adversarial.common.AdversarialTrainer.logits_gen_is_high`.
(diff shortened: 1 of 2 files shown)


src/imitation/algorithms/adversarial/common.py
```diff
@@ -29,7 +29,7 @@ def compute_train_stats(
 
     Args:
         disc_logits_gen_is_high: discriminator logits produced by
-            `DiscrimNet.logits_gen_is_high`.
+            `algorithms.adversarial.common.AdversarialTrainer.logits_gen_is_high`.
```
AdamGleave, line 32: Do we need the fully qualified name? AdversarialTrainer is defined in this very file.

src/imitation/scripts/train_adversarial.py
```diff
@@ -76,7 +76,8 @@ def train_adversarial(
     """Train an adversarial-network-based imitation learning algorithm.
 
     Checkpoints:
-        - DiscrimNets are saved to `f"{log_dir}/checkpoints/{step}/discrim/"`,
+        - AdversarialTrainer train and test RewardNets are saved to 
+           `f"{log_dir}/checkpoints/{step}/reward_train.pt"` (and similarly for test)
```
AdamGleave, line 80: Perhaps
```suggestion
           `f"{log_dir}/checkpoints/{step}/reward_{train,test}.pt"`
```

AdamGleave (commented):
There's a minor lint issue (whitespace) causing checks to fail.

Couple of small suggestions

AdamGleave (approved):
LGTM

PR #475, opened 2022-07-14
Fixes Wrong type passed into _torchify_array on adversarial/common.py and other minor type issues

### LINKED ISSUE #469: Wrong type passed into _torchify_array on adversarial/common.py
In the snippet below:
https://github.com/HumanCompatibleAI/imitation/blob/2a383ebfeef1753a9538435cb35e7b1f7f0ddb80/src/imitation/algorithms/adversarial/common.py#L563-L566
the variable `labels_expert_is_one` is indeed an NDArray; however, `log_policy_act_prob` is already (afaik) a torch tensor. There is no need to torchify it, and it gives a type error since `_torchify_array` only takes in np.ndarray objects (or None).

I see two courses of action:
1. Remove the torchify call after making sure that log_policy_act_prob is indeed always a tensor (seems the most logical thing to do IMO)
2. Only overload the type of _torchify_array (I think torch.as_tensor is idempotent) / change function body + signature to do nothing if the input is already a tensor, and call the function anyways there and potentially in other places (seems "safer" but overly cautious and unclean).

Happy to provide a fix for this whatever way is preferred by others.

Rocamonde:
Fixes #469, as well as other minor type issues in algorithms.adversarial.common.py

### DIFF
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -221,10 +221,13 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
         self.gen_algo.set_logger(self.logger)
 
         if gen_train_timesteps is None:
-            gen_train_timesteps = self.gen_algo.get_env().num_envs
+            gen_algo_env = self.gen_algo.get_env()
+            assert gen_algo_env is not None
+            self.gen_train_timesteps = gen_algo_env.num_envs
             if hasattr(self.gen_algo, "n_steps"):  # on policy
-                gen_train_timesteps *= self.gen_algo.n_steps
-        self.gen_train_timesteps = gen_train_timesteps
+                self.gen_train_timesteps *= self.gen_algo.n_steps
+        else:
+            self.gen_train_timesteps = gen_train_timesteps
 
         if gen_replay_buffer_capacity is None:
             gen_replay_buffer_capacity = self.gen_train_timesteps
@@ -432,7 +435,7 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
         self,
         obs_th: th.Tensor,
         acts_th: th.Tensor,
-    ) -> th.Tensor:
+    ) -> Optional[th.Tensor]:
         """Evaluates the given actions on the given observations.
 
         Args:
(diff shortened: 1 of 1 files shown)


src/imitation/algorithms/adversarial/common.py
```diff
@@ -451,6 +454,12 @@ def _get_log_policy_act_prob(
             )
         elif isinstance(self.policy, sac_policies.SACPolicy):
             gen_algo_actor = self.policy.actor
+            if gen_algo_actor is None:
```
AdamGleave, line 457: I think it can only happen if `SACPolicy._build` hasn't been called yet: https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/sac/policies.py#L292 But that should never happen as `SACPolicy.__init__` always calls `self._build` at the end: https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/sac/policies.py#L289

So I think this is just a false positive from your type checker? I'd add an `assert gen_algo_actor is not None` to quieten it and move on.

AdamGleave (commented):
Linting is failing with some minor issues.

Tests were screaming that `log_policy_act_prob` is not a Tensor. Looks like we're converting it to NumPy in `_get_log_policy_act_prob` which is bizarre as that's an unnecessary conversion (we convert it back) and is against what the method claims in it's type signature. I've pushed a fix for that, let's see if it breaks anything else...

Sorry our CI didn't run these tests before I triggered it, not sure what went wrong there, let me know if this happens again. You can also run the tests locally using `pytest -n <num_cores> -vv tests/`, and I like to do that for the most relevant file (e.g. `pytest -vv tests/algorithms/test_adversarial.py`) to have a quicker feedback loop.

AdamGleave:
Your assert doesn't handle the case that log_policy_act_prob is none which is why tests are continuing to fail for off-policy algorithms.

Rocamonde:
I get all tests passing. Had to make a few tweaks to get to run the tests. Mentioned them in Issue #476. Do take a look, I think it should be good to merge.

Rocamonde:
How can there be such a massive drop in the code coverage? I barely changed the file

AdamGleave:
> How can there be such a massive drop in the code coverage? I barely changed the file

I'm only seeing 1 line more missed in the codecov report, I think it's just a high % change because it's considering it relative to the small diff. I can override this check.

AdamGleave (approved):
LGTM

PR #477, opened 2022-07-15
Add Support for Apple Silicon installation

### LINKED ISSUE #476: Experiments Compatibility with Apple Silicon
Currently, the experiments module is incompatible with Apple Silicon.

`brew install gdate` is not a known formula on Apple Silicon (and most likely under Intel as well, as I could not find it onlne), and `brew install coreutils` has to be used. 

Additionally, `common.sh` exports the GNU_GETOPT path to be `/usr/local/opt/gnu-getopt/bin/getopt`, but in Apple Silicon brew packages are no longer on `/usr/local/opt` but under `/opt/homebrew/opt/`. So a check to detect the architecture of the processor is necessary.

Rocamonde:
Fixes #476.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -158,3 +158,5 @@ dmypy.json
 /tests/testdata/reward_models
 # Produced by examples/quickstart.sh
 /quickstart/
+
+.vscode/
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 """Setup for imitation: a reward and imitation learning library."""
 
 from setuptools import find_packages, setup
+from setuptools.command.install import install
 
 import src.imitation  # pytype: disable=import-error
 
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -48,7 +49,29 @@ def get_readme() -> str:
         return f.read()
 
 
+class InstallCommand(install):
+
+    def run(self):
+        """Run the install command."""
+        install.run(self)
+        from sys import platform
```
AdamGleave, line 57: Is there a reason the imports are nested? We usually go for top level in line with https://google.github.io/styleguide/pyguide.html#313-imports-formatting

setup.py
```diff
@@ -48,7 +49,29 @@ def get_readme() -> str:
...
+        import warnings
+
+        if platform == "darwin":
+            warnings.warn(
+                "Installation of important packages for macOS is required. "
+                "Scripts in the experiments folder will likely not run without these packages: "
+                "gnu-getopt, parallel, coreutils. They can be installed with Homebrew by running "
+                "`brew install gnu-getopt parallel coreutils`."
+                "See https://brew.sh/ for installation instructions.",
+                category=Warning,
```
AdamGleave, line 67: `Warning` is the base category; is there something more specific we can use? The default of `warnings.warn` of `UserWarning` seems preferable but may not be the best option.

AdamGleave (commented):
Some minor suggestions.

Apart from those and lint issues caught by the CI, seems good.

Running `ci/code_checks.sh` will run the linter, and is often useful to do this pre-commit even if CI was working for you as it'll run faster locally.

setup.py
Rocamonde: My bad, that was Copilot. Will change now.

setup.py
Rocamonde: We can create a custom class called `CompatibilityWarning` or something of the kind, or just use UserWarning. I don'
t see any other more fitting default class.

AdamGleave: `UserWarning` seems OK, making a new class is probably overkill

.gitignore
```diff
@@ -158,3 +158,5 @@ dmypy.json
 /tests/testdata/reward_models
 # Produced by examples/quickstart.sh
 /quickstart/
+
+.vscode/
```
AdamGleave, line 162: Missing a newline.

AdamGleave (approved):
LGTM

PR #480, opened 2022-07-20
Allow warm-starting through pre-trained policies for fine-tuning

### LINKED ISSUE #376: Loading pre-trained/warmstarted policies for fine-tuning
At some point, we'll need a feature for loading a pre-trained/warmstarted policy for continuous training. 

This feature is currently required in [`imitation.scripts.train_preference_comparisons`](https://github.com/HumanCompatibleAI/imitation/blob/dacb2425e9d19d57e318578a414b50cb98ead647/src/imitation/scripts/train_preference_comparisons.py#L101), and will be required in `imitation.scripts.train_rl`.

To load an existing policy for further training, we at least need to have the following features:
- enabling the setting of `vec_normalize.training=True`.
- No need for introducing a NormalizePolicy like what's done in [`serialize`](https://github.com/HumanCompatibleAI/imitation/blob/dacb2425e9d19d57e318578a414b50cb98ead647/src/imitation/policies/serialize.py#L168).

Also, there are ways to simplify `imitation.policies.serialize`:
- We could refactor policy to algorithm where appropriate. The naming of `policy` and `algorithm` in `imitation.policies.serialize` is kind of ambiguous. e.g. in [`load_policy()`](https://github.com/HumanCompatibleAI/imitation/blob/dacb2425e9d19d57e318578a414b50cb98ead647/src/imitation/policies/serialize.py#L210), "ppo" is regarded as one candidate of policy types, while PPO is an `OnPolicyAlgorithm` in SB3.
- Policy loading should be independent with choices of algorithms. e.g. SAC and PPO could load the same policy. The current [method](https://github.com/HumanCompatibleAI/imitation/blob/dacb2425e9d19d57e318578a414b50cb98ead647/src/imitation/policies/serialize.py#L197) use the class method of PPO and then extract its policy attribute. But the class method was defined in [BaseAlgorithm](https://stable-baselines3.readthedocs.io/en/master/_modules/stable_baselines3/common/base_class.html#BaseAlgorithm.load). It seems fine to get rid of additional specifiers of algorithm choice.

The above more or less serves as some notes for me. Happy to give it a go.

Rocamonde:
Fixes #376. Added corresponding tests to ensure that loading policies works as desired.

### DIFF
--- a/src/imitation/scripts/config/train_imitation.py
+++ b/src/imitation/scripts/config/train_imitation.py
@@ -60,8 +60,11 @@ def defaults(
                 "final",
             ),
         )
-    agent_path = None
+    agent_path = None  # Path to load agent from, optional.
 
+    # Specify arguments below if agent_path for pre-trained policy is not provided.
+    policy_cls = None  # Type of a Stable Baselines3 policy architecture.
+    policy_kwargs = None  # Keyword arguments for policy constructor.
 
 @train_imitation_ex.named_config
 def mountain_car():
--- a/src/imitation/scripts/config/train_imitation.py
+++ b/src/imitation/scripts/config/train_imitation.py
@@ -42,6 +42,7 @@ def config():
     )
     agent_path = None  # Path to load agent from, optional.
 
+
 @train_imitation_ex.config
 def defaults(
     common,
--- a/src/imitation/scripts/config/train_imitation.py
+++ b/src/imitation/scripts/config/train_imitation.py
@@ -66,6 +66,7 @@ def defaults(
     policy_cls = None  # Type of a Stable Baselines3 policy architecture.
     policy_kwargs = None  # Keyword arguments for policy constructor.
 
+
 @train_imitation_ex.named_config
 def mountain_car():
     common = dict(env_name="MountainCar-v0")
--- a/src/imitation/scripts/config/train_imitation.py
+++ b/src/imitation/scripts/config/train_imitation.py
@@ -60,6 +60,7 @@ def defaults(
                 "final",
             ),
         )
+    agent_path = None
 
 
 @train_imitation_ex.named_config
--- a/src/imitation/scripts/config/train_imitation.py
+++ b/src/imitation/scripts/config/train_imitation.py
@@ -40,7 +40,7 @@ def config():
         expert_policy_type=None,  # 'ppo', 'random', or 'zero'
         total_timesteps=1e5,
     )
-
+    agent_path = None  # Path to load agent from, optional.
 
 @train_imitation_ex.config
 def defaults(
@@ -60,11 +60,6 @@ def defaults(
                 "final",
             ),
         )
-    agent_path = None  # Path to load agent from, optional.
-
-    # Specify arguments below if agent_path for pre-trained policy is not provided.
-    policy_cls = None  # Type of a Stable Baselines3 policy architecture.
-    policy_kwargs = None  # Keyword arguments for policy constructor.
 
 
 @train_imitation_ex.named_config
--- a/src/imitation/scripts/train_imitation.py
+++ b/src/imitation/scripts/train_imitation.py
@@ -13,6 +13,8 @@ from imitation.data import rollout
 from imitation.policies import serialize
 from imitation.scripts.common import common, demonstrations, train
 from imitation.scripts.config.train_imitation import train_imitation_ex
+from stable_baselines3.common.utils import get_device
+import torch as th
 
 logger = logging.getLogger(__name__)
 
@@ -22,6 +24,7 @@ def make_policy(
     venv: vec_env.VecEnv,
     policy_cls: Type[policies.BasePolicy],
     policy_kwargs: Mapping[str, Any],
+    policy_path: Optional[str],
 ) -> policies.BasePolicy:
     """Makes policy.
 
@@ -29,6 +32,8 @@ def make_policy(
         venv: Vectorized environment we will be imitating demos from.
         policy_cls: Type of a Stable Baselines3 policy architecture.
         policy_kwargs: Keyword arguments for policy constructor.
+        policy_path: Path to serialized policy. If provided, then load the
+            policy from this path. Otherwise, make a new policy.
 
     Returns:
         A Stable Baselines3 policy.
@@ -43,7 +48,15 @@ def make_policy(
                 "lr_schedule": utils.get_schedule_fn(1),
             },
         )
-    policy = policy_cls(**policy_kwargs)
+    if policy_path is not None:
+        
+        device = get_device("auto")
+        saved_variables = th.load(policy_path, map_location=device)
+        policy = policy_cls(**saved_variables["data"], **policy_kwargs)
+        policy.load_state_dict(saved_variables["state_dict"])
+        policy.to(device)
+    else:
+        policy = policy_cls(**policy_kwargs)
     logger.info(f"Policy network summary:\n {policy}")
     return policy
 
@@ -88,6 +101,7 @@ def train_imitation(
     bc_train_kwargs: Mapping[str, Any],
     dagger: Mapping[str, Any],
     use_dagger: bool,
+    agent_path: Optional[str],
 ) -> Mapping[str, Mapping[str, float]]:
     """Runs DAgger (if `use_dagger`) or BC (otherwise) training.
 
@@ -96,6 +110,8 @@ def train_imitation(
         bc_train_kwargs: Keyword arguments passed through to `BC.train` method.
         dagger: Arguments for DAgger training.
         use_dagger: If True, train using DAgger; otherwise, use BC.
+        agent_path: Path to directory containing pre-trained agent for warm start. If None, train from
+            scratch.
 
     Returns:
         Statistics for rollouts from the trained policy and demonstration data.
--- a/src/imitation/scripts/train_imitation.py
+++ b/src/imitation/scripts/train_imitation.py
@@ -13,8 +13,6 @@ from imitation.data import rollout
 from imitation.policies import serialize
 from imitation.scripts.common import common, demonstrations, train
 from imitation.scripts.config.train_imitation import train_imitation_ex
-from stable_baselines3.common.utils import get_device
-import torch as th
 
 logger = logging.getLogger(__name__)
 
@@ -105,15 +103,15 @@ def train_imitation(
         bc_train_kwargs: Keyword arguments passed through to `BC.train` method.
         dagger: Arguments for DAgger training.
         use_dagger: If True, train using DAgger; otherwise, use BC.
-        agent_path: Path to directory containing pre-trained agent for warm start. If None, train from
-            scratch.
+        agent_path: Path to directory containing pre-trained agent for warm start. If
+            None, train from scratch.
 
     Returns:
         Statistics for rollouts from the trained policy and demonstration data.
     """
     custom_logger, log_dir = common.setup_logging()
     venv = common.make_venv()
-    imit_policy = make_policy(venv)
+    imit_policy = make_policy(venv, policy_path=agent_path)
 
     expert_trajs = None
     if not use_dagger or dagger["use_offline_rollouts"]:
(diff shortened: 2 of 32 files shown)


src/imitation/scripts/config/train_imitation.py
```diff
@@ -60,6 +60,7 @@ def defaults(
                 "final",
             ),
         )
+    agent_path = None
```
AdamGleave, line 63:
```suggestion
    agent_path = None  # Path to load agent from, optional.
```

src/imitation/scripts/train_imitation.py
```diff
@@ -7,7 +7,7 @@
 from sacred.observers import FileStorageObserver
 from stable_baselines3.common import policies, utils, vec_env
 
-from imitation.algorithms.bc import BC
+from imitation.algorithms.bc import BC, reconstruct_policy
```
AdamGleave, line 10: Import modules not classes/methods: https://google.github.io/styleguide/pyguide.html#22-imports

This is an existing issue in this code, but one we should probably fix not exacebrate.

tests/test_scripts.py
```diff
@@ -376,6 +436,35 @@ def test_train_adversarial(tmpdir, named_configs, command):
...
 
 
+@pytest.mark.parametrize("command", ("airl", "gail"))
+def test_train_adversarial_warmstart(tmpdir, command):
+    named_configs = ["cartpole"] + ALGO_FAST_CONFIGS["adversarial"]
+    config_updates = {
+        "common": dict(log_root=tmpdir),
+        "demonstrations": dict(rollout_path=CARTPOLE_TEST_ROLLOUT_PATH),
+    }
+    run = train_adversarial.train_adversarial_ex.run(
```
AdamGleave, line 446: I think `train_adversarial` can be warmstarted from the expert in `CARTPOLE_TEST_POLICY_PATH` without needing to train AIRL/GAIL then warmstart from that policy, which might simplify things.

src/imitation/scripts/train_imitation.py
```diff
@@ -22,13 +22,16 @@ def make_policy(
...
     policy_kwargs: Mapping[str, Any],
+    policy_path: Optional[str],
 ) -> policies.BasePolicy:
     """Makes policy.
 
     Args:
         venv: Vectorized environment we will be imitating demos from.
         policy_cls: Type of a Stable Baselines3 policy architecture.
         policy_kwargs: Keyword arguments for policy constructor.
+        policy_path: Path to serialized policy. If provided, then load the
```
AdamGleave, line 33: If this is specified then `policy_cls` and `policy_kwargs` are ignored. Worth either documenting, or adding input validation to check that `policy_{cls,kwargs}` match the deserialized policy.

AdamGleave (commented):
Thanks for the PR! Overall looks pretty good, and appreciate the additional test cases. A few minor suggestions -- please request re-review when addressed.

src/imitation/scripts/train_imitation.py
Rocamonde: Not sure if all the policy kwargs are always stored as attributes in the class that can be checked by looping and doing `getattr`. I think it's not obvious that incompatible kwargs should raise an error rather than replace the ones from the saved policy, so we should probably not support it and raise a ValueError when both are provided

tests/test_scripts.py
Rocamonde: If I use the CARTPOLE_TEST_POLICY_PATH, tests don't pass with a horizon of 10, getting `AssertionError: No updates (need at least 512 timesteps, have only total_timesteps=10)!`. Thoughts?

src/imitation/scripts/train_imitation.py
Rocamonde: Actually there is a problem with this; there is already a `bc` in the namespace, the function defined to call the command for behavioral cloning training from the script. (around line 175). I'll temporarily use `bc_algo` as the module alias to be able to import it, but feel free to suggest a better name.

Rocamonde:
Hi @AdamGleave, included all your suggestions, except for a change in the test case where doing so introduces an issue (see above). Should otherwise be good to merge. Let me know your thoughts!

src/imitation/scripts/config/train_imitation.py
```diff
@@ -60,7 +60,11 @@ def defaults(
                 "final",
             ),
         )
-    agent_path = None
+    agent_path = None  # Path to load agent from, optional.
+
+    # Specify arguments below if agent_path for pre-trained policy is not provided.
+    policy_cls = None  # Type of a Stable Baselines3 policy architecture.
```
AdamGleave, line 66: We now have two `policy_cls` and `policy_kwargs`, one at the top-level and one in `train`, which is confusing:

```
$ python -m imitation.scripts.train_imitation print_config 
INFO - train_imitation - Running command 'print_config'
INFO - train_imitation - Started
Configuration (modified, added, typechanged, doc):
  agent_path = None                  # Path to load agent from, optional.
  policy_cls = None                  # Type of a Stable Baselines3 policy architecture.
  policy_kwargs = None               # Keyword arguments for policy constructor.
  [... omitted]
  train:
    n_episodes_eval = 50             # Num of episodes for final mean ground truth return
    policy_cls = <class 'imitation.policies.base.FeedForward32Policy'>    # Training
    policy_kwargs:
      features_extractor_class = <class 'imitation.policies.base.NormalizeFeaturesExtractor'>
      features_extractor_kwargs:
        normalize_class = <class 'imitation.util.networks.RunningNorm'>
INFO - train_imitation - Completed after 0:00:00
```

If you want to change the defaults for `train.{policy_cls,policy_kwargs}` you should be able to do this via a Sacred config hook: https://sacred.readthedocs.io/en/stable/apidoc.html#sacred.Experiment.config_hook

But do you actually want it to default to None...? This would mean the default (None agent_path, and None policy_cls/policy_kwargs) wouldn't run. You could provide a named_config to set them to `None`.

(Sacred doesn't make this particular use case easy, unfortunately.)

src/imitation/scripts/train_imitation.py
```diff
@@ -20,22 +20,31 @@
...
+            Specify only if policy_path is not specified.
         policy_kwargs: Keyword arguments for policy constructor.
+            Specify only if policy_path is not specified.
         policy_path: Path to serialized policy. If provided, then load the
             policy from this path. Otherwise, make a new policy.
+            Specify only if policy_cls and policy_kwargs are not specified.
 
     Returns:
         A Stable Baselines3 policy.
     """
```
AdamGleave, line 41: Linter is complaining you do not document the `ValueError` raises.

src/imitation/scripts/train_imitation.py
```diff
@@ -20,19 +20,31 @@
 @train_imitation_ex.capture(prefix="train")
 def make_policy(
     venv: vec_env.VecEnv,
-    policy_cls: Type[policies.BasePolicy],
-    policy_kwargs: Mapping[str, Any],
+    policy_cls: Optional[Type[policies.BasePolicy]],
+    policy_kwargs: Optional[Mapping[str, Any]],
+    policy_path: Optional[str],
```
AdamGleave, line 25: Given this is already a capture function, if you just made this `agent_path` not `policy_path` (or vice-versa updated the config to use `policy_path` not `agent_path`) then you could remove the `agent_path` paramter from `train_imitation` entirely.

AdamGleave (changes requested):
The `policy_cls` and `policy_kwargs` parameters are inadvertently duplicated (and I think the duplicates are ignored).

Also some more minor linting stuff.

src/imitation/scripts/train_imitation.py
```diff
@@ -100,9 +109,9 @@ def train_imitation(
 
     Args:
         bc_kwargs: Keyword arguments passed through to `bc.BC` constructor.
-        bc_train_kwargs: Keyword arguments passed through to `BC.train` method.
+        bc_train_kwargs: Keyword arguments passed through to `bc.train` method.
```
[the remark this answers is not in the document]
Rocamonde: Yes, this was due to some replacing that I did to allow both the module and the function `bc()` to coexist, forgot to check the match case box.

src/imitation/scripts/config/train_imitation.py
Rocamonde: Sorry, I did not know that policy_cls and policy_kwargs had defaults that were injected in common/train. IMO I find that the peculiar compositional nature of sacred makes it a bit hard to see everything that is going on and getting injected through decorators. I now understand your point: if you provide a policy path and the other values are automatically injected, it will always raise with my current approach. The reason I changed the default to None is to have function overloads. I still think that we should at least warn the user that the arguments are getting ignored, if enforcing it is not viable.

Rocamonde: Is there a way to only inject the default values that are in common/train if the path is not specified?  (and have them default to None if the path is specified). That seems like the best solution. Otherwise we can just replace the value error with a warning for now.

src/imitation/scripts/train_imitation.py
Rocamonde: This actually does not work out-of-the-box as one might think, because `make_policy()` has the `"train"` prefix, which means that top-level config is not available in that scope as it is with `train_imitation()`. I think I could fix this by moving the defaults to `common.train`, but this would break the pattern of passing `agent_path` as a top-level config to other algorithms. What do you think?

Rocamonde: Rolling this back makes all tests pass again. Let me know if you can think of a workaround, or we can keep it this way for now.

tests/test_scripts.py
AdamGleave: Huh, where's the horizon of 10 coming from? CartPole should have 500 horizon by default.

src/imitation/scripts/train_imitation.py
AdamGleave: Oh I see, yeah missed that it's capturing the `train` prefix. You *can* capture at the top-level and have one of the arguments you capture be `train`, which would then be a dict containing `policy_path`, `policy_cls`, etc. But I'm not sure that's better than the status quo: your call.

src/imitation/scripts/config/train_imitation.py
AdamGleave: I don't think you ever updated this, was that intentional?

AdamGleave: Also I think this belongs in `config` not `defaults`

src/imitation/scripts/config/train_imitation.py
AdamGleave: > Sorry, I did not know that policy_cls and policy_kwargs had defaults that were injected in common/train. IMO I find that the peculiar compositional nature of sacred makes it a bit hard to see everything that is going on and getting injected through decorators.

Yes, the modularity has its costs. The `print_config` command is a useful way to figure out everything that's present.

> I still think that we should at least warn the user that the arguments are getting ignored, if enforcing it is not viable.

Yes agreed there should be some kind of warning and ideally an error.

> The problem is that since defaults are injected outside the function call, it is not possible to tell from within the function whether the value is a default (so ignoring is fine) or if the user is providing it (so warn/raise). It would be nice if there were a way to only inject defaults if the user does not provide a policy path. Otherwise, we can throw a warning for now whenever a path is specified, and find a better solution later.

It's possible to do this but tricky.

If we wanted to move the `agent_path` into `common/train` things would be easy. But except for BC/DAgger, we don't *want* to preload a policy -- we want to preload from an RL algorithm. So, that change wouldn't make sense.

One way to do it would be change `common/train.py` to have `policy_cls = None` be the default, then add a config hook in `common/train.py` that checks for if `agent_path` is present and non-None. This would be breaking the abstraction a bit (`common/train.py` now cares about things outside the ingredient) but would work.

You could also just change the default to be `None` and then have defaults added in the scripts as needed -- which right now is just in `common/rl.py` and `train_imitation.py`. The downside is it's harder to figure out what policy class is being used from just inspecting the config

AdamGleave (approved):
LGTM.

The `policy_cls` and `policy_kwargs` defaults could be handled better -- but there's no great way of doing this, so up to you if you want to try and address this further in this PR or defer it to later.

PR #481, opened 2022-07-20
Regularization API for preference comparisons

### LINKED ISSUE #461: [Preference Comparison] L2 regularization with dynamic regularization coefficient
The [DRLHP paper](https://arxiv.org/pdf/1706.03741.pdf) states in section 2.2.3 that:
> A fraction of 1/e of the data is held out to be used as a validation set for each predictor. We use `2 regularization and adjust the regularization coefficient to keep the validation loss between 1.1 and 1.5 times the training loss. In some domains we also apply dropout for regularization.

A similar approach was used in follow-up work by [Ibarz et al](https://arxiv.org/pdf/1811.06521.pdf), described in a bit more detail in section A.3:
> A fraction of 1/e of the data is held out to be used as a validation set. We use L2-
regularization of network weights with the adaptive scheme described in Christiano et al. (2017): the L2-regularization weight increases if the average validation loss is more than 50% higher than the average training loss, and decreases if it is less than 10% higher (initial weight 0.0001, multiplicative rate of change 0.001 per learning step).

Currently preference comparisons does not support any form of adaptive L2 regularization, although it does support a constant weight decay in [CrossEntropyRewardTrainer](https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/algorithms/preference_comparisons.py#L695). This has a similar effect to constant L2 regularization. 

We should add support since it shows a modest performance improvement in ablations (e.g. section 3.3 of DRLHP paper showed significant improvements in swimmer, qbert and seaquest though little difference in other environments). Moreover as it was part of the original paper it is good to have support to be able to replicate it and perform our own ablations. Furthermore, I suspect regularization may also improve the robustness of learned reward functions aiding in transfer, although they don't evaluate that.

Concretely, I think the tasks would be:
  - Support L2 coefficient (or weight decay coefficient) being a variable rather than a constant.
  - Add train/validation split of dataset.
  - Compute validation loss.
  - Add logic to dynamically adjust the L2 coefficient based on validation loss.

Rocamonde:
Fixes #461.

 - [x] Support seeding via generator
 - [x] Implement base class update params
 - [x] Add regularization param change to logger
 - [x] Add loss_regularize
 - [x] Add weight_regularize
 - [x] Add support for passing general regularization classes in preference_comparisons
 - [x] Add tests


AdamGleave:
Heads up we'll probably have some conflicts once https://github.com/HumanCompatibleAI/imitation/pull/460 gets merged, although shouldn't be too hard to resolve.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -758,16 +765,28 @@ def _loss(
             rews2 = self._rewards(trans2)
             probs[i] = self._probability(rews1, rews2)
         # TODO(ejnnr): Here and below, > 0.5 is problematic
-        # because getting exactly 0.5 is actually somewhat
-        # common in some environments (as long as sample=False or temperature=0).
-        # In a sense that "only" creates class imbalance
-        # but it's still misleading.
+        #  because getting exactly 0.5 is actually somewhat
```
AdamGleave, line 768: Why the additional whitespace?

Rocamonde: TODO detectors in e.g. vscode and PyCharm only detect that the lines below are part of the TODO above if you indent it. Otherwise it understands it as a separate comment

Rocamonde:
(pulled all commits from #460 so that we can start integrating that into this issue)

Rocamonde:
Added initial skeleton for the regularization module. Let me know your thoughts. Seems like weight decay has to act before `loss.backward()`, but L2 regularization / arbitrary loss penalties have to go before. So I moved the call to `loss.backward()` inside the regularizer, and replaced that call with `.regularize(loss)`. Then each regularizer implementation computes the final loss and final gradients accordingly.

Rocamonde:
Sounds good, thanks for doing that. I suspected it was possible but wasn't sure how to.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -889,15 +892,33 @@ def _make_data_loader(self, dataset: PreferenceDataset) -> data_th.DataLoader:
 
     def _train(self, dataset: PreferenceDataset, epoch_multiplier: float = 1.0) -> None:
         """Trains for `epoch_multiplier * self.epochs` epochs over `dataset`."""
-        dataloader = self._make_data_loader(dataset)
+        if self.regularizer:
+            val_length = int(len(dataset) * self.val_split)
+            train_length = len(dataset) - val_length
+            train_dataset, val_dataset = data_th.random_split(dataset, lengths=[train_length, val_length])
```
AdamGleave, line 898: Support seeding (via `generator`)?

src/imitation/reg/__init__.py
```diff
@@ -0,0 +1,86 @@
+"""Implements a variety of reguarlization techniques for NN weights."""
+
+import abc
+from typing import Dict
+
+import torch as th
+from torch import optim
+
+
+class Regularizer:
```
AdamGleave, line 10: this should inherit from `abc.ABC` given it has abstract methods

AdamGleave (commented):
Thanks for the draft implementation!

Right now `Regularize` seems to be doing two different things:
  - Adaptive regularization via `update_params`, implemented in `WeightDecayRegularizer`.
  - Support for regularizing with a variable scale factor, implemented in `regularize`.

I'm inclined to split these out. We might want to try different adaptive regularization schemes (the one in Ibarz; something more continuous) with weight decay, or potentially even with L2 regularization.

One option would just be to replace the `update_params` method in `Regularizer` ABC with a `set_scale` method, that gets called by some adaptive code. Or `Regularizer` could call a callback.

The adaptive logic could then be moved into something more akin to a `Scheduler` ABC (and we could also support annealing etc in the future).

WDYT?

Rocamonde:
Agreed, that’s what I had in mind from the beginning, but since I wasn’t sure how to implement the abstraction exactly I hard-coded the update params function. The idea was to not make it an abstract method and define it in the ABC, where the parameters are updated by calling a callable passed as an init. But wasn’t sure how general to make the type signature of this callable so I wrote a sketch of a specific method first to see how that looked.

Rocamonde:
> Right now `Regularize` seems to be doing two different things:
> 
> * Adaptive regularization via `update_params`, implemented in `WeightDecayRegularizer`.
> * Support for regularizing with a variable scale factor, implemented in `regularize`.
> 
> I'm inclined to split these out. 

Agreed, that’s what I had in mind from the beginning, but since I wasn’t sure how to implement the abstraction exactly I hard-coded the `update_params` function. The idea was to not make it an abstract method and define it in the ABC, where the parameters are updated by calling a callable passed to `__init__()`. But wasn’t sure how general to make the type signature of this callable so I wrote a sketch of a specific method first to see how that looked.

src/imitation/reg/__init__.py
Rocamonde: Thanks for spotting this! Didn't remember seeing this in the linter

src/imitation/algorithms/preference_comparisons.py
Rocamonde: Thanks, yes. This was in the original version, but got over-written when I pulled the changes from `reward_ensemble`.

Rocamonde:
I guess the uncertainty came from the following points:

1. Should all regularization parameter updates only take in validation and training loss? Is there anything else that we should include? I think that the idea of the methods we are thinking about is dynamically updating the regularization params based on validation loss, but maybe more general regularization techniques follow completely different approaches. I am inclined to keep it limited to this for now, but worth giving it a think.
2. Will our regularization techniques only have one parameter or potentially multiple parameters that get updated? If there are multiple parameters, we would have to iterate. But then we might want to keep different tuning hyperparameters separate for each regularization parameter (e.g. the update rate or the initial value). I don't want to introduce unnecessary abstraction to support this if we are only ever likely to support regularization techniques that have only one parameter. Also, if there are multiple parameters, I am inclined to constrain them to be separately "updateable" and iterate over the update loop within the ABC, but maybe there is a use case for updating them simultaneously?
3. Should the regularization methods be composable? i.e. easy to apply multiple simultaneously, or to add e.g. L2 and L1 loss by just adding/stacking them in some way, rather than having to create a separate method. I am again inclined to not allow this for now, as I have the impression that this would introduce a lot of abstraction that might go unused.

WDYT?

Rocamonde:
Two other things that come to mind:

1. Do we want to update the parameters after each batch or after the whole epoch? Currently I implemented the latter, and I take a cumulative validation loss that is what gets passed every time.
2. Should we include the hyperparams and final params in the logger / serialization dump? Seems important for reproducibility.

AdamGleave:
> I guess the uncertainty came from the following points:

1) I think it's fine to have adaptive regularization take in only validation and training loss. We can add new use cases later down the line as needed. It's going to be very hard to support all possible use cases up front and I don't think we should try.

> 2. Will our regularization techniques only have one parameter or potentially multiple parameters that get updated? If there are multiple parameters, we would have to iterate. But then we might want to keep different tuning hyperparameters separate for each regularization parameter (e.g. the update rate or the initial value). I don't want to introduce unnecessary abstraction to support this if we are only ever likely to support regularization techniques that have only one parameter. Also, if there are multiple parameters, I am inclined to constrain them to be separately "updateable" and iterate over the update loop within the ABC, but maybe there is a use case for updating them simultaneously?

Think fine to keep it to only support a single parameter for now. We can add support for multiple parameters without breaking backward compatibility, so no need to engineer it in now.

> 3. Should the regularization methods be composable? i.e. easy to apply multiple simultaneously, or to add e.g. L2 and L1 loss by just adding/stacking them in some way, rather than having to create a separate method. I am again inclined to not allow this for now, as I have the impression that this would introduce a lot of abstraction that might go unused.

Hm good question. I struggle to see a case where we'd really want multiple regularization techniques. But the way it's currently architected might make it hard to add support for this later. I think we could add support for this by methods defining some `loss_regularize` (that takes in a loss and does some transformation to it, e.g. adding a penalty term) and `weight_regularize` (that updates the parameter groups in `self.optimizer` directly ala for weight decay), then having some class that lets us chain these together. But this seems like it'd add a fair bit of implementation complexity, so I'm inclined to agree with you that it's not worth supporting right now.

> Do we want to update the parameters after each batch or after the whole epoch? Currently I implemented the latter, and I take a cumulative validation loss that is what gets passed every time.

Updating it once per epoch seems reasonable for now. However, if we had very big datasets we might want to update it more than once per epoch, so ideally this would be a design decision that wouldn't be too hard to change.

> Should we include the hyperparams and final params in the logger / serialization dump? Seems important for reproducibility.

Definitely worth logging these regularization parameters over time: it'll be a useful metric to track.

I don't right now see the case for including them in the serialized policy / reward network. For restarting training from a checkpoint, it'd be useful to have the final params (history not required). But we largely don't support checkpointing now (we lose e.g. the momentum parameters from Adam) -- it'd be a feature we'd like to have in the future, but no need to complexify this PR with that.

Rocamonde:
Apart from adding logging, I think this is now implementing all the required features for an initial version. pytype does not notice, but I get an error from Pylance saying that:
```
Argument of type "Subset[Unknown]" (train_dataset) cannot be assigned to parameter "dataset" of type "PreferenceDataset" in function "_make_data_loader"
  "Subset[Unknown]" is incompatible with "PreferenceDataset"
```

Which is technically correct, the types Subset and PreferenceDataset are both subtypes of DataSet, but not subtypes of each other. In fact, `_make_data_loader` could use a more general type afaiu. Should I change this?

Rocamonde:
I have no idea what is going on here, I am getting errors on CircleCI that seem so weird and don't show up on my local machine. Before I was getting an indentation error, now I'm getting a sphinx error, neither of which were caught by `ci/code_checks.sh`.

AdamGleave:
> I have no idea what is going on here, I am getting errors on CircleCI that seem so weird and don't show up on my local machine. Before I was getting an indentation error, now I'm getting a sphinx error, neither of which were caught by `ci/code_checks.sh`.

Probably a version difference. Does `pip freeze` on your machine match up with "print installed package" output on CircleCI? Could also try "rerunning job with SSH" on CircleCI, SSHing into the machine, and playing around with it?

We do need CI to pass before merge, but I'll start reviewing without.

src/imitation/regularization/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+
+        Args:
+            lambda_: The current value of the lambda.
+            train_loss: The loss on the training set.
+            val_loss: The loss on the validation set.
+
+        Returns:
+            The new value of the lambda.
+        """
+        # TODO(juan) does this division below work? they are tensors.
```
AdamGleave, line 235: Adding an assert that they're scalars (or at least are scalar after squeezing) seems good.

src/imitation/regularization/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+class ConstantParamScaler:
+    """A dummy param scaler implementation to use as default."""
+
+    def __call__(
+        self,
+        lambda_: float,
+        train_loss: th.Tensor,
+        val_loss: th.Tensor,
+    ) -> float:
+        return lambda_
```
AdamGleave, line 254: Nit: consider `del train_loss, val_loss` to document that train_loss, val_loss are intentionally unused

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -905,8 +927,30 @@ def __init__(
         self.optim = th.optim.AdamW(
             self._model.parameters(),
             lr=lr,
-            weight_decay=weight_decay,
         )
+        self.seed = seed
+        self.val_split = val_split
+
+        # TODO(juan) this could accept an arbitrary regularizer function
```
AdamGleave, line 934: I think we probably want to do that? It seems weird that we implement `L2Regularizer` etc but it's not used anywhere in our current codebase.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -905,8 +927,30 @@ def __init__(
...
+        #  in the future if we wanted, with some changes in the __init__ arguments.
+        self.regularizer: Optional[regularization.Regularizer] = (
+            regularization.WeightDecayRegularizer(
+                initial_lambda=weight_decay,
+                optimizer=self.optim,
+                update_params_fn=weight_decay_updater
+                or regularization.ConstantParamScaler(),
+                logger=self.logger,
+            )
+            if weight_decay > 0
```
AdamGleave, line 944: Very long if-else statement makes it hard to understand flow. Could just set regularizer to None by default and then override it `if weight_decay > 0`.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -905,8 +927,30 @@ def __init__(
...
+            else None
+        )
+        if self.val_split == 0 and self.regularizer is not None:
```
AdamGleave, line 947: You could strengthen this to `(self.val_split == 0) == (self.regularizer is not None)` which will catch both this error and also the case where `val_split != 0` but regularizer is None. (Though maybe you want a different error message for that.)

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -919,16 +963,43 @@ def _make_data_loader(self, dataset: PreferenceDataset) -> data_th.DataLoader:
...
 
+            if self.regularizer:
+                # TODO(juan) my type checker complains because it cannot infer the type
+                #  of val_dataloader by checking the logic. Is there a better way
+                #  to do this, or should we leave the cast?
+                val_dataloader = cast(data_th.DataLoader, val_dataloader)
+                for fragment_pairs, preferences in val_dataloader:
+                    loss = self._training_inner_loop(fragment_pairs, preferences)
+                    val_loss += loss.item()
+                self.regularizer.update_params(train_loss, val_loss)
```
AdamGleave, line 1001: To check semantics: `train_loss` and `val_loss` are summed across entire dataset? 

Guess that's fine, scale doesn't really matter given you're just doing relative comparison between train and val loss.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -919,16 +963,43 @@ def _make_data_loader(self, dataset: PreferenceDataset) -> data_th.DataLoader:
...
+                train_loss += loss.item()
+                if self.regularizer:
+                    self.regularizer.regularize(loss)
                 self.optim.step()
 
+            if self.regularizer:
+                # TODO(juan) my type checker complains because it cannot infer the type
+                #  of val_dataloader by checking the logic. Is there a better way
+                #  to do this, or should we leave the cast?
+                val_dataloader = cast(data_th.DataLoader, val_dataloader)
```
AdamGleave, line 997: Can it follow it if you `assert val_dataloader is not None` (should work if it's inferred type as `Optional[data_th.DataLoader]`)?

That's a bit less aggressive than a cast in that if you earlier on do something like `val_dataloader = "foo"` it'll still fail type checking.

AdamGleave (changes requested):
Overall design for Regularizer looks good. My only real API critique is I think there should be a way to get the final loss (probably just by returning it) for use in e.g. logging. The other major points are:
  1. L1/L2Regularizer aren't actually used anywhere -- add support for running them?
  2. Add some test cases!

Most other suggested changes are minor.

src/imitation/regularization/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+        loss.backward()
+        for group in self.optimizer.param_groups:
+            for param in group["params"]:
+                param.data = th.add(param.data, self._regularize_weight(param, group))
+
+
+class L2Regularizer(LossRegularizer):
+    """Applies L2 regularization to a loss function."""
+
+    def _regularize_loss(self, loss: th.Tensor) -> Union[float, th.Tensor]:
```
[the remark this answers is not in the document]
Rocamonde: Function signatures are contravariant so this is actually okay, but if you want I can fix it above as well.

src/imitation/regularization/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+        """Returns the loss penalty.
+
+        Calculates the squared L2 norm of the weights in the optimizer,
+        and returns a scaled version of it as the penalty.
+
+        Args:
+            loss: The loss to regularize.
+
+        Returns:
+            The scaled L2 norm of the weights.
```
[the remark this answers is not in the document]
Rocamonde: Lol that was copilot, sorry.

src/imitation/regularization/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        penalty = 0
+        for group in self.optimizer.param_groups:
+            for param in group["params"]:
+                penalty += th.sum(param.data.pow(2))
+        return self.lambda_ * penalty
+
+
+class L1Regularizer(LossRegularizer):
```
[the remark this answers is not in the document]
Rocamonde: .pow(ord) makes sense to me, that's literally the inverse of how the Lp norm is defined.

src/imitation/regularization/__init__.py
Rocamonde: Sounds good.

src/imitation/regularization/__init__.py
Rocamonde: wdym after squeezing?

Rocamonde: Also, how do you actually check the rank of a tensor. Google suggests dimension but dimension is mathematically not the same as the rank, even though this might be the folk term for it in the ML community.

src/imitation/regularization/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+        """Initialize the regularizer.
+
+        Args:
+            optimizer: The optimizer to which the regularizer is attached.
+            initial_lambda: The initial value of the regularization parameter.
+            update_params_fn: A callable object that takes in the current lambda and
+                the train and val loss, and returns the new lambda.
+            logger: The logger to which the regularizer will log its parameters.
+        """
+        self.optimizer: optim.Optimizer = optimizer
```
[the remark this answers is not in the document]
Rocamonde: Doesn't matter for type checkers AFAIK but happy to change it to be more compliant.

src/imitation/regularization/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+
+    @abc.abstractmethod
+    def regularize(self, loss: th.Tensor) -> None:
+        """Abstract method for performing the regularization step.
+
+        Args:
+            loss: The loss to regularize.
+        """
+
+    def update_params(self, train_loss, val_loss) -> None:
```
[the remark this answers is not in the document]
Rocamonde: Should be tensors.

src/imitation/regularization/__init__.py
```diff
@@ -0,0 +1,254 @@
...
+        not the regularized loss itself.
+
+        Args:
+            loss: The loss function to which the regularization term is added.
+        """
+        ...
+
+    def regularize(self, loss: th.Tensor) -> None:
+        """Add the regularization term to the loss and compute gradients."""
+        regularized_loss = th.add(loss, self._regularize_loss(loss))
```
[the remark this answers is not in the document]
Rocamonde: We can return it, then log it too. I'll do this, let me know if you have a better idea

src/imitation/algorithms/preference_comparisons.py
Rocamonde: I was not a big fan of using assert in that pattern. However, I had not really thought about cast vs assert much. I'm now of the opinion that asserting is 'less evil' than asserting whenever assert can be used, i.e. when this is literals or class instances. It doesn't work with type vars / generics / protocols, where we should use cast.

Rocamonde: I'm surprised type checkers cannot infer that this is not none, given what other impressive things they can do. This is a general pattern that they are consistently bad at.

Rocamonde: In fact Pylance continues to complain after I add the assert, but that's just because Pylance is not very good.

src/imitation/algorithms/preference_comparisons.py
Rocamonde: Yes, that's what I'm doing, I was not sure which approach to follow. If you have a preference for the other, let me know. I might have to change this anyways given the other issue/PR on batch size.

src/imitation/regularization/__init__.py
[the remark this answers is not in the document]
Rocamonde: Actually this was fine in this particular line, but the fix was required anyways as it was leading to errors in other places. It's fixed now.

src/imitation/algorithms/preference_comparisons.py
Rocamonde: Smart use of XNOR, but maybe not so obvious to read for some people, and I agree that a different error message is in order

Rocamonde: Added a separate case for this

Rocamonde:
I wonder what actually happens to the loss gradient when adding an L1 norm, since it's not differentiable. Does pytorch compute subgradients?

src/imitation/regularization/__init__.py
```diff
@@ -0,0 +1,254 @@
+"""Implements a variety of reguarlization techniques for NN weights."""
```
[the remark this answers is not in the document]
Rocamonde: Why is the folder called util and not utils? (perhaps a similar reason Americans say math and not maths? hehe)

src/imitation/algorithms/preference_comparisons.py
Rocamonde: I added that. It seems excessive to add a protocol and a wrapper (or split initialization of the class into two through separate methods) to first initialize the class with its appropriate regularization algorithm parameters (e.g. "p" in the Lp norm) and then another to pass python objects e.g. optimizer and logger instance, which are only accessible within the RewardTrainer. So we just pass a type as an argument, which isn't great but fairly usable and simpler IMO.

Rocamonde:
Added fixes for all the comments, except for moving the file to utils, still thinking about whether that makes sense. I agree that a separate package is overkill at the moment, but maybe it makes sense as a root module on `imitation.regularizers` instead or similar.

src/imitation/regularization/__init__.py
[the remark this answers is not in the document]
AdamGleave: I thought parameters were contravariant and return types covariant?

src/imitation/regularization/__init__.py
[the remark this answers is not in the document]
AdamGleave: If you want to rename it to `utils` I won't object, but best a separate PR ;)

src/imitation/regularization/__init__.py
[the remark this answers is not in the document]
AdamGleave: returning it SGTM

src/imitation/regularization/__init__.py
AdamGleave: By squeezing I mean `th.squeeze` -- removing any length-1 dimensions. So we might want to treat a tensor of shape `(1,)` as a scalar. (Maybe -- depends on the context.)

`dim` is the number of indices needed to index the tensor, which I think lines up with the mathematical notion of rank (as opposed to the dimension of the space it's embedded in), but I'm rusty on tensors in maths.

AdamGleave:
Thanks for the updates! 

Moving `regularization/__init__.py` to some sub-module in `regularization` seems  fine to me as an alternative to `util` (or `utils`...). Main question is whether we expect to add multiple things to the `regularization` package? Introducing a new package seems overkill for a single module, but if we expect to add things to the `regularization` package in the future (or even split up the current module into multiple ones) then it'd be good to have.

AdamGleave:
> I wonder what actually happens to the loss gradient when adding an L1 norm penalty, since it's not differentiable. Does pytorch compute subgradients? @AdamGleave

Yeah, it uses subgradients at non-differentiable points. For added chaos, different implementations of the same mathematical function can return different subgradients -- e.g. https://github.com/pytorch/pytorch/issues/7172

Rocamonde:
I have just realized that the regularization parameter updates in preference comparisons on the reward trainer are called within the outer training loop of the preference comparisons class. Therefore, we are only dumping the final and the mean values of the regularization parameter, instead of the value at each step. This is only specific to the preference comparisons implementation and is not an issue with general usage of regularization, but it would be good to find a way to dump the data more granularly for the reward trainer. AFAIK `.dump()` takes an integer step that corresponds to the PreferenceComparsions iteration, so overloading this with the step of the trainer would probably result in data overridden.

Other minor improvements to be considered: 
- [ ] Add `dump()` after `record()` and check that the historic data is logged in the CSV correctly.

src/imitation/regularization/updaters.py
```diff
@@ -0,0 +1,139 @@
...
+            The new value of the lambda.
+
+        Raises:
+            ValueError: If the loss on the validation set is not a scalar.
+            ValueError: if lambda_ is zero (will result in no scaling).
+            ValueError: if lambda_ is not a float.
+        """
+        # assert that the tensors val_loss and train_loss are both scalars
+        try:
+            assert isinstance(val_loss, float) or (
```
AdamGleave, line 96: Why the indirection: assert then trap it and raise ValueError? Note `assert` is disabled when Python runs in optimized mode `-O` so it's not just a distinction between `AssertionError` and `ValueError` (although optimized mode is so rarely used the importance of this is minor).

One alternative would be to define a helper function that does this input validation, and call it for both `train_loss` and `val_loss`

src/imitation/regularization/updaters.py
```diff
@@ -0,0 +1,139 @@
...
+                isinstance(val_loss, th.Tensor) and val_loss.dim() == 0
+            )
+            assert isinstance(train_loss, float) or (
+                isinstance(train_loss, th.Tensor) and train_loss.dim() == 0
+            )
+        except AssertionError as exc:
+            raise ValueError("val_loss and train_loss must be scalars") from exc
+        if np.finfo(float).eps > lambda_:
```
AdamGleave, line 104: Nit: this will also trigger if `lambda_` is negative. We do want to catch that of course but this error message would be misleading in that case.

src/imitation/regularization/updaters.py
```diff
@@ -0,0 +1,139 @@
...
+            raise ValueError(
+                "lambda_ must not be zero. Make sure that you're not "
+                "scaling the value of lambda down too quickly or passing an "
+                "initial value of zero to the lambda parameter.",
+            )
+        if not isinstance(lambda_, float):
+            raise ValueError("lambda_ must be a float")
+
+        eps = np.finfo(float).eps
+        if train_loss < eps and val_loss < eps:
```
AdamGleave, line 114: This assumes `train_loss` and `val_loss`. Which it should be, but probably worth explicitly checking for, given that for most ML algorithms you can shift loss up/down by a constant value and not change things (whereas this algorithm treats zero as special because it takes the ratio of losses).

AdamGleave (commented):
I like the new regularization package -- splitting up the core regularization logic vs updaters makes a lot of sense.

Right now `BasicRewardTrainer.__init__` is doing a lot of input validation for regularization and other assorted parameter mushing (e.g. default values). By default I'd lean towards moving this logic to the `regularizers` package where possible, so that when we add regularization in future algorithms there is no need to duplicate code, and we keep a consistent codebase. In particular, it seems like:
  - We can roll line 1209 setting a default `lambda_updater` into `regularize.Regularizer.__init__` (or even remove it entirely if we make `None` a legitimate updater, that's semantically equivalent to `ConstantParamScaler`).
  - We can move the check that `reg_lambda_update` is specified or `reg_lambda` is non-zero on line 1189 into `Regularizer.__init__` as well.

The checks involving the train-test split are harder, but could perhaps be moved into a helper method.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1114,20 +1125,95 @@ def __init__(
                 on the fly by specifying an `epoch_multiplier` in `self.train()`
                 if longer training is desired in specific cases).
             lr: the learning rate
-            weight_decay: the weight decay factor for the reward model's weights
-                to use with ``th.optim.AdamW``. This is similar to but not equivalent
-                to L2 regularization, see https://arxiv.org/abs/1711.05101
+            seed: the random seed to use for splitting the dataset into training
+                and validation.
             custom_logger: Where to log to; if None (default), creates a new logger.
+            reg_class: the regularization class to use. If not specified, no
```
AdamGleave, line 1131: Why have `preference_comparisons` construct the regularizer, rather than have the caller construct it and pass it in? Both conventions have their pros/cons. We've leaned towards the latter so far in this code (e.g. `model` and `loss` must both be constructed in caller), and by defauld I'd lean towards keeping it consistent unless there's a clear benefit to deviating.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1138,30 +1224,101 @@ def _make_data_loader(self, dataset: PreferenceDataset) -> data_th.DataLoader:
...
+                train_loss = 0.0
+                for fragment_pairs, preferences in dataloader:
+                    self.optim.zero_grad()
+                    loss = self._training_inner_loop(
+                        fragment_pairs,
+                        preferences,
+                        mode=f"{prefix}/train",
+                    )
+                    train_loss += loss.item()
+                    if self.regularizer:
```
AdamGleave, line 1276: We're not doing `loss.backward()` here. So if `self.regularizer` is `None`, we never make a backward call?

Unless I'm missing something, this seems like a major bug. We should probably add a test to catch this...

This also makes me more unhappy about `regularize` being responsible for calling `backward()` though I don't see a good alternative admittedly.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1138,30 +1224,101 @@ def _make_data_loader(self, dataset: PreferenceDataset) -> data_th.DataLoader:
...
         loss = output.loss
-        self.logger.record("loss", loss.item())
+        self.logger.record(self._get_logger_key(mode, "loss"), loss.item())
         for name, value in output.metrics.items():
-            self.logger.record(name, value.item())
+            self.logger.record(self._get_logger_key(mode, name), value.item())
         return loss
 
+    def _get_logger_key(self, mode: Optional[str], key: str) -> str:
+        if mode is None:
```
AdamGleave, line 1318: Can we remove this once https://github.com/HumanCompatibleAI/imitation/blob/09670d399b5f410524a0e3d5efb3e1b8f043ebd1/src/imitation/util/logger.py#L126 gets merged? If so, might be worth adding a TODO or opening an issue to this effect.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1138,30 +1224,101 @@ def _make_data_loader(self, dataset: PreferenceDataset) -> data_th.DataLoader:
...
+            if key.startswith(f"mean/reward/epoch-{epoch_num}"):
+                val = self.logger.name_to_value[key]
+                new_key = key.replace(f"mean/reward/epoch-{epoch_num}", "reward/final")
+                self.logger.record(new_key, val)
 
     def _training_inner_loop(
         self,
         fragment_pairs: Sequence[TrajectoryPair],
         preferences: np.ndarray,
+        mode: Optional[str] = None,
```
AdamGleave, line 1308: Nit: I'd prefer "prefix" over "mode" as it's not really a pure mode (train vs eval) since also includes the epoch number (and training is in the same "mode" across epochs, exactly the same code just different weights).

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1189,35 +1351,57 @@ def __init__(
                 on the fly by specifying an `epoch_multiplier` in `self.train()`
                 if longer training is desired in specific cases).
             lr: the learning rate
-            weight_decay: the weight decay factor for the reward model's weights
-                to use with ``th.optim.AdamW``. This is similar to but not equivalent
-                to L2 regularization, see https://arxiv.org/abs/1711.05101
-            seed: seed for the internal RNG used in bagging
+            seed: the random seed to use for splitting the dataset into training
```
AdamGleave, line 1354: `seed` and `rng_seed` are rather confusing names. Aren't all seeds used to seed random number generators (RNGs)? `seed` also suggests some more primacy or importance ove the longer name `rng_seed`, but they seem similarly important to me?

I'd give them clearer names: `bagging_seed` and `split_seed`, for example.

Is there a reason to even separate them? Why not just use the same seed for both?

tests/test_regularization.py
```diff
@@ -0,0 +1,436 @@
...
+        assert scaler(lambda_, train_loss, val_loss) == lambda_
+        val_loss = th.tensor(1.0)
+        assert scaler(lambda_, train_loss, val_loss) == lambda_ * (1 + scaling_factor)
+        val_loss = 1.0
+        assert scaler(lambda_, train_loss, val_loss) == lambda_ * (1 + scaling_factor)
+
+
+def test_interval_param_scaler_raises(interval_param_scaler):
+    scaler = interval_param_scaler
+    with pytest.raises(ValueError, match="val_loss and train_loss must be scalars"):
```
AdamGleave, line 97: Nit: because both train and val loss are illegal arguments, this error will trigger (and test will pass) if input validation checks *either* argument. Might want to split test cases to use legal argument for one not the other?

Not as relevant but it also seems odd to have `lambda_ = 0.0` the whole time as that's also illegal.

tests/test_regularization.py
```diff
@@ -0,0 +1,436 @@
...
+        scaler(th.tensor(1.0), 1.0, 1.0)  # type: ignore
+    with pytest.raises(ValueError, match="lambda_ must not be zero.*"):
+        scaler(0.0, 1.0, 1.0)
+
+
+def test_interval_param_scaler_init_raises():
+    # this validates the value of scaling_factor.
+    with pytest.raises(
+        ValueError,
+        match=r"scaling_factor must be in \(0, 1\) within machine precision.",
```
AdamGleave, line 111: Could assign this to a variable and reuse it rather than having to copy-paste it.

Heck you could even assign the whole `pytest.raises(...)` to a context manager though that's maybe sufficiently counterintuitive to be not worth the lines saved.

tests/test_regularization.py
```diff
@@ -0,0 +1,436 @@
...
+    ):
+        # cannot be exactly one, as when lambda is scaled down
+        # this brings it to zero.
+        updaters.IntervalParamScaler(1.0, (0.1, 0.9))
+
+    # an interval obviously needs two elements only.
+    with pytest.raises(
```
AdamGleave, line 139: You might enjoy the library `hypothesis` for testing things like this. Tests as written are fine, just might save you time in future for checking invariants like this.

tests/test_regularization.py
```diff
@@ -0,0 +1,436 @@
...
+    assert regularizer.lambda_ == initial_lambda
+    assert (
+        hierarchical_logger.default_logger.name_to_value["regularization_lambda"]
+        == initial_lambda
+    )
+    regularizer.update_params(train_loss, val_loss)
+    expected_lambda_value = interval_param_scaler(initial_lambda, train_loss, val_loss)
+    assert regularizer.lambda_ == expected_lambda_value
+    assert expected_lambda_value != initial_lambda
+    # TODO(juan) where's the historic data for this? should check the
```
AdamGleave, line 223: As I think you know at this point, if you don't `dump()` the values just get overwritten. So I think there's nothing to check here...

tests/test_regularization.py
```diff
@@ -0,0 +1,436 @@
...
+@pytest.mark.parametrize(
+    "train_loss",
+    [
+        th.tensor(10.0),
+        th.tensor(1.0),
+        th.tensor(0.1),
+    ],
+)
+@pytest.mark.parametrize("p", [1, 2, 3])
+def test_lp_regularizer(
```
AdamGleave, line 370: I appreciate the thoroughness of this test! And a good use of fixtures.

src/imitation/regularization/updaters.py
Rocamonde: Yeah you're right, I was not convinced by writing it this way, but somehow ended up leaving it. I guess I found it easier to write a bunch of checks and write only one error message, rather than split it up.

src/imitation/regularization/updaters.py
Rocamonde: I think you meant it assumes they are "positive" (you probably missed the word). yes, sure, will do.

tests/test_regularization.py
Rocamonde: Fair. Haha hope we don't get the AI optimizing for passing the checks only for the tensor `th.Tensor([3.0, 4.0])` ;)

Rocamonde: Actually, there's a cool library that I saw recently that tries to touch on that by having test values be automatically generated from assumptions (e.g. integer): https://hypothesis.readthedocs.io/en/latest/quickstart.html. And the test values are different every time.

Rocamonde: Oh lol, it seems like we already use it.

tests/test_regularization.py
Rocamonde: I could write a separate test and parametrize it, but I think commenting why each set of values raises is useful enough to leave it this way.

tests/test_regularization.py
Rocamonde: Haha okay. I just mentioned it on a comment above, seems like you already know about it. I don't think it's worth re-writing now, but will certainly considering using it in the future.

tests/test_regularization.py
Rocamonde: Yeah when I wrote this code I didn't really understand how SB3 loggers worked very well, now I understand them better.

src/imitation/algorithms/preference_comparisons.py
Rocamonde: Lol yes when I made this change I didn't want to touch stuff that someone else had added (I merged incoming changes from master) so I thought that would be an acceptable fix. Actually thinking about what the code is doing, having the same seed for both is probably more sensible.

Even though the mypy PR is trying to replace everything with RandomStates directly, but since the point of the PR is to improve type hints, I think it's fine if there are some inconsistencies still, those should be addressed and cleaned up on a separate PR.

Rocamonde: Actually the reason why I did this is because they have different defaults (0 and None), and that's because Generator.manual_seed seems to require an integer. But I can fix this by not passing a generator if no seed is passed.

src/imitation/algorithms/preference_comparisons.py
Rocamonde: The reason this is not possible currently is because the regularizer takes a logger and an optimizer, which is not accessible outside the scope of the initializer. You could inject them later, but then you might as well pass the class and construct the instance in `__init__()`.

Rocamonde: I've created a factory class that takes care of all the input validation, let me know your thoughts.

src/imitation/algorithms/preference_comparisons.py
Rocamonde: The problem is that you apparently have to call backward before regularizing if you use weight decay so that the change in weights does not affect the gradient. Potential solutions:

1. Keep as is, add an `else: loss.backward()`
2. Do 1 but rename `.regularize()` to `.regularize_and_backward()` to make sure we don't miss this
3. Split `.regularize()` into `.pre_regularize()` and `.post_regularize()`, calling manually the `.backward()` method in the middle.

We should add the test to catch this anyways, but super confused as to why this is not failing. If there is no gradient in the tensors, how is the optimizer doing gradient descent?

src/imitation/regularization/regularizers.py
```diff
@@ -0,0 +1,189 @@
...
+        self.optimizer = optimizer
+        self.lambda_ = initial_lambda
+        self.lambda_updater = lambda_updater
+        self.logger = logger
+
+        self.logger.record("regularization_lambda", self.lambda_)
+
+    @abc.abstractmethod
+    def regularize(self, loss: th.Tensor) -> R:
+        """Abstract method for performing the regularization step.
```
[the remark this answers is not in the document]
Rocamonde: See the other comment you left for a discussion.

src/imitation/regularization/updaters.py
```diff
@@ -0,0 +1,139 @@
...
+            ValueError: if the tolerable interval is negative or not a proper interval.
+        """
+        eps = np.finfo(float).eps
+        if not (eps < scaling_factor < 1 - eps):
+            raise ValueError(
+                "scaling_factor must be in (0, 1) within machine precision.",
+            )
+        if len(tolerable_interval) != 2:
+            raise ValueError("tolerable_interval must be a tuple of length 2")
+        if not (0 <= tolerable_interval[0] < tolerable_interval[1]):
```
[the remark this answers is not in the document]
Rocamonde: I think you mean "higher", that's what the ratio corresponds to, and you normally increase regularization when the training loss is low and validation loss is high.

I guess you have a point, but then you might intuitively also require tolerable_interval[0]` to be `<=1` since why would you reduce regularization if the training loss is lower than validation? However, this is what the original paper does -- they set the tolerable interval from 1.1 to 1.5 IIRC, meaning that when losses are approximately the same you bring regularization down. I guess I can imagine an algorithm that when this happens does the opposite, and it's not totally obvious to me why that would never make sense and therefore we should disallow it by design.

src/imitation/algorithms/preference_comparisons.py
AdamGleave: OTOH 2 seems like the best solution. I think 3 could be better if there's something we'd want to do (other than backward the loss) in between "pre" and "post" regularization -- but otherwise it's just adding extra cruft, and we can still forget to backward.

The grad parameter is still present without backward, just zero (set by `zero_grad()`). So won't be caught by tests unless we explicitly check for loss going down.

src/imitation/regularization/updaters.py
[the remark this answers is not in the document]
AdamGleave: Thanks, you're right I was interpreting the ratio in the opposite direction. I do find it odd that `tolerable_interval[0] > 1` -- I guess it's some kind of anti-hysteresis measure. Given this, fine to leave it as-is, we should be able to replicate DRLHP even if we don't necessarily agree with their parameter choices.

src/imitation/algorithms/preference_comparisons.py
Rocamonde: Is there a way to add to all our tests where we use an optimizer that the loss is going down? This feels like a pretty big bug that we should systematically test for.

AdamGleave: But the loss doesn't always go down -- at least not in SGD, for small batch sizes.

You can absolutely test this but you have to be careful to train for long enough, set sufficiently conservative thresholds, etc. Don't think it's something we can just tack on as boilerplate.

src/imitation/regularization/regularizers.py
```diff
@@ -13,22 +14,66 @@
...
+    the regularizer factory to create the regularizer.
+
+    This two-step process separates the configuration of the regularization
+    algorithm from additional "operational" parameters. This is useful because it
+    solves two problems:
+        1. The end-user does not have access to the optimizer and logger when
+        configuring the regularization algorithm.
+        2. Validation of the configuration is done outside the network constructor.
+
+    It also allows re-using a same regularizer factory for multiple networks.
```
AdamGleave, line 41:
```suggestion
    It also allows re-using the same regularizer factory for multiple networks.
```

src/imitation/regularization/regularizers.py
```diff
@@ -38,14 +83,71 @@ def __init__(
...
+                "result in no regularization.",
+            )
+
+        if val_split is not None and (
+            not isinstance(val_split, float)
+            or np.allclose(val_split, 0.0)
+            or val_split <= 0
+            or val_split >= 1
+        ):
+            raise ValueError("val_split must be a float strictly between 0 and 1.")
```
AdamGleave, line 102:
```suggestion
            raise ValueError(f"val_split = {val_split} must be a float strictly between 0 and 1.")
```

AdamGleave (commented):
Factory approach seems good! It's an easy to use API. The actual implementation is a little trickier to follow over direct instantiation. But overall I think it's the least-bad option.

The main alternative would be to just make the callee to preference comparisons responsible for passing in a non-empty `custom_logger` and create the optimizer. This isn't crazy -- the scripts are already creating a custom logger, and making the callee create the optimizer does at least give flexibility as to what optimization class to use. But it seems like it's adding a lot of friction for people who want to just programatically use the Python API, whereas the regularization factory is easy to use.

Another approach would have been to just let you instantiate the regularizer without specifying optimizer or logger, and have this be something that's set after instantiation (with some check that they have been set before calling `.regularize()`). But this seems more error prone: it's nice to have instantiation mean the object is actually ready to use.

There's a few small comments outstanding (e.g. adding tests), but once those addressed I think we should be good to go -- just request a re-review once it's ready to look at again.

Rocamonde:
Added all the requested changes, except the tests for the `.backward()`, which I moved off to a new issue, #562 . This is because I read through the tests and I found this is a problem that we should address more generally and probably deserves a separate PR. We only really check for progress in 3 algorithms as far as I could find.

AdamGleave (approved):
LGTM. Note I made a couple of minor comments to adjust phrasing of docstring -- please review those and merge if happy.

Issue #462, opened 2022-07-07
[Preference Comparison] Active learning from ensemble


AdamGleave:
The original [deep RL from human preferences paper](https://arxiv.org/pdf/1706.03741.pdf) uses an ensemble of reward models. It then selects queries for comparison that have the highest disagreement between models, a proxy for epistemic uncertainty. Specifically, section 2.2.4 states:

> We decide how to query preferences based on an approximation to the uncertainty in the reward function estimator, similar to Daniel et al. (2014): we sample a large number of pairs of trajectory segments of length k, use each reward predictor in our ensemble to predict which segment will be preferred from each pair, and then select those trajectories for which the predictions have the highest variance across ensemble members.

The ablation in section 3.3 (random queries) shows active learning significantly improves results in some environments (e.g. swimmer, qbert). However, it makes little difference in most environments, and even hurts performance in a handful (e.g. breakout). Nonetheless, it seems useful to have support for as it does often improve performance and is a useful starting point for more sophisticated active learning approaches.

https://github.com/HumanCompatibleAI/imitation/pull/460 adds support for reward ensembles. Given this, I think the remaining work is "just" to add support for using that uncertainty for query selection during preference comparison. I think the natural way to add support for this is to implement a new `Fragmenter` that performs active selection, rather than the current uniform random [`RandomFragmenter`](https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/algorithms/preference_comparisons.py#L339).

PR #482, opened 2022-07-20
Adding active selection support using epistemic uncertainty for reward ensemble (#462)

taufeeque9:
This pull request adds the `ActiveSelectionFragmenter` class for active learning with three supported uncertainty variants - logit, probability, and label.  It also refactors `CrossEntropyRewardLoss` by creating the `PreferencePredictor` class that can be wrapped on `RewardNet` to create a model that predicts the preference probability given a fragment pair.


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -317,6 +317,144 @@ def _get_trajectories(
...
+        Returns:
+            The reward given by the network(s) for all the transitions.
+            Shape: (num_transitions, ) for Single reward network and
+                   (num_transitions, num_networks) for ensemble of networks.
+        """
+        state = transitions.obs
+        action = transitions.acts
+        next_state = transitions.next_obs
+        done = transitions.dones
+        if self.is_ensemble:
```
AdamGleave, line 408: Why do we go via `predict_processed_all` for ensemble but `self.model` for no ensemble?

In particular I'm worried that `predict_processed_all` might apply normalization, which we do not want when computing probabilities (only for RL training).

@levmckinney I think you disabled forward on ensemble so might have some thoughts here on what the right thing to do is.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -317,6 +317,144 @@ def _get_trajectories(
...
+            return self.model(*preprocessed)
+
+    def probability(
+        self,
+        rews1: th.Tensor,
+        rews2: th.Tensor,
+    ) -> th.Tensor:
+        """Computes the Boltzmann rational probability that the first trajectory is best.
+
+        Args:
```
AdamGleave, line 422: What are the shapes -- is it batch size first or ensemble member first?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -317,6 +317,144 @@ def _get_trajectories(
...
+                matrix for ensemble models and array for non-ensemble models.
+            rews2: array/matrix of rewards for the second trajectory fragment.
+                matrix for ensemble models and array for non-ensemble models.
+
+        Returns:
+            The softmax of the difference between the (discounted) return of the
+            first and second trajectory.
+        """
+        # TODO(lev) this could be extracted into its own class.
+        if self.is_ensemble:
```
AdamGleave, line 433:
```suggestion
expected_dims = 2 if self.is_ensemble else 1
assert rews1.ndim == rews2.ndim == expected_dims
```

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -448,6 +586,99 @@ def __call__(
         return list(zip(iterator, iterator))
 
 
+class ActiveSelectionFragmenter(Fragmenter):
+    """Sample fragments of trajectories based on active selection.
+
+    Actively picks the fragment paris with the highest uncertainty (variance)
```
AdamGleave, line 592:
```suggestion
    Actively picks the fragment pairs with the highest uncertainty (variance)
```

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -448,6 +586,99 @@ def __call__(
...
+        Raises:
+            ValueError: `uncertainty_on` not in logit|probability|label.
+        """
+        super().__init__(custom_logger=custom_logger)
+        self.preference_predictor = preference_predictor
+        self.base_fragmenter = base_fragmenter
+        self.fragment_sample_factor = fragment_sample_factor
+        # self.preference_predictor = PreferencePredictor()
+        if not (uncertainty_on in ["logit", "probability", "label"]):
+            raise ValueError(f"{uncertainty_on} not supported.")
```
AdamGleave, line 625: Consider mentioning in error message which values are supported

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -448,6 +586,99 @@ def __call__(
...
+            custom_logger: Where to log to; if None (default), creates a new logger.
+
+        Raises:
+            ValueError: `uncertainty_on` not in logit|probability|label.
+        """
+        super().__init__(custom_logger=custom_logger)
+        self.preference_predictor = preference_predictor
+        self.base_fragmenter = base_fragmenter
+        self.fragment_sample_factor = fragment_sample_factor
+        # self.preference_predictor = PreferencePredictor()
```
AdamGleave, line 623: unneeded?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -448,6 +586,99 @@ def __call__(
...
+
+    def __call__(
+        self,
+        trajectories: Sequence[TrajectoryWithRew],
+        fragment_length: int,
+        num_pairs: int,
+    ) -> Sequence[TrajectoryWithRewPair]:
+        # sample a large number (self.fragment_sample_factor*num_pairs)
+        # of fragments from all the trajectories
+        fragment_pairs = self.base_fragmenter(
```
AdamGleave, line 636: It's a bit opaque what the different arguments are doing; consider calling using the keyword argument format, or defining a variable for the `int(...)` to self-docmuent.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -448,6 +586,99 @@ def __call__(
...
+            trajectories,
+            fragment_length,
+            int(self.fragment_sample_factor * num_pairs),
+        )
+        var_estimates = []
+        for i, fragment in enumerate(fragment_pairs):
```
AdamGleave, line 642: `i` doesn't seem to be used anywhere

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -448,6 +586,99 @@ def __call__(
...
+            frag1, frag2 = fragment
+            trans1 = rollout.flatten_trajectories([frag1])
+            trans2 = rollout.flatten_trajectories([frag2])
+            if self.uncertainty_on == "logit":
+                with th.no_grad():
+                    # NOTE: Should we first sum the rewards
```
AdamGleave, line 648: I'd lean towards summing the rewards to compute returns and then taking the variance, as otherwise differently shaped but otherwise identical ensemble members would show up as having high variance.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -448,6 +586,99 @@ def __call__(
...
+                    # and then take the var (curr approach)
+                    # or first take the var and then sum the rewards?
+                    _, rew1_var = self._reward_moments(trans1)
+                    _, rew2_var = self._reward_moments(trans2)
+                var_estimates.append(rew1_var + rew2_var)
```
AdamGleave, line 653: I think we might actually want the variance of the *difference* in rewards.

For example, suppose we have a 2-member ensemble, with member 1 identical to member 0 but with a constant shift of +100.

The *difference* in their returns is always the same, and would have variance zero. But the variance of their predictions would be quite high.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -448,6 +586,99 @@ def __call__(
...
+        fragment_pairs = self.base_fragmenter(
+            trajectories,
+            fragment_length,
+            int(self.fragment_sample_factor * num_pairs),
+        )
+        var_estimates = []
+        for i, fragment in enumerate(fragment_pairs):
+            frag1, frag2 = fragment
+            trans1 = rollout.flatten_trajectories([frag1])
+            trans2 = rollout.flatten_trajectories([frag2])
```
AdamGleave, line 645: This function is getting long -- maybe factor out the logic to compute variance under logit/probability/label into helper function(s)?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -954,7 +1141,7 @@ def _training_inner_loop(
             losses.append(output.loss)
             metrics.append(output.metrics)
         losses = th.stack(losses)
-        loss = losses.mean()
+        loss = losses.sum()
```
AdamGleave, line 1144: Good catch (otherwise effectively learning rate per member declines with # of members)

src/imitation/scripts/common/reward.py
```diff
@@ -47,6 +47,8 @@ def normalize_output_running():
 @reward_ingredient.named_config
 def reward_ensemble():
     net_cls = reward_nets.RewardEnsemble  # noqa: F841
```
AdamGleave, line 49: For multi-line configs we normally drop `# noqa: F841` and just add a `locals()` line to the end of the function which also quietens flake8. Hacky, admittedly, but probably still less bad than peppering every line with noqa.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -735,14 +966,13 @@ def forward(
...
         """
-        probs = th.empty(len(fragment_pairs), dtype=th.float32)
-        for i, fragment in enumerate(fragment_pairs):
-            frag1, frag2 = fragment
-            trans1 = rollout.flatten_trajectories([frag1])
-            trans2 = rollout.flatten_trajectories([frag2])
-            rews1 = self._rewards(model, trans1)
-            rews2 = self._rewards(model, trans2)
-            probs[i] = self._probability(rews1, rews2)
+        preference_predictor = PreferencePredictor(
```
AdamGleave, line 969: `PreferencePredictor` is created inside `CrossEntropyRewardLoss` (once for each call to `forward()`) but is created externally and passed through to `ActiveSelectionFragmenter`. Is there a reason to have a different convention here? Either seem fine to me but I'd expect it to be simpler to understand the code if we stick to one.

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -177,6 +181,18 @@ def train_preference_comparisons(
         seed=_seed,
         custom_logger=custom_logger,
     )
+    if active_selection:
+        preference_predictor = preference_comparisons.PreferencePredictor(
```
AdamGleave, line 185: Related to a previous comment in `preference_comparisons.py`: I find it odd that we have to create `PreferencePredictor` here but only for `ActiveSelectionFragmenter`

AdamGleave (commented):
Thanks for the PR -- looks like the key features are there and I appreciate the refactoring as you go (e.g. with `PreferencePredictor`).

Left some comments, mostly minor, although there's one more substantive one regarding whether to use `predict_processed_all` or not for ensemble.

Please do request another review once you've addressed the comments, also feel free to request another review as you make other changes as I know this is still a draft PR.

src/imitation/algorithms/preference_comparisons.py
levmckinney: Also, `predict_processed_all` returns NumPy arrays that have no gradients attached to them. This should not work for training right?

For now, I would recommend `rews = th.stack([member(state, action, next_state, done) for member in self.model.memeber], dim=-1)`. Second option we could bring back the forward_all method as a shorthand for this. Third option, we could make the reward `ensemble.forward` return a tensor of shape `(batch_size, num_members)` with all the rewards.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -317,6 +317,144 @@ def _get_trajectories(
...
+            rews1: array/matrix of rewards for the first trajectory fragment.
+                matrix for ensemble models and array for non-ensemble models.
+            rews2: array/matrix of rewards for the second trajectory fragment.
+                matrix for ensemble models and array for non-ensemble models.
+
+        Returns:
+            The softmax of the difference between the (discounted) return of the
+            first and second trajectory.
+        """
+        # TODO(lev) this could be extracted into its own class.
```
levmckinney, line 432: This todo seems to be resolved.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -448,6 +586,99 @@ def __call__(
...
+                    rews1 = self.preference_predictor.rewards(trans1)
+                    rews2 = self.preference_predictor.rewards(trans2)
+                    probs = (
+                        self.preference_predictor.probability(rews1, rews2)
+                        .cpu()
+                        .numpy()
+                    )
+                if self.uncertainty_on == "probability":
+                    var_estimates.append(probs.var())
+                else:  # uncertainty_on is label
```
levmckinney, line 665: This seems like it will just be capturing aleatoric uncertainty. Is that what we want?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -929,7 +1116,7 @@ def __init__(
         """Create an ensemble trainer."""
         if not isinstance(model, reward_nets.RewardEnsemble):
             raise TypeError(
-                f"RewardEnsemble expected by RewardEnsembleTrainer not {type(model)}.",
```
levmckinney, line 932: Fixed in upstream pull request.

src/imitation/algorithms/preference_comparisons.py
levmckinney: Should I fix this upstream?

src/imitation/algorithms/preference_comparisons.py
levmckinney, line 969: Make sure preference predictor passes gradients through from the reward network since it's used to calculate the loss.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -317,6 +317,144 @@ def _get_trajectories(
     return trajectories
 
 
+class PreferencePredictor(nn.Module):
```
levmckinney, line 320: Might call this `PreferenceModel` instead of preference predictor since it is effectively a likelihood model.

src/imitation/algorithms/preference_comparisons.py
levmckinney: When I hear variance on logits that's what I think of as well since the difference in the reward that gets plugged into the soft max not the return it's self.

levmckinney (changes requested):
Thanks for extracting out the probability model that seems like the right call!

src/imitation/algorithms/preference_comparisons.py
taufeeque9: Yes, that's correct. When I wrote this, I was trying to use the `reward_moments` method in `RewardEnsemble` and hence ended up summing the variance for the two fragments separately. Thanks for pointing it out!

src/imitation/algorithms/preference_comparisons.py
taufeeque9: I think you should fix this upstream as the change is more relevant there.

src/imitation/algorithms/preference_comparisons.py
taufeeque9: The PEBBLE paper's implementation seems to be calculating the variance over the probabilities given by the ensemble members. So that's why we have added 'probability' as one of the options for `uncertainty_on`. 
@AdamGleave Can you confirm if the implementation is correct?

src/imitation/algorithms/preference_comparisons.py
taufeeque9: `PreferenceModel` sounds better!

src/imitation/algorithms/preference_comparisons.py
taufeeque9: The preference predictor does pass the gradient through the reward network. It just doesn't pass the gradient through the RewardEnsemble since predict_processed_all returns a Numpy array. But this is not a bug since the `EnsembleTrainer` goes through the member networks and calculates the loss for them separately. Hence, for each of the member networks, the gradient will be passed through the loss computation.

5_train_preference_comparisons.ipynb
```diff
@@ -0,0 +1,194 @@
+{
```
AdamGleave, line 1: This notebook should not be committed at the top-level (belongs in examples/)?

taufeeque9: Sorry, got added by mistake. I have removed it.

src/imitation/algorithms/preference_comparisons.py
AdamGleave: I think `abs(0.5 - probs)` would be aleatoric uncertainty and `probs.var()` is epistemic.

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -177,20 +181,33 @@ def train_preference_comparisons(
         seed=_seed,
         custom_logger=custom_logger,
     )
+    preference_model = preference_comparisons.PreferenceModel(
+        **cross_entropy_loss_kwargs,
```
levmckinney, line 185: Think we should rename these to `preference_model_kwargs`

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -981,16 +1187,26 @@ def __init__(
...
         self,
         fragment_pairs: Sequence[TrajectoryPair],
         preferences: np.ndarray,
     ) -> th.Tensor:
+        assert len(fragment_pairs) == preferences.shape[0]
         losses = []
         metrics = []
-        for member in self._model.members:
-            output = self.loss.forward(member, fragment_pairs, preferences)
+        for member_idx in range(self._model.num_members):
```
levmckinney, line 1200: Is bagging the batch what we want to do here? Or should we be sampling independent batches from the dataset for each ensemble member?

AdamGleave: Independent batches does seem better. The standard bagging approach would be to subsample the entire dataset -- so if the batch size `len(fragment_pairs)` is entire dataset we recover that. But with smaller batch sizes the bagged batches will end up quite correlated with one another.

In the interests of wrapping this up soon, I'd be fine with merging the PR with the current bagging approach and just opening an issue to track it, then improve it later.

AdamGleave (changes requested):
Thanks this looks significantly improved and I think is almost ready to merge. Most of my comments were fairly minor. Please request re-review once changes addressed.

Sorry for being slow getting to this, will aim to be more responsive in future!

src/imitation/algorithms/preference_comparisons.py
taufeeque9: Yes, I'll add the correct implementation of bagging over the entire dataset through another PR. 

The current version is sort of an intermediate option of no bagging vs. bagging over the entire dataset. It shouldn't perform worse than no bagging IMO.

levmckinney: Agreed. Also, I'd be happy to write the improved bagging code latter today.

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -187,20 +191,33 @@ def train_preference_comparisons(
...
+        model=reward_net,
+    )
+    if active_selection:
+        fragmenter = preference_comparisons.ActiveSelectionFragmenter(
+            preference_model=preference_model,
+            base_fragmenter=fragmenter,
+            fragment_sample_factor=transition_oversampling,
```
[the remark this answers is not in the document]
taufeeque9: I have added a new `active_selection_oversampling` variable to set the extra factor of fragment pairs to sample for active selection. 
Just one question: Would `active_selection_oversampling` > `transition_oversampling` make sense? In this case, we are asking for more fragment pairs than available, and so there will be repeated fragments.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -355,45 +355,74 @@ def __init__(
...
+    ):
+        """Computes the preference probability of the first fragment for all pairs.
 
-    def forward(self, fragment_pairs: Sequence[TrajectoryPair]):
-        """Computes the preference probability of first fragment for all pairs.
+        Note: This function passes the gradient through for non-ensemble models.
+              For an ensemble model, this function should not be used for loss
+              calculation. It can be used in case when passing the gradient is not
+              required like during active selection and inference time.
+              The EnsembleTrainer passes each member through this function instead of
```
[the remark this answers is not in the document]
taufeeque9: Yes, that is correct. I have updated the sentence to make it clearer.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -641,50 +680,59 @@ def __call__(
...
+                    rews2,
+                )
+                .cpu()
+                .numpy()
+            )
+            if self.uncertainty_on == "probability":
+                var_estimate = probs.var()
+            else:  # uncertainty_on is label
+                preds = (probs > 0.5).astype(np.float32)
+                # probability estimate of Bernoulli random variable
```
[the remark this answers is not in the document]
taufeeque9: I'm treating it as a Bernoulli random variable since the random variable we are capturing is the preferred fragment (between fragment0 and fragment1). So if 60% of the ensemble member networks prefer fragment1 over fragment0 (i.e., they assign `prob` > 0.5), then `preds` will be an array of 60% 1s and 40% 0s. Finally, `prob_estimate` would be the Bernoulli probability estimate of preferring fragment1 over fragment0 which will be 60%. 

Does this make sense, or am I confusing/mixing things up?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -317,6 +317,181 @@ def _get_trajectories(
...
+        Returns:
+            The reward given by the network(s) for all the transitions.
+            Shape - (num_transitions, ) for Single reward network and
+            (num_transitions, num_networks) for ensemble of networks.
+        """
+        state = transitions.obs
+        action = transitions.acts
+        next_state = transitions.next_obs
+        done = transitions.dones
+        if self.is_ensemble:
```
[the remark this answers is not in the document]
taufeeque9: The first branch is called in `ActiveSelectionFragmenter` on lines 707 and 708.

levmckinney (commented):
LGTM

levmckinney (approved):

src/imitation/scripts/train_preference_comparisons.py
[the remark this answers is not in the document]
AdamGleave: Hm, I'm a bit rusty on the code, so not sure this is correct (@levmckinney do you want to check my reasoning?) But I think that what's going on is:
  1. `transition_oversampling` controls how many more timesteps we sample from the environment than the number of fragment pairs * fragment length we generate. Even if we had `transition_oversampling < 1` (not usually desirable), we wouldn't necessarily get exact duplicates as we (a) will still sample different pairs, even if we've seen each element in that pair before in another comparison; (b) a trajectory of length M splits into M-N fragments of length N, so as long as fragment length is smaller than trajectory length there's actually a lot of possible (but highly correlated!) fragments.
  2. `active_selection_oversampling` controls how many fragment pairs we sample from the generated fragments for each one we actually label.

So, if we hypothetically did used the `ActiveSelectionFragmenter` but with a totally random active selection criteria, I think this would be the same as just using `RandomFragmenter` but with `transition_oversampling` set to `transition_oversampling / active_oversampling`.

If this is right, then maybe we should multiply `transition_oversampling` by `active_oversampling` to decouple these two hyperparameters?

src/imitation/algorithms/preference_comparisons.py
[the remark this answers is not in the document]
AdamGleave: This makes sense, thanks for explanation, I was thinking of the Bernoulli random variable as being based on `probs` rather than `preds` being an estimate of a latent Bernoulli that the ensemble as a whole is modeling. Thanks for clearing this up!

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -377,31 +377,37 @@ def forward(
...
+              `ensemble_member_index`.
 
         Args:
             fragment_pairs: batch of pair of fragments.
-            ensemble_member_index: index of member network in ensemble model
+            ensemble_member_index: index of member network in ensemble model.
+                If the model is an ensemble of networks, this cannot be None.
+
+        Raises:
+            ValueError: `uncertainty_on` not in logit|probability|label.
```
AdamGleave, line 397: I think this is meant to be about `ensemble_member_index` instead? (That's the only ValueError I see in this function)

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -713,20 +729,15 @@ def variance_estimate(self, rews1, rews2) -> float:
...
-                .cpu()
-                .numpy()
-            )
+            probs = self.preference_model.probability(rews1, rews2)
+            probs = probs.cpu().numpy()
+            assert probs.shape == (self.preference_model.model.num_members,)
             if self.uncertainty_on == "probability":
                 var_estimate = probs.var()
-            else:  # uncertainty_on is label
+            elif self.uncertainty_on == "label":  # uncertainty_on is label
```
AdamGleave, line 740: Inclined to add a catch all `else: assert False`  though mostly just to make it self-documenting (even without this it'd still error, as `var_estimate` would not be defined, just in a less intuitive way)

tests/algorithms/test_preference_comparisons.py
```diff
@@ -498,12 +499,53 @@ def test_active_fragmenter_uncertainty_on_not_supported_error(venv, random_fragm
...
+        match="`ensemble_member_index` required for ensemble models",
+    ):
+        ensemble_preference_model([])
+
+
+def test_active_fragmenter_uncertainty_on_not_supported_error(
+    ensemble_preference_model,
+    random_fragmenter,
+):
+    with pytest.raises(ValueError):
```
AdamGleave, line 536: add match? (it can include a wildcard for brevity.)

tests/algorithms/test_preference_comparisons.py
```diff
@@ -498,12 +499,53 @@ def test_active_fragmenter_uncertainty_on_not_supported_error(venv, random_fragm
...
+            fragment_sample_factor=2,
+            uncertainty_on="uncertainty_on",
+        )
+
+
+def test_active_selection_raises_error_when_initialized_without_an_ensemble(
+    preference_model,
+    random_fragmenter,
+):
     with pytest.raises(ValueError):
```
AdamGleave, line 549: add match?

AdamGleave (approved):
LGTM apart from a few very minor comments. Nice work, I particularly liked the use of fixtures in the test cases!

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -110,10 +113,16 @@ def train_preference_comparisons(
         save_preferences: if True, store the final dataset of preferences to disk.
         agent_path: if given, initialize the agent using this stored policy
             rather than randomly.
-        cross_entropy_loss_kwargs: passed to CrossEntropyRewardLoss
+        preference_model_kwargs: passed to CrossEntropyRewardLoss
```
AdamGleave, line 116: Should this say passed to `PreferenceModel`?

AdamGleave:
CodeCov is passing now after the merge (which I screwed up but fixed). I think CodeCov sometimes gets confused when the base isn't the most recent version of master.

src/imitation/scripts/train_preference_comparisons.py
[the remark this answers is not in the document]
levmckinney: > So, if we hypothetically did used the ActiveSelectionFragmenter but with a totally random active selection criteria, I think this would be the same as just using RandomFragmenter but with transition_oversampling set to transition_oversampling / active_oversampling.

In the case of just using the `RandomFragmenter` with a transition over sample of `transition_oversampling / active_oversampling`. You would have more corolations between the fragments. Apart from that, I agree that the two cases you have outlined behave basically the same.

Let's consider the design decision of how to handle these two hyperparameters in more detail.

## Case 1: We **don't** multiply `tranision_oversampling` by `active_oversampling`.
In this case we calculate how many timesteps to sample from the environment like so,
```python
num_steps = math.ceil(
    self.transition_oversampling * 2 * num_pairs * self.fragment_length,
)
```
Then as a user of the library I can easily control in some sense how “hard we are looking for” surprising pairs in the pool of trajectories we collect. In addition, `active_oversampling` is decoupled from the number of transitions we actually collect from the environment.

## Case 2: We **do** multiply `tranision_oversampling` by `active_oversampling`.
In this case we calculate how many timesteps to sample from the environment like so,
```python
num_steps = math.ceil(
    self.transition_oversampling * self.active_oversampling * 2 * num_pairs * self.fragment_length,
)
```
The amount of correlation among the fragments would be kept to a minimum since we would always sample more from the environment. In addition, the end user might not have to worry about tuning the `self.transition_oversampling` at all. However, now `active_oversampling` also controls how many steps in the environment we are taking which feels a little weird to me.

Thus, I think we should not be multiply them together. We should emphasize that `transition_oversampling` controls how correlated the fragment pairs will be. And `active_oversampling` controls approximately how many of the possible fragment pairs we consider.

Perhaps a better way to decouple these are to replace `active_oversampling` with something like `fraction_of_pairs_to_consider` which controls what fraction of the total set of possible pairs we actually evaluate and rank.

Issue #386, opened 2022-01-19
Official support for Windows & Mac


AdamGleave:
Given a lot of people use Windows or Mac for development, it'd be nice if `imitation` could reliably support this. We don't use any Linux/BSD-specific features. Moreover, PyTorch and even MuJoCo support Windows and Mac, so this should require minimal (if any) porting.

To make sure we don't break it in the future, it'd be good to add continuous integration testing for these platforms. CircleCI supports [Mac and Windows executors](https://circleci.com/docs/2.0/executor-intro/), so I think it's just a matter of adding versions of the current [`unit-test` job](https://github.com/HumanCompatibleAI/imitation/blob/dbd26c889c36d70bf0eef42dd44f050a7d328541/.circleci/config.yml#L127) for these platforms. The complicating factor will be that we'll need to setup the binary environment to support what we need -- probably we just want to skip MuJoCo and all the nasty graphics dependency it pulls in? Might want to run an abbreviated version of the test suite on these anyway, both because CircvleCI [pricing](https://circleci.com/product/features/resource-classes/) is a lot higher for Windows/Mac and to not slow down the CI test too much.

Rocamonde:
Related: #477 and #476.

AdamGleave:
We now have CI on Windows, everything is working apart from some bash scripts (that are not a core part of the library) and the need to enable developer mode to get symlinks to work (needed by some of our scripts).

Only thing remaining is to add Mac CI (which I expect should be easier as POSIX compliant), and document Windows/Mac installation.

PR #487, opened 2022-07-23
Windows support on CircleCI (#386)

taufeeque9:
Added a windows-test workflow to CircleCI. 
Still need to fix ipynb errors on Windows.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,10 +84,10 @@ commands:
           keys:
             - v5-dependencies-{{ checksum "setup.py" }}
 
-      # - run:
-      #     name: install python
-      #     command: choco install python --version=3.8
-      #     shell: powershell.exe
+      - run:
+          name: install python
+          command: choco install --side-by-side -y python --version 3.8
+          shell: powershell.exe
 
       - run:
           name: install virtualenv
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,7 @@ commands:
       # Download and cache dependencies
       - restore_cache:
           keys:
-            - v5-dependencies-{{ checksum "setup.py" }}
+            - v6-unix-dependencies-{{ checksum "setup.py" }}
 
       - run:
           name: install dependencies
@@ -60,7 +60,7 @@ commands:
       - save_cache:
           paths:
             - /venv
-          key: v5-dependencies-{{ checksum "setup.py" }}
+          key: v6-unix-dependencies-{{ checksum "setup.py" }}
 
       - run:
           name: install imitation
@@ -71,10 +71,16 @@ commands:
           command: pip freeze --all
 
   dependencies-windows:
-    # You must still manually update the Docker image if any
-    # binary (non-Python) dependencies change.
     description: "Check out and update Python dependencies."
     steps:
+      # Note: The error preference is set to Stop in powershell which only applies to
+      # cmdlets. For exes, will have to manually check $LastErrorCode as done for pytest.
+      - run:
+          name: set error preference for ps
+          command: |
+            Add-Content -Path $PSHOME/Profile.ps1 -Value "`$ErrorActionPreference='Stop'"
+          shell: powershell.exe
+
       - checkout
 
       # Download and cache dependencies
@@ -233,6 +239,10 @@ jobs:
     executor:
       name: win/default
       size: xlarge
+      # using bash.exe as the default shell because te codecov/upload command 
+      # does not work if we default to powershell.exe as it internally uses a bash script. 
+      # Moreover it is not possible to specify the shell just for that command.
+      shell: bash.exe
 
     steps:
       - dependencies-windows
@@ -243,11 +253,11 @@ jobs:
       # still better than nothing.
       - restore_cache:
           keys:
-            - v8-win-pytest-cache-{{ checksum "setup.py" }}
+            - v7-pytest-cache-{{ checksum "setup.py" }}
             # This prefix-matched key restores the most recent pytest cache with any
             # checksum. If that cached policy happens to be still loadable, we avoid
             # retraining. See https://circleci.com/docs/2.0/caching/#restoring-cache
-            - v8-win-pytest-cache-
+            - v7-pytest-cache-
           paths:
             - .pytest_cache
 
@@ -259,6 +269,7 @@ jobs:
             pytest -n auto --cov=venv\Lib\site-packages\imitation `
             --cov=tests --junitxml=\tmp\test-reports\junit.xml -k 'not run_example_sh_scripts' `
             -vv tests\ --ignore tests\test_experiments.py ` --ignore tests\test_scripts.py
+            # manually checking (required for exes) and returning if pytest gives error code other than 0
             if ($LASTEXITCODE -ne 0) { throw "pytest failed" }
             mv .coverage .coverage.imitation
             coverage combine  # rewrite paths from virtualenv to src/
@@ -267,7 +278,7 @@ jobs:
       - codecov/upload
 
       - save_cache:
-          key: v8-win-pytest-cache-{{ checksum "setup.py" }}
+          key: v7-pytest-cache-{{ checksum "setup.py" }}
           paths:
             - .pytest_cache
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,9 +273,9 @@ jobs:
           command: |
             .\venv\Scripts\activate
             # Xdummy-entrypoint.py: starts an X server and sets DISPLAY, then runs wrapped command.
-            python ci\Xdummy-entrypoint.py pytest -n ${NUM_CPUS} --cov=/venv/lib/python3.10/site-packages/imitation `
-            --cov=tests --junitxml=/tmp/test-reports/junit.xml `
-            -vv tests/
+            python pytest -n ${NUM_CPUS} --cov=venv\Lib\site-packages\imitation `
+            --cov=tests --junitxml=\tmp\test-reports\junit.xml `
+            -vv tests\
             mv .coverage .coverage.imitation
             coverage combine  # rewrite paths from virtualenv to src/
           shell: powershell.exe
@@ -288,15 +288,15 @@ jobs:
             - .pytest_cache
 
       - store_artifacts:
-          path: /tmp/test-reports
+          path: \tmp\test-reports
           destination: test-reports
 
       - store_test_results:
-          path: /tmp/test-reports
+          path: \tmp\test-reports
           unit-test:
 
       - store_artifacts:
-          path: /tmp/resource-usage
+          path: \tmp\resource-usage
           destination: resource-usage
 
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,7 +82,7 @@ commands:
       # Download and cache dependencies
       - restore_cache:
           keys:
-            - v5-dependencies-{{ checksum "setup.py" }}
+            - v5-win-dependencies-{{ checksum "setup.py" }}
 
       - run:
           name: install python
@@ -104,7 +104,7 @@ commands:
       - save_cache:
           paths:
             - .\venv
-          key: v5-dependencies-{{ checksum "setup.py" }}
+          key: v5-win-dependencies-{{ checksum "setup.py" }}
 
       - run:
           name: install imitation
@@ -250,11 +250,11 @@ jobs:
       # still better than nothing.
       - restore_cache:
           keys:
-            - v7-pytest-cache-{{ checksum "setup.py" }}
+            - v7-win-pytest-cache-{{ checksum "setup.py" }}
             # This prefix-matched key restores the most recent pytest cache with any
             # checksum. If that cached policy happens to be still loadable, we avoid
             # retraining. See https://circleci.com/docs/2.0/caching/#restoring-cache
-            - v7-pytest-cache-
+            - v7-win-pytest-cache-
           paths:
             - .pytest_cache
 
@@ -272,7 +272,7 @@ jobs:
       - codecov/upload
 
       - save_cache:
-          key: v7-pytest-cache-{{ checksum "setup.py" }}
+          key: v7-win-pytest-cache-{{ checksum "setup.py" }}
           paths:
             - .pytest_cache
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,6 +72,11 @@ commands:
           name: print installed packages
           command: pip freeze --all
 
+  # codecov-upload-windows:
+  #   shell: bash.exe
+  #   steps:
+  #     - codecov/upload
+
   dependencies-windows:
     # You must still manually update the Docker image if any
     # binary (non-Python) dependencies change.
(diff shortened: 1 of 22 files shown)


.circleci/config.yml
```diff
@@ -69,6 +70,57 @@ commands:
           name: print installed packages
           command: pip freeze --all
 
+  dependencies-windows:
+    # You must still manually update the Docker image if any
```
AdamGleave, line 74: are we using Docker with windows?

.circleci/config.yml
```diff
@@ -69,6 +70,57 @@ commands:
...
+
+      # Download and cache dependencies
+      - restore_cache:
+          keys:
+            - v6-win-dependencies-{{ checksum "setup.py" }}
+
+      - run:
+          name: install python
+          command: |
+            $ErrorActionPreference = "Stop"  # exit immediately on any error
```
AdamGleave, line 88: Is this needed when there's only one command?

.circleci/config.yml
```diff
@@ -69,6 +70,57 @@ commands:
...
+      - run:
+          name: install dependencies
+          # Only create venv if it's not been restored from cache
+          command: if (-not (Test-Path venv)) { .\ci\build_and_activate_venv.ps1 -venv venv }
+          shell: powershell.exe
+
+      - save_cache:
+          paths:
+            - .\venv
+          key: v6-win-dependencies-{{ checksum "setup.py" }}
```
AdamGleave, line 106: For symmetry consider changing the Linux cache to `v6-unix-dependencies-{{ checksum "setup.py" }}`

.circleci/config.yml
```diff
@@ -69,6 +70,57 @@ commands:
...
+      - restore_cache:
+          keys:
+            - v6-win-dependencies-{{ checksum "setup.py" }}
+
+      - run:
+          name: install python
+          command: |
+            $ErrorActionPreference = "Stop"  # exit immediately on any error
+            choco install --side-by-side -y python --version 3.8
+          shell: powershell.exe
```
AdamGleave, line 90: Is it necessary to specify this? https://circleci.com/docs/using-windows#specifying-a-shell-with-the-windows-executor suggests powershell is the default

.circleci/config.yml
```diff
@@ -177,9 +229,76 @@ jobs:
           path: /tmp/resource-usage
           destination: resource-usage
 
+  unit-test-windows:
+    executor:
+      name: win/default
+      shell: bash.exe
```
AdamGleave, line 235: why do we default to bash.exe but then override and use powershell.exe in many of the following?

.circleci/config.yml
```diff
@@ -177,9 +229,76 @@ jobs:
...
+    steps:
+      - dependencies-windows
+
+      # - run:
```
AdamGleave, line 239: Should either port this to Windows or delete it. Probably fine to delete it. (Useful auxiliary debugging information if test go awry, but we can add it if/when needed.)

.circleci/config.yml
```diff
@@ -177,9 +229,76 @@ jobs:
...
+      #       done
+      #     background: true
+
+      # Note: The setup.py checksum in the cache name ensures that we retrain our expert
+      # policies, which are cached in the pytest cache, whenever setup.py changes. This
+      # is just a rough heuristic to decide when the experts should be retrained but
+      # still better than nothing.
+      - restore_cache:
+          keys:
+            - v8-win-pytest-cache-{{ checksum "setup.py" }}
```
AdamGleave, line 257: Does this need to be Windows specific? Expert policies should be saved in same format regardless of underlying OS.

If this does need to be separate, then should change other one to start `v8-unix-pytest-cache-` for symmetry.

.circleci/config.yml
```diff
@@ -177,9 +229,76 @@ jobs:
...
+            - v8-win-pytest-cache-
+          paths:
+            - .pytest_cache
+
+      - run:
+          name: run tests
+          command: |
+            $ErrorActionPreference = "Stop"  # exit immediately on any error
+            .\venv\Scripts\activate
+            # Xdummy-entrypoint.py: starts an X server and sets DISPLAY, then runs wrapped command.
```
AdamGleave, line 270: We don't use XDummy, so remove this?

.circleci/config.yml
```diff
@@ -177,9 +229,76 @@ jobs:
...
+            pytest -n 4 --cov=venv\Lib\site-packages\imitation `
```
AdamGleave, line 271: You might actually be able to use `-n auto` -- we avoided using that for Linux because the Docker executor exposed all the CPUs on the underlying machine but throttled us to a subset. I suspect this won't be an issue on Windows VM.

If we do need to specify number of CPUs manually, then consider adding a NUM_CPUS environment variable -- that way we only specify it in one place, and can update if/when we change the executor.

.circleci/config.yml
```diff
@@ -177,9 +229,76 @@ jobs:
...
+            --cov=tests --junitxml=\tmp\test-reports\junit.xml -k 'not run_example_sh_scripts' `
+            -vv tests\ --ignore tests\test_experiments.py ` --ignore tests\test_scripts.py
+            if ($LASTEXITCODE -ne 0) { throw "pytest failed" }
```
AdamGleave, line 274: Do we need this as well as ``$ErrorActionPreference = "Stop"``?

.circleci/config.yml
```diff
@@ -190,3 +309,7 @@ workflows:
       - unit-test:
           context:
           - docker-hub-creds
+
```
AdamGleave, line 312: Whitespace between `test` and `windows-test` but not between `version: 2` and `test`. I'm OK with either, but let's be consistent.

.circleci/config.yml
```diff
@@ -177,9 +229,76 @@ jobs:
...
+      - store_artifacts:
+          path: \tmp\resource-usage
+          destination: resource-usage
+
+
 workflows:
   version: 2
-  test:
+  test:
```
AdamGleave, line 301: spurious whitespace at end of this

.circleci/config.yml
```diff
@@ -177,9 +229,76 @@ jobs:
...
+            - .pytest_cache
+
+      - run:
+          name: run tests
+          command: |
+            $ErrorActionPreference = "Stop"  # exit immediately on any error
+            .\venv\Scripts\activate
+            # Xdummy-entrypoint.py: starts an X server and sets DISPLAY, then runs wrapped command.
+            pytest -n 4 --cov=venv\Lib\site-packages\imitation `
+            --cov=tests --junitxml=\tmp\test-reports\junit.xml -k 'not run_example_sh_scripts' `
```
AdamGleave, line 272: I'm pretty sure \tmp\ is not the canonical thing to use here, but if it works then :shrug: More concise than `%systemdrive%\Windows\Temp` for sure...

src/imitation/algorithms/dagger.py
```diff
@@ -634,18 +634,30 @@ def train(
 
             for traj in trajectories:
                 self._logger.record_mean(
-                    "dagger/mean_episode_reward",
+                    util.get_universal_path_from_posix("dagger/mean_episode_reward"),
```
AdamGleave, line 637: These are keys, not paths!

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1048,8 +1048,12 @@ def train(
                     self.dataset,
                     epoch_multiplier=epoch_multiplier,
                 )
-            reward_loss = self.logger.name_to_value["mean/reward/loss"]
-            reward_accuracy = self.logger.name_to_value["mean/reward/accuracy"]
+            reward_loss = self.logger.name_to_value[
+                util.get_universal_path_from_posix("mean/reward/loss")
```
AdamGleave, line 1052: also a key, not path. Was this causing a test failure beforehand?

src/imitation/scripts/analyze.py
```diff
@@ -123,7 +126,13 @@ def gather_tb_directories() -> dict:
...
 
                 tb_symlink = osp.join(symlinks_dir, run_name)
-                os.symlink(tb_src_dir, tb_symlink)
+                try:
+                    os.symlink(tb_src_dir, tb_symlink)
+                except OSError as e:
+                    if os.name == "nt":  # Windows
+                        print("Exception occurred while creating symlink.")
+                        print("Please ensure that Developer model is enabled.")
+                    raise e
```
AdamGleave, line 135: print could be easily be missed (goes to stdout not stderr, some execution environments will catch exception and log them but not stdout/stderr, etc). Would favour adding this message directly to exception. See https://stackoverflow.com/questions/9157210/how-do-i-raise-the-same-exception-with-a-custom-message-in-python for methods. Simplest is something like:

```raise OSError("Exception occurred while creating symlink ....") from e```

File: src/imitation/scripts/analyze.py [hunk 17]
```diff
@@ -123,7 +126,13 @@ def gather_tb_directories() -> dict:
                 os.makedirs(symlinks_dir, exist_ok=True)
 
                 tb_symlink = osp.join(symlinks_dir, run_name)
-                os.symlink(tb_src_dir, tb_symlink)
+                try:
```
AdamGleave, line 129: Since this symlink-plus-warning is used in two places in the code, could consider factoring it out into a utility function, though it's simple enough not sure it's worth it. (Would be if we use it in 3 or more places.)

src/imitation/util/sacred.py
```diff
@@ -99,7 +99,13 @@ def build_sacred_symlink(log_dir: types.AnyPath, run: sacred.run.Run) -> None:
...
     # Use relative paths so we can mount the output directory at different paths
     # (e.g. when copying across machines).
-    symlink_path.symlink_to(target_path, target_is_directory=True)
+    try:
+        symlink_path.symlink_to(target_path, target_is_directory=True)
+    except OSError as e:
+        if os.name == "nt":  # Windows
+            print("Exception occurred while creating symlink.")
+            print("Please ensure that Developer model is enabled.")
+        raise e
```
AdamGleave, line 108: Same comment as above about `raise ... from e`

tests/util/test_logger.py
```diff
@@ -102,24 +103,72 @@ def test_name_to_value(tmpdir):
 
     with hier_logger.accumulate_means("foo"):
         hier_logger.record("A", 1)
-        assert hier_logger.name_to_value["raw/foo/A"] == 1
-        assert hier_logger.name_to_count["raw/foo/A"] == 0
-        assert hier_logger.name_to_excluded["raw/foo/A"] is None
+        assert (
+            hier_logger.name_to_value[util.get_universal_path_from_posix("raw/foo/A")]
```
AdamGleave, line 107: keys not directories

AdamGleave (commented):
Thanks for adding support for this! Mostly looks good. Main comment is that a lot (possibly all?) uses of `util.get_universal_path_from_posix` seem to be for logger keys, not directories. I don't think there's any need to rewrite these -- were you seeing errors without rewriting these?

Other suggestions are pretty minor and mostly about simplifying CircleCI config where possible (and/or documenting where it's not possible to do so).

AdamGleave:
Ah I see that the tests were failing if you didn't rewrite all the keys, so it makes sense why you made that change. However, these keys were never intended to be paths -- they're just identifiers in the logger. The problem is we were being stupid and using `os.path.join` inside `util/logger.py`. I've modified that file to always use "/" as a separator for keys.

AdamGleave:
OK, only two tests failing now. The notebook test failures were because the logging keys were being rewritten to be too long, and is now resolved. I think I've pushed a fix for the examples script failing that was because it was using wrong Python interpreter.

Not sure about the remaining one: perhaps time on Windows is less fine-grained and we should add an eps?

.circleci/config.yml
taufeeque9: The codecov/upload command does not work if we default to powershell.exe as it internally uses a bash script. I tried to find whether it can be used with powershell.exe as the default shell, but there is no documentation for Windows. There are Github issues on their repo regarding the use of Windows but couldn't find anything concrete from that. 
It is also not possible to specify the shell for the codecov/upload command. It is run in the default shell of the job. So the easiest solution was to make bash.exe the default shell.

.circleci/config.yml
taufeeque9: Since I have provided bash.exe as the default shell (required for codecov/upload), I need to specify powershell.exe explicitly for commands that require it.
The virtual environment can only be activated through powershell.exe. So all the commands that require virtual env are run in powershell.

.circleci/config.yml
taufeeque9: As far as I understand, `$ErrorActionPreference = "Stop"` works when the error is generated through a powershell command (cmdlet), but it does not work when the error is generated through exe files. In such cases, we have to manually check through the $LASTEXITCODE variable whether the command returned with an error or not. 
So we definitely require the if condition. However, I'm not sure how helpful is setting `$ErrorActionPreference = "Stop"`.

You can look up more about this issue in the accepted answer of [this](https://stackoverflow.com/questions/9948517/how-to-stop-a-powershell-script-on-the-first-error) StackOverflow question.

.circleci/config.yml
taufeeque9: No, we are using the Windows executor provided by CircleCI orbs. I'll remove that comment.

.circleci/config.yml
taufeeque9: I don't think it is needed, but I added it anyway since it doesn't do any harm. Should I remove it?

.circleci/config.yml
taufeeque9: I wasn't sure if it would be OS-specific or not, so I decided to cache it separately for now.
I can remove it.

.circleci/config.yml
taufeeque9: I see that you removed `shell: bash.exe`, and the download codecov uploader step in CircleCI is failing because of it. 
Should I make the default shell bash again?

AdamGleave: Thanks for the explanation and sorry for breaking this with my commit. Yes, please do revert that part of my change. Can you add a comment explaining why we're defaulting to `bash` as I expect this to be confusing in the future?

.circleci/config.yml
AdamGleave: I lean towards removing it as we do not include it in any of the following one-liner commands.

There is a case for keeping it as defensive coding in case we add multiple commands later, but in that case I think we should do this elsewhere as well.

We could perhaps make Stop the default and eliminate the need for this in every script block? https://github.com/hashicorp/packer/issues/6449#issuecomment-402586809 has a method of doing this, though it seems like it needs admin permissions (perhaps there is a local-user version of this too though).

.circleci/config.yml
AdamGleave: OK, in that case definitely keep this line, and I'm fine having the ErrorActionPreference = "Stop" as well in case we have any errors in (future) cmdlet's too.

.circleci/config.yml
AdamGleave: It shouldn't be OS specific, so let's use the same keys and can revert if something breaks. (If it does break we should understand why... as we want our policies to be shareable across platforms. But that can be a new PR.)

.circleci/config.yml
taufeeque9: I checked the link's command through CircleCI's ssh option, and it seems to be working. I'm guessing CircleCI provides shell with Admin privileges. 
So I have added the command from the link in the config.

.circleci/config.yml
taufeeque9: Added a comment in the config.

AdamGleave (approved):
LGTM. One minor suggestion.

The test that was failing is flaky (confirmed via `pytest --flake-finder -k test_trainer_no_crash -vv tests/algorithms/test_preference_comparisons.py` on the Windows CircleCI box). I've submitted a PR upstream https://github.com/DLR-RM/stable-baselines3/pull/979 to fix this, and changed `setup.py` to use that hotfix until it gets merged.

AdamGleave (approved):
LGTM

PR #490, opened 2022-07-24
Version 0.3.0 release

AdamGleave:


Rocamonde:
LGTM. Ideally we'd integrate the bump version number and the github release into one, so that we don't ever forget to change the version number on releases.

AdamGleave:
> Ideally we'd integrate the bump version number and the github release into one

Yes agreed, it is possible to trigger GitHub releases via GitHub workflows but it looked involved enough I didn't want to set it up right now. Feel free to make an issue for this.

PR #495, opened 2022-07-26
Improve Windows CI

AdamGleave:
Unignore `tests/test_scripts.py` and install `ffmpeg` so the tests pass.

Where skips still are required (bash shell scripts that haven't been ported to Windows), skip in the test rather than in `.circleci/config.yml`. This has the benefit of: (a) documenting where tests do/don't work in the tests (easier to discover, and will be retained even if tests renamed); (b) skipping expected-to-fail tests on local test runs, not just CI builds.

I was running into the SB3 bug fixed in https://github.com/DLR-RM/stable-baselines3/pull/979 with `test_scripts.py` so I've swapped our SB3 version to `master` on Windows. This feels like the least-evil solution. Flipping it everywhere would break our PyPI automation. We can remove this hack as soon as the next SB3 release is out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,9 +89,10 @@ commands:
             - v6-win-dependencies-{{ checksum "setup.py" }}
 
       - run:
-          name: install python
+          name: install python and binary dependencies
           command: |
             choco install --side-by-side -y python --version 3.8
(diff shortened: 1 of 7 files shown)


levmckinney (approved):
LGTM

Issue #500, opened 2022-07-28
Fix test PyPI automation


AdamGleave:
Currently we have a GitHub workflow that uploads to test PyPI on every commit; and to real PyPI on the release. The idea of the test PyPI automation is to check the automation is working and that the package is accepted by PyPI on each commit. In particular this'll catch things like `setup.py` depending on things not in PyPI.

However, this automation is sometimes failing because of version number conflicts. This is because we're automatically generating versions with `setuptools_scm`, which for pre-release appends a `.devN` number. This seems to be guessing the N based on how many commits relative to the last release. But a linear number N that's incremented can't handle there being multiple branches. Usually setuptools_scm handles this with a local scheme including things like a GitHub commit hash -- but test PyPI doesn't accept that, so we've disabled it with `no-local-version`.

I think we either need to remove the test PyPI automation (simplest, but loses some checks), or hack `setuptools_scm` to include the local scheme but in a way that PyPI accepts (e.g. change `-` to `_`).

I'm surprised there isn't a good general solution to this. Some relevant issues: https://github.com/pypa/setuptools_scm/issues/455 and https://github.com/pypa/setuptools_scm/issues/342

Rocamonde:
I guess my first question would be, why are we using pre-release versions? We probably shouldn't do that if we can avoid it, as this makes it hard for other people to use our library (we're forcing them to use unstable software which might be incompatible with their code or other packages they use).

AdamGleave:
I don't quite understand this question. This is for *test PyPI*, not production PyPI. No user will ever see these packages, unless they go out of their way to point pip at test PyPI. It's solely to test that the automated pip release workflow works correctly.

Rocamonde:
Apologies, I thought the pre-release versions came from the dependencies, not from the package itself, as it also seemed surprising to me that there is not already a solution for this issue. I will investigate.

Rocamonde:
Is there a way to check if the installer is being run for test pypi purposes or for regular pypi purposes? Because the benefit of this "bug" is that unversioned releases will never get published, which is actually a very important feature.

Rocamonde:
Actually there's another problem with this, which is that any other version format is not PEP 440 compliant, so the wheel does not compile, which is a requirement for pyproject.toml-based projects. I don't think testpypi can be configured to allow local versions. 

Possible solutions: 

1. Remove
[...]

PR #505, opened 2022-07-28
Dropping support for python 3.7

levmckinney:
Drops support for python version 3.7.


AdamGleave:
Seems reasonable. The PyPI automation failing is not your fault and is tracked in https://github.com/HumanCompatibleAI/imitation/issues/500

setup.py
```diff
@@ -161,7 +161,6 @@ def run(self):
         "License :: OSI Approved :: MIT License",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
```
levmckinney, line 163: Are we only supporting 3.8 or do we also support 3.9 and 3.10?

levmckinney: `stable-baselines` says it to supports up to 3.9.

AdamGleave (approved):
LGTM.

We support later Python versions, I've run it with 3.10 successfully, we should add it to `setup.py`.

We're not testing other versions in CI but it'd be pretty weird for later versions to break, Python aims to be backward compatible, the main issues I've run into with later Python versions has been sometimes binary dependencies don't exist for a particular wheel (e.g. TF 1.15 only goes up to 3.7 support).

PR #515, opened 2022-08-02
circleci: Remove MuJoCo

tomtseng:
This PR simplifies the CI tests by removing the MuJoCo requirement, as suggested by https://github.com/HumanCompatibleAI/imitation/pull/511#pullrequestreview-1058076927.

## Testing

Check CI to make sure no tests are newly failing or passed.

[`master`](https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/2134/workflows/5cf3e8ff-ace1-486f-b274-a10eb9e6253c/jobs/5394):
```
========= 2475 passed, 588 skipped, 4156 warnings in 179.25s (0:02:59) =========
```

[This PR](https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/2148/workflows/5b6bc64e-0237-48d6-aaf5-c1d647e79f17/jobs/5455):
```
========= 2475 passed, 588 skipped, 4155 warnings in 146.21s (0:02:26) =========
```


AdamGleave:
The test `test_parallel_train_adversarial_custom_env` was previously being skipped on Windows CI because there's no MuJoCo on our Windows CI executor. But is now not being skipped and failing.

It's worth briefly trying to figure out why it's failing (you could rerun job with SSH from CircleCI instance and SSH in to troubleshoot). My guess is Ray is either just broken on Windows, or that the scripts are failing because of Windows not supporting symlinks and this error is just being hidden by Ray. If can't figure it out in an hour then OK just skipping it without getting to bottom of this (search the tests/ code for `os.name == "nt"` to see how we've handled this elsewhere).

AdamGleave (commented):
LGTM apart from test failure on Windows

tomtseng:
After wrangling with the failing test for a bit, I'm just going to skip it since I can't figure out exactly what's going on.

The problem is that `ray.init()` hangs in `test_parallel_train_adversarial_custom_env`. But it seems to mostly fail when running concurrently with other `test_parallel*` tests. E.g., besides `pytest -n auto` failing, `pytest -n auto -k test_parallel` also fails. OTOH, running just the failing test alone does not fail, and running the test suite serially with `pytest` (no `-n` flag) also does not fail. Maybe `ray` doesn't behave well when there are several processes trying to use `ray` independently? 

An alternative to skipping the test could be to make the CI test job run only the `test_parallel*` tests without `-n`. If we see similar tests failing on Windows in the future, this may be a good idea — not sure why this issue is only showing up on `test_parallel_train_adversarial_custom_env` currently

AdamGleave (approved):
LGTM. You have one lint error (missing trailing comma).

`@pytest.mark.skipif` decorator is an alternative, but I think given the long reason for skipping the current code is probably cleaner

Issue #512, opened 2022-08-01
Standard deviation of loss not logging in ensemble trainer


levmckinney:
The `dist_loss` and `loss_std` metrics in ensemble trainer are not being recorded properly to tensorboard.

PR #516, opened 2022-08-03
Improved logging for preference comparison

levmckinney:
This pull request fixes a bug in the logging code for the `EnsembleTrainer` #512. It adds logging of the loss of the ground truth reward on the training data when the ground truth reward is available. This should allow users to more easily check how well the reward model is fitting the training data.

### DIFF
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -995,17 +995,13 @@ class EnsembleTrainer(BasicRewardTrainer):
         loss = losses.sum()
 
         self.logger.record("loss", loss.item())
-        self.logger.record("loss_std", loss.std().item())
-        # Note here we are returning all the losses not just the mean
-        # This will give us a histogram
-        self.logger.record("dist_loss", losses.detach().cpu().numpy())
+        self.logger.record("loss_std", losses.std().item())
 
         # Turn metrics from a list of dictionaries into a dictionary of
-        # tensors. Again this should give us a histogram in tensorboard.
+        # tensors.
         metrics = {k: th.stack([di[k] for di in metrics]) for k in metrics[0]}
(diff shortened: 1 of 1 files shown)


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -765,10 +780,17 @@ def forward(
...
+        if gt_reward_available:
+            metrics["gt_reward_loss"] = th.nn.functional.binary_cross_entropy(
+                gt_probs,
+                preferences_th,
+            )
+
         return LossAndMetrics(
             loss=th.nn.functional.binary_cross_entropy(probs, preferences_th),
-            metrics={"accuracy": accuracy.detach().cpu()},
+            metrics=metrics,
```
AdamGleave, line 793: Previously we had `.detatch().cpu()`, now we're just returning raw tensors I think. Is this supported?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -975,17 +997,13 @@ def _training_inner_loop(
         loss = losses.sum()
 
         self.logger.record("loss", loss.item())
-        self.logger.record("loss_std", loss.std().item())
-        # Note here we are returning all the losses not just the mean
-        # This will give us a histogram
-        self.logger.record("dist_loss", losses.detach().cpu().numpy())
```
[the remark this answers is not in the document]
levmckinney: It doesn't seem to be working, so I decided to remove it.

src/imitation/algorithms/preference_comparisons.py
levmckinney: This could cause crashes if we are training on the GPU or lead to a memory leak. Good catch. I just fixed it.

AdamGleave (approved):
LGTM

Issue #507, opened 2022-07-31
Dependencies should be fixed to major versions


ernestum:
In lots of cases we suddenly get in trouble since dependencies are specified without a maximum version while a new major release comes out.
Examples:
- https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/2090/workflows/49b7d3d9-59bd-40a3-89e6-8d2c1498cc0a/jobs/5218?invite=true#step-110-43
- https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/2088/workflows/a138a7b9-e0ea-40b4-b4fb-845ece17a819/jobs/5212/parallel-runs/0/steps/0-113

As @Rocamonde suggested, we should fix this by restricting the major version number of package dependencies.

ernestum:
I followed the discussion in #509 and there were some discussions on how to chose the trade off between constraining package versions more tightly which results in a more stable build pipeline and, on the other side, having less restrictions for our users at the cost of a more fragile pipeline.

I think the way to deal with the issue depends a lot on the context. 
Libraries should constrain their dependency version as little as possible, such that their users are not bothered with version conflicts. Of course broken versions should be excluded and for dependencies with [semantic versioning](https://semver.org/), the major version could be pinned.
Applications on the other hand should pin their dependencies as tightly as possible to make sure they always run.
Their dependencies should be upgraded manually for security reasons or when newer features are required.
This is nicely summarized in [this SO answer](https://stackoverflow.com/a/44938662).

In our context, `imitation` is obviously a library, but the tests and documentation have more the nature of an application.
Therefore I would vote to pin the versions of our docs and test requirements tightly and keep the install requirements as loose as possible and leave a comment in the `setup.py` to ensure future changes to the dependencies adhere to this principle.
What do you think @tomtseng @Rocamonde @AdamGleave ?

PR #517, opened 2022-08-03
Pin test and doc requirements

ernestum:
This is a proposal to solve #507.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ else:
 
 # pinned to 0.21 until https://github.com/DLR-RM/stable-baselines3/pull/780 goes
 # upstream.
-GYM_VERSION_SPECIFIER="==0.21.0"
+GYM_VERSION_SPECIFIER = "==0.21.0"
 
 # Note: the versions of the test and doc requirements should be tightly pinned to known
 #   working versions to make our CI/CD pipeline as stable as possible.
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ from setuptools.command.install import install
 IS_NOT_WINDOWS = os.name != "nt"
 
 PARALLEL_REQUIRE = ["ray[debug,tune]>=1.13.0"]
-PYTYPE = ["pytype"] if IS_NOT_WINDOWS else []
+PYTYPE = ["pytype==2022.7.26"] if IS_NOT_WINDOWS else []
 if IS_NOT_WINDOWS:
     # TODO(adam): use this for Windows as well once PyPI is at >=1.6.1
     STABLE_BASELINES3 = "stable-baselines3>=1.6.0"
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,10 @@ else:
         "https://github.com/DLR-RM/stable-baselines3.git@master"
     )
 
+# pinned to 0.21 until https://github.com/DLR-RM/stable-baselines3/pull/780 goes
+# upstream.
+GYM_VERSION_SPECIFIER="==0.21.0"
+
 # Note: the versions of the test and doc requirements should be tightly pinned to known
 #   working versions to make our CI/CD pipeline as stable as possible.
 TESTS_REQUIRE = (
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -103,11 +106,12 @@ def run(self):
...
+    #   for our users.
     install_requires=[
-        # If you change gym version here, change it in "mujoco" below too.
-        # pinned to 0.21 until https://github.com/DLR-RM/stable-baselines3/pull/780
-        # goes upstream.
-        "gym[classic_control]==0.21.0",
+        "gym[classic_control]" + GYM_VERSION_SPECIFIER,
         "matplotlib",
         "numpy>=1.15",
         "torch>=1.4.0",
```
tomtseng, line 117: should we also pin the packages listed in `PYTYPE = ["pytype"]` and in `extras_require["dev"]`?

tomtseng (commented):
PR looks good to me, but I'll let someone more involved with `imitation` approve this

setup.py
```diff
@@ -20,48 +20,51 @@
...
-        "seals",
-        "black[jupyter]",
-        "coverage",
-        "codecov",
-        "codespell",
-        "darglint",
-        "filelock",
-        # TODO(adam): remove pin once flake8-isort fixed:
-        #  https://github.com/gforcada/flake8-isort/issues/115
+        "seals==0.1.2",
```
AdamGleave, line 31: I think seals shouldn't make breaking changes (and if it does we know who to complain to...)

```suggestion
        "seals~=0.1.2",
```

setup.py
```diff
@@ -20,48 +20,51 @@
...
-        # remove pin once https://github.com/jupyter/jupyter_client/issues/637 fixed
-        "jupyter-client<7.0",
-        "pandas",
-        "pytest",
-        "pytest-cov",
-        "pytest-notebook",
-        "pytest-xdist",
-        "scipy>=1.8.0",
-        "wandb",
+        "flake8-blind-except==0.2.1",
```
AdamGleave, line 39:
```suggestion
        "flake8-blind-except~=0.2.1",
```

Perhaps I'm missing something: does ~= interact poorly with 0.x.y versions?

setup.py
```diff
@@ -20,48 +20,51 @@
...
-        "flake8-blind-except",
-        "flake8-builtins",
-        "flake8-commas",
-        "flake8-debugger",
-        "flake8-docstrings",
-        "flake8-isort",
-        "hypothesis",
-        "ipykernel",
-        "jupyter",
-        # remove pin once https://github.com/jupyter/jupyter_client/issues/637 fixed
```
AdamGleave, line 44: Do we want to lose this comment? Admittedly we don't want to remove the pin any longer, but I think we can move to the 7.x series once it's fixed.

setup.py
AdamGleave: I agree that pinning pytype is good (although because they use a YYYY.MM.DD release format not semantic versioning it unfortunately may have to be an == not >=).

I think it's fine to leave the `"dev"` requires unpinned as not used in our CI.

In general I'd rather leave things unpinned except when needed for CI stability. Pinning forces us to manually update them which adds maintenance burden.

AdamGleave (approved):
LGTM. Some small suggestions to switch `==` to `~=` in a couple of places, and add pin for pytype.

setup.py
ernestum: True, I pinned it with `==` to the current version.

setup.py
ernestum: You are right. I added it back as a TODO

setup.py
ernestum: 0.x.y versions are not considered stable [according to semver](https://semver.org/#spec-item-4), therefore I fixed all packages with such a version to an exact version.

setup.py
ernestum: I would rather propose releasing seals `1.0.0` then. Having a `0.x.x` release signals, that the API might change at any time.
See the [semver spec](https://semver.org/#spec-item-4) for details.

AdamGleave (approved):
LGTM. Fair point about semantic version specifier, will try and get a v1 seals out at some point.

ernestum:
Totally unclear to my why the pipeline fails. Coverage went down just because our dependency versions are fixed? This is very strange. I removed those three lines just to see what happens. Codecov still claims that the coverage is down but when I go on the page itself and look at the stats for the branch, it claims that it was going up. Is codecov buggy?

I also tried to investigate into the issues uploading to TestPyPI but short of just removing that workflow, I have no clue how to solve the issue.

AdamGleave:
Test PyPI issue is tracked in https://github.com/HumanCompatibleAI/imitation/issues/500 -- we need to modify how the version number is automatically generated.

AdamGleave:
I can also just override the codecov requirement to merge without that being satisfied, it doesn't seem like a significant difference.

PR #518, opened 2022-08-05
preference_comparisons: Save untransposed trajectories if observations are images

### LINKED ISSUE #486: Dealing with inconstent conventions for env observations: (h,w,c) vs (c,h,w)
Atari environments in gym return observations in (h,w,c) format. In stable baselines 3, policies are implemented in pytorch, which assumes (c,h,w) observations. As a result, stable baselines 3 has a few places where observations are transposed: in a wrapper applied to an environment when it's fed into a learning algorithm (see lines 163 and 224 of `stable_baselines3/common/base_class.py`), in some methods of policies (see lines 335 and 243 of `stable_baselines3/common/policies.py`), and possibly in other places. This means that when writing code, it is not always obvious whether this transposition has been applied to an environment.

For instance, the `__init__` method of `AgentTrainer` in `preference_comparisons.py` calls `self.algorithm.get_env()` to set the environment. Because `algorithm` must be a `BaseAlgorithm`, it wraps its input venv in a wrapper that transposes observations as necessary before returning that venv in the `get_env` method. 

This means that in the notebook `imitation/examples/5_train_preference_comparisons.ipynb`, if the environment is changed to be an Atari environment, `reward_net` is originally trained on inputs with format (c,h,w), but when placed in a `RewardVecEnvWrapper`, is called on inputs with format (h,w,c). This will likely cause poor performance, or in the case of a CNN reward net, a type error.

To deal with problems like this, the best solution is likely a wrapper around reward functions to detect when this transposition needs to be made and make it before it's fed into the naive reward function.

tomtseng:
Issue #486: When an environment has image observations with shape `(h,w,c)`, SB3 algorithms will internally transpose the environment to get `(c,h,w)` observations. In `preference_comparisons.py`, we save `(c,h,w)` trajectories in a buffer and train a reward model, which the user probably initialized assuming untransposed `(h,w,c)` observations, on the transposed trajectories in the buffer. 

Changes:
- Add `venv` argument to `AgentTrainer` so that `AgentTrainer` can wrap its `BufferingWrapper` around the original untransposed `venv`. This makes us save `(h,w,c)` trajectories in the buffer and train reward models in the correct observation space.

Fixes #486.

## Testing

- unit tests pass
- `examples/5_train_preference_comparisons.ipynb` doesn't crash if I run all the cells
- `examples/5_train_preference_comparisons.ipynb` doesn't crash if I run all the cells using the `gym[box2d]` env `CarRacing-v0`, which has `(h,w,c)` observations


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -176,7 +186,7 @@ def __init__(
         self.algorithm.set_env(self.venv)
         policy_callable = rollout._policy_to_callable(
             self.algorithm,
-            self.venv,
+            self.algorithm.get_env(),
```
tomtseng, line 189: we can't use `self.venv` here or else `_policy_to_callable` will complain about observation spaces not matching up between `self.algorithm`'s transposed observation space vs. `self.venv`'s untransposed observation space

src/imitation/scripts/common/reward.py
```diff
@@ -150,7 +150,7 @@ def make_reward_net(
             for _ in range(ensemble_size)
         ]
 
-        reward_net = net_cls(venv.observation_space, venv.observation_space, members)
+        reward_net = net_cls(venv.observation_space, venv.action_space, members)
```
tomtseng, line 153: typo? this was causing a test to fail

dfilan:
FYI I have a related PR where I write a wrapper around reward functions to deal with this: https://github.com/HumanCompatibleAI/imitation/pull/519

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -24,6 +24,7 @@
 import torch as th
 from scipy import special
 from stable_baselines3.common import base_class, type_aliases, vec_env
+from stable_baselines3.common.utils import check_for_correct_spaces
```
AdamGleave, line 27: Violates section 2.2 of style guide https://google.github.io/styleguide/pyguide.html#22-imports

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -146,26 +148,30 @@ def __init__(
...
-
-        Raises:
-            ValueError: `algorithm` does not have an environment set.
         """
         self.algorithm = algorithm
         # NOTE: this has to come after setting self.algorithm because super().__init__
         # will set self.logger, which also sets the logger for the algorithm
         super().__init__(custom_logger)
         if isinstance(reward_fn, reward_nets.RewardNet):
+            check_for_correct_spaces(
```
AdamGleave, line 157: :+1 good sanity check!

src/imitation/algorithms/preference_comparisons.py
AdamGleave: Thanks for flagging. Perhaps turn this into a one-line comment in the code as others may wonder about this in the future?

This is a bit awkward but I don't think we can do much to improve this without modifying SB3.

src/imitation/scripts/common/reward.py
AdamGleave: Uhh yeah this makes more sense!

tests/algorithms/test_preference_comparisons.py
```diff
@@ -409,3 +417,45 @@ def test_exploration_no_crash(agent, reward_net, fragmenter, custom_logger):
...
+    trajectories = agent_trainer.sample(2)
+    assert len(trajectories) > 0
+    assert all(
+        trajectory.obs.shape[1:] == venv.observation_space.shape
+        for trajectory in trajectories
+    )
+
+
+def test_agent_trainer_sample_image_observations():
+    # SB3 algorithms may internally rearrange the channel dimension in environments with
```
AdamGleave, line 437: Should this comment be a docstring rather than a comment? (About what the test does, not how it does it.)

AdamGleave (approved):
LGTM, just a few minor suggested changes.

Are we confident this is just an issue with preference comparison? I could imagine similar issues with `train_adversarial` populating its generator. I know you're largely jumping off of `imitation` end of today, but it might be worth at least opening an issue to track it if we think there might be problems elsewhere so that it doesn't get dropped.

tomtseng:
I grepped for `RewardVecEnvWrapper` and `BufferingWrapper` — in both `DensityAlgorithm` and `AdversarialTrainer` they're being applied `venv` given directly as an `__init__` argument rather than being applied to `get_env()`.

AdamGleave:
> Re whether this is happening elsewhere: I grepped for `RewardVecEnvWrapper` and `BufferingWrapper` — in both `DensityAlgorithm` and `AdversarialTrainer` they're being applied `venv` given directly as an `__init__` argument rather than being applied to `get_env()`.

OK, sounds good.

Long-term we should add tests for these algorithms on image-based environments with CNNs, but that's a more natural fit for https://github.com/HumanCompatibleAI/imitation/pull/519 (or even a follow-up to it to avoid making the PR too large), so don't want that to block this one.

AdamGleave:
The codecov is failing for a silly reason and is only a small drop so I'm happy to ignore that, though I think I'll need to be the one to do the merge in that case as I think it requires admin privileges to override. So, let me know when it's ready to merge.

norabelrose (approved):
This looks good to me! I'm still a little sad that we're having to add `venv` as an extra argument to `AgentTrainer`, but it's not a big deal and I don't have a proposal that Pareto dominates this one.

PR #522, opened 2022-08-06
Optimize EMANorm by removing for loop over a batch

taufeeque9:
## Description

Kept the incremental version of `EMANorm` in `update_stats_incremental`. Implemented the exact algorithm for computing EMA without any for loops. The runtime of the function has decreased by a factor of ~10. Now, the order of runtime is the same as that of `RunningNorm`.  

Also, updated the code to compute EMA at the very first step too. Earlier, simple mean and average were computed in the first step.

## Testing

Added a new test to compare the batch computation of EMA with the original incremental version of EMA. Also removed one test case (`test_running_norm_identity_train`) for EMA since it is not applicable for EMA. Earlier the test was passing since we were computing simple mean and variance at the first step.


src/imitation/util/networks.py
```diff
@@ -155,7 +155,7 @@ def __init__(
 
         self.decay = decay
 
-    def update_stats(self, batch: th.Tensor) -> None:
+    def update_stats_incremental(self, batch: th.Tensor) -> None:
```
AdamGleave, line 158: Do we expect this to be used for any purpose other than testing? If not, I'm inclined to pull it out of here. You could e.g. make a class in the test file that subclasses `EMANorm` and overrides `update_stats` in this way, or even just create two instances of `EMANorm` and use `mock.patch` to replace the `update_stats` method in one of them with this.

src/imitation/util/networks.py
```diff
@@ -165,26 +165,67 @@ def update_stats(self, batch: th.Tensor) -> None:
...
+        b_size = batch.shape[0]
+        batch = batch.reshape(b_size, -1)
+        # Shuffle the batch since we don't don't want to bias the mean
+        # towards data that appears latter in the batch
+        perm = th.randperm(b_size)
+        batch = batch[perm]
+        # geometric progession of decay: decay^(N-1),...,decay,1
+        weights = th.vander(th.tensor([self.decay]), N=b_size).T
+        alpha = 1 - self.decay
+        if self.count == 0:
```
AdamGleave, line 206: FWIW using a simple mean/variance on the first batch seemed OK to me. All data in the batch we can treat as having arrived at the same time, so there's no reason to weight some data in the batch more than others. This is effectively what the randomization is trying to prevent (by not *consistently* weighting some indices more than others), but for the first batch a simple average also works for this.

However I don't feel particularly strongly about this either way, it does seem nice to be consistent. Any thoughts @levmckinney?

src/imitation/util/networks.py
```diff
@@ -165,26 +165,67 @@ def update_stats(self, batch: th.Tensor) -> None:
...
         self.count += b_size
 
+    def update_stats(self, batch: th.Tensor) -> None:
+        """Update `self.running_mean` and `self.running_var` in batch mode.
+
+        Args:
+            batch: A batch of data to use to update the running mean and variance.
+        """
+        b_size = batch.shape[0]
+        batch = batch.reshape(b_size, -1)
```
AdamGleave, line 198: Do we want to flatten the array apart from the batch dimension? I think the [forward](https://github.com/HumanCompatibleAI/imitation/blob/254b5f9d70654110e41ad08f90d2bac1b013c9fa/src/imitation/util/networks.py#L89) method is assuming `running_mean` and `running_variance` will have shape `batch.shape[1:]` so this flattening might break that code.

src/imitation/util/networks.py
```diff
@@ -165,26 +165,67 @@ def update_stats(self, batch: th.Tensor) -> None:
...
+        # Shuffle the batch since we don't don't want to bias the mean
+        # towards data that appears latter in the batch
+        perm = th.randperm(b_size)
+        batch = batch[perm]
+        # geometric progession of decay: decay^(N-1),...,decay,1
+        weights = th.vander(th.tensor([self.decay]), N=b_size).T
+        alpha = 1 - self.decay
+        if self.count == 0:
+            weights[1:] *= alpha
```
AdamGleave, line 207: Might be worth noting that sum(weights) = 1 after this. Immediate from the sum of finite GP formula, but a bit non-obvious at first.

src/imitation/util/networks.py
```diff
@@ -165,26 +165,67 @@ def update_stats(self, batch: th.Tensor) -> None:
...
+            self.running_mean = th.sum(weights * batch, dim=0)
+            if b_size > 1:
+                self.running_var = th.sum(weights * (batch**2), dim=0)
+                self.running_var -= self.running_mean**2
+        else:
+            running_weight = self.decay * weights[0]
```
AdamGleave, line 213: Math is a bit hard to follow, consider adding comments/docstring describing what this is doing.

tests/util/test_networks.py
```diff
@@ -192,3 +187,41 @@ def test_input_validation_on_ema_norm():
...
+        th.random.manual_seed(42)
+        random_tensor = th.randn(64, 256)
+        for _ in range(100):
+            # calculating EMA of a moving distribution
+            random_tensor += 1
+            norm_for_batch(random_tensor)
+
+        # seeding torch's random module again so that the random permutation
+        # is exactly the same for both incremental and batch method
+        th.random.manual_seed(42)
```
AdamGleave, line 207: Almost identical to code block above -- could factor into helper method and just pass a different method to evaluate (`norm_for_batch` vs `norm_for_incremental.update_stats_incremental`).

AdamGleave:
Thanks for the PR! Changes look reasonable to me at a high level, my suggestions are fairly minor apart and largely to do with improving clarity.

I'm tagging @levmckinney to take a look at this as he wrote the original version. In particular I think we were a bit dissatisfied with details of the old implementation (e.g. the random permutation part is a bit hacky), but went that route largely due to simplicity. It's possible after this refactor there's an alternative that's cleaner. If not then happy to stick with the current semantics (I expect in practice these implementation differences won't really matter anyway).

src/imitation/util/networks.py
taufeeque9: I updated the code for the first batch because of consistency reasons. 

Randomizing the batch before EMA updates did feel weird to me since randomizing counters with what we want to do with EMA (giving more weight to recently seen examples). A better approach could be to assign weights batch-wise. That is, give more weight to recently seen batches (by using the exponential decay) while keeping the same weight for all examples within a batch. For example, if the batch size is N and we have seen 4 batches till now, the weights for all the examples would be (recent examples on right):
(d^3)/N, ..., (d^3)/N, a(d^2)/N, ..., a(d^2)/N, ad/N, ..., ad/N, a/N, ..., a/N (weights sum to 1)
where I'm using ... to represent iteration over a batch of N examples (having equal weight).

This algorithm would be equivalent to doing EMA updates over the batch mean and EMA updates with the batch variance.

If this sounds better to you too, then I can implement this algorithm which wouldn't require randomization within a batch. The code would also be considerably easier to understand.

src/imitation/util/networks.py
taufeeque9: Oh I didn't want to flatten the array. I wrongly assumed that. I actually wanted to do this: 
 if len(batch.shape) == 1: batch = batch.reshape(b_size, 1)

Basically, to ensure that the batch has at least two dimensions (one of the test cases were passing batch containing only one dimension).

I'll fix it.

src/imitation/util/networks.py
levmckinney: > This algorithm would be equivalent to doing EMA updates over the batch mean and EMA updates with the batch variance.

I'm fairly sure this is false or at least not true in all cases. 

First this algorithm is Ill defined when we have a batch size of 1. This is actually quite plausible when we are doing RL training in an expensive environment. Second, I tried to do something along these lines before. If I'm understanding your notation correctly, you're proposing something like the following,

Data $\{x_{k,j}\}$ first index is for batch second index is for element in batch.
Batch size: $b$
Number of batches:  $m$
Decay rate $\gamma$ and learning rate $\alpha = 1 - \gamma$
Let $w_{m, k} = \begin{cases} \frac{\gamma^{m - k}\alpha}{b} , 1 \leq k \leq m \\ \frac{\gamma^m}{b}, k = 0\end{cases}$  These two indices represent the largest batch index and the batch index.



$\mu_{m} = \mathbb E_{m}\left[x\right]$
$\sigma^2_{m} = \mathbb E_{m} [(x - \mu_{bm})^2]$

Now we can show that $\sigma^2_{m} = \mathbb E_{m} [(x - \mu_{bm})^2]$ is not the same as the EMA norm of the sample variances.

Consider the following counter example we have two batches {(0, 0), (1,1)}, $\gamma = \alpha = 1/2$.
Then computing $\sigma^2$ by doing EMA on the _batch sample variances_  gives $\sigma^2_1 = 0$. On the other hand, using the above approach gives $\sigma_{1,1} = \mathbb E_{1,1} [(x - \mu_{11})^2]$. First, we can compute $\mu_{1, 1} = w_{1,0}*0 + w_{1, 0}*0 + w_{1, 1}*1 + w_{1,1}*1 = \gamma\alpha = 1/4$. Thus finally, $\sigma_{11} = \mathbb E_{11} [(x - \mu_{11})^2] = \mathbb E_{11} [(x - 1/4)^2]  = w_{02}*\frac{1}{16} + w_{02}*\frac{1}{16} + w_{12}\frac{9}{16} + w_{12}*\frac{9}{16} = \frac{\gamma}{16} + \frac{\alpha\gamma 9}{16} =  \frac{11}{64}$

levmckinney: Coming back to this again I think I've finally figured out the correct incremental algorithm. I've attached in bellow with a proof of correctness,

[Incremental EMA with batches.pdf](https://github.com/HumanCompatibleAI/imitation/files/9278587/Incremental.EMA.with.batches.pdf)

taufeeque9: Actually, I incorrectly wrote EMA over variance in the last comment. I meant to write that EMA would be equivalent as if we are doing updates on batch mean, and the EMV would be similar to what I have currently implemented in the code (the batch version of original incremental EMV), just that it would become a lot simpler if we take equal weights in a batch.

I looked at the pdf, and that is exactly what I was proposing. Just to confirm, I think in the pseudo-code, the subscript should be (k-1) instead of (m-1) everywhere.

src/imitation/util/networks.py
taufeeque9: Going ahead with the first approach.

src/imitation/util/networks.py
taufeeque9: Added comments and reshuffled some lines to make the code easy to follow.

src/imitation/util/networks.py
levmckinney: Yes it should be k - 1.

levmckinney: This won't be numerically stable https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance.

taufeeque9: If we first multiply the examples by their weights and then compute the sum, that should be numerically stable, right? For simple mean and variance, the weights would be 1/N, where N is the total number of examples.

If the above is true, then the current code should be numerically stable.

levmckinney: I believe this will be better

image

I'll get you the proof for this one soon

levmckinney: Last one because I think it's nice!
image
I'm a bit less confident this one is correct.

AdamGleave:
Thanks for putting the PDF together Lev!

I don't follow the 2nd step in lemma 2, where the denominator switches from
1-alpha to 1-w_{m,m}. I thought w_{m,m} was alpha/b -- so wouldn't these
only agree in the case b=1?

taufeeque9:
@AdamGleave Yes, alpha = b * w_{m,m}. So the denominator should in step 2 should be (1 - b * w_{m, m}). This factor of b would go away in step 5 because the weights sum to 1/b (in the proof this mistake cancels out with the mistake you mentioned).

src/imitation/util/networks.py
levmckinney: Created an overleaf document https://www.overleaf.com/project/62f292899a3ec6a55359da7e

tests/util/test_networks.py
```diff
@@ -189,39 +261,42 @@ def test_input_validation_on_ema_norm():
     networks.EMANorm(128, decay=0.05)
 
 
-def test_ema_norm_batch_correctness():
-    norm_for_incremental = networks.EMANorm(256, 0.5)
-    norm_for_batch = networks.EMANorm(256, 0.5)
-    norm_for_incremental.train(), norm_for_batch.train()
+def run_ema_norm(norm, random_tensor):
```
AdamGleave, line 264: Is this function used anywhere?

tests/util/test_networks.py
```diff
@@ -192,3 +259,44 @@ def test_input_validation_on_ema_norm():
...
+@pytest.mark.parametrize("decay_within_batch", [False, True])
+@pytest.mark.parametrize("decay", [0.5, 0.99])
+def test_ema_norm_batch_correctness(decay_within_batch, decay):
+    norm_for_incremental = EMANormIncremental(
+        num_features=256,
+        decay=decay,
+        decay_within_batch=decay_within_batch,
+    )
+    norm_for_batch = networks.EMANorm(256, decay, decay_within_batch=decay_within_batch)
+    random_tensor = th.randn(64, 256)
```
AdamGleave, line 281: Consider parameterizing the shape? In particular might be useful to try a tensor that has length 1 (just batch dimension), as you've got some logic to special-case that in `EMANorm`, would be good to check it works.

levmckinney:
I've attached the PDF giving the proofs for the update rule: [Incremental_batch_EMA_and_EMV.pdf](https://github.com/HumanCompatibleAI/imitation/files/9364938/Incremental_batch_EMA_and_EMV.pdf)
The latex source code can be found [here](https://gist.github.com/levmckinney/fed0b4feb125ba201301dcf5298c1fe2)

tests/util/test_networks.py
```diff
@@ -12,7 +12,79 @@
...
+                self.running_mean = batch[0]
+                self.running_var = th.zeros_like(self.running_mean, dtype=th.float)
+                start_index = 1
+
+            for i in range(start_index, b_size):
+                diff = batch[i, ...] - self.running_mean
+                incr = alpha * diff
+                self.running_mean += incr
+                self.running_var = self.decay * (self.running_var + diff * incr)
+        else:
```
levmckinney, line 67: I think this would be slightly more intuitive and easier to read using algorithm 3 rather than 2 from the pdf. That's because it basically boils down to using the incremental EMA and EMV update rules on the batch means but adding a correction factor for the batch variance.

levmckinney (commented):
LGTM. I just have a slight preference for using algorithm 3 over 2 since it's a bit easier to intuitively understand.

AdamGleave (approved):
LGTM.

Note I made one small change to add back the identity test.

yawen-d:
I'm a bit not sure what a good way should be to test the correctness or convergence of the EMA running stats. I'm mainly unsure about

1. How to set the default decay rate?

We don't have a way to customize the decay rate of EMANorm yet. See below:
https://github.com/HumanCompatibleAI/imitation/blob/a6c79e9378a9a2d571bcf11c40846c6db86f3b3f/src/imitation/rewards/reward_nets.py#L434

Most likely I think we can directly set a reasonable value and not change it afterwards. The current default decay rate is set to be 0.99, which may be too low.
https://github.com/HumanCompatibleAI/imitation/blob/a6c79e9378a9a2d571bcf11c40846c6db86f3b3f/src/imitation/util/networks.py#L138

Consider an environment of horizon length == 1000 (which is the case for all environments in DM Control), 0.99^1000=0.00004317, so the EMA already ignores early transitions within a single trajectory. I currently set a default decay=0.9997 based on the intuition to put weights on the past 10 trajectories because 0.9997^(1000*10)=0.04976466. This isn't precisely how EMA works, but it gives a general idea of how much weights we set on the past trajectories.
Screenshot 2022-08-23 at 15 48 15

2. How to set reasonable values of `rtol` and `atol` in `th.testing.assert_close()`?
For example, if I change the default decay rate to 0.9997 (see ), some of the current tests (e.g. `test_parameters_converge`) won't pass, although this could be alleviated by 
 increasing `num_samples`.

In summary, there are two decision-relevant takeaways:

1. How do we set the default decay rate of EMANorm?
    - My current guess is to set 0.9997 based on my intuition, but still lack a more principled way of setting this.
3. How do we design the tests in `tests/util/test_networks.py` when the results could be sensitive to the decay rate? (e.g. test_running_norm_eval_fixed; test_parameters_converge; test_ema_norm_batch_correctness)
    - My current guess is to set reasonable stats, such as
        - Set `num_samples` to be >1000 instead of 100
        - Perhaps no need to test unreasonable decay rates (e.g. 0.5)
        - Not sure how to set `rtol` and `atol`

levmckinney (approved):
Algorithm 3 looks correct to me

taufeeque9:
Thanks for the analysis Yawen! I also had my doubts regarding setting the default decay rate to 0.99. However, I think it is a reasonable decay rate, and any rate above 0.99 might not be suitable.

In the code @yawen-d referenced, there's a slight problem. The weights should also be multiplied by $\alpha$ (which is 1 - $\gamma$) for all the elements except the first one. So the weights for EMA will be (for N+1 examples): $\alpha, \alpha \cdot \gamma, \alpha \cdot \gamma^2, ..., \alpha \cdot \gamma ^ {N-1}, \gamma ^ N$. All these weights sums to $1$ for all $N>0$. So the correct code for plotting the weights is:
```python
num_timesteps = 1000
x = np.linspace(0, num_timesteps, num_timesteps+1)
for decay in decay_list:
    y = (decay**x)
    y[:-1] *= 1-decay
    assert abs(y.sum()-1) < 1e-6, "weights (y) should sum to 1."
    plt.plot(x, y, label=f'{decay}')
```
With the above code, we get the following plots
 


As can be seen from the first plot, even a decay rate of 0.999 gives a very high weight to the first example seen by EMANorm. We can also check that if decay = 1, then all the weight is given to the first example, and all the subsequent examples are not given any weight (since alpha becomes 0).  Therefore decay values extremely close to 1 are undesirable. This is the reason why the tests in #538 are failing since the default decay of 0.9997 is very high.

From the second plot, we can also see that decay rates close to or lesser than 0.9 are also not good since they give significantly lesser weight to previously seen examples. Therefore, weights in the range 0.97-0.992 seem reasonable for horizon lengths of 1000.

taufeeque9:
For testing the correctness of `EMANorm`, I implemented Algorithm 1 from [Incremental_batch_EMA_and_EMV.pdf](https://github.com/HumanCompatibleAI/imitation/files/9364938/Incremental_batch_EMA_and_EMV.pdf) in the class `EMANormIncremental`. However, since Algorithm 1 is numerically unstable, the `test_ema_norm_batch_correctness` fails for decay = 0.5 and num_samples = 1000 as shown by @yawen-d in #538. So I have now implemented Algorithm 2 in `EMANormIncremental` to test the implementation of Algorithm 3 used in `EMANorm`.

AdamGleave (approved):
Change LGTM. I agree that testing algorithm 3 with algorithm 2 is not ideal given their similarity (could have overlapping bugs) but given we also have the identity_train check this seems OK.

On the decay rate, we want there to be weight on things from many episodes back I think (normalizing reward within an episode seems too fast), but that probably seems better discussed in detail in Yawen's PR.

yawen-d:
Thanks for the response! I would be happy to continue this discussion in this issue #540.

AdamGleave:
Ah, I guess this error gets "canceled out" later on: the weights sum to 1/b
not 1, so if we just change the 2nd step to be 1 - b*w_{m,m} then
everything works out.

Issue #526, opened 2022-08-15
Remove unused ` log_root` in `imitation.scripts.common`?


yawen-d:
## Problem

In `imitation.scripts.common`, the user is expected to specify a directory path to store the training statistics or artifacts by `common.log_root` or `common.log_dir`.

It seems that `log_root` is [not used anywhere](https://github.com/HumanCompatibleAI/imitation/search?q=log_root) in the source code except for `experiment/*.sh` and `tests/scripts/test_scripts.py`.

In [src/imitation/scripts/common/common.py](https://github.com/HumanCompatibleAI/imitation/blob/b8a4c369b3d6ee861ac84618dab1adcbaa537b26/src/imitation/scripts/common/common.py), `log_root` is also not used by any captured function.

https://github.com/HumanCompatibleAI/imitation/blob/41dafea0369a83894cde248fd649292be593638c/src/imitation/scripts/common/common.py#L22

This could potentially confuse the users and store unnecessary test data in the default paths during unit tests.

## Proposed Solution

- Remove `common.log_root` from `common`.
- Replace `log_root` with `log_dir` in
    - [experiments/dagger_benchmark.sh](https://github.com/HumanCompatibleAI/imitation/blob/47abd983dc4879b148d8b574226cab7ba0ea92ab/experiments/dagger_benchmark.sh), 
    - [experiments/imit_benchmark.sh](https://github.com/HumanCompatibleAI/imitation/blob/47abd983dc4879b148d8b574226cab7ba0ea92ab/experiments/imit_benchmark.sh), 
    - [experiments/transfer_learn_benchmark.sh](https://github.com/HumanCompatibleAI/imitation/blob/8efa8011545f00a86ee23f9e187f0db4a4367c8c/experiments/transfer_learn_benchmark.sh)
    - [tests/scripts/test_scripts.py](https://github.com/HumanCompatibleAI/imitation/blob/41dafea0369a83894cde248fd649292be593638c/tests/scripts/test_scripts.py)
    - [experiments/bc_benchmark.sh](https://github.com/HumanCompatibleAI/imitation/blob/47abd983dc4879b148d8b574226cab7ba0ea92ab/experiments/bc_benchmark.sh)
    - [experiments/rollouts_from_policies.sh](https://github.com/HumanCompatibleAI/imitation/blob/2ea8aaa025d22ec0e5c7cea337a3fff9b3117d9f/experiments/rollouts_from_policies.sh)
[...]

PR #527, opened 2022-08-15
Specify `log_root` in `imitation.scripts.common` to replace default `log_root`

yawen-d:
## Description

Address #526.

## Testing

Existing tests already suffice.

### DIFF
--- a/src/imitation/scripts/common/common.py
+++ b/src/imitation/scripts/common/common.py
@@ -49,7 +49,7 @@ def hook(config, command_name, logger):
     if config["common"]["log_dir"] is None:
         env_sanitized = config["common"]["env_name"].replace("/", "_")
         log_dir = os.path.join(
-            "output",
+            config["common"]["log_root"] or "output",
             command_name,
             env_sanitized,
             util.make_unique_timestamp(),
--- a/src/imitation/scripts/common/common.py
+++ b/src/imitation/scripts/common/common.py
@@ -48,8 +48,9 @@ def hook(config, command_name, logger):
     updates = {}
     if config["common"]["log_dir"] is None:
         env_sanitized = config["common"]["env_name"].replace("/", "_")
+        log_root = config["common"]["log_root"] or "output"
         log_dir = os.path.join(
-            config["common"]["log_root"] or "output",
+            log_root,
             command_name,
             env_sanitized,
             util.make_unique_timestamp(),
--- a/src/imitation/scripts/common/common.py
+++ b/src/imitation/scripts/common/common.py
@@ -1,5 +1,6 @@
 """Common configuration elements for scripts."""
 
+import contextlib
 import logging
 import os
 from typing import Any, Mapping, Sequence, Tuple, Union
@@ -126,6 +127,7 @@ def setup_logging(
     return custom_logger, log_dir
 
 
+@contextlib.contextmanager
 @common_ingredient.capture
 def make_venv(
     _seed,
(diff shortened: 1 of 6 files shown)


src/imitation/scripts/common/common.py
```diff
@@ -49,7 +49,7 @@ def hook(config, command_name, logger):
     if config["common"]["log_dir"] is None:
         env_sanitized = config["common"]["env_name"].replace("/", "_")
         log_dir = os.path.join(
-            "output",
+            config["common"]["log_root"] or "output",
```
AdamGleave, line 52: Should maybe define this as a separate variable since the logic is a little complex?

yawen-d: yep

AdamGleave:
Tests are now failing on Windows because `TemporaryDirectory` used in https://github.com/HumanCompatibleAI/imitation/blob/yawen-d/fix/add-log_root/tests/scripts/test_scripts.py#L830 tries to remove the directory after the script returns, but there's still an open file handle to Monitor. This isn't a problem on Unix that allows files to be unlinked that are still open by a process, but is a problem on Windows. The file handle is still open because we never actually call `venv.close()`. I've modified the scripts to close this using a `try:`, `finally:` block similar to what we were already doing in `eval_policy`.

Worried what I've done is a bit overkill, it is annoying that most of the logic is now in a `try:` indent block, we could also just have `venv.close()` outside `try-finally` and accept it won't get cleaned up on exceptions.

Rocamonde:
> Worried what I've done is a bit overkill, it is annoying that most of the logic is now in a `try:` indent block, we could also just have `venv.close()` outside `try-finally` and accept it won't get cleaned up on exceptions.

Another option is to create a context manager that automatically closes the environment and then enter it using a `with` statement. If this is done in many places across the repo, it could be a good use case for abstracting it out.

src/imitation/scripts/train_rl.py
```diff
@@ -96,55 +96,66 @@ def train_rl(
...
+                rollout_save_n_timesteps,
+                rollout_save_n_episodes,
+            )
+            types.save(save_path, rollout.rollout(rl_algo, venv, sample_until))
+        if policy_save_final:
+            output_dir = os.path.join(policy_dir, "final")
+            serialize.save_stable_model(output_dir, rl_algo)
+
+        # Final evaluation of expert policy.
+        result = train.eval_policy(rl_algo, venv)
```
Rocamonde, line 154: This is not necessary, you can return, and the finally still gets called anyways. (e.g. https://stackoverflow.com/questions/19805654/python-try-finally-block-returns)

Rocamonde (commented):
LGTM, apart from the minor style suggestion.

Rocamonde (approved):

AdamGleave:
Good suggestion to use context manager instead. I've changed `common.make_venv` to be a context manager and updated the scripts to use `with` instead of `try:`, `finally:`.

I debated whether to turn `util.make_vec_env` (which `common.make_venv` is a thin wrapper around) into a context manager but ultimately decided against. Mandating use of context manager seemed like it constrains end-user code more than we really want. An alternative would be to have `util.make_vec_env` return an object that can be used directly or via a context-manager, ala `open`, but I think this would require some hacky wrapping of `VecEnv` objects to add `__enter__` and `__exit__` methods -- probably not worth it.

Rocamonde (approved):
LGTM.

src/imitation/scripts/train_imitation.py
```diff
@@ -119,52 +119,55 @@ def train_imitation(
...
+            )
+            # TODO(adam): add checkpointing to DAgger?
+            save_locations = model.save_trainer()
+            print(f"Model saved to {save_locations}")
+        else:
+            bc_trainer.train(**bc_train_kwargs)
+            # TODO(adam): add checkpointing to BC?
+            bc_trainer.save_policy(policy_path=osp.join(log_dir, "final.th"))
+
+        imit_stats = train.eval_policy(imit_policy, venv)
```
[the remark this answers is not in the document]
AdamGleave: If I was just doing `return imit_stats` in the next line I'd agree. But we also have code computing stats from expert rollouts that's outside the context manager (but part of return value). I left it outside to minimize the surface of the code inside the context manager. This doesn't particularly matter, I can see an argument for the other way, but it seems nice to draw a distinction between code that does vs doesn't need the venv as a form of self-documentation.

PR #528, opened 2022-08-16
Fix dev versioning system to allow test PyPI releases

### LINKED ISSUE #500: Fix test PyPI automation

Rocamonde:
Fixes #500.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,10 +2,10 @@
 requires = ["setuptools>=45", "setuptools_scm[toml]>=6.2"]
 build-backend = "setuptools.build_meta"
 
-[tools.setuptools_scm]
-# Disable local scheme to allow uploads to Test PyPI.
-# See https://github.com/pypa/setuptools_scm/issues/342
-local_scheme = "no-local-version"
+# [tool.setuptools_scm]
+# # Disable local scheme to allow uploads to Test PyPI.
+# # See https://github.com/pypa/setuptools_scm/issues/342
+# local_scheme = "imitation.version:get_version"
 
 [tool.black]
 target-version = ["py38"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,10 @@
 requires = ["setuptools>=45", "setuptools_scm[toml]>=6.2"]
 build-backend = "setuptools.build_meta"
 
+# TODO(juan): we've commented this out because currently
+# there's no way (that we could find in the docs) to point
+# the local_scheme and version_scheme to a function from the
+# pyproject.toml file, so setup.py has to be used.
 # [tool.setuptools_scm]
 # # Disable local scheme to allow uploads to Test PyPI.
 # # See https://github.com/pypa/setuptools_scm/issues/342
--- a/setup.py
+++ b/setup.py
@@ -157,6 +157,14 @@ def get_local_version(version: "ScmVersion", time_format="%Y%m%d") -> str:
 
     Local versions are still present if there are uncommitted changes (if the tree
     is dirty), in which case the current date is added to the version.
+
+    Args:
+        version: The version object given by setuptools_scm, calculated
+            from the git repository.
+        time_format: The format to use for the date.
+
+    Returns:
+        The formatted local version string to use for the package.
     """
     return version.format_choice(
         "",
--- a/setup.py
+++ b/setup.py
@@ -3,10 +3,13 @@
 import os
 import warnings
 from sys import platform
+from typing import TYPE_CHECKING
 
 from setuptools import find_packages, setup
 from setuptools.command.install import install
-from setuptools_scm import version as scm_version
+
+if TYPE_CHECKING:
+    from setuptools_scm.version import ScmVersion
 
 IS_NOT_WINDOWS = os.name != "nt"
 
@@ -56,6 +59,7 @@ TESTS_REQUIRE = (
         "pytest-xdist~=2.5.0",
         "scipy~=1.9.0",
         "wandb==0.12.21",
+        "setuptools_scm~=7.0.5",
     ]
     + PARALLEL_REQUIRE
     + PYTYPE
@@ -94,7 +98,9 @@ class InstallCommand(install):
             )
 
 
-def get_version(version: scm_version.ScmVersion) -> str:
+def get_version(version: "ScmVersion") -> str:
+    from setuptools_scm import version as scm_version
+
     if version.node:
         version.node = str(int(version.node.lstrip("g"), 16))
     if version.exact:
@@ -106,7 +112,7 @@ def get_version(version: scm_version.ScmVersion) -> str:
         )
 
 
-def get_local_version(version: scm_version.ScmVersion, time_format="%Y%m%d") -> str:
+def get_local_version(version: "ScmVersion", time_format="%Y%m%d") -> str:
     return version.format_choice(
         "",
         "+d{time:{time_format}}",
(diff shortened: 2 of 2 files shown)


Rocamonde:
Hooray! https://test.pypi.org/project/imitation/0.3.2.dev30933807/

Rocamonde:
Damn, the `setuptools_scm` library needs to be imported during setup, but the function is defined at the `setup.py` module level, so it gets imported before the installer installs the library, giving an error. Seems overkill to add this as a dependency

AdamGleave:
Is it just a type error? If so could add `setuptools_scm` as a test-only dependency.

Rocamonde:
Actually it's also needed to call the default version guessing script, even though we could just copy this manually from source... This library feels poorly designed.

Rocamonde:
Oh wait, I can just only import the library if we are type checking, then import it within the function scope.

setup.py
```diff
@@ -93,12 +98,34 @@ def run(self):
             )
 
 
+def get_version(version: "ScmVersion") -> str:
+    from setuptools_scm import version as scm_version
```
AdamGleave, line 102: We normally avoid imports inside methods. You have a good reason to do this, but can you add a comment to explain why?

setup.py
```diff
@@ -93,12 +98,34 @@ def run(self):
             )
 
 
+def get_version(version: "ScmVersion") -> str:
```
AdamGleave, line 101: Add docstring explaining format of version?

setup.py
```diff
@@ -93,12 +98,34 @@ def run(self):
...
+    if version.exact:
+        return version.format_with("{tag}")
+    else:
+        return version.format_next_version(
+            scm_version.guess_next_version,
+            fmt="{guessed}.dev{node}",
+        )
+
+
+def get_local_version(version: "ScmVersion", time_format="%Y%m%d") -> str:
```
AdamGleave, line 115: Add docstring explaining what this does. (Was non-obvious to me without looking up format_choice; I see this is something added if working directory is dirty.)

setup.py
```diff
@@ -93,12 +98,34 @@ def run(self):
...
+        "",
+        "+d{time:{time_format}}",
+        time_format=time_format,
+    )
+
+
 setup(
     cmdclass={"install": InstallCommand},
     name="imitation",
     # Disable local scheme to allow uploads to Test PyPI.
```
AdamGleave, line 126: This comment is no longer accurate!

AdamGleave (changes requested):
Thanks for fixing this! Happy with the solution at a high-level, few minor requested changes.

setup.py
```diff
@@ -93,12 +98,34 @@ def run(self):
             )
 
 
+def get_version(version: "ScmVersion") -> str:
+    from setuptools_scm import version as scm_version
+
+    if version.node:
+        version.node = str(int(version.node.lstrip("g"), 16))
```
AdamGleave, line 105: Where's the g coming from (G for git?)

pyproject.toml
```diff
@@ -2,10 +2,10 @@
 requires = ["setuptools>=45", "setuptools_scm[toml]>=6.2"]
 build-backend = "setuptools.build_meta"
 
-[tools.setuptools_scm]
-# Disable local scheme to allow uploads to Test PyPI.
-# See https://github.com/pypa/setuptools_scm/issues/342
-local_scheme = "no-local-version"
+# [tool.setuptools_scm]
```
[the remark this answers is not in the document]
Rocamonde: Yes, that's right. We should fix this at some point, but I don't see this happening in the near future unless the setuptools_scm decided to support our feature request.

setup.py
Rocamonde: Yes, it appends "g" before the first few characters of the commit hash if it comes from git (and does something else if using mercurial, ugh).

setup.py
Rocamonde: Apologies, will document. Spent so long reading the docs for this thing that forgot other people don't actually know how this works 😂

AdamGleave:
Linter is complaining about docstring format -- ping me when that's fixed and I'll do another review.

Rocamonde:
Fixed the linting issues, but now getting a separate error which is certainly unrelated as I have not touched code that affects tests in this way:

```TypeError: run() got an unexpected keyword argument 'loggers'```

on `./imitation/scripts/parallel.py:120`:


```python
ray.tune.run(
                trainable,
                config=search_space,
                name=run_name,
                local_dir=local_dir,
                loggers=ray_loggers,
                resources_per_trial=resources_per_trial,
                sync_config=ray.tune.syncer.SyncConfig(upload_dir=upload_dir),
            )
```

Rocamonde:
Yes, they just released ray 2.0.0. 10 hours ago. I'll pin the version and create an issue.

Rocamonde:
Getting these errors on CircleCI which do not show up on my linting:

```
setup.py:147:1: DAR201 Missing "Returns" in Docstring: - return
setup.py:147:1: DAR101 Missing parameter(s) in Docstring: - time_format
setup.py:147:1: DAR101 Missing parameter(s) in Docstring: - version
```

Will fix these now, but any thoughts on the reason behind inconsistent error reporting?

Rocamonde:
Getting errors on CircleCI that are noticed inline by vscode but that do not get pointed out when running `./ci/code_checks.sh`. Any thoughts on why?

```
setup.py:147:1: DAR201 Missing "Returns" in Docstring: - return
setup.py:147:1: DAR101 Missing parameter(s) in Docstring: - time_format
setup.py:147:1: DAR101 Missing parameter(s) in Docstring: - version
```

AdamGleave:
> Getting errors on CircleCI that are noticed inline by vscode but that do not get pointed out when running `./ci/code_checks.sh`. Any thoughts on why? This is a pretty annoying issue, as I can't really go through every file visually, plus it breaks the principle of having local CI checks before pushing.
> 
> ```
> setup.py:147:1: DAR201 Missing "Returns" in Docstring: - return
> setup.py:147:1: DAR101 Missing parameter(s) in Docstring: - time_format
> setup.py:147:1: DAR101 Missing parameter(s) in Docstring: - version
> ```

Do you have `flake8-darglint` installed? If you run `flake8 setup.py` directly does it replicate the error?

The only difference I'm aware of between CI and ci/code_checks.sh is that we have some logic in `ci/code_checks.sh` to only run darglint on files that were changed in a commit (as running darglint on everything is very slow; OK on CI, but not really on each commit hook).

AdamGleave (approved):
LGTM. I made some cosmetic changes so please review before merging.

PR #529, opened 2022-08-17
Implemented the ability to train rewards in preference comparison against multiple policies 

levmckinney:
## Description

This pull request adds the `MixtureOfTrajectoryGenerators` class which allows preference comparison to be trained on data from multiple sources. This includes the ability to train against multiple agents and train against a mixture of data from an agent and a dataset.

To facilitate this I have made modifications to `HeirachicalLogger` class allowing for arbitrary prefixes to be added so that different trajectory generators logs don't clobber each other.

## Testing

Changes to the `HeirachicalLogger` class have been unit tested.  `MixtureOfTrajectoryGenerators` has also been united test. I've also included new integration tests for the `train_preference_comparison.py` script to test the use case of having multiple policies.


AdamGleave:
@levmckinney let me know if there's any particular areas you'd like my feedback on, I'll aim to try and do a high-level review but may not have time to look at it in detail.

@Rocamonde can you do a style/technical review on Monday?

src/imitation/util/logger.py
```diff
@@ -108,10 +127,12 @@ def accumulate_means(self, subdir: types.AnyPath) -> Generator[None, None, None]
         if self.current_logger is not None:
             raise RuntimeError("Nested `accumulate_means` context")
 
+        name = types.path_to_str(name)
```
Rocamonde, line 130: This doesn't really check that name is a path if the input type is a string. Why aren't we using Pathlib?

src/imitation/util/logger.py
```diff
@@ -108,10 +127,12 @@ def accumulate_means(self, subdir: types.AnyPath) -> Generator[None, None, None]
...
+        subdir = os.path.join(*self._prefixes, name)
```
Rocamonde, line 131: Again, pathlib is the reccommended way to handle this IMO (I'm sure there's a ton of other places where os.path is used and if we were to use pathlib more generally it should probably be addressed in a separate PR.)

src/imitation/util/logger.py
```diff
@@ -72,27 +74,44 @@ def _update_name_to_maps(self) -> None:
...
+    def accumulate_means(self, name: types.AnyPath) -> Generator[None, None, None]:
         """Temporarily modifies this HierarchicalLogger to accumulate means values.
 
         During this context, `self.record(key, value)` writes the "raw" values in
-        "{self.default_logger.log_dir}/{subdir}" under the key "raw/{subdir}/{key}".
-        At the same time, any call to `self.record` will also accumulate mean values
-        on the default logger by calling
-        `self.default_logger.record_mean(f"mean/{subdir}/{key}", value)`.
+        "{self.default_logger.log_dir}/{prefix}/{name}" under the key
+        "raw/{prefix}/{name}/{key}". At the same time, any call to `self.record` will
```
Rocamonde, line 98: It is my understanding based on your code that multiple prefixes might be used at a given time, and that they are composed. If that is the case, I find that this is not very clear/consistent with the docstring

src/imitation/util/logger.py
```diff
@@ -72,27 +74,44 @@ def _update_name_to_maps(self) -> None:
         self.name_to_excluded = self._logger.name_to_excluded
 
     @contextlib.contextmanager
-    def accumulate_means(self, subdir: types.AnyPath) -> Generator[None, None, None]:
+    def add_prefix(self, prefix: str) -> Generator[None, None, None]:
```
Rocamonde, line 77: If the prefix is removed after leaving the context, and only one prefix is supported as input, why are you adding support for a list of prefixes? How and why would I use multiple prefixes? (Is the expectation that I should enter nested prefix contexts, as that might be quite hard to read and understand in practice, e.g. if this happens in different files or function calls.) I also think this prefix/name idea should be explained with an examlpe in e.g. the class-level docstring.

Rocamonde:
When reading through the logging code, I see we are maintaining a somewhat advanced and general logging infrastructure for hierarchical logging. I'm surprised there's not an already existing way to handle this without a bespoke implementation. What are other people doing? (i.e. in similar ML projects) Are our logging needs very specific and different to the way this is normally done? It might be interesting to spin off the logging module into e.g. an SB3 PR, so people don't have to reinvent the wheel every time.

src/imitation/util/logger.py
Rocamonde: If I'm understanding correctly how this is supposed to be used, entering this should be disallowed if one is using an `accumulate_means` context, otherwise that would mess with the path where the rest of the logs are being recorded. Do you agree and if so do you think you can add a way to throw an error?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1283,9 +1363,8 @@ def train(
             # at the end of training (where the reward model is presumably best)
             if i == self.num_iterations - 1:
                 num_steps += extra_timesteps
-            with self.logger.accumulate_means("agent"):
-                self.logger.log(f"Training agent for {num_steps} timesteps")
-                self.trajectory_generator.train(steps=num_steps)
+            self.logger.log(f"Training agent for {num_steps} timesteps")
```
Rocamonde, line 1366: If I understand correctly, entering `accumulate_means` is moved off to the trajectory generator, and calls to `log` do not use the prefix/name system as it's not a recorded quantity. Is this correct?

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -155,22 +157,43 @@ def train_preference_comparisons(
             relabel_reward_fn=relabel_reward_fn,
         )
 
-    if trajectory_path is None:
+    if num_agents < 1:
```
Rocamonde, line 160: Once you're at it, you might as well check it's an integer. No reason to handle custom error case for range but then let a "more obscure" error arise if float.

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -155,22 +157,43 @@ def train_preference_comparisons(
...
+        raise ValueError("num_agents must be at least 1!")
+
+    def make_agent_trainer(seed: Optional[int] = None):
```
Rocamonde, line 163: Now lines 151-158 aren't being used?

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -155,22 +157,43 @@ def train_preference_comparisons(
...
         )
+
+    if trajectory_path is None and num_agents == 1:
+        trajectory_generator = make_agent_trainer()
         # Stable Baselines will automatically occupy GPU 0 if it is available. Let's use
         # the same device as the SB3 agent for the reward model.
         reward_net = reward_net.to(trajectory_generator.algorithm.device)
+        allow_save_policy = True
+    elif trajectory_path is None and num_agents > 1:
+        members = [make_agent_trainer(_seed * i) for i in range(num_agents)]
```
Rocamonde, line 188: I suppose this is a way to generate pseudo-iid copies of the agent trainer. Might be a silly observation, but what if the seed is zero? Unlikely that someone would set that, but if they did they would start getting very bizarre errors and results without an apparent cause.

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -155,22 +157,43 @@ def train_preference_comparisons(
...
+        trajectory_generator = preference_comparisons.MixtureOfTrajectoryGenerators(
+            members=members,
+            custom_logger=custom_logger,
+        )
+        reward_net = reward_net.to(members[0].algorithm.device)
+        allow_save_policy = False
```
Rocamonde, line 194: I would move setting this flag and also line 186 just below the if/elif/else clause (as is the case in line 196)

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -299,6 +302,83 @@ def logger(self, value: imit_logger.HierarchicalLogger):
...
+        Args:
+            members: Individual trajectory generators that will make up the ensemble.
+            custom_logger: Custom logger passed to super class.
+
+        Raises:
+            ValueError: if members is empty.
+        """
+        if len(members) == 0:
+            raise ValueError(
+                "MixtureOfTrajectoryGenerators requires at least one member!",
```
Rocamonde, line 326: Surely at least two members? Not too important, but at least throw a warning, as that's likely a mistake on the usage side.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -299,6 +302,83 @@ def logger(self, value: imit_logger.HierarchicalLogger):
...
+
+        Returns:
+            A list of sampled trajectories with rewards (which should
+            be the environment rewards, not ones from a reward model).
+        """
+        n = len(self.members)
+        # Approximately evenly partition work.
+        d = steps // n
+        r = steps % n
+        i = np.random.randint(n)
```
Rocamonde, line 346: You're not setting the seed for this / using a RandomState so this it not reproducible. Maybe that is important

Rocamonde (commented):
Some initial comments. Have not looked at the tests yet. Let me know what you think and when/if you implement any changes from my suggestions, and I can take another look.

src/imitation/util/logger.py
```diff
@@ -72,27 +74,44 @@ def _update_name_to_maps(self) -> None:
...
-        "{self.default_logger.log_dir}/{subdir}" under the key "raw/{subdir}/{key}".
-        At the same time, any call to `self.record` will also accumulate mean values
-        on the default logger by calling
-        `self.default_logger.record_mean(f"mean/{subdir}/{key}", value)`.
+        "{self.default_logger.log_dir}/{prefix}/{name}" under the key
+        "raw/{prefix}/{name}/{key}". At the same time, any call to `self.record` will
+        also accumulate mean values on the default logger by calling
+        `self.default_logger.record_mean(f"mean/{prefix}/{name}/{key}", value)`.
 
         During the context, `self.record(key, value)` will write the "raw" values in
```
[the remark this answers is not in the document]
levmckinney: I've clarified the doc string.

src/imitation/util/logger.py
levmckinney: Good point switched name to a string and opened an issue regarding path lib.

src/imitation/util/logger.py
levmckinney: Issue opened regarding path lib.

src/imitation/util/logger.py
levmckinney: I've clarified this in the updates to the doc string.

src/imitation/util/logger.py
levmckinney: > If I'm understanding correctly how this is supposed to be used, entering this should be disallowed if one is using an `accumulate_means` context, otherwise that would mess with the path where the rest of the logs are being recorded. Do you agree and if so do you think you can add a way to throw an error?

Agreed, I've added a runtime error for this case.

src/imitation/algorithms/preference_comparisons.py
levmckinney: That is correct.

src/imitation/scripts/train_preference_comparisons.py
levmckinney: Good catch! I guess there was a merge here that slipped by me.

src/imitation/scripts/train_preference_comparisons.py
levmckinney: Good point made it additive instead.

src/imitation/algorithms/preference_comparisons.py
levmckinney: I could imagine valid use cases, but they seem against the spirit of the class. I'll increase this.

src/imitation/algorithms/preference_comparisons.py
levmckinney: I've just removed the RNG entirely since I don't think it's really necessary.

AdamGleave:
> When reading through the logging code, I see we are maintaining a somewhat advanced and general logging infrastructure for hierarchical logging. I'm surprised there's not an already existing way to handle this without a bespoke implementation. What are other people doing? (i.e. in similar ML projects) Are our logging needs very specific and different to the way this is normally done? It might be interesting to spin off the logging module into e.g. an SB3 PR, so people don't have to reinvent the wheel every time.

What we are doing is kind of weird -- you don't normally have several RL trainers that all need to log to the same place -- but it could come up in some other applications, e.g. population-based training. You could look at how `rllib` handles this.

src/imitation/util/logger.py
levmckinney: I've added a doctest documenting how to use it.

levmckinney:
> @levmckinney let me know if there's any particular areas you'd like my feedback on, I'll aim to try and do a high-level review but may not have time to look at it in detail.

@AdamGleave Could you look at the changes to the HierarchicalLogger?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -299,6 +302,95 @@ def logger(self, value: imit_logger.HierarchicalLogger):
...
+class MixtureOfTrajectoryGenerators(TrajectoryGenerator):
+    """A collection of trajectory generators merged together."""
+
+    members: Sequence[TrajectoryGenerator]
+
+    def __init__(
+        self,
+        members: Sequence[TrajectoryGenerator],
+        custom_logger: Optional[imit_logger.HierarchicalLogger] = None,
+        share_training_steps: bool = True,
```
levmckinney, line 314: @AdamGleave do you think its more natural to have this default to True or False?

AdamGleave: Sorry for slow response, somehow missed this question. My inclination is to default to True but this is weakly held.

src/imitation/util/logger.py
```diff
@@ -41,6 +41,49 @@ class HierarchicalLogger(sb_logger.Logger):
...
+    ...     # record the key value pair (loss, 1.0) to path `dir`
+    ...     # at step 1.
+    ...     logger.record("loss", 1.0)
+    ...     logger.dump(step=1)
+    ...     with logger.accumulate_means("dataset"):
+    ...         # record the key value pair `("raw/dataset/entropy", 5.0)` to path
+    ...         # `dir/raw/dataset` at step 100
+    ...         logger.record("entropy", 5.0)
+    ...         logger.dump(step=100)
+    ...         # record the key value pair `("raw/dataset/entropy", 5.0)` to path
```
AdamGleave, line 57:
```suggestion
    ...         # record the key value pair `("raw/dataset/entropy", 6.0)` to path
```

AdamGleave (commented):
The changes to `HiearchicalLogger` LGTM -- I have not reviewed the other files.

AdamGleave:
@Rocamonde I think Lev requested re-review, could you take another look soon? Thanks!

Rocamonde (changes requested):

Rocamonde:
Left some comments. Also, I'm getting some typing errors locally, specifically on https://github.com/HumanCompatibleAI/imitation/blob/07fa8305f3dbc74f7db95cde051273012456fc5f/src/imitation/scripts/train_preference_comparisons.py#L271-L271

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -146,31 +152,52 @@ def train_preference_comparisons(
...
+
+        if trajectory_path is None and num_agents == 1:
+            trajectory_generator = make_agent_trainer()
             # Stable Baselines will automatically occupy GPU 0 if it is available.
             # Let's use the same device as the SB3 agent for the reward model.
             reward_net = reward_net.to(trajectory_generator.algorithm.device)
-        else:
+        elif trajectory_path is None and num_agents > 1:
+            members = [make_agent_trainer(_seed + i) for i in range(num_agents)]
+            trajectory_generator = preference_comparisons.MixtureOfTrajectoryGenerators(
```
Rocamonde, line 193: This variable is going unused because it's being redefined in https://github.com/HumanCompatibleAI/imitation/blob/07fa8305f3dbc74f7db95cde051273012456fc5f/src/imitation/scripts/train_preference_comparisons.py#L205-L205 An error in the type checker in https://github.com/HumanCompatibleAI/imitation/blob/07fa8305f3dbc74f7db95cde051273012456fc5f/src/imitation/scripts/train_preference_comparisons.py#L271-L271
is likely related to this issue.

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -133,6 +134,9 @@ def train_preference_comparisons(
             be allocated to each iteration. "hyperbolic" and "inverse_quadratic"
             apportion fewer queries to later iterations when the policy is assumed
             to be better and more stable.
+        share_training_steps_among_agents: If True (default), when training with
```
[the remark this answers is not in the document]
levmckinney: I made this the default behavior for the following reason. Often in RL research papers we want to have a specific budget of environment interactions. In this context, I felt having the default behavior respect that budget would make the most sense. However, I do agree, this could be handled at the level of the scripts and reduce the complexity of the business logic a bit.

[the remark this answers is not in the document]
Rocamonde: That makes sense, thanks for your clarification. I just figured that for a case of e.g. 7 agents, 2113 steps instead of 2114 (a multiple of 7) would not really make a difference, but maybe it does to some users.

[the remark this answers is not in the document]
levmckinney: > I just figured that for a case of e.g. 7 agents, 2113 steps instead of 2114 (a multiple of 7) would not really make a difference, but maybe it does to some users.

I don't think this would make a difference to most users. I'm mainly concerned with what the _default_ behavior. Furthermore, I believe it should be to split training steps equally amongst the agents. Thus, the `num_steps` records the actually budget of environment interactions across all agents.

We could add this in the script level. However, I suspect that down stream users of the core API will also just end up implementing something similar to `num_steps//num_agents` themselves.

Rocamonde (approved):
The new changes LGTM. Do complete the docstring (see comment) before merging. Others might want to have a chance to take a look too before then.

src/imitation/util/logger.py
```diff
@@ -41,8 +41,57 @@ class HierarchicalLogger(sb_logger.Logger):
...
+    ...         logger.dump(step=200)
+    ...     # record the key value pair `("mean/dataset/entropy", 5.5)` to path
+    ...     # `dir` at step 1.
+    ...     logger.dump(step=1)
+    ...     with logger.add_prefix("foo"), logger.accumulate_means("bar"):
+    ...         # record the key value pair ("raw/foo/bar/entropy", 42.0) to path
+    ...         # `dir/raw/foo/bar` at step 2000
+    ...         logger.record("biz", 42.0)
+    ...         logger.dump(step=2000)
+    ...     # record the key value pair `("mean/foo/bar/entropy", 42.0)` to path
```
taufeeque9, line 69: The key in the code is "biz" whereas the comments imply that the key should be "entropy".

src/imitation/scripts/train_preference_comparisons.py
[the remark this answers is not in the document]
AdamGleave: Having default behavior preserve total training timesteps budget sems good to me. Integer division `//` vs something more clever that handles remainders seems unimportant -- we usually have hundreds of thousands of timesteps and single-digit numbers of agents.

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -155,31 +161,55 @@ def train_preference_comparisons(
...
+            raise ValueError("num_agents must be a positive integer!")
+
+        def make_agent_trainer(seed: Optional[int] = None):
+            if agent_path is None:
+                agent = rl_common.make_rl_algo(
+                    venv,
+                    relabel_reward_fn=relabel_reward_fn,
+                )
+            else:
+                agent = rl_common.load_rl_algo_from_path(
```
AdamGleave, line 175: There'll be hilariously little diversity between agents in this case, but not much we can do there. (Support loading *different* agents I guess? But that seems overkill for what's a rare use case.)

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -155,31 +161,55 @@ def train_preference_comparisons(
...
+            members = [make_agent_trainer(_seed + i) for i in range(num_agents)]
+            trajectory_generator = preference_comparisons.MixtureOfTrajectoryGenerators(
+                members=members,
+                custom_logger=custom_logger,
+                share_training_steps=share_training_steps_among_agents,
+            )
+            # Again using the same device as the SB3 agent
+            reward_net = reward_net.to(members[0].algorithm.device)
         else:
+            single_agent = False
```
AdamGleave, line 211: Not single agent in the sense that it's... zero agent? This is a bit counterintuitive.

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -251,7 +281,7 @@ def save_callback(iteration_num):
                         "checkpoints",
                         f"{iteration_num:04d}",
                     ),
-                    allow_save_policy=bool(trajectory_path is None),
+                    allow_save_policy=single_agent,
```
AdamGleave, line 284: We can't save checkpoints if there's multiple agents? That's a little sad.

AdamGleave (approved):
Left some minor comments, but overall looks good. Only did a shallow review. Let me know if there's anything specific I should pay closer attention to.

Rocamonde:
What's the status of this? I borrowed from this PR partially in other PRs that have been merged to avoid blockage / reinventing the wheel, so it probably requires some re-writing.

AdamGleave:
@levmckinney, what's outstanding here? I know you're busy now, but I can likely find someone else to finish the PR off.

However, IIRC the policy ensemble didn't help performance much, so perhaps it's served its purpose (of running those experiments) and we should just cherry-pick any helpful features from it then close this PR.

levmckinney:
> @levmckinney, what's outstanding here? I know you're busy now, but I can likely find someone else to finish the PR off.
> 
> However, IIRC the policy ensemble didn't help performance much, so perhaps it's served its purpose (of running those experiments) and we should just cherry-pick any helpful features from it then close this PR.

I think we can close this. The main thing we might want to cherry-pick out is the changes to the hierarchical logger. However, that only really matters if we are planning to move forward with this logging framework.

PR #534, opened 2022-08-23
Improve typing: better type hints, increased type safety, mypy support

### LINKED ISSUE #532: Add mypy as a second type-checker
## Problem
 Pytype is a good inference-based type checker. However, this often leads to lots of false negatives when it cannot infer the type. Mypy is a good second pair of eyes that looks only at type annotations, but will likely introduce lots of false positives.

## Solution
Add mypy, but not enforcing this in CI initially if it produces a lot of false positives.

## Possible alternative solutions
Alternative type checkers can be suggested. Mypy seems to be extremely well supported and popular, so it seems to be the default choice.

Rocamonde:
## Description

Fixes #532.

## Features
- [ ] Ensures all classes have type annotations in their attributes
- [x] Enforces no mypy type errors on new diffs from PRs
- [x] Fixes 200+ type errors and several exposed bugs as a result
- [x] Redesigns internal random number generation conventions to use modern `np.random.Generator` and require explicit RNG passing to all instances that require random numbers if and only if RNG is being used.
- [x] Disallows `# type: ignore` that does not specify the specific rule to be ignored, e.g. `# type: ignore[return-value]`.


Makefile
```diff
@@ -0,0 +1,40 @@
+SRC_FILES=src/ tests/ experiments/ examples/ docs/conf.py setup.py
```
AdamGleave, line 1: Do we need this in addition to `ci/code_checks.sh` and `.circleci/config.yml`?

With some work could perhaps be a replacement for `ci/code_checks.sh`, though it's missing some features right now (e.g. doing darglint but only on the diff).

It might actually be best to move to https://pre-commit.com/, I think set up right we could use that both locally and on CI (though this should be done in a separate PR if we do that!) Nora made a pre-commit config that's similar to what we do in `imitation` here: https://github.com/HumanCompatibleAI/go_attack/blob/main/.pre-commit-config.yaml

setup.py
```diff
@@ -52,6 +52,7 @@
         # TODO: upgrade jupyter-client once
         #  https://github.com/jupyter/jupyter_client/issues/637 is fixed
         "jupyter-client~=6.1.12",
+        "mypy~=0.791",
```
AdamGleave, line 55: So far we've used `==` for things that are in `0.x` version (as under SemVer this allows breaking changes). I'm a bit unsure on the merits of it, but I think we should at least be consistent within the file. Additionally `~=0.791` actually allows anything within `0.x` (there's no major/minor branch) so is quite permissive.

src/imitation/algorithms/adversarial/airl.py
```diff
@@ -69,7 +70,7 @@ def logits_expert_is_high(
         action: th.Tensor,
         next_state: th.Tensor,
         done: th.Tensor,
-        log_policy_act_prob: th.Tensor,
+        log_policy_act_prob: Optional[th.Tensor] = None,
```
AdamGleave, line 73: lol I'm amazed pytype did not catch this, very sad

src/imitation/algorithms/adversarial/common.py
```diff
@@ -212,6 +225,7 @@ def __init__(
             self.venv_wrapped = venv
             self.gen_callback = None
         else:
+            # TODO(juan) this is a bug; venv is going unused.
             venv = self.venv_wrapped = reward_wrapper.RewardVecEnvWrapper(
```
AdamGleave, line 229: Yeah think we can just delete `venv` here and a few lines above (though note `self.venv_wrapped` is being used, so it's mostly a cosmetic issue).

src/imitation/algorithms/adversarial/common.py
```diff
@@ -227,6 +241,8 @@ def __init__(
             assert gen_algo_env is not None
             self.gen_train_timesteps = gen_algo_env.num_envs
             if hasattr(self.gen_algo, "n_steps"):  # on policy
+                # TODO(juan) this looks like a bug; could not find n_steps
```
AdamGleave, line 244: It's defined in `OnPolicyAlgorithm` in SB3, in `__init__`. The tests would break if this was missing.

src/imitation/algorithms/adversarial/common.py
```diff
@@ -240,6 +256,7 @@ def __init__(
 
     @property
     def policy(self) -> policies.BasePolicy:
+        # TODO(juan) either change the return type to optional or add an assertion.
```
AdamGleave, line 259: probably an assertion? policy should always be non-None unless the algorithm is not fully initialized

src/imitation/algorithms/adversarial/common.py
```diff
@@ -429,9 +447,18 @@ def train(
...
+        ...
+
+    @overload
+    def _torchify_array(self, ndarray: None) -> None:
+        ...
+
     def _torchify_array(self, ndarray: Optional[np.ndarray]) -> Optional[th.Tensor]:
         if ndarray is not None:
             return th.as_tensor(ndarray, device=self.reward_train.device)
+        return None
```
AdamGleave, line 461: Does `mypy` insist on this? Kind of sad (no return should be semantically equivalent to returning None and I think is more Pythonic)

AdamGleave (commented):
Thanks for working on this, think it'll be really useful when done! Skimmed through and left a few comments, but haven't checked most of the files as seems like still a work-in-progress.

Makefile
Rocamonde: Having a makefile is incredibly useful to run only specific commands without having to retype all the paths and settings every time, e.g. `make format` to format the file, or `make formatcheck` to check just formatting (and not typing). If I know typing passed before, I might want to only check formatting if the change I made is only e.g. a comma. 

Also, the makefile already prints out the commands that are run, like the options we have on the code checks bash script.

I think that we should 
1. Have a Makefile that contains a pre-commit command that properly replaces `./ci/code-checks.sh`
2. Automatically run the pre-commit command using the pre-commit config that you suggested
3. Point circleCI to the makefile if possible

For now I think there's no harm in leaving the Makefile in as that's what I use all the time instead of ./ci/code-checks.sh tbh.

setup.py
Rocamonde: Sure. I guess the reason I did that is if mypy adds support for some of the stuff I had to hardcode, we could get it right away. But we would have to remove the `#type ignore` comments / the `cast`s anyways, so you're probably right.

src/imitation/algorithms/adversarial/airl.py
Rocamonde: I think pytype lets you replace the supertype's function signature in the subclass without complaining, I saw a bunch of other errors like this by mypy that pytype never flagged. (Even though many times the only difference was the variable names, not the types. But this is still a problem if you call them as keyword arguments. We didn't, hence why there were no errors. But still better to be consistent).

src/imitation/algorithms/adversarial/common.py
Rocamonde: I can add a type ignore or customize the errors in the config file, but yes, by default it wants you to have explicit returns for legibility. I tend to prefer the more pythonic way, but I don't disagree it's more readable.

Rocamonde:
Yes, there are still things left to do. But for the most part, a second pair of eyes would be needed; there are some non-trivial decisions to make to fix some type errors that require a good understanding of what the code is meant to do and I don't feel comfortable making those decisions on my own with no feedback.

src/imitation/algorithms/adversarial/common.py
Rocamonde: Okay. Then we shouldn't accept as gen_algo any BaseAlgorithm, but only an OnPolicyAlgorithm. This is why mypy was complaining then.

src/imitation/algorithms/adversarial/common.py
AdamGleave: OK, fine to leave as-is, better than a type ignore.

src/imitation/algorithms/adversarial/common.py
AdamGleave: This is optional, we have a code-path to handle other algorithms. I don't object to changing the if statement to `isinstance(self.gen_algo, OnPolicyAlgorithm)` although it's less Pythonic (duck-typing).

Makefile
AdamGleave: Having the checks defined in three places is going to be very brittle every time we change the CI checks. By all means use the Makefile personally but I'd rather it not be merged until it can replace `ci/code-checks.sh`.

I'm not sure I understand your concrete suggestion to use a `Makefile` and `pre-commit`. Isn't the Makefile redundant ifw we use `pre-commit`? Note you can specify a single hook to run https://pre-commit.com/#pre-commit-run so you'd still have support for just doing formatting, or just type-checking, etc.

AdamGleave:
> Yes, there are still things left to do. But for the most part, a second pair of eyes would be needed; there are some non-trivial decisions to make to fix some type errors that require a good understanding of what the code is meant to do and I don't feel comfortable making those decisions on my own with no feedback.

OK, I'll add it to my list to review.

src/imitation/data/rollout.py
```diff
@@ -88,11 +93,11 @@ def finish_trajectory(
         del self.partial_trajectories[key]
         out_dict_unstacked = collections.defaultdict(list)
         for part_dict in part_dicts:
-            for key, array in part_dict.items():
-                out_dict_unstacked[key].append(array)
+            for key_, array in part_dict.items():
```
AdamGleave, line 106: I'd rather `k` than `key_` though it doesn't really matter.

src/imitation/data/rollout.py
```diff
@@ -335,6 +340,7 @@ def generate_trajectories(
         may be collected to avoid biasing process towards short episodes; the user
         should truncate if required.
     """
+    rng = rng or np.random.RandomState()
```
AdamGleave, line 353: As I discussed before with MCE IRL, this is a semantic change that means we no longer benefit from global seeding. In TabularPolicy it did not matter so much, that code path was barely even used. But rollouts are a pretty core part of it, and this change will make our experiments less replicalbe (previously setting the global Sacred seed to the same value should be largely enough.)

Main options seem to either be to just ignore the type error (e.g. add ugly cast), as `np.random` does in fact implement all the methods of `RandomState` that we need, or change callsites to `rollout` to take in an explicit `rng` (this is the cleaner and more replicable thing long-term, since global sources of randomness are easily polluted anyway).

src/imitation/data/rollout.py
```diff
@@ -358,9 +364,11 @@ def generate_trajectories(
     #
     # To start with, all environments are active.
     active = np.ones(venv.num_envs, dtype=bool)
+    assert isinstance(obs, np.ndarray)
```
AdamGleave, line 377: Add comment explaining this is because we don't support dict/tuple observations?

AdamGleave (commented):
Big PR so going to have to do the reviews in batches. Any file I've left comments on I've fully reviewed, and I'm up to `policies/base.py` under the alphabetic sorting by GitHub.

tests/policies/test_policies.py
```diff
@@ -161,7 +162,10 @@ def test_normalize_features_extractor(obs_space: gym.Space) -> None:
 
     for i in range(10):
         obs = th.as_tensor([obs_space.sample()])
-        obs = preprocessing.preprocess_obs(obs, obs_space)
+        # TODO(juan) the cast below is because preprocess_obs has too general a type.
```
AdamGleave, line 165: Agreed -- if you make the PR in SB happy to review it.

tests/policies/test_replay_buffer_wrapper.py
```diff
@@ -32,21 +32,20 @@ def make_algo_with_wrapped_buffer(
...
-        buffer_size=buffer_size,
-    )
     rl_algo = rl_cls(
         policy=policy_cls,
         policy_kwargs=dict(),
         env=venv,
         seed=42,
-        **rl_kwargs,
-    )
+        # we ignore the type below because sb3 has a bug (forgot to put Type[...])
```
AdamGleave, line 40: Make this a TODO?

Good to have some easy way to find all these things we need to strip out once the upstream bugs get fixed, and grepping for TODO(juan) seems like a decent way to do it.

tests/rewards/test_reward_fn.py
```diff
@@ -7,7 +7,7 @@
 
 OBS = np.random.randint(0, 10, (64, 100))
 ACTS = NEXT_OBS = OBS
-DONES = np.zeros(64, dtype=np.bool)
+DONES = np.zeros(64, dtype=np.bool)  # type: ignore
```
AdamGleave, line 10: This fixes it and I think is the intended NumPy use:
```suggestion
DONES = np.zeros(64, dtype=np.bool_)
```

tests/rewards/test_reward_nets.py
```diff
@@ -541,7 +538,11 @@ def test_wrappers_pass_on_kwargs(
         env_2d.observation_space,
         env_2d.action_space,
     )
-    basic_reward_net.predict_processed = mock.Mock(return_value=np.zeros((10,)))
+    # TODO(juan) reassigning a method is very bad practice. I added
```
AdamGleave, line 541: You can use `with mock.patch:`

tests/util/test_wb_logger.py
```diff
@@ -87,7 +87,11 @@ def finish(self):
 mock_wandb = MockWandb()
 
 
-@mock.patch.object(wandb, "__init__", mock_wandb.__init__)
+# we ignore the type below as one should technically not access the
+# __init__ method directly but only by creating an instance.
+
```
AdamGleave, line 92: Why the two line whitespace? This makes it unclear that the comment is for line 94.

AdamGleave (changes requested):
Finished reviewing now. Please request re-review once comments addressed and remaining type issues fixed (or all of them you can figure out without too much work; happy to take a look at a few obscure ones).

src/imitation/algorithms/adversarial/common.py
Rocamonde: Oh sorry, I changed the type of this in the args. I'll fix this.

Makefile
Rocamonde: Okay, I didn't really read the docs of this tool properly. I was under the impression that you could only run it automatically and not manually. Yes, I agree 3 places for the checks is a bad idea. I just like the modularity of the `Makefile` which a shell script does not immediately provide without a lot of code overhead.

src/imitation/algorithms/adversarial/common.py
```diff
@@ -103,13 +113,18 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
...
+    venv_wrapped: vec_env.VecEnvWrapper
+
     def __init__(
         self,
         *,
         demonstrations: base.AnyTransitions,
         demo_batch_size: int,
         venv: vec_env.VecEnv,
-        gen_algo: base_class.BaseAlgorithm,
+        gen_algo: on_policy_algorithm.OnPolicyAlgorithm,
```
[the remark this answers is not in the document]
Rocamonde: Haha yes you're right, just commented on the other box about this. I'll change that back.

src/imitation/algorithms/density.py
```diff
@@ -121,6 +134,13 @@ def _set_demo_from_batch(
         next_obs_b: Optional[np.ndarray],
     ) -> None:
         next_obs_b = next_obs_b or itertools.repeat(None)
+        # TODO(juan) I think this assumes that if next_obs_b is an array,
```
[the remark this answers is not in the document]
Rocamonde: That sounds good. I think you meant `assert obs_b.shape[1:] == self.venv.observation_space.shape` and/or `assert next_obs_b.shape[1:] == self.venv.observation_space.shape`?

src/imitation/algorithms/density.py
```diff
@@ -222,12 +242,16 @@ def _preprocess_transition(
...
+    #  if the previous names are better suited.
     def __call__(
         self,
-        obs_b: np.ndarray,
-        act_b: np.ndarray,
-        next_obs_b: np.ndarray,
-        dones: np.ndarray,
+        state: np.ndarray,
+        action: np.ndarray,
+        next_state: np.ndarray,
```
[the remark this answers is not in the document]
Rocamonde: Thanks, that was useful. I think you meant in line 225; will change.

src/imitation/algorithms/mce_irl.py
```diff
@@ -162,8 +165,7 @@ def __init__(
         assert isinstance(action_space, gym.spaces.Discrete), "action not tabular"
         # What we call state space here is observation space in SB3 nomenclature.
         super().__init__(observation_space=state_space, action_space=action_space)
-        self.rng = rng or np.random
-        self.pi = None
+        self.rng = rng or np.random.RandomState()
```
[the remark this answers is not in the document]
Rocamonde: I think the issue with doing `np.random` is that you're mixing a module type with a class instance type, and type checkers don't like that. Is there a way to grab the global seed when calling `RandomState()`?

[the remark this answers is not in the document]
Rocamonde: It seems that relying on the global seed is actually discouraged, and that setting it is now "legacy" behavior.

src/imitation/algorithms/mce_irl.py
```diff
@@ -176,10 +178,14 @@ def set_pi(self, pi: np.ndarray) -> None:
     def _predict(self, observation: th.Tensor, deterministic: bool = False):
         raise NotImplementedError("Should never be called as predict overridden.")
 
-    def forward(self, observation: th.Tensor, deterministic: bool = False):
+    def forward(  # type: ignore[override]
```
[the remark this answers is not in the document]
Rocamonde: Yes, I raised this issue in SB3. It does technically violate the LSP as the parent class could take any args and kwargs, whereas the subclasses can only take certain args. https://github.com/pytorch/pytorch/issues/29099 https://github.com/python/mypy/issues/8795 https://github.com/DLR-RM/stable-baselines3/issues/1027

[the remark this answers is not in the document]
Rocamonde: Added this as an issue on imitation (#553).

src/imitation/data/rollout.py
Rocamonde: I am inclined to go for the second.

src/imitation/algorithms/mce_irl.py
```diff
@@ -234,7 +232,8 @@ def predict(  # type: ignore[override]
                 actions.append(self.rng.choice(len(dist), p=dist))
 
         timesteps += 1  # increment timestep
-        return np.array(actions), timesteps
+        state = (timesteps,)
```
Rocamonde, line 235: This is a breaking change in the API. I updated the tests to reflect the new usage (have to extract the zeroth element from state). What is the reason behind this? We always return a tuple of length 1 anyways. Is this for compatibility with other objects / functions?

Makefile
Rocamonde: I think there's stuff in the Makefile that does not really make sense to include in the pre-commit hook, such as frequently-used development scripts, e.g. make clean removes cache and other folders that are not committed to the repo. In this example, this is not a useful pre-commit action as git will ignore them anyways, but it's quite useful to have a `make clean` action. 

Probably worth moving this discussion off to another issue though, I don't think it should be a particularly hard thing to discuss but it seems useful to document the decision on its own.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -270,30 +275,35 @@ def sample(self, steps: int) -> Sequence[types.TrajectoryWithRew]:
...
-    @TrajectoryGenerator.logger.setter
+            # We call _get_trajectories separately on agent_trajs and exploration_trajs
+            # and then just concatenate. This could mean we return slightly too many
+            # transitions, but it gets the proportion of exploratory and agent
+            # transitions roughly right.
+            trajectories.extend(list(exploration_trajs))
+        return trajectories
+
+    # Type ignore due to https://github.com/python/mypy/issues/5936
+    @TrajectoryGenerator.logger.setter  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
Rocamonde: Thought about doing that, decided not to in the end, as I was hesitant since we might forget to remove that once that's fixed. But you're right that it's cleaner until then and even if that stays it's not necessarily bad style.

src/imitation/rewards/reward_nets.py
```diff
@@ -10,6 +10,7 @@
 from torch import nn
 
 from imitation.util import networks, util
+from imitation.util.networks import BaseNorm
```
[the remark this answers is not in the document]
Rocamonde: Apologies, it slipped my mind to fix it after I auto-imported it.

src/imitation/rewards/reward_nets.py
```diff
@@ -82,20 +83,29 @@ def preprocess(
         del state, action, next_state, done  # unused
 
         # preprocess
-        state_th = preprocessing.preprocess_obs(
-            state_th,
-            self.observation_space,
-            self.normalize_images,
+        state_th = cast(
```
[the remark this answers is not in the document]
Rocamonde: It might be worth adding an SB3 PR to allow type-safe polymorphism through generics.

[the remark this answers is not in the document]
Rocamonde: I mean, to be fair, we aren't checking anywhere in the class that the spaces we are given are array spaces. Maybe we should do that? Otherwise it seems unwarranted to have the cast.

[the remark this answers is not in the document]
Rocamonde: Oh damn, this is fixed in the new gym release (they made Space a generic), but we aren't using it due to SB3.

src/imitation/rewards/reward_nets.py
```diff
@@ -375,21 +385,21 @@ def __init__(
...
-        full_build_mlp_kwargs.update(
-            {
-                # we do not want these overridden
-                "in_size": combined_size,
-                "out_size": 1,
-                "squeeze_output": True,
-            },
+        self.mlp = networks.build_mlp(
+            hid_sizes=(32, 32),
+            **kwargs,
```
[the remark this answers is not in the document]
Rocamonde: You're correct. this was the way I was building the arguments in a separate dict before so the args below would override them, but that's not happening anymore as function calls only accept a single value for each kwarg.

src/imitation/scripts/analyze.py
```diff
@@ -49,14 +49,15 @@ def _gather_sacred_dicts(
...
+    sacred_dicts_list = []
 
     for sacred_dir in sacred_dirs:
         try:
-            sacred_dicts.append(sacred_util.SacredDicts.load_from_dir(sacred_dir))
+            sacred_dicts_list.append(sacred_util.SacredDicts.load_from_dir(sacred_dir))
         except json.JSONDecodeError:
             warnings.warn(f"Invalid JSON file in {sacred_dir}", RuntimeWarning)
 
+    sacred_dicts: Iterable = sacred_dicts_list
```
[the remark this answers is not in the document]
Rocamonde: I get the pros and cons of each. It's hard to visually keep track of type changes and it can lead to confusion. A fix could be forcing a new annotation every time you change the type to aid clarity, but this defeats type inference. Unlikely they will change this though, I'm sure there was an extensive debate about this issue at some point ;)

src/imitation/scripts/common/reward.py
```diff
@@ -85,7 +84,7 @@ def _make_reward_net(
     venv: vec_env.VecEnv,
     net_cls: Type[reward_nets.RewardNet],
     net_kwargs: Mapping[str, Any],
-    normalize_output_layer: Optional[Type[nn.Module]],
+    normalize_output_layer: Optional[Type[networks.BaseNorm]],
```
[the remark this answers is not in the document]
Rocamonde: If we ever want to do that we can always pull the important public API into a protocol and then have our current BaseNorm implement it.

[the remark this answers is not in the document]
Rocamonde: Actually the problem with doing that is that protocols can only inherit from protocol classes, and nn.Module is not a protocol, so we would have to re-define all the nn.Module methods we need in our custom protocol. That's quite silly, I'm surprised that it's not possible to do this.

src/imitation/scripts/parallel.py
```diff
@@ -192,14 +192,10 @@ def inner(config: Mapping[str, Any], reporter) -> Mapping[str, Any]:
         ex.observers = [FileStorageObserver("sacred")]
 
         # Apply base configs to get modified `named_configs` and `config_updates`.
-        named_configs = []
-        named_configs.extend(base_named_configs)
-        named_configs.extend(run_kwargs["named_configs"])
+        named_configs = [*base_named_configs, *run_kwargs["named_configs"]]
```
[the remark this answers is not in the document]
Rocamonde: I don't mind using addition instead of unpacking. I have a slight preference for unpacking as it's an operation that has become quite intuitive to me, but I understand that for most people the addition operator is probably the most obvious.

src/imitation/util/networks.py
```diff
@@ -205,7 +205,7 @@ def build_mlp(
     dropout_prob: float = 0.0,
     squeeze_output: bool = False,
     flatten_input: bool = False,
-    normalize_input_layer: Optional[Type[nn.Module]] = None,
+    normalize_input_layer: Optional[Type[BaseNorm]] = None,
```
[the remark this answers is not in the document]
Rocamonde: It's a private class so they might replace it anytime, but at the same time I think it's easier than defining a custom type that is the union of all possible normalization layers that are built into pytorch. Unless you think only a few are relevant and we can easily enumerate those.

tests/rewards/test_reward_fn.py
Rocamonde: Yes, that is correct. I learned this later but forgot to go back. Thanks for pointing it out. Another option is to do `np.dtype('bool')`.

src/imitation/data/rollout.py
Rocamonde: To be fair, I don't see how relying on the global seed makes anything more replicable. The global seed is drawn from /dev/random just like the local seed when creating an instance of `RandomState` AFAIK. The only benefit is that you can fix the global seed when you start the program, but I don't see that we do that anywhere except for one test.

Rocamonde: After thinking about this, I opted for the design decision to force all functions to require explicit random states everywhere. Then the user can decide on the high-level API whether to use a fixed seed or to take it from /dev/random every time the script is run. This explicit consent ensures that users know what parts of their code are using random number generators, and whether this behavior is being controlled or not. It also allows us to explicitly choose for each test whether we want a fixed or a random seed, through a fixture that I have added. 

Let me know if you disagree with this, we can have a more detailed discussion. I know it's a fairly big change tangential to the PR, but in a sense ensuring type safety is strongly related to ensuring that default arguments are not passed tacitly when they shouldn't be, and seeds are a very good case of that IMO.

Rocamonde:
lol, I wrote formatting with three t's. Clearly did not format my commit message.

I have now added support for controlled seeding for the entire API, except for the Buffer and ReplayBuffer class, for which I couldn't be bothered at 1.30am. I managed to get 800 failing tests to zero, pretty happy with how that turned out.

I added an ingredient called seeding which propagates the seed and random state to all experiments, which then inject those values down to all API methods. So there is a unique source of truth for controlling randomness, which is the seed in the ingredient config. By default it's fixed to 0, but this can be set to None, which effectively draws the seed from `/dev/random`.

This was a bit of a detour from the original PR but I think it has been well worth it, as now all randomness in imitation is controlled by default, and can be made "truly" random by explicitly setting a null seed when calling API methods.

I also made all tests have fixed seeds by default, except those that were explicitly using a random seed. @AdamGleave if you disagree with a specific test being controlled for and would like it to be actually random in every CI run, just leave a comment when you review and I'll change it back. I did not have an informed opinion on which tests to fix and which to make random, so my intention was to get your feedback on that.

src/imitation/algorithms/mce_irl.py
```diff
@@ -376,6 +386,10 @@ def set_demonstrations(self, demonstrations: MCEDemonstrations) -> None:
         # Demonstrations are either trajectories or transitions;
         # we must compute occupancy measure from this.
         if isinstance(demonstrations, Iterable):
+            # TODO(juan) this is wrong; if this is a container (list-like)
```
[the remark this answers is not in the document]
Rocamonde: This is solved in  https://github.com/HumanCompatibleAI/imitation/pull/534/commits/0a1b06a6e2741e461ecb7ab4e9de783455b91b46

tests/util/test_wb_logger.py
Rocamonde: That was a stupid PEP rule but I agree, we should override that here.

tests/rewards/test_reward_nets.py
Rocamonde: This is gone in the latest commit to master so marking this as resolved.

tests/data/test_types.py
```diff
@@ -27,7 +26,8 @@
 LENGTHS = [0, 1, 2, 10]
 
 
-def _check_1d_shape(fn: Callable[[np.ndarray], Any], length: float, expected_msg: str):
+def _check_1d_shape(fn: Callable[[np.ndarray], Any], length: int, expected_msg: str):
+    assert isinstance(length, int)
```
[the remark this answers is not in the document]
Rocamonde: Because this was a float before (incorrectly), and I just replaced the type. I added an assertion to see if my assumption of the true type was correct, and forgot to remove it.

tests/algorithms/test_preference_comparisons.py
```diff
@@ -22,10 +22,12 @@
 
 
 @pytest.fixture
-def venv():
+def venv(random_state_fixed):
+    random_state = random_state_fixed
```
AdamGleave, line 26: Why not just pass this in directly on line 30?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -54,13 +56,18 @@ def agent(venv):
 
 
 @pytest.fixture
-def random_fragmenter():
-    return preference_comparisons.RandomFragmenter(seed=0, warning_threshold=0)
+def random_fragmenter(random_state_fixed):
+    random_state = random_state_fixed
+    return preference_comparisons.RandomFragmenter(
+        random_state=random.Random(util.make_seeds(random_state)),
```
AdamGleave, line 62: Why not just pass this directly in on line 62?

If that makes line too long then could split it up as `seeds = util.make_seeds(random_state)`.

Otherwise these assignments just seem vacuous.

tests/algorithms/test_preference_comparisons.py
```diff
@@ -54,13 +56,18 @@ def agent(venv):
...
+        warning_threshold=0,
+    )
 
 
 @pytest.fixture
-def agent_trainer(agent, reward_net, venv):
-    return preference_comparisons.AgentTrainer(agent, reward_net, venv)
+def agent_trainer(agent, reward_net, venv, random_state_fixed):
+    random_state = random_state_fixed
```
AdamGleave, line 69: As before, why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -78,10 +85,12 @@ def _check_trajs_equal(
         assert traj1.terminal == traj2.terminal
 
 
-def test_mismatched_spaces(venv, agent):
+def test_mismatched_spaces(venv, agent, random_state_fixed):
+    random_state = random_state_fixed
```
AdamGleave, line 89: You guessed it... why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -186,7 +200,9 @@ def test_trainer_no_crash(
     random_fragmenter,
     custom_logger,
     schedule,
+    random_state_fixed,
 ):
+    random_state = random_state_fixed
```
AdamGleave, line 205: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -204,7 +221,8 @@ def test_trainer_no_crash(
     assert 0.0 < result["reward_accuracy"] <= 1.0
 
 
-def test_reward_ensemble_trainer_raises_type_error(venv):
+def test_reward_ensemble_trainer_raises_type_error(venv, random_state_fixed):
+    random_state = random_state_fixed
```
AdamGleave, line 225: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -229,14 +248,17 @@ def test_correct_reward_trainer_used_by_default(
     reward_net,
     random_fragmenter,
     custom_logger,
+    random_state_fixed,
 ):
+    random_state = random_state_fixed
```
AdamGleave, line 253: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -258,7 +280,9 @@ def test_init_raises_error_when_trying_use_improperly_wrapped_ensemble(
     venv,
     random_fragmenter,
     custom_logger,
+    random_state_fixed,
 ):
+    random_state = random_state_fixed
```
AdamGleave, line 285: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -279,11 +303,19 @@ def test_init_raises_error_when_trying_use_improperly_wrapped_ensemble(
...
 
-def test_discount_rate_no_crash(agent_trainer, venv, random_fragmenter, custom_logger):
+def test_discount_rate_no_crash(
+    agent_trainer,
+    venv,
+    random_fragmenter,
+    custom_logger,
+    random_state_fixed,
+):
+    random_state = random_state_fixed
```
AdamGleave, line 318: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -305,14 +337,23 @@ def test_discount_rate_no_crash(agent_trainer, venv, random_fragmenter, custom_l
...
 
 
-def test_synthetic_gatherer_deterministic(agent_trainer, random_fragmenter):
-    gatherer = preference_comparisons.SyntheticGatherer(temperature=0)
+def test_synthetic_gatherer_deterministic(
+    agent_trainer,
+    random_fragmenter,
+    random_state_fixed,
+):
+    random_state = random_state_fixed
```
AdamGleave, line 352: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -373,11 +414,12 @@ def test_preference_dataset_errors(agent_trainer, random_fragmenter):
         dataset.push(fragments, preferences)
 
 
-def test_preference_dataset_queue(agent_trainer, random_fragmenter):
+def test_preference_dataset_queue(agent_trainer, random_fragmenter, random_state_fixed):
+    random_state = random_state_fixed
```
AdamGleave, line 418: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -389,11 +431,17 @@ def test_preference_dataset_queue(agent_trainer, random_fragmenter):
...
 
 
-def test_store_and_load_preference_dataset(agent_trainer, random_fragmenter, tmp_path):
+def test_store_and_load_preference_dataset(
+    agent_trainer,
+    random_fragmenter,
+    tmp_path,
+    random_state_fixed,
+):
+    random_state = random_state_fixed
```
AdamGleave, line 440: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -415,11 +463,14 @@ def test_exploration_no_crash(
     venv,
     random_fragmenter,
     custom_logger,
+    random_state_fixed,
 ):
+    random_state = random_state_fixed
```
AdamGleave, line 468: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -441,7 +493,9 @@ def test_active_fragmenter_discount_rate_no_crash(
     random_fragmenter,
     uncertainty_on,
     custom_logger,
+    random_state_fixed,
 ):
+    random_state = random_state_fixed
```
AdamGleave, line 498: Why not pass it in directly?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -587,13 +643,17 @@ def test_agent_trainer_sample(venv, agent_trainer):
...
     """Test `AgentTrainer.sample()` in an image environment.
 
     SB3 algorithms may rearrange the channel dimension in environments with image
     observations, but `sample()` should return observations matching the original
     environment.
+
+    Args:
+        random_state_fixed: Random state (with a fixed seed).
     """
+    random_state = random_state_fixed
```
AdamGleave, line 656: Why not pass it in directly?

tests/policies/test_exploration_wrapper.py
```diff
@@ -15,14 +15,15 @@ def make_wrapper(random_prob, switch_prob):
     venv = util.make_vec_env(
         "seals/CartPole-v0",
         n_envs=1,
+        random_state=np.random.RandomState(),
```
AdamGleave, line 18: Should this be using a fixture?

tests/policies/test_exploration_wrapper.py
```diff
@@ -15,14 +15,15 @@ def make_wrapper(random_prob, switch_prob):
...
     )
     return (
         exploration_wrapper.ExplorationWrapper(
             policy_callable=constant_policy,
             venv=venv,
             random_prob=random_prob,
             switch_prob=switch_prob,
-            seed=0,
+            random_state=np.random.RandomState(0),
```
AdamGleave, line 26: Should this be using a fixture?

tests/policies/test_policies.py
```diff
@@ -82,7 +103,7 @@ def _test_serialize_identity(env_name, model_cfg, tmpdir):
         venv,
         n_timesteps=1000,
         deterministic_policy=True,
-        rng=np.random.RandomState(0),
+        random_state=np.random.RandomState(0),
```
AdamGleave, line 106: Should this be using a fixture?

tests/policies/test_policies.py
```diff
@@ -94,7 +115,7 @@ def _test_serialize_identity(env_name, model_cfg, tmpdir):
...
```
AdamGleave, line 118: Should this be using a fixture?

tests/rewards/test_reward_nets.py
```diff
@@ -143,10 +160,22 @@ def _make_env_and_save_reward_net(env_name, reward_type, tmpdir):
...
 @pytest.mark.parametrize("env_name", ENVS)
 @pytest.mark.parametrize("reward_type", DESERIALIZATION_TYPES)
-def test_reward_valid(env_name, reward_type, tmpdir):
+def test_reward_valid(env_name, reward_type, tmpdir, random_state_fixed):
     """Test output of reward function is appropriate shape and type."""
-    venv = util.make_vec_env(env_name, n_envs=1, parallel=False)
-    venv, tmppath = _make_env_and_save_reward_net(env_name, reward_type, tmpdir)
+    random_state = random_state_fixed
+    # TODO(juan) the line below is not being used?
+    venv = util.make_vec_env(
```
AdamGleave, line 167: huh, yeah, think you can delete it!

tests/algorithms/test_mce_irl.py
```diff
@@ -240,36 +240,37 @@ def test_tabular_policy():
     pi = np.stack(
         [np.eye(2), 1 - np.eye(2)],
     )
-    rng = np.random.RandomState(42)
+    random_state = np.random.RandomState(42)
```
AdamGleave, line 243: Use fixture instead?

I don't think the test needs the seed to be 42. I hope not!

tests/algorithms/test_mce_irl.py
```diff
@@ -240,36 +240,37 @@ def test_tabular_policy():
...
         pi=pi,
-        rng=rng,
+        random_state=random_state,
     )
 
     states = np.array([0, 1, 1, 0, 1])
     actions, timesteps = tabular.predict(states)
     np.testing.assert_array_equal(states, actions)
-    np.testing.assert_equal(timesteps, 1)
+    np.testing.assert_equal(timesteps[0], 1)
```
AdamGleave, line 254: thanks for fixing this :face_exhaling:

AdamGleave (commented):
I like making `random_state` explicit, thanks for doing this!

Note tests and linting are currently failing. I'd suggest installing a commit hook for the linting, you'll catch it faster than waiting for the CI, especially if you make the commit hook just look at the diff. While I don't mind reviewing code that's still rough around the edges it will overall speed things up for both of us if the obvious errors are fixed first.

For the tests, I'd default to using `random_state`, reserving `random_state_fixed` only for special cases.

We should strive to have our tests pass no matter what seed we use. There's two categories under which they might fail:
  1. There's a bug and things happen to work under certain seeds but not others.
  2. The correctly implemented algorithm really does have very seed-dependent performance, and for a few seeds it'll fail.
For 1) ideally we'd write tests such that they'd catch the bug no matter the initial seed -- e.g. do a bunch of iterations of it so no matter the initial state one of them is likely to fail. For 2) we'd ideally write the tests such that they check for statistically significant results, and have adequate power to catch ones that really do fail.

I used "ideally" a lot above. In practice running many iterations may be too computationally expensive for us to want in CI. Or we might just mess up an implementation. Either way, some of our tests may end up being seed dependent.

If a test fails for x% of seeds and we fix the seeds, there's an (1-x)% chance that it *always* passes and we never notice. If we randomize the seed, it'll be a flaky test and fail for x% of runs. Flaky tests are bad---but tests that are consistent false positives are worse. At least a flaky test is a call to arms to see what's going wrong.

Given this I'd suggest changing all `random_state_fixed` to `random_state`, then run a flaky test finder. For ones that are flaky, we can look into what's going on in a case-by-case basis. There might be a few cases where `random_state_fixed` is right: I'm sympathetic to using a fixed seed for the `*_expert_trajectories` fixture for example, as expert trajectories can vary in how high-performing they are and it seems like just adding noise to penalize downstream algorithms for some sub-par input data.

There were a lot of instances of `random_state = random_state_fixed`. I assume this was to make it easier to switch to `random_state` later? These should all be deleted before merge, and switch to either passing `random_state` or `random_state_fixed` argument in directly.

AdamGleave:
Took a quick look at some of the tests that were failing due to `get_first_iter_element`. Implementing this method right is actually quite subtle. My original comment was mistaken: you can't just always return `itertools.chain([first_element], iterator)` because if it's a sequence you want to return the sequence back so you can iterate over it repeatedly.

The challenge is there's no well-defined type hierarchy that describes "iterables you can iterate over multiple times". For example, `range(n)` can be iterated over multiple times, but is decidedly not a list or tuple -- or even a sequence. After a little research ([this article](https://opensource.com/article/18/3/loop-better-deeper-look-iteration-python) was useful although overly verbose), I think the Pythonic thing to do is just check if `iter(iterable) == iterable`. If so, iterable was already an iterator, and we need to do the chain thing. If not, we can return iterable directly.

This seems to work and I added a test for it, but please do check (and consider adding more tests), we've both made at least one mistake here so I think it's a tricky area to get right.

AdamGleave:
Just noticed mypy is still not run in the CircleCI config. Making `mypy` type check pass is pointless if we don't enforce it. So should for sure add support for that before merging.

Rocamonde:
Very interesting subtlety. Will read the docs on python data structures more carefully tomorrow.

> 
> My original comment was mistaken: you can't just always return itertools.chain([first_element],
> iterator) because if it's a sequence you want to return the sequence back
> so you can iterate over it repeatedly.
> 

Yes, I already realized of this which is why I checked for tuple and list subclasses. The mistake was not to check for a general sequence object. I guess my post-hoc rationalization would be that I did not expect anyone to use any other sequence type.

Your interpretation seems correct: iterators return a pointer to themselves when you call `iter` on them. So by definition they are iterables. This is strange at first sight, because they are in principle very distinct data types: the iterator is the data stream and the iterable is the object that produces the stream(s), which can be ephemeral or persistent.

My basic understanding is that it’s for polymorphic support. It makes sense that if you try to get an iterator from an iterator you just get the object itself, at which point it satisfies the protocol of an iterable. An example of where this is obviously useful is when you pass an iterator to a for loop, as what python actually does is construct an iterator from the object passed, enter a while loop that fetches the next element each time on the iterator, and exits the loop when the iterator is exhausted. (`iter()` also works for sequences in general even if `__iter__()` is not implemented as python wraps the structure by getting the length of the object with `__len__()`, constructs an iterator from range(length), then uses `__getitem__()` to fetch the element at the right index.)

Any sensible data structure that is an iterable but not an iterator (so not an ephemeral stream) should return a fresh iterator every time the `__iter__()` method is called. In theory, I could define a data structure that is an iterable, not an iterator, and that returns iterators when `__iter__()` is called such that the sequences extracted from them are not identical (e.g. by exhausting them, removing or adding items, or sorting them). But this cannot really be prevented, as there's no guarantee that Python provides to guard against this, and is just never going to happen anyways. The only way would be to only allow sequences and iterators as types, which I guess we could also do since that's all we're ever going to need. But then at that point we can follow the principle of duck typing and trust that the user implements a sequence-like iterable in the sense of providing reproducible iterators on every __iter__() call.

Rocamonde:
Yes, I have not done this yet because I wanted to use visual CI checks as an indication of whether I had broken something, until I had fixed all the mypy errors, then add mypy. Happy to add mypy now as only a few errors are left and it's the whole goal anyways.

src/imitation/algorithms/mce_irl.py
Rocamonde: sounds good. Marking this as resolved then.

Rocamonde:
> There were a lot of instances of `random_state = random_state_fixed`. I assume this was to make it easier to switch to `random_state` later? These should all be deleted before merge, and switch to either passing `random_state` or `random_state_fixed` argument in directly.

That's right, it's to make it easier to change later. IDEs have refactoring tools, but thought that removing a line and changing the function signature would be faster and safer to avoid errors.

Makefile
Rocamonde: #554

tests/policies/test_policies.py
Rocamonde: Unfortunately there's no easy way to do this since python does not allow excluding types (so tensor is returned for any space except Dict), and the version of gym currently used in SB3 does not have support for generic spaces, so we cannot specify the types of elements in the observation space (i.e. what is returned by `sample()`).  Left this on https://github.com/DLR-RM/stable-baselines3/issues/1065

ci/clean_notebooks.py
```diff
@@ -0,0 +1,33 @@
+import json
```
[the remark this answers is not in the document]
Rocamonde: Added support for that:
```
(venv)  ~/git/HumanCompatibleAI/imitation   add-mypy ±  ./ci/clean_notebooks.py --check
Checking /Users/rocamonde/git/HumanCompatibleAI/imitation/examples/1_train_bc.ipynb
Checking /Users/rocamonde/git/HumanCompatibleAI/imitation/examples/7_train_density.ipynb
Checking /Users/rocamonde/git/HumanCompatibleAI/imitation/examples/4_train_airl.ipynb
Checking /Users/rocamonde/git/HumanCompatibleAI/imitation/examples/3_train_gail.ipynb
Checking /Users/rocamonde/git/HumanCompatibleAI/imitation/examples/2_train_dagger.ipynb
Traceback (most recent call last):
  File "./ci/clean_notebooks.py", line 70, in <module>
    clean_notebook(file, check_only=check_only)
  File "./ci/clean_notebooks.py", line 41, in clean_notebook
    raise UncleanNotebookError(f'Notebook {file} has metadata')
UncleanNotebookError: Notebook /Users/rocamonde/git/HumanCompatibleAI/imitation/examples/2_train_dagger.ipynb has metadata
```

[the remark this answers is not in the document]
Rocamonde: Also added it to the CI.

ci/clean_notebooks.py
```diff
@@ -0,0 +1,33 @@
...
+import pathlib
+
+import nbformat
+
+
+def clean_notebook(file: pathlib.Path):
```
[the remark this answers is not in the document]
Rocamonde: Much improved in the new revision.

src/imitation/util/util.py
```diff
@@ -208,6 +241,45 @@ def safe_to_tensor(numpy_array: np.ndarray, **kwargs) -> th.Tensor:
...
+
+    """
+    if obj is None:
+        # We ignore the type due to https://github.com/google/pytype/issues/445
+        return None  # type: ignore[bad-return-type]
+    elif isinstance(obj, np.ndarray):
+        return obj
+    else:
+        if warn:
+            warnings.warn(f"Converted tensor {obj} to numpy array.")
```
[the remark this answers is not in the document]
Rocamonde: LOL. Okay. Will fix.

src/imitation/util/util.py
```diff
@@ -236,3 +308,26 @@ def tensor_iter_norm(
...
+    Returns:
+        A tuple containing the first element of the iterable, and a fresh iterable
+        with all the elements.
+    """
+    if isinstance(iterable, (list, tuple)):
+        return iterable[0], iterable
+    else:
+        try:
+            first_element = next(iter(iterable))
+            return first_element, itertools.chain([first_element], iterable)
```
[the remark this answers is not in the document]
Rocamonde: I think we've discussed this pretty extensively by now :) Glad we found a good solution to this.

src/imitation/util/util.py
```diff
@@ -236,3 +308,26 @@ def tensor_iter_norm(
     # = sum(x**ord for x in tensor for tensor in tensor_iter)**(1/ord)
     # = th.norm(concatenated tensors)
     return th.norm(norm_tensor, p=ord)
+
+
+def get_first_iter_element(iterable: Iterable[T]) -> Tuple[T, Iterable[T]]:
```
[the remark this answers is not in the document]
Rocamonde: Thanks!

src/imitation/algorithms/base.py
```diff
@@ -113,6 +114,11 @@ def __setstate__(self, state):
 
 TransitionMapping = Mapping[str, Union[np.ndarray, th.Tensor]]
 TransitionKind = TypeVar("TransitionKind", bound=types.TransitionsMinimal)
+# TODO(juan) AnyTransitions is used in many places but TransitionKind
+#  is just a type variable. If what we want is to allow for subclasses of
+#  types.TransitionsMinimal, we should use that type directly, as that is already
```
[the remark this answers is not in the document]
Rocamonde: I have, and it doesn't. Replaced it as suggested.

src/imitation/util/util.py
```diff
@@ -208,6 +241,45 @@ def safe_to_tensor(numpy_array: np.ndarray, **kwargs) -> th.Tensor:
...
+        obj: torch tensor object to convert to numpy array
+        warn: if True, warn if the object is not already a numpy array. Useful for
+            warning the user of a potential performance hit if a torch tensor is
+            not the expected input type.
+
+    Returns: object converted to numpy array
+
+    """
+    if obj is None:
+        # We ignore the type due to https://github.com/google/pytype/issues/445
```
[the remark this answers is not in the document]
Rocamonde: Thanks for the tip, just added that.

src/imitation/scripts/common/seeding.py
```diff
@@ -0,0 +1,26 @@
+"""Single source of truth for seeding random number generators in experiments."""
```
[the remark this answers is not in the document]
Rocamonde: I was going to do that, but that forces other ingredients and experiments to depend on common, and thought it might not be desirable. Since you're okay with it, I'll change it.

src/imitation/util/networks.py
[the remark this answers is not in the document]
Rocamonde: In a new revision I add the type AnyNorm to this file to solve this issue, let me know if you think this is an acceptable solution.

tests/policies/test_policies.py
Rocamonde: I think actually here it might sense that this be manually set, since we are also seeding the environment to a specific value, and that might have something to do with it?

Rocamonde: Actually here it makes sense as we want to make sure that when fixing the seed you get the same result if you run it multiple times. if we use a fixture, the state of the RNG is going to change from one run to another (we would have to pass separate rngs, but I don't think it's worth writing a hack to support this using fixtures).

tests/policies/test_policies.py
Rocamonde: See above.

tests/data/test_rollout.py
```diff
@@ -251,18 +295,19 @@ def test_compute_returns(gamma):
...
 
 
-def test_generate_trajectories_type_error():
+def test_generate_trajectories_type_error(rng):
     venv = vec_env.DummyVecEnv([functools.partial(TerminalSentinelEnv, 1)])
     sample_until = rollout.make_min_episodes(1)
     with pytest.raises(TypeError, match="Policy must be.*got <class 'str'> instead"):
         rollout.generate_trajectories(
-            "strings_are_not_valid_policies",
+            "strings_are_not_valid_policies",  # type: ignore
```
AdamGleave, line 303: `ci/check_typeignore.py` will complain about this

tests/policies/test_replay_buffer_wrapper.py
```diff
@@ -29,47 +29,53 @@ def make_algo_with_wrapped_buffer(
...
         policy_kwargs=dict(),
         env=venv,
         seed=42,
-        **rl_kwargs,
-    )
+        # TODO(juan) we ignore the type below due to
+        #  https://github.com/DLR-RM/stable-baselines3/issues/1039
+        #  PR fixing this has been merged to master,
+        #  remove the type ignore in the next sb3 release.
+        replay_buffer_class=ReplayBufferRewardWrapper,  # type: ignore
```
AdamGleave, line 45: `ci/check_typeignore.py` will complain

tests/policies/test_replay_buffer_wrapper.py
```diff
@@ -29,47 +29,53 @@ def make_algo_with_wrapped_buffer(
...
+        replay_buffer_kwargs=dict(
+            replay_buffer_class=replay_buffer_class,
+            reward_fn=zero_reward_fn,
+        ),
+        buffer_size=buffer_size,
+    )  # type: ignore
```
AdamGleave, line 51: `ci/check_typeignore.py` will complain

tests/policies/test_replay_buffer_wrapper.py
```diff
@@ -29,47 +29,53 @@ def make_algo_with_wrapped_buffer(
...
+def test_invalid_args(rng):
     with pytest.raises(
         TypeError,
         match=r".*unexpected keyword argument 'replay_buffer_class'.*",
     ):
+        # we ignore the type because we are intentionally
+        # passing the wrong type for the test
         make_algo_with_wrapped_buffer(
-            rl_cls=sb3.PPO,
+            rl_cls=sb3.PPO,  # type: ignore
```
AdamGleave, line 63: `ci/check_typeignore.py` will complain

AdamGleave (changes requested):
Thanks for seeing this through. There's a few issues that I think need to be addressed before merge, but I'd like to get it merged ASAP before the codebase changes further underneath us, so let's do some combination of (a) you addressing the changes and me reviewing them quickly, (b) postponing what we can to future PRs.

I've left a bunch of comments, most of them pretty minor.

The main thing that stands out is:
  - ci/check-typeignore.py: good idea but seems semi-abandoned? It's not actually applied anywhere in `.circleci/config.yml` and perhaps because of this there's lots of `# type: ignore` present in the codebase that it complains about. I suggest you *delete* it from this PR and make a new PR that introduces that, adds it to `.circleci/config.yml` and cleans up the remaining errors. WDYT?
  - `ci/clean_notebooks.py` needs to be refactored a little. This is another thing that probably would have been easier had it been in its own PR. But it's so close to being finished now it's probably easiest to finish in this PR.
  - You use `types.path_to_str` in a bunch of places that end up as pathlib objects, like `pathlib.Path(types.path_to_str(x))`. This is a bit absurd, especially when `x` could sometimes start out as a pathlib! It probably makes sense to add a `path_to_pathlib` helper method and replace many of the uses of that there. However, I'm OK ignoring this so long as we make sure to clean it up in 

Test coverage is fine, happy to ignore the patch coverage issue given that project coverage has increased.

ci/clean_notebooks.py
```diff
@@ -1,33 +1,108 @@
...
+        print(f"Cleaned {file}")
 
-    # Write the notebook
-    with open(file, 'w') as f:
-        nbformat.write(nb, f)
 
+if __name__ == "__main__":
+    # if the argument --check has been passed, check if the notebooks are clean
+    # otherwise, clean them in-place
+    import argparse
```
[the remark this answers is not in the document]
Rocamonde: If you use the code as an import and not as command line, there would be no need to import argparse. I don't have a strong preference for this, as I don't anticipate we will be using this much through imports, if at all.

ci/code_checks.sh
```diff
@@ -22,6 +23,7 @@ fi
 if [ "${skipexpensive:-}" != "true" ]; then
   echo "Type checking"
   pytype -j auto "${SRC_FILES[@]}"
+  mypy --show-error-codes "${SRC_FILES[@]}"
```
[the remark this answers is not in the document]
Rocamonde: Yes, but it incentivizes users to fix those errors? Since code_checks.sh is only for local development. 

To be fair, I guess you're right and we should just open a PR for that (or for each file) as I don't think anyone will actually read the error messages.

We will replace it with precommit / makefile soon anyways.

src/imitation/algorithms/mce_irl.py
```diff
@@ -176,14 +176,16 @@ def set_pi(self, pi: np.ndarray) -> None:
         self.pi = pi
 
     def _predict(self, observation: th.Tensor, deterministic: bool = False):
-        raise NotImplementedError("Should never be called as predict overridden.")
+        raise NotImplementedError(  # pragma: no cover
```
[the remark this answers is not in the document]
Rocamonde: Yeah probably not, good catch

src/imitation/policies/serialize.py
```diff
@@ -50,24 +51,27 @@ def load_stable_baselines_model(
...
     if vec_normalize_path.exists():
         raise FileExistsError(
             "Outdated policy format: we do not support restoring normalization "
             "statistics from '{vec_normalize_path}'",
         )
 
-    return cls.load(path, env=venv, **kwargs)
+    # TODO(juan) remove the type ignore when this SB3 PR gets merged
+    #  and released:
+    #  https://github.com/DLR-RM/stable-baselines3/pull/1043
```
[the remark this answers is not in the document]
Rocamonde: I think it has, thanks for catching this.

src/imitation/util/util.py
```diff
@@ -136,16 +141,49 @@ def make_env(i, this_seed):
...
+    """Generate n random seeds from a random state.
+
+    Args:
+        rng: The random state to use to generate seeds.
+        n: The number of seeds to generate.
+
+    Returns:
+        A list of n random seeds.
+    """
+    seeds: List[int] = rng.integers(0, (1 << 31) - 1, (n or 1,)).tolist()
```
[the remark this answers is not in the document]
Rocamonde: I think it's a bit strange to request zero seeds from a seed-generating function, but I do agree that the logical result should be an empty list, not a list with one seed, or alternatively throw an error.

src/imitation/scripts/config/train_rl.py
```diff
@@ -6,7 +6,11 @@
 
 train_rl_ex = sacred.Experiment(
     "train_rl",
-    ingredients=[common.common_ingredient, train.train_ingredient, rl.rl_ingredient],
+    ingredients=[
```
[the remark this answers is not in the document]
Rocamonde: It didn't, it's a side-effect of having added too many elements when we temporarily created a new ingredient, then flake8 forces adding a comma at the end, then when you remove the elements black does not reformat the file back due to the magic comma. I think this is actually quite undesirable, and I have been annoyed at flake8's expectation of a final comma everywhere triggering a reformat from black, even if the elements enumerated fit in one line (but not the surrounding brackets, e.g. square or round).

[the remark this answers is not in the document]
Rocamonde: For example:
```python
func_call(list_of_long, arguments_that_all, fit_in_one_line, but_not_also, the_fn_name)
```
black will reformat this to:
```python
func_call(
     list_of_long, arguments_that_all, fit_in_one_line, but_not_also, the_fn_name
)
```
flake8 will require a trailing comma:
```python
func_call(
     list_of_long, arguments_that_all, fit_in_one_line, but_not_also, the_fn_name,
)
```
but then black breaks the entire line:
```python
func_call(
     list_of_long, 
     arguments_that_all, 
     fit_in_one_line, 
     but_not_also, 
     the_fn_name,
)
```
this can be super unreadable sometimes and make the code too vertically long. It also has the problem that if you remove elements from this call (e.g. optionals):
```python
func_call(
     list_of_long, 
)
```
neither flake8 nor black will complain/reformat this.

.circleci/config.yml
```diff
@@ -246,6 +246,10 @@ jobs:
           name: ipynb-check
           command: ./ci/clean_notebooks.py --check ./docs/tutorials
 
+      - run:
+          name: typeignore-check
+          command: ./ci/check_typeignore.py ./src/ ./tests/
```
AdamGleave, line 251: Why not `${SRC_FILES}`? Do we not want to apply this to examples, setup.py, etc? I could see us wanting to be more lax in example code, just some chance we fail to apply it to places we actually want it, especially as we update SRC_FILES over time.

ci/check_typeignore.py
```diff
@@ -0,0 +1,67 @@
...
+@click.command()
+@click.option("--root-dir", type=click.Path(exists=True, file_okay=False, dir_okay=True), default="src")
+def main(root_dir: str):
+    """Checks that no files in our source code have a "#type: ignore" comment without explicitly indicating the
+    reason for the ignore. This is to ensure that we don't accidentally ignore errors that we should be fixing."""
+    files = get_files_to_check(pathlib.Path(root_dir))
+    try:
+        check_files(files)
+    except ValueError as e:
+        print(e)
```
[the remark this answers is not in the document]
AdamGleave: This conversation was resolved but still just doing `print(e)`?

If there's a reason you want to keep it as-is then that's OK but good to leave a comment explaining why.

In general good to avoid resolving comments unless the underlying issue is definitely addressed. Otherwise it makes code reviews a lot harder (I can't just look at the diff since last review and outstanding comments, actually have to track what the issues from the last review were).

ci/clean_notebooks.py
```diff
@@ -38,27 +42,28 @@ def clean_notebook(file: pathlib.Path, check_only=False) -> None:
...
-                raise UncleanNotebookError(f"Notebook {file} has execution count")
-            cell["execution_count"] = None
-            was_dirty = True
         if cell["cell_type"] == "code" and not cell["source"]:
             if check_only:
                 raise UncleanNotebookError(f"Notebook {file} has empty code cell")
             nb.cells.remove(cell)
             was_dirty = True
+        for field, default in fields_defaults:
+            if cell.get(field) != default:
```
AdamGleave, line 57: There's a slight semantic change here which I think wasn't intended. Previously you were OK with `"field" not in cell"` -- but now for outputs/metadata you add the default. This has caused the notebooks to have `"outputs": []` even for text entries that don't usually produce any output.

Probably want to change this to something like `if field in cell and cell.get(field) != default`? Or have a list of permissible values as well as defaults, like:

```python
for field, default, allowed in fields_defaults:
    if cell.get(field) not in allowed:
        # ....
```

where you could include `None` in `allowed` for `"outputs"` even while keeping `[]` as the default.

docs/tutorials/1_train_bc.ipynb
```diff
@@ -3,6 +3,7 @@
   {
    "cell_type": "markdown",
    "metadata": {},
+   "outputs": [],
```
AdamGleave, line 6: Lots of changes like this, was that intended/desired? If not you may want to revert this and other notebooks.

AdamGleave (approved):
LGTM! There's a couple minor changes outlined in comments, most substantive of one is avoiding adding some spurious `"output": []` in notebooks (I think spurious -- if there's a good reason for it happy to change my mind). But I don't think this needs my review again unless you end up making substantial changes.

Thanks for seeing this PR through to the end :)

ci/check_typeignore.py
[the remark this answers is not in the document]
Rocamonde: Apologies about this, I re-wrote this a few times and in my mind I thought it had been resolved. Looking at it now I see I didn't leave a comment about it. 

My opinion was that since all the error messages that can arise are the same and come from the same line, printing the trace back doesn't really add much information, only if it's a bug in the script. 

Do you agree?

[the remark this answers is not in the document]
AdamGleave: Ah, you're imagining this is just the ValueError raised in `check_file`?

I think that's a bit of an assumption -- `ValueError` is used all over Python standard library, it could be raised for a reason you didn't expect.

If you made a custom exception (maybe inheriting from `ValueError` similar to in the CI notebook cleaning) and just caught and printed that I'd be OK with it.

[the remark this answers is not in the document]
Rocamonde: I thought I was using a custom exception? Must have accidentally overwritten the changes. I'll add that in and mark it as resolved if you're happy with that.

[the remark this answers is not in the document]
AdamGleave: Yep, changing to custom exception and marking this as resolved is OK by me.

Rocamonde:
@AdamGleave your comment on ValueError is actually on a previous commit -- if you check the current branch, I have this:

https://github.com/HumanCompatibleAI/imitation/blob/4620ef6ccf88aa931db20cf4e9e12eb1c281a5ee/ci/check_typeignore.py#L81-L84

That's why I marked it as resolved previously

docs/tutorials/1_train_bc.ipynb
Rocamonde: Not intended, this should not be the case, thanks for spotting it.

ci/clean_notebooks.py
Rocamonde: I think the "correct" way to format is to have an empty list for code cells, no entry for any other cell. Not sure if the best way to do this is by allowing different defaults by different cell types. What you're suggesting achieves the same result when using a notebook executor because md cells never add the "outputs" key, but it would not roll it back automatically from e.g. the state it is now.

Rocamonde: Came up with a solution for this, it is somewhat different to what we've discussed, let me know if you want to take a look before I merge.

ci/clean_notebooks.py
```diff
@@ -35,44 +62,56 @@ def clean_notebook(file: pathlib.Path, check_only=False) -> None:
...
+                    )
+                del cell[key]
+                was_dirty = True
+            else:
+                cell_structure = structure[cell["cell_type"]][key]
+                if cell_structure["do"] == "keep":
+                    continue
+                elif cell_structure["do"] == "default":
+                    default_value = cell_structure["value"]
+                    if cell[key] != default_value:
```
AdamGleave, line 90: `"default"` feels like a bit of a misnomer here given you're forcing it to always have that value. That's not a default value, that's the only legal value. Perhaps rename it to something like `"constant"`?

AdamGleave (approved):
LGTM once tests pass -- one suggestion on naming

AdamGleave:
> @AdamGleave your comment on ValueError is actually on a previous commit -- if you check the current branch, I have this:

Ah, that explains it, sorry about that.

PR #535, opened 2022-08-23
Remove loggers option from Ray 2.0

### LINKED ISSUE #533: Incompatible with ray 2.0.0.
## Bug description
Due to the ray 2.0.0. release, code that uses ray is breaking. (to be expected in a major release). Currently #528 will pin the version to 1.x that we were using before, but a decision has to be made whether to upgrade to 2.0.0 (presumably desirable) and make the necessary changes (thus dropping 1.x. support?) or keep the pin for now.

## Steps to reproduce
Clone the repo and run the tests.

Rocamonde:
## Description

Fixes #533.

## Testing

This PR solves broken tests; existing tests suffice.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ if TYPE_CHECKING:
 
 IS_NOT_WINDOWS = os.name != "nt"
 
-PARALLEL_REQUIRE = ["ray[debug,tune]~=1.13.0"]
+PARALLEL_REQUIRE = ["ray[debug,tune]~=2.0.0"]
 PYTYPE = ["pytype==2022.7.26"] if IS_NOT_WINDOWS else []
 if IS_NOT_WINDOWS:
     # TODO(adam): use this for Windows as well once PyPI is at >=1.6.1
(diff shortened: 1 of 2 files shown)


Rocamonde:
@AdamGleave I think the tweak is very minor, do take a look when you can. This is making CI to currently fail in all branches, so worth getting merged as a priority IMO so CI checks are reliable.

AdamGleave (commented):
I'm confused what this has to do with Ray 2.0; isn't the setup.py on this branch still pinned to 1.13?

Rocamonde:
Yes, sorry, I forgot to do that. Will change now.

Rocamonde:
Done. Should be good to review, but no longer urgent given that the version pin is on master now (didn't realize before).

AdamGleave (approved):
LGTM. Maybe we will want to disable the loggers again given the comments seemingly dire warning? But let's see if it annoys people, in general I'm in favor of just logging everything in case it's useful, even if it seems duplicated.

Rocamonde:
I would guess the reason it got removed as an option is because by default it does not log anything and there's another way to customize logging? I tried to find it but don't know how Ray works very well and it seems like they changed up the API quite a bit in the v2 release.

Also agree with the principle you suggest, filtering logs is normally a lesser evil than not logging something, within reason.

PR #537, opened 2022-08-23
Migrate to pathlib

### LINKED ISSUE #531: Switch to using pathlib
Why aren't we using Pathlib?

_Originally posted by @Rocamonde in https://github.com/HumanCompatibleAI/imitation/pull/529#discussion_r951410908_

Rocamonde:
## Description

Fixes #531.

Migrate all path manipulation operations to be done under the pathlib package, a safe and robust way of handling paths accross platforms.

- [x] remove `os.path`
- [x] search any objects working as paths that are `str` and convert them to `pathlib.Path`
- [x] search for `osp`

## Design Principles

 - Pass around `pathlib.Path` objects only (except in some special places) accross internal layers of the API
 - Allow `pathlib.Path`, `str`, `bytes`, or `path-like` and convert with validation to `pathlib.Path` at the highest API level (user level)
 - Paths in config dictionaries to remain stored as strings (for now, until someone checks what these things actually do)
 - Force helper functions to receive absolute paths, construct relative paths explicitly using cwd only at the highest level that is viable.
 - Force all helper functions to either create a directory or check if it exists before operating on the path

The reason for enforcing these checks is:
- [...]: For consistency. If we are going to use `Path` objects, we should use them everywhere that is viable, unless there is a good reason for it
- [...]: Forcing a user to pass a pathlib.Path object can be annoying and seems like an unnecessary requirement. We can have a function that normalizes string paths and makes the necessary checks.
- [...]: Not sure whether those config objects are used in other places that don't accept Path objects, and checking seems complicated, so external help would be valuable
- [...]: Functions that are not managing the working environment should not be making assumptions about where folders should be stored.  It is also easier to debug IO errors if paths are absolute, as one can add a checkpoint and `cd` into the folder directly.
- [...] Explicitly checking if a folder exists or creating is self-documenting of the role of the function.

## Testing

~Current tests should suffice, if the migration is done properly nothing should break. No new features are being added.~

- [ ] Add tests where it is viable to check whether helper functions using paths are disallowing relative paths


Rocamonde:
`imitation.scripts.common.demonstrations:guess_expert_dir()` is trying to assert that the data dir is absolute, but this is not the case (it's a relative dir). When looking up where this path is constructed and whether the relative dir is the current working directory, I found something weird. 

The dagger object passed below:
https://github.com/HumanCompatibleAI/imitation/blob/81d159483f26b764463ebeac4509e0466a612d7a/src/imitation/scripts/train_imitation.py#L99-L108
contains an `expert_policy_path` key that points to `data/expert_models/cartpole_0/policies/final`, but this path never exists during the lifecycle of the tests (not relative to the cwd or to any other directory). The actual location of the policy is in `tests/testadata/`. This is injected through the `TEST_DATA_PATH` when the policy is actually accessed, but this strange path is still getting constructed. 


1. It seems valuable to ensure that paths are absolute wherever possible, as it allows spotting errors like these. 
2. I could force the path to be constructed relative to the cwd somewhere, so that the assertion passes, but this seems strange because I don't actually know what this path would be for or if it would be used in some circumstance/setting.

Rocamonde:
Something closely related to this is what happens below:
https://github.com/HumanCompatibleAI/imitation/blob/81d159483f26b764463ebeac4509e0466a612d7a/src/imitation/scripts/common/demonstrations.py#L37-L49

A similar path is constructed when `config["demonstrations"]["rollout_path"] is None`, but this never happens because this is where the correct path gets injected from the `TEST_DATA_PATH`.

I presume this is some sort of fallback-setting feature that tries to guess where the files are, but cannot find the test corresponding to this feature.

Rocamonde:
I changed the base to mypy because I expect the type improvements and fixes on the mypy PR to really help this PR. I also expect that there would be significant merge conflicts, and would be better to resolve as we go along to avoid chaos.

src/imitation/scripts/common/common.py
```diff
@@ -45,19 +46,20 @@ def update_log_format_strs(log_format_strs, log_format_strs_additional):
...
     if config["common"]["log_dir"] is None:
         env_sanitized = config["common"]["env_name"].replace("/", "_")
-        log_root = config["common"]["log_root"] or "output"
-        log_dir = os.path.join(
-            log_root,
-            command_name,
-            env_sanitized,
-            util.make_unique_timestamp(),
+        assert isinstance(env_sanitized, str)
+        config_log_root = types.parse_optional_path(config["common"]["log_root"])
```
AdamGleave, line 55: This is quite hard to follow. Can we simplify it? E.g. (perhaps this doesn't work verbatim because of type checking?):
```python
log_root = config["common"]["log_root"] or "output"
log_root = types.parse_path(log_root)
```

src/imitation/scripts/common/common.py
```diff
@@ -45,19 +46,20 @@ def update_log_format_strs(log_format_strs, log_format_strs_additional):
...
+        log_root = (
+            config_log_root
+            if config_log_root
+            else types.parse_path("output")  # relative to cwd
         )
-        updates["log_dir"] = log_dir
+        log_dir = log_root / command_name / env_sanitized / util.make_unique_timestamp()
+        updates["log_dir"] = str(log_dir)
```
AdamGleave, line 62: Do we need to do a `str` conversion here? I think everything that takes `log_dir` can accept `Pathlib` as well, though we might need to update type annotations in e.g. `common.make_log_dir` accordingly (not that type checkers will be smart enough to catch that code path).

src/imitation/scripts/eval_policy.py
```diff
@@ -105,13 +104,15 @@ def eval_policy(
         )
 
     if rollout_save_path:
-        types.save(rollout_save_path.replace("{log_dir}", log_dir), trajs)
+
+        types.save(log_dir / rollout_save_path.lstrip("{log_dir}/"), trajs)
```
AdamGleave, line 108: `lstrip()` strips *characters* not substring replacement, I don't think this is what you want:

```python
>>> rollout_save_path='{log_dir}/{log}/{dir}/log/dir/dddrrriiibar'
>>> rollout_save_path.lstrip("{log_dir}/")
'bar'
```

src/imitation/scripts/train_adversarial.py
```diff
@@ -142,22 +141,22 @@ def train_adversarial(
             venv=venv,
             demonstrations=expert_trajs,
             gen_algo=gen_algo,
-            log_dir=log_dir,
+            log_dir=str(log_dir),
```
AdamGleave, line 144: why is this needed? `AdversarialTrainer` takes `log_dir: types.AnyPath`

tests/scripts/test_scripts.py
```diff
@@ -53,7 +53,7 @@
     train_rl,
 ]
 
-TEST_DATA_PATH = pathlib.Path("tests/testdata")
+TEST_DATA_PATH = types.parse_path("tests/testdata")
```
AdamGleave, line 56: What was wrong with just using `pathlib.Path` here?

tests/scripts/test_scripts.py
```diff
@@ -601,7 +601,7 @@ def test_transfer_learning(tmpdir: str) -> None:
     Args:
         tmpdir: Temporary directory to save results to.
     """
-    tmpdir_path = pathlib.Path(tmpdir)
+    tmpdir_path = types.parse_path(tmpdir)
```
AdamGleave, line 604: What was wrong with just using `pathlib.Path` here given `tmpdir` is guaranteed to be a `str`?

tests/scripts/test_scripts.py
```diff
@@ -650,7 +650,7 @@ def test_preference_comparisons_transfer_learning(
         tmpdir: Temporary directory to save results to.
         named_configs_dict: Named configs for preference_comparisons and rl.
     """
-    tmpdir_path = pathlib.Path(tmpdir)
+    tmpdir_path = types.parse_path(tmpdir)
```
AdamGleave, line 653: What was wrong with just using `pathlib.Path` here given `tmpdir` is guaranteed to be a `str`?

tests/scripts/test_scripts.py
```diff
@@ -791,7 +791,7 @@ def test_parallel_arg_errors(tmpdir):
 
 
 def _generate_test_rollouts(tmpdir: str, env_named_config: str) -> pathlib.Path:
-    tmpdir_path = pathlib.Path(tmpdir)
+    tmpdir_path = types.parse_path(tmpdir)
```
AdamGleave, line 794: What was wrong with just using `pathlib.Path` here given `tmpdir` is guaranteed to be a `str`?

tests/scripts/test_scripts.py
```diff
@@ -863,7 +863,7 @@ def _run_train_bc_for_test_analyze_imit(run_name, sacred_logs_dir, log_dir):
     ),
 )
 def test_analyze_imitation(tmpdir: str, run_names: List[str], run_sacred_fn):
-    sacred_logs_dir = tmpdir_path = pathlib.Path(tmpdir)
+    sacred_logs_dir = tmpdir_path = types.parse_path(tmpdir)
```
AdamGleave, line 866: What was wrong with just using `pathlib.Path` here given `tmpdir` is guaranteed to be a `str`?

AdamGleave (changes requested):
Thanks for making this change, seems like a significant improvement to our path handling.

Most of my comments are nits. The one substantive one is to actually add some tests for `types.parse_path` -- shouldn't be too hard.

tests/scripts/test_scripts.py
Rocamonde: parse_path also makes it absolute wrt cwd.

tests/scripts/test_scripts.py
Rocamonde: Same as others.

src/imitation/scripts/eval_policy.py
Rocamonde: Okay yes this is terrible, thanks for pointing it out, I thought it literally matched the entire string.

Rocamonde: Weird that this didnt make any test fail.

src/imitation/data/types.py
```diff
@@ -44,11 +44,87 @@ def dataclass_quick_asdict(obj) -> Dict[str, Any]:
...
+
+    All resulting paths are resolved, absolute paths. If `allow_relative` is True,
+    then relative paths are allowed as input, and are resolved relative to the
+    current working directory, or relative to `base_directory` if it is
+    specified.
+
+    Args:
+        path: The path to parse. Can be a string, bytes, or `os.PathLike`.
+        allow_relative: If True, then relative paths are allowed as input, and
+            are resolved relative to the current working directory. If false,
```
AdamGleave, line 62:
```suggestion
            are resolved relative to the current working directory. If False,
```

src/imitation/data/types.py
```diff
@@ -44,11 +44,87 @@ def dataclass_quick_asdict(obj) -> Dict[str, Any]:
...
```
AdamGleave, line 116:
```suggestion
            are resolved relative to the current working directory. If False,
```

src/imitation/data/types.py
```diff
@@ -420,7 +496,9 @@ def save(path: AnyPath, trajectories: Sequence[Trajectory]):
         ValueError: If the trajectories are not all of the same type, i.e. some are
             `Trajectory` and others are `TrajectoryWithRew`.
     """
-    p = pathlib.Path(path_to_str(path))
+    if isinstance(path, bytes):
```
AdamGleave, line 499: Why not use `parse_path` here?

src/imitation/scripts/common/common.py
AdamGleave: Bump here -- question seems unaddressed

AdamGleave (approved):
LGTM, some small suggestions

src/imitation/scripts/common/common.py
Rocamonde: Added the suggestion.

PR #539, opened 2022-08-24
Optimize macOS CI

### LINKED ISSUE #536: Speed up Mac OS CI
The Mac OS CI executor is currently taking 4x longer than our Linux executor. This seems to be for two main reasons: 
  1. The `brew install` step alone takes 9m, to install just a handful of packages. We might be able to cache this to speed things up. I also question whether some of the packages are necessary.
  2. It's a smaller executor than the others (medium and not x-large). Unfortunately CircleCI only goes up to large for MacOS and that's not available to us without upgrading to an annual plan. I think the best option is to run multiple MacOS executors in parallel and shard the tests across them using e.g. [pytest-shard](https://pypi.org/project/pytest-shard/)

Rocamonde:
Fixes #536.

### DIFF
--- a/tests/scripts/test_scripts.py
+++ b/tests/scripts/test_scripts.py
@@ -10,11 +10,12 @@ import filecmp
 import os
 import pathlib
 import pickle
+import platform
 import shutil
 import sys
 import tempfile
 from collections import Counter
-from typing import List, Mapping, Optional
+from typing import Dict, List, Mapping, Optional
 from unittest import mock
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


Rocamonde:
Why are we installing `ffmpeg` in the CI? That's not what's documented in our own docs. We are however not installing parallel.

Rocamonde:
Removing ffmpeg brings the CI time down by almost 10 minutes, and only causes 2 tests to fail (2 failed, 2548 passed), which is 0.07% of the test suite. We are also currently skipping 588 tests (not sure why). I say we skip those two when on macOS. This will already bring down the macOS CI time by 50%.

tests/scripts/test_scripts.py
```diff
@@ -413,13 +414,21 @@ def test_train_rl_sac(tmpdir):
...
-    {"videos": True},
-    {"videos": True, "video_kwargs": {"single_video": False}},
+# check if platform is macos
+
+EVAL_POLICY_CONFIGS: List[Dict] = [
     {"reward_type": "zero", "reward_path": "foobar"},
     {"rollout_save_path": "{log_dir}/rollouts.pkl"},
 ]
 
+if platform.system() != "Darwin":
```
AdamGleave, line 424: Slight preference for `sys.platform` over `platform.system()` -- on Linux `platform.system()` runs uname and returns fine-grained information which is a bit overkill here. But doesn't really matter: will be same as sys.platform on non-Linux systems anyway I think. https://stackoverflow.com/questions/4553129/when-to-use-os-name-sys-platform-or-platform-system

AdamGleave (approved):
Thanks for working on this. I agree `ffmpeg` is not worth it for 9m delay in tests! We could try just downloading their static build https://evermeet.cx/ffmpeg/ -- I think brew is so slow because it's building stuff -- but not clearly worth it for just a couple of tests.

I think most of the tests being skipped are because we parameterize tests, and some of the parameter combinations are invalid so we check for that and skip them.

PR #525, opened 2022-08-11
Add developer guide

taufeeque9:
## Description

Add a basic developer guide to describe the implementation of the library.

PR #541, opened 2022-08-27
Replace imitation environments with seals

### LINKED ISSUE #501: Add back in `imitation.envs.examples` support to scripts
https://github.com/HumanCompatibleAI/imitation/pull/484 removes the import to `imitation.envs.examples` from `src/imitation/scripts/__init__.py` to workaround https://github.com/sphinx-doc/sphinx/issues/9069

Some options:
  1. Move this example env code out of the repo. It was never a great fit for `imitation` anyway. Perhaps put it in a new contrib repo, or in `seals`.
  2. Workaround the Sphinx bug, e.g. by a relative import or perhaps importing a specific function to call `register()` on.
  3. Document an alternative way to use the example environments. I think something like `env_name=imitation.envs.examples:imitation/Random-v0` might just work? (`gym.make` has some magic to import modules.)

Rocamonde:
## Description

Fixes #501. (Depends on https://github.com/HumanCompatibleAI/seals/pull/58).


AdamGleave:
Note will need to remove mentions of `envs` from the docs (especially once https://github.com/HumanCompatibleAI/imitation/pull/525/ is merged)

Rocamonde:
Tests are passing now. The errors were quite silly:

- The first was that the base POMDP was incrementing the number of steps after the done bool had been computed, so it was always running one step over the horizon.
- The second was that in the notebook examples, I was passing the same environment instance to each parallel vector environment instead of creating a new one.

I think that the seals pull (https://github.com/HumanCompatibleAI/seals/pull/58) should be good to merge, except perhaps for adding some tests. @AdamGleave do you have any suggestions as to what tests we could add? Once we merge that, we replace the version pointer on this PR and we can merge this too.

docs/algorithms/mce_irl.rst
```diff
@@ -13,6 +13,8 @@ Detailed example notebook: :doc:`../tutorials/6_train_mce`
 
     from functools import partial
 
+    from seals import base_envs as envs
```
AdamGleave, line 16: I tend to prefer avoiding renaming imports (one less name to track) unless it's an established convention (e.g. `numpy as np`) or would be overly long otherwise. Does that or another reason apply here?

src/imitation/testing/envs.py
```diff
@@ -1,35 +0,0 @@
-"""Helper methods for tests of custom Gym environments.
-
-This is used in the imitation test suite and may also be useful for users of this
-library.
-"""
-
-import gym
-
-
-def test_model_based(env: gym.Env) -> None:
```
AdamGleave, line 10: Do we have an analogous test in `seals`? If not, perhaps worth adding? It's a rather simple test admittedly but is testing some non-trivial invariants.

AdamGleave (changes requested):
Mostly looks fine -- satisfying how much code we were able to delete :) Definitely was the right move to migrate to `seals`.

I'm not sure about the `from seals import base_envs as envs`, OTOH would prefer it to just be `from seals import base_envs` as `envs` seems sufficiently generic it could easily cause clashes down the line and is another name to keep track of (what's `envs`? oh, it's `seals.base_envs`) but I'm happy to hear arguments to the contrary.

Other suggestion is to beef up the seals testing with some of the code removed in this PR. Obviously not something to resolve in this PR, but I'd feel better if we either decide against that or open an issue/PR in seals to track that before we close this PR and forget about it.

docs/algorithms/mce_irl.rst
Rocamonde: Changed it.

src/imitation/testing/envs.py
Rocamonde: Added an issue for this.

AdamGleave (approved):
LGTM

Issue #530, opened 2022-08-22
Improve bagging for ensemble training


levmckinney:
Is bagging the batch what we want to do for ensemble training. I think we should be sampling independent batches from the dataset for each ensemble member.

_Originally posted by @levmckinney in https://github.com/HumanCompatibleAI/imitation/pull/482#discussion_r948382938_

AdamGleave:
@taufeeque9 you and Lev should figure out which one of you is going to work on this -- don't want you to both spend time on this.

taufeeque9:
I can work on it if @levmckinney  hasn't started it yet. 

The plan is to use [`torch.utils.data.RandomSampler`](https://pytorch.org/docs/stable/data.html#torch.utils.data.RandomSampler) with `replacement=True` in the `_make_data_loader` function when preparing the `DataLoader`. We can further make mini-batches of the data from `RandomSampler` by using [`torch.utils.data.BatchSampler`](https://pytorch.org/docs/stable/data.html#torch.utils.data.BatchSampler).

PR #542, opened 2022-08-29
Improve bagging method for ensemble training (#530)

taufeeque9:
## Description

Improved the bagging method to train ensemble members by sampling examples over the entire preference dataset instead of the earlier implementation of sampling over fixed batches.  Also refactored `PreferenceModel`, `CrossEntropyRewardLoss`, `EnsembleTrainer`, and `_make_reward_trainer`.  

The ensemble member models are trained sequentially using different samples of the dataset. I also tried to train them simultaneously by running the training method in multiple threads. However, this doesn't give much gain and, in fact, increases the overall training time for more than 3 ensemble members. Training simultaneously using multiple processes is too much of a hassle as we cannot directly pass the `_train` method to processes as they require pickling the `self` object. So I have currently let the member models train sequentially, and we can try training them in parallel in another PR if required.

One thing to check is the logging behavior of loss and other metrics for the ensemble members. As far as I understand, the sequential training of the member models will overwrite the logs of the model by the last member model that is trained. But the accumulated mean metrics should be calculated correctly across all the member models.

## Testing

I have separately tested that the sampler returns different indexes across the member model for `bagging_dataset`. No new test case added.


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1198,55 +1198,54 @@ def __init__(
         Raises:
             TypeError: if model is not a RewardEnsemble.
         """
-        if not isinstance(model, reward_nets.RewardEnsemble):
+        if not preference_model.is_ensemble:
             raise TypeError(
-                f"RewardEnsemble expected by EnsembleTrainer not {type(model)}.",
+                "Preference Model of a RewardEnsemble expected by EnsembleTrainer.",
```
AdamGleave, line 1203:
```suggestion
                "PreferenceModel of a RewardEnsemble expected by EnsembleTrainer.",
```

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1198,55 +1198,54 @@ def __init__(
...
+        if seed:
+            self.rng = self.rng.manual_seed(seed)
 
-        self.logger.record("loss", loss.item())
-        self.logger.record("loss_std", losses.std().item())
+    @RewardTrainer.logger.setter
+    def logger(self, custom_logger):
+        self._logger = custom_logger
+        for member_trainer in self.member_trainers:
+            member_trainer._logger = custom_logger
```
AdamGleave, line 1235: We have a setter in `RewardTrainer` so I think no need to access the private variable.
```suggestion
            member_trainer.logger = custom_logger
```

tests/algorithms/test_preference_comparisons.py
```diff
@@ -202,22 +202,22 @@ def test_trainer_no_crash(
...
         discount_factor=0.9,
         threshold=50,
     )
-    loss = preference_comparisons.CrossEntropyRewardLoss(preference_model)
+    loss = preference_comparisons.CrossEntropyRewardLoss()
 
     with pytest.raises(
         TypeError,
-        match=r"RewardEnsemble expected by EnsembleTrainer not .*",
+        match=r"Preference Model of a RewardEnsemble expected by EnsembleTrainer.",
```
AdamGleave, line 217:
```suggestion
        match=r"PreferenceModel of a RewardEnsemble expected by EnsembleTrainer.",
```

AdamGleave (approved):
LGTM, I like the refactor, just left a few suggestions.

As you flagged on Slack, this will result in a different bagging dataset each time train gets called because the dataset continually grows. I don't see a great way around this, so am happy to stick with this implementation for now, and revisit in the future if the empirical results from this method are not strong enough. But wanted to include it in the PR conversation for posterity.

src/imitation/algorithms/preference_comparisons.py
levmckinney, line 1203:
```suggestion
                "PreferenceModel of a RewardEnsemble expected by EnsembleTrainer.",
```

tests/algorithms/test_preference_comparisons.py
```diff
@@ -202,22 +202,22 @@ def test_trainer_no_crash(
     assert 0.0 < result["reward_accuracy"] <= 1.0
 
 
-def test_reward_ensemble_trainer_raises_type_error(venv):
```
levmckinney, line 205: I think the `RewardEnsemble` is still raising the type error in this implementation.

taufeeque9:
The ensemble model now logs the loss and the metrics for each member and reports the mean and the standard deviation of the average of the final epoch metrics. Also added a function to change the `max_length` for truncation of log keys.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1316,6 +1319,34 @@ def _train(self, dataset: PreferenceDataset, epoch_multiplier: float = 1.0) -> N
...
             self.member_trainers[member_idx]._train(bagging_dataset, epoch_multiplier)
+            keys = list(self.logger.name_to_value.keys())
+            for key in keys:
+                val = self.logger.name_to_value[key]
+                if key.startswith("mean/reward/epoch-"):
+                    new_key = key.replace(
+                        "mean/reward",
+                        f"mean/reward/member-{member_idx}",
+                    )
+                    self.logger.record(new_key, val)
```
AdamGleave, line 1330: I'm confused by this code. It seems like we're recording things to a new key (based on the member index), but:
  1. Only the last value logged. (Maybe that's OK, if there's only a single timestep, that is `self.logger.dump` is never called inside `_train`?)
  2. We don't remove the old key.

Why not log with the right key in the first place? You could use the `add_prefix` method from Lev's PR https://github.com/HumanCompatibleAI/imitation/blob/4df85518203e5e83f34864d31ff807813f2a259c/src/imitation/util/logger.py (not yet merged, but you could always cherry-pick the logging commits into a separate PR, get that PR approved & merged, then go back and use it here). Alternatively you could follow the approach Rocamonde took with the regularization PR and pass around a prefix explicitly like in https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/algorithms/preference_comparisons.py#L1223

src/imitation/util/logger.py
```diff
@@ -3,23 +3,70 @@
 import contextlib
 import datetime
 import os
+import sys
 import tempfile
 from typing import Any, Dict, Generator, Optional, Sequence, Tuple, Union
 
-import stable_baselines3.common.logger as sb_logger
+from stable_baselines3.common.logger import (
```
AdamGleave, line 10: Style guide violation, we only import modules not classes (except for `typing`): https://google.github.io/styleguide/pyguide.html#22-imports

src/imitation/util/logger.py
```diff
@@ -3,23 +3,70 @@
...
+    """
+    os.makedirs(log_dir, exist_ok=True)
+    if _format == "stdout":
+        return HumanOutputFormat(sys.stdout, max_length=max_length)
+    elif _format == "log":
+        return HumanOutputFormat(
+            os.path.join(log_dir, f"log{log_suffix}.txt"),
+            max_length=max_length,
+        )
+    elif _format == "json":
```
AdamGleave, line 51: These and following lines are identical to that in `sb_logger.make_output_format` when using `HumanOutputFormat`. We could make code more DRY by having e.g.

```
else:
    return sb_logger.make_output_format(_format, log_dir, log_suffix)
```

WDYT?

tests/util/test_logger.py
```diff
@@ -21,13 +22,36 @@ def _csv_to_dict(csv_path: str) -> dict:
...
+    all_line_dicts = []
+    with open(json_path, "r") as f:
+        for line in f.readlines():
+            line_dict = json.loads(line)
+            all_line_dicts.append(line_dict)
+    all_keys = set().union(*[list(line_dict.keys()) for line_dict in all_line_dicts])
+    all_keys = list(all_keys)
+    for line_dict in all_line_dicts:
+        for key in all_keys:
+            if key in line_dict:
```
AdamGleave, line 36: Could replace if-else by `line_dict.get(key, "")`

tests/util/test_logger.py
```diff
@@ -21,13 +22,36 @@ def _csv_to_dict(csv_path: str) -> dict:
...
 
+def _json_to_dict(json_path: str) -> dict:
+    result = defaultdict(list)
+    all_line_dicts = []
+    with open(json_path, "r") as f:
+        for line in f.readlines():
+            line_dict = json.loads(line)
+            all_line_dicts.append(line_dict)
+    all_keys = set().union(*[list(line_dict.keys()) for line_dict in all_line_dicts])
+    all_keys = list(all_keys)
```
AdamGleave, line 33: Not necessary? (You can iterate over a set.)

tests/util/test_logger.py
```diff
@@ -21,13 +22,36 @@ def _csv_to_dict(csv_path: str) -> dict:
     return result
 
 
+def _json_to_dict(json_path: str) -> dict:
+    result = defaultdict(list)
+    all_line_dicts = []
+    with open(json_path, "r") as f:
+        for line in f.readlines():
```
AdamGleave, line 29: Could replace with list comprehension: `all_line_dicts = [json.loads(line) for line in f.readlines()]`.

tests/util/test_logger.py
```diff
@@ -21,13 +22,36 @@ def _csv_to_dict(csv_path: str) -> dict:
     return result
 
 
+def _json_to_dict(json_path: str) -> dict:
```
AdamGleave, line 25: Took me a couple of minutes to figure out what this function does, consider adding a (brief) docstring?

AdamGleave (changes requested):
Thanks for making these changes. Appreciate the test case, and adding support for different `max_length`.

I'm confused by the way you're adding prefixes for log keys. Left some suggestions for a couple of ways that seem more intuitive to me, although there may be a reason I'm missing why we need to do it another way.

Other comments pretty minor. Please request re-review once addressed.

src/imitation/util/logger.py
taufeeque9: Thanks! I wasn't aware of this one.

src/imitation/util/logger.py
taufeeque9: Great suggestion! Should've thought of that.

src/imitation/algorithms/preference_comparisons.py
taufeeque9: I have currently used Rocamonde's approach to log. Once #529 gets merged, we can switch the current code to using the `add_prefix` method in Lev's PR.

AdamGleave (approved):
LGTM. Thanks for implementing this!

Issue #543, opened 2022-08-30
`ShapedRewardNet` has no effect when calling `predict_processed`


levmckinney:
## Bug description
Calling `predict_processed` on a `ShapedRewardNet` has no effect since it bypasses the shaping applied in the forward call.

## Steps to reproduce
Create a shaped reward network. Observe that the shaping function never gets called when calling the `predict_processed` method.

First mentioned by @dfilan in https://github.com/HumanCompatibleAI/imitation/pull/519#issuecomment-1231004118
> Turns out no potentials were being covered by tests, since the code was calling the predict_processed method of the ShapedRewardNets, which in turn called the base network's predict_processed, which called the base network's forward - not the ShapedRewardNet's forward. Should have a fix up in a few minutes.

dfilan:
Fixed this in the CNN reward functions PR, but should probably make it its own PR. FYI my fix caused some tests to fail, which I think is not my fault but I'm not totally sure.

dfilan:
Also my bad, it was `predict` not `predict_processed`.

levmckinney:
> Also my bad, it was `predict` not `predict_processed`.

I think this bug should apply to both.

dfilan:
Indeed - but it makes more sense for `ShapedRewardNet` to have a `predict` method than a `predict_processed` method.

PR #544, opened 2022-08-30
Fix bug in shaped reward networks

levmckinney:
## Description

Addresses #543 

## Testing
- [x] Add test for shaping actually being applied during predict and predict processed.
- [x] Add test for `ValueError` being raised when you try to shape a wrapped reward net.

### DIFF
--- a/src/imitation/rewards/reward_nets.py
+++ b/src/imitation/rewards/reward_nets.py
@@ -483,18 +483,27 @@ class ShapedRewardNet(RewardNetWrapper):
 
         Args:
             base: the base reward net to which the potential shaping
-                will be added.
+                will be added. Shaping must be applied directly to the raw reward net.
+                See error below.
             potential: A callable which takes
                 a batch of states (as a PyTorch tensor) and returns a batch of
                 potentials for these states. If this is a PyTorch Module, it becomes
                 a submodule of the ShapedRewardNet instance.
             discount_factor: discount factor to use for the potential shaping.
+
+        Raises:
+            ValueError: if the base network is an instance of RewardNetWrapper.
         """
         super().__init__(
             base=base,
         )
         self.potential = potential
         self.discount_factor = discount_factor
+
+        if not isinstance(base, RewardNetWrapper):
+            # Doing this could cause confusing errors like normalization
+            # not being applied.
+            raise ValueError("Shaping cannot be applied to wrapped reward nets.")
 
     def forward(
         self,
@@ -532,6 +541,52 @@ class ShapedRewardNet(RewardNetWrapper):
         )
         assert final_rew.shape == state.shape[:1]
         return final_rew
+
+    def predict_th(
+        self,
+        state: np.ndarray,
+        action: np.ndarray,
+        next_state: np.ndarray,
+        done: np.ndarray,
+    ) -> th.Tensor:
+        __doc__ = super().forward.__doc__  # noqa: F841
+        with networks.evaluating(self):
+            # switch to eval mode (affecting normalization, dropout, etc)
+
+            state_th, action_th, next_state_th, done_th = self.preprocess(
+                state,
+                action,
+                next_state,
+                done,
+            )
+            with th.no_grad():
+                rew_th = self(state_th, action_th, next_state_th, done_th)
+
+            assert rew_th.shape == state.shape[:1]
+            return rew_th
+
+    def predict(
+        self,
+        state: np.ndarray,
+        action: np.ndarray,
+        next_state: np.ndarray,
+        done: np.ndarray,
+    ) -> np.ndarray:
+        __doc__ = super().forward.__doc__  # noqa: F841
+        rew_th = self.predict_th(state, action, next_state, done)
+        return rew_th.detach().cpu().numpy().flatten()
+
+    def predict_processed(
+        self,
+        state: np.ndarray,
+        action: np.ndarray,
+        next_state: np.ndarray,
+        done: np.ndarray,
+        **kwargs,
+    ) -> np.ndarray:
+        __doc__ = super().forward.__doc__  # noqa: F841
+        del kwargs
+        return self.predict(state, action, next_state, done)
 
 
 class BasicShapedRewardNet(ShapedRewardNet):
--- a/src/imitation/rewards/reward_nets.py
+++ b/src/imitation/rewards/reward_nets.py
@@ -542,51 +542,9 @@ class ShapedRewardNet(RewardNetWrapper):
         assert final_rew.shape == state.shape[:1]
         return final_rew
 
-    def predict_th(
-        self,
-        state: np.ndarray,
-        action: np.ndarray,
-        next_state: np.ndarray,
-        done: np.ndarray,
-    ) -> th.Tensor:
-        __doc__ = super().forward.__doc__  # noqa: F841
-        with networks.evaluating(self):
-            # switch to eval mode (affecting normalization, dropout, etc)
-
-            state_th, action_th, next_state_th, done_th = self.preprocess(
-                state,
-                action,
-                next_state,
-                done,
-            )
-            with th.no_grad():
-                rew_th = self(state_th, action_th, next_state_th, done_th)
-
-            assert rew_th.shape == state.shape[:1]
-            return rew_th
-
-    def predict(
-        self,
-        state: np.ndarray,
-        action: np.ndarray,
-        next_state: np.ndarray,
-        done: np.ndarray,
-    ) -> np.ndarray:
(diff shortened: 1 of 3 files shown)


src/imitation/rewards/reward_nets.py
```diff
@@ -533,6 +542,52 @@ def forward(
         assert final_rew.shape == state.shape[:1]
         return final_rew
 
+    def predict_th(
```
AdamGleave, line 545: This looks copy-pasted from `RewardNet`, can we just do `predict_th = RewardNet.predict_th`? I think this works because the `self` will still be different between them. Similar for the following methods.

tests/rewards/test_reward_nets.py
```diff
@@ -531,25 +535,50 @@ def test_add_std_reward_wrapper(
...
-@pytest.mark.parametrize("make_wrapper", MAKE_BASIC_REWARD_NET_WRAPPERS)
+def test_shaped_reward_net(
+    zero_reward_net: MockRewardNet,
+    numpy_transitions: NumpyTransitions,
+):
+    def potential(x: th.Tensor):
+        return th.full((x.shape[0],), 10, device=x.device)
+
+    shaped = reward_nets.ShapedRewardNet(zero_reward_net, potential, 0.9)
+    shaped_rew = th.full((10,), -1, dtype=th.float32)
```
AdamGleave, line 546: add comment explaining why we should expect reward to be `-1`? (I had to think about it for a minute, and many people are less familiar with potential shaping than me.)

AdamGleave (changes requested):
I'd like to reduce code duplication but otherwise this looks fine. Please request re-review once addressed.

src/imitation/rewards/reward_nets.py
levmckinney: This does not work since when doing this it does not automatically curry the method, so you would have to manually pass in self.

dfilan: Can you write something like `def predict_th(self, s, a, s_) = RewardNet.predict_th(self, s, a, s_)`?

AdamGleave: I just made a commit changing it along the lines I suggested and the tests still pass. Feel free to revert if me/the test cases are missing something, but could you give an example of where it breaks?

levmckinney: Okay I think this works. I was adding it as an attribute in the `__init__` method. Which causes errors.

levmckinney: I was working on a slightly different solution. That I think might be a bit more general.

Add two subclasses of `RewardNetWrapper`: `PredictProcessedWrapper` and `ForwardWrapper`. `PredictProcessedWrapper` requires you to override `predict_processed` and defaults to calling the base RewardNet for everything. `ForwardWrapper` requires you to override `forward` cannot be placed on top of a `PredictProcessedWrapper` wrapper and uses pythons default method resolution to call the super class.

Then `ShapedRewardNet` would be a `ForwardWrapper` and all of our other current wrappers would be subclasses of `PredictProcessedWrapper`.

src/imitation/rewards/reward_nets.py
```diff
@@ -213,11 +213,10 @@ def dtype(self) -> th.dtype:
 
 
 class RewardNetWrapper(RewardNet):
-    """An abstract RewardNet wrapping a base network.
+    """Abstract class representing a wrapper that modifies a `RewardNet`'s functionality.
 
-    Note: The wrapper will default to forwarding calls to `device`, `forward`,
-        `preproces`, `predict`, and `predict_processed` to the base reward net unless
-        explicitly overridden in a subclases.
+    In general `RewardNetWrapper`s should either override `ForwardWrapper`
```
dfilan, line 218: is 'override' the right word here? should it instead be 'subclass'? (real question)

src/imitation/rewards/reward_nets.py
dfilan: FWIW I quite like this idea.

src/imitation/rewards/reward_nets.py
levmckinney: It should be "subclass". You create a subclass of a superclass and then _override_ its methods.

tests/rewards/test_reward_nets.py
```diff
@@ -43,10 +43,17 @@ def _potential(x):
     testing_reward_nets.make_ensemble,
 ]
 
+MakeForwardWrapper = Callable[[reward_nets.RewardNet], reward_nets.ForwardWrapper]
+MAKE_PREDICT_PROCESSED_WRAPPERS = [
+    lambda base: reward_nets.NormalizedRewardNet(base, networks.RunningNorm),
+]
```
dfilan, line 49: why is `MakeForwardWrapper` grouped with `MAKE_PREDICT_PROCESSED_WRAPPERS` and vice versa?

src/imitation/rewards/reward_nets.py
```diff
@@ -240,6 +239,46 @@ def __init__(
     def base(self) -> RewardNet:
         return self._base
 
+
+class ForwardWrapper(RewardNetWrapper):
```
AdamGleave, line 243: Do we still want to forward device and dtype? I could see this going either way. Subclasses should probably aim to make them consistent between `base` and the main class anyway I guess.

src/imitation/rewards/reward_nets.py
```diff
@@ -240,6 +239,46 @@ def __init__(
...
+
+
+class PredictProcessedWrapper(RewardNetWrapper):
+    """An abstract RewardNetWrapper that changes the behavior of predict_processed.
+
+    Subclasses should override `predict_processed`. Implementations
+    should pass along `kwargs` to the `base` reward nets `predict_processed` method.
+
+    Note: The wrapper will default to forwarding calls to `device`, `forward`,
+        `preproces` and `predict` to the base reward net unless
```
AdamGleave, line 278:
```suggestion
        `preprocess` and `predict` to the base reward net unless
```

tests/rewards/test_reward_nets.py
AdamGleave: :+1:

tests/rewards/test_reward_nets.py
```diff
@@ -531,30 +520,89 @@ def test_add_std_reward_wrapper(
...
+
+    base.dtype = th.float32
+
+    predict_processed_wrapper._base = base
+    # Check method calls
+    for attr, call_with, return_value in [
+        ("forward", torch_transitions, th.zeros(10)),
+        ("predict_th", numpy_transitions, np.zeros(10)),
+        ("predict", numpy_transitions, np.zeros(10)),
+        ("predict_processed", numpy_transitions, np.zeros(10)),
```
AdamGleave, line 593: I'm confused why this line passes. `PredictProcessedWrapper` does not have a concrete implementation of `predict_processed`, so shouldn't line 597 fail with this?

levmckinney: We're never actually creating instances of `PredictProcessedWrapper` directly since it's an abstract class. The fixture `predict_processed_wrapper` is parametrized, so it actually tests all subclasses of `PredictProcessedWrapper` listed in `MAKE_PREDICT_PROCESSED_WRAPPER`.

Thus, this test just checks that the subclasses of `PredictProcessedWrapper` maintain certain properties. For example calling down to the base classes `predict_processed` method when calling its own `predict_processed` method. Whether it is unreasonable to enforce these for all future subclasses is another question.

dfilan:
FWIW I'm kind of confused re: what about `ForwardWrapper` makes it so that it primarily changes the behaviour of `forward`, and why it looks so different from the code for `PredictProcessedWrapper`.

src/imitation/rewards/reward_nets.py
dfilan: Have addressed this.

levmckinney:
> FWIW I'm kind of confused re: what about `ForwardWrapper` makes it so that it primarily changes the behaviour of `forward`, and why it looks so different from the code for `PredictProcessedWrapper`.

Since the forward wrapper inherits from `RewardNet` without implementing forward, it still an abstract class. Thus, anyone subclassing `ForwardWrapper` will have to implement `forward` if they want to be able to create concrete instances of the class.

The reason it looks so different is that: for the `ForwardWrapepr` we don't want all the methods calls on instances of `ForwardWrapper` to be pass through to their `self.base`.  On the other hand, for the `PredictProcessedWrapper` we generally do.

AdamGleave (approved):
LGTM.

Issue #545, opened 2022-08-30
Account for finite horizon in EMANorm


levmckinney:
## Problem
As pointed out by @yawen-d and @taufeeque9 in https://github.com/HumanCompatibleAI/imitation/pull/522#issuecomment-1224985517 and  https://github.com/HumanCompatibleAI/imitation/pull/522#issuecomment-1225642692 respectively, the current implementation of EMANorm puts too much weight on the first batch, particularly when $\gamma$ is very close to 1.

## Solution
To address this we can use a similar approach to pandas `adjust` parameter in their implementation of ewma https://github.com/pandas-dev/pandas/issues/8861. Currently, EMANorm updates its internal perameters using a fixed learning rate $w_{m,m} = \alpha$. Instead, we can make the learning rate start large and slowly decay to $\alpha$ over time by setting it to be 
$$w_{m,m} = \left(\sum_{k=0}^m \gamma^k\right)^{-1}$$.

 This will reduce how much weight we are putting on the first batch and ensure that each batch has weight proportional to $\gamma^i$ where i is how many batches in the past it is.

Batch EMA and EMV still work with this weighting. I've rewritten out the proofs here [Incremental_batch_EMA_and_EMV.pdf](https://github.com/HumanCompatibleAI/imitation/files/9456540/Incremental_batch_EMA_and_EMV.pdf), latex source code [Incremental batch EMA and EMV.zip](https://github.com/HumanCompatibleAI/imitation/files/9456542/Incremental.batch.EMA.and.EMV.zip).

## Possible alternative solutions
Scrap `EMANorm` and stick to using `RunningNorm`.

PR #546, opened 2022-08-30
EMANorm adjusting learning rate

levmckinney:
## Description

Addresses #545

## Testing

There is already testing for EMANorm though it seems like it will need to be updated to address the new update rule.

### DIFF
--- a/tests/util/test_networks.py
+++ b/tests/util/test_networks.py
@@ -101,7 +101,7 @@ def test_running_norm_identity_train(normalization_layer: Type[networks.BaseNorm
     running_norm.train()  # stats will change in eval mode
     normalized = th.Tensor([-1, -1, -1, -1, 1, 1, 1, 1])  # mean 0, variance 1
     for _ in range(10):
-        th.testing.assert_allclose(
+        th.testing.assert_close(
             running_norm.forward(normalized),
             normalized,
             rtol=0.05,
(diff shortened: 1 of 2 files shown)


tests/util/test_networks.py
```diff
@@ -18,8 +18,8 @@
 ]
 
 
-class EMANormIncremental(networks.EMANorm):
-    """EMA Norm with every new instance added incrementally in a for loop."""
+class EMANormAlgorithm2(networks.EMANorm):
+    """EMA Norm using algorithm 2."""
```
AdamGleave, line 22: Cite the PDF doc? Algorithm 2 is meaningless as a reference without it.

AdamGleave (approved):
LGTM apart from one small suggestion. This adjustment is quite elegant. I did not verify that the derivation of Algorithm 3 still holds when the weights are variable but the proofs looked reasonable.

@taufeeque9 you've written a lot of the surrounding code, do you mind taking a quick look at this stage?

taufeeque9 (approved):
LGTM. 
One minor concern: we initialize `running_mean` and `running_var` as `torch.empty` tensors, which means they have uninitialized data. On the first iteration, we add to these tensors the delta tensors which contain the negative of the respective tensors. So logically, it wouldn't create any problems. But would it be cleaner to initialize the tensors with zeros instead?

levmckinney:
> LGTM. One minor concern: we initialize `running_mean` and `running_var` as `torch.empty` tensors, which means they have uninitialized data. On the first iteration, we add to these tensors the delta tensors which contain the negative of the respective tensors. So logically, it wouldn't create any problems. But would it be cleaner to initialize the tensors with zeros instead?

Good point, don't want to be messing with utilized data. I'll initialize them to be zeros.

Issue #514, opened 2022-08-02
Expand algorithm documentation


AdamGleave:
The algorithm documentation is currently just converted directly from docstrings and is a method-by-method description, e.g. for [behavioral cloning](https://imitation.readthedocs.io/en/latest/algorithms/bc.html). It would be more user friendly to include some high-level description of the algorithm and example usage, either as a separate entry in the guide (Sphinx RST) or perhaps as module-level docstrings.

AdamGleave:
This is fixed now in various PRs.

PR #548, opened 2022-09-01
Add example usage for algorithms in documentation

taufeeque9:
## Description

This PR works on #514 by adding examples for algorithms taken from their respective notebooks in `imitation/examples/`. The examples are shortened by removing arguments that do not override the default values and extra code that's not central for demonstrating the use of the algorithm. Some of the hyperparameters are also changed to reduce the runtime of the examples. 

Side note: The example for density-based reward modeling loads the rollouts from the pickle file in the repo. We may want to remove that and generate the rollouts on-the-fly to make the example self-contained.

## Testing

The examples are tested using `doctest` which is now also part of CI. The `doctest` command is also able to test any example usage lines written in the docstring of any function. For example, [these lines](https://github.com/HumanCompatibleAI/imitation/blob/de36306397610f2db08ded0c921a5b35731b63f8/src/imitation/testing/reward_improvement.py#L28-L35) are now also tested with `doctest` in CI.

Since the docs are built in the `lint` job on CI, I have added the command to test the docs there. This increases the time for lint by around one and a half minutes. We may want to switch it to a unit-test job for syntactic consistency or to reduce the overall time for CI testing.


docs/algorithms/airl.rst
```diff
@@ -14,6 +14,59 @@ Notes
...
+    from imitation.rewards.reward_nets import BasicShapedRewardNet
+    from imitation.util.networks import RunningNorm
+
+    env = gym.make("seals/CartPole-v0")
+    expert = PPO(policy=MlpPolicy, env=env)
+    expert.learn(1000)
+
+    rollouts = rollout.rollout(
+        expert,
+        DummyVecEnv([lambda: RolloutInfoWrapper(gym.make("seals/CartPole-v0"))] * 5),
```
AdamGleave, line 42: I know this is just copied from the notebooks, but I think it'd be better to use `util.make_vec_env` instead here. Both I think `util.make_vec_env("seals/CartPole-v0", n_envs=5)` is simpler (and introduces a useful utility method), and *more correct* as it will seed each of the environments separately and deterministically.

I'd suggest making that change in both the examples and source notebook.

docs/algorithms/airl.rst
```diff
@@ -14,6 +14,59 @@ Notes
...
+        rollout.make_sample_until(min_timesteps=None, min_episodes=60),
+    )
+
+    venv = DummyVecEnv([lambda: gym.make("seals/CartPole-v0")] * 8)
```
AdamGleave, line 46: ditto

docs/algorithms/density.rst
```diff
@@ -2,6 +2,51 @@
...
+    density_trainer = db.DensityAlgorithm(
+        venv=env,
+        demonstrations=rollouts,
+        rl_algo=imitation_trainer,
+    )
+    density_trainer.train()
+
+    novice_stats = density_trainer.test_policy()
+    pprint.pprint(novice_stats)
+    novice_stats_im = density_trainer.test_policy(true_reward=False, n_trajectories=1)
```
AdamGleave, line 34: Why do we specify `n_trajectories` here but not on line 32 or line 42?

docs/algorithms/density.rst
```diff
@@ -2,6 +2,51 @@
...
+    pprint.pprint(novice_stats_im)
+
+    density_trainer.train_policy(100)
+
+    good_stats = density_trainer.test_policy(n_trajectories=1)
```
AdamGleave, line 39: Seems like similar logic to to 32-35 -- can we factor some of this out into a function?

docs/algorithms/gail.rst
```diff
@@ -12,6 +12,61 @@ Notes
...
+    expert = PPO(policy=MlpPolicy, env=env, n_steps=64)
+    expert.learn(1000)
+
+    rollouts = rollout.rollout(
+        expert,
+        DummyVecEnv([lambda: RolloutInfoWrapper(gym.make("seals/CartPole-v0"))] * 5),
+        rollout.make_sample_until(min_timesteps=None, min_episodes=60),
+    )
+
+    venv = DummyVecEnv([lambda: gym.make("seals/CartPole-v0")] * 8)
```
AdamGleave, line 44: Similar comment to AIRL: replace with `util.make_vec_env`

docs/algorithms/gail.rst
```diff
@@ -12,6 +12,61 @@ Notes
...
+    from imitation.rewards.reward_nets import BasicRewardNet
+    from imitation.util.networks import RunningNorm
+
+    env = gym.make("seals/CartPole-v0")
+    expert = PPO(policy=MlpPolicy, env=env, n_steps=64)
+    expert.learn(1000)
+
+    rollouts = rollout.rollout(
+        expert,
+        DummyVecEnv([lambda: RolloutInfoWrapper(gym.make("seals/CartPole-v0"))] * 5),
```
AdamGleave, line 40: Similar comment to AIRL: replace with `util.make_vec_env`

docs/algorithms/preference_comparisons.rst
```diff
@@ -2,6 +2,94 @@
 Preference comparisons
 ======================
 
+The preference comparisons algorithms learns a reward function from the preferences between pairs of 
+trajectories. The reward function is learnt by gathering the preference between trajectories and
```
AdamGleave, line 6:
```suggestion
trajectories. The reward function is learnt by gathering the preferences between trajectories and 
```

docs/algorithms/preference_comparisons.rst
```diff
@@ -2,6 +2,94 @@
 Preference comparisons
 ======================
 
+The preference comparisons algorithms learns a reward function from the preferences between pairs of
```
AdamGleave, line 5: I found this explanation a bit circular: "to predict the trajectory that will get a higher reward" --> aren't we learning the reward? It does make sense but you'd need to explain the difference between learned reward vs latent reward (if human preference comparisons) / ground truth reward (if synthetic).

How about something like:

```
The preference comparison algorithm learns a reward function from preferences between pairs of
trajectories. The comparisons are modeled as being generated from a Bradley-Terry (or Boltzmann rational) model, where the probability of preferring trajectory A over B is proportional to the exponential of the difference between the return of trajectory A minus B. In other words, the difference in returns forms a logit for a binary classification problem, and accordingly the reward function is trained using a cross-entropy loss to predict the preference comparison.

File: docs/algorithms/preference_comparisons.rst [hunk 9]
```diff
@@ -2,6 +2,94 @@
 Preference comparisons
 ======================
 
+The preference comparisons algorithms learns a reward function from the preferences between pairs of 
+trajectories. The reward function is learnt by gathering the preference between trajectories and 
+training a reward network with the cross entropy loss to predict the trajectory that will get a higher reward.
+
+Notes
+-----
+- Preference comparisons paper: `Deep Reinforcement Learning from Human Preferences <https://arxiv.org/pdf/1706.03741.pdf>`_
```
AdamGleave, line 11:
```suggestion
- Our implementation is based on the `Deep Reinforcement Learning from Human Preferences <https://arxiv.org/pdf/1706.03741.pdf>`_ algorithm.
```

AdamGleave (commented):
Thanks for adding this! Looks pretty nearly ready, left a few minor suggestions. Please re-request review once addressed.

docs/algorithms/density.rst
taufeeque9: This is taken as it is from the example notebook. I have added the `n_trajectories` argument on all such lines for now.

docs/algorithms/preference_comparisons.rst
```diff
@@ -2,6 +2,94 @@
...
+        num_iterations=5,
+        fragmenter=fragmenter,
+        preference_gatherer=gatherer,
+        reward_trainer=reward_trainer,
+        seed=0,
+        initial_epoch_multiplier=1,
+    )
+    pref_comparisons.train(total_timesteps=5_000, total_comparisons=200)
+
+    learned_reward_venv = RewardVecEnvWrapper(venv, reward_net.predict)
```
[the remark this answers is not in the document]
taufeeque9: I have made the change. Please check if the current code looks correct.

AdamGleave (approved):
LGTM, thanks for addressing the requested changes.

PR #558, opened 2022-09-12
Clarify error message

PavelCz:
## Description

This is a very minor change, but this tripped me up. I wrongly assumed `n_steps` would just be set to `batch_size`. Hard to see that this assumption is wrong, since e.g. `print_config` doesn't show you `n_steps`.


AdamGleave:
I like this change in principle. Lint is failing due to line being too long. You might want to assign the error message to a temporary variable, or split it across lines using `("foo " "bar")` syntax.

PavelCz:
@AdamGleave Fixed

AdamGleave (approved):
LGTM

PR #564, opened 2022-09-14
Fix n_rounds is not an integer error

taufeeque9:
This fix explicitly converts `n_rounds` to `int` as it can be of `float` type if any of the operands are float. For example, when we set `total_timesteps = 1e5`. The float type gives an error on line 422 in `range` as it requires an integer.

Alternatively, we can also ensure that `total_timesteps` and `gen_train_timesteps` are integers by explicitly converting them if given in the scientific notation.

### DIFF
--- a/src/imitation/scripts/train_adversarial.py
+++ b/src/imitation/scripts/train_adversarial.py
@@ -107,7 +107,7 @@ def train_adversarial(
         `rollout_stats()` on the expert demonstrations.
     """
     # This allows to specify total_timesteps and checkpoint_interval in scientific
-    # notation, which is interpreted as a float by sacred.
+    # notation, which is interpreted as a float by python.
     total_timesteps = int(total_timesteps)
     checkpoint_interval = int(checkpoint_interval)
 
--- a/src/imitation/scripts/train_adversarial.py
+++ b/src/imitation/scripts/train_adversarial.py
@@ -106,6 +106,11 @@ def train_adversarial(
         "monitor_return" key). "expert_stats" gives the return value of
         `rollout_stats()` on the expert demonstrations.
     """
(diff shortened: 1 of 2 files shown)


taufeeque9:
> How about making `train_adversarial` script do `int(total_timesteps)`? I think that should fix this issue and any others that are downstream of `total_timesteps` being the wrong type.

That should also fix the error. As many of the config files use the scientific notation for `total_timesteps`, should I make this change in all of the script files like `train_imitation`, `train_adversarial`, `train_preference_comparisons`, etc?

AdamGleave:
> That should also fix the error. As many of the config files use the scientific notation for total_timesteps, should I make this change in all of the script files like train_imitation, train_adversarial, train_preference_comparisons, etc?

Making the change in all the scripts sounds good to me, thanks!

ernestum:
Looks like this PR was abandoned. I took the liberty to implement what you agreed upon.

ernestum:
@AdamGleave would you review this?

src/imitation/scripts/train_adversarial.py
```diff
@@ -106,6 +106,11 @@ def train_adversarial(
         "monitor_return" key). "expert_stats" gives the return value of
         `rollout_stats()` on the expert demonstrations.
     """
+    # This allows to specify total_timesteps and checkpoint_interval in scientific
+    # notation, which is interpreted as a float by sacred.
```
AdamGleave, line 110: I think it's Python not Sacred that's responsible for treating it as a float.

AdamGleave (approved):
LGTM.

Are no changes needed in `train_imitation.py`?

ernestum:
> Are no changes needed in `train_imitation.py`?

No should have mentioned it. In `train_imitation.py` it is already cast to `int`.

Issue #566, opened 2022-09-14
Remove ntfy[slack] from project dependencies


Rocamonde:
## Problem
1. We are currently not using the `ntfy[slack]` dependency anywhere
2. It adds 100+ packages of additional installation overhead on macOS when installing the `dev` extra

## Solution
Remove the dependency.

PR #567, opened 2022-09-14
Remove ntfy[slack] dependency from setup.py

Rocamonde:
## Description

Solves #566 by removing the unused `ntfy[slack]` dependency in `dev` mode.

## Testing

Tests passing after removing dependency.


AdamGleave (approved):
LGTM

PR #569, opened 2022-09-19
Introduce hypothesis testing to BC tests

ernestum:
## Description

This PR introduces hypothesis tests for BC and does some general cleanup such as introducing a GIVEN, WHEN, THEN structure where appropriate  and renaming tests to make it clear what they test.

## Testing

Na, I did not test the tests ...

### DIFF
--- a/tests/algorithms/test_bc.py
+++ b/tests/algorithms/test_bc.py
@@ -3,77 +3,197 @@
 import dataclasses
 import os
 
+import gym
+import hypothesis
+import hypothesis.strategies as st
 import pytest
 import torch as th
 from stable_baselines3.common import evaluation, vec_env
 from torch.utils import data as th_data
 
+from conftest import get_expert_trajectories
 from imitation.algorithms import bc
 from imitation.data import rollout, types
+from imitation.data.wrappers import RolloutInfoWrapper
 from imitation.testing import reward_improvement
 from imitation.util import logger, util
 
 
-@pytest.fixture(params=[32, 50])
-def batch_size(request):
-    return request.param
-
-
-@pytest.fixture(params=["data_loader", "ducktyped_data_loader", "transitions"])
-def expert_data_type(request):
-    return request.param
-
-
-class DucktypedDataset:
-    """Used to check that any iterator over Dict[str, Tensor] works with BC."""
-
-    def __init__(self, transitions: types.TransitionsMinimal, batch_size: int):
-        """Builds `DucktypedDataset`."""
-        self.trans = transitions
-        self.batch_size = batch_size
-
-    def __iter__(self):
-        for start in range(0, len(self.trans) - self.batch_size, self.batch_size):
-            end = start + self.batch_size
-            d = dict(obs=self.trans.obs[start:end], acts=self.trans.acts[start:end])
-            d = {k: util.safe_to_tensor(v) for k, v in d.items()}
-            yield d
-
-
-@pytest.fixture
-def trainer(
+def make_expert_trajectory_loader(
+    pytestconfig,
     batch_size,
-    cartpole_venv,
     expert_data_type,
-    custom_logger,
-    cartpole_expert_trajectories,
+    env_name: str,
+    max_num_trajectories: int = -1,
 ):
-    trans = rollout.flatten_trajectories(cartpole_expert_trajectories)
+    trajectories = get_expert_trajectories(pytestconfig, env_name)
+    transitions = rollout.flatten_trajectories(
+        trajectories[: max(max_num_trajectories, batch_size)]
+    )
     if expert_data_type == "data_loader":
-        expert_data = th_data.DataLoader(
-            trans,
+        return th_data.DataLoader(
+            transitions,
             batch_size=batch_size,
             shuffle=True,
             drop_last=True,
             collate_fn=types.transitions_collate_fn,
         )
     elif expert_data_type == "ducktyped_data_loader":
-        expert_data = DucktypedDataset(trans, batch_size)
+
+        class DucktypedDataset:
+            """Used to check that any iterator over Dict[str, Tensor] works with BC."""
+
+            def __init__(self, transitions: types.TransitionsMinimal, batch_size: int):
+                """Builds `DucktypedDataset`."""
+                self.trans = transitions
+                self.batch_size = batch_size
+
+            def __iter__(self):
+                for start in range(
+                    0, len(self.trans) - self.batch_size, self.batch_size
+                ):
+                    end = start + self.batch_size
+                    d = dict(
+                        obs=self.trans.obs[start:end], acts=self.trans.acts[start:end]
+                    )
+                    d = {k: util.safe_to_tensor(v) for k, v in d.items()}
+                    yield d
+
+        return DucktypedDataset(transitions, batch_size)
     elif expert_data_type == "transitions":
-        expert_data = trans
+        return transitions
     else:  # pragma: no cover
         raise ValueError(expert_data_type)
 
+
+@pytest.fixture
+def cartpole_bc_trainer(
+    pytestconfig,
+    cartpole_venv,
+    cartpole_expert_trajectories,
+):
     return bc.BC(
         observation_space=cartpole_venv.observation_space,
         action_space=cartpole_venv.action_space,
+        batch_size=50,
+        demonstrations=make_expert_trajectory_loader(
+            pytestconfig,
+            50,
+            "transitions",
+            "seals/CartPole-v0",
+        ),
+        custom_logger=None,
+    )
+
+
+# Note: we don't use the Mujoco envs here because mujoco is not installed on CI.
+envs = st.sampled_from(["Pendulum-v1", "seals/CartPole-v0"])
+batch_sizes = st.integers(min_value=1, max_value=50)
+env_numbers = st.integers(min_value=1, max_value=10)
+loggers = st.sampled_from([None, logger.configure()])
+expert_data_types = st.sampled_from(
+    ["data_loader", "ducktyped_data_loader", "transitions"],
+)
+
+
+@st.composite
+def bc_train_args(draw):
+    args = dict(
+        on_epoch_end=draw(st.sampled_from([None, lambda: None])),
+        on_batch_end=draw(st.sampled_from([None, lambda: None])),
+        log_interval=draw(st.integers(500, 10000)),
+        log_rollouts_n_episodes=draw(st.sampled_from([-1, 2])),
+        progress_bar=draw(st.booleans()),
+        reset_tensorboard=draw(st.booleans()),
+    )
+    duration_measure = draw(st.sampled_from(["n_batches", "n_epochs"]))
+    duration = draw(st.integers(1, 3))
+    args[duration_measure] = duration
+    return args
+
+
+@hypothesis.given(
+    batch_size=batch_sizes,
+    env_name=envs,
+    num_envs=env_numbers,
+    custom_logger=loggers,
+    expert_data_type=expert_data_types,
+)
+# Setting the deadline to none since during the first runs, the expert trajectories must
+# be computed. Later they can be loaded from cache much faster.
+@hypothesis.settings(deadline=None)
+def test_smoke_bc_creation(
+    batch_size,
+    env_name,
+    num_envs,
+    custom_logger,
+    expert_data_type,
+    pytestconfig,
+):
+    # GIVEN
+    env = vec_env.DummyVecEnv(
+        [lambda: RolloutInfoWrapper(gym.make(env_name)) for _ in range(num_envs)],
+    )
+    bc.BC(
+        observation_space=env.observation_space,
+        action_space=env.action_space,
         batch_size=batch_size,
-        demonstrations=expert_data,
+        demonstrations=make_expert_trajectory_loader(
+            pytestconfig,
+            batch_size,
+            expert_data_type,
+            env_name,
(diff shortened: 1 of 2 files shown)


src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+from imitation.util import util
+
+
+def generate_expert_trajectories(
+    env_id: str,
+    num_trajectories: int,
+) -> Sequence[types.TrajectoryWithRew]:
+    """Generate expert trajectories for the given environment.
+
+    Note: will just pull a pretrained policy from the huggingface model hub.
```
AdamGleave, line 26:
```suggestion
    Note: will just pull a pretrained policy from the Hugging Face model hub.
```

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Returns:
+        A list of trajectories with rewards.
+    """
+    env = util.make_vec_env(
+        env_id,
+        post_wrappers=[lambda e, _: wrappers.RolloutInfoWrapper(e)],
+    )
+    trajectories = rollout.rollout(
+        serialize.load_policy("ppo-huggingface", env, env_name=env_id),
```
AdamGleave, line 40: Nit but I'd rather introduce an intermediate variable `policy` to hold this, although only one line it's doing something non-trivial here.

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+        env,
+        rollout.make_sample_until(min_episodes=num_trajectories),
+    )
+    env.close()
```
AdamGleave, line 44: Consider putting this in a `try:`, `finally:` block?

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+    Args:
+        cache_path: A path to the folder to be used as cache for the expert
+            trajectories.
+        env_id: The environment to generate trajectories for.
+        num_trajectories: The number of trajectories to generate.
+
+    Returns:
+        A list of trajectories with rewards.
+    """
+    environment_cache_path = pathlib.Path(cache_path) / hfsb3.EnvironmentName(env_id)
```
AdamGleave, line 64: The `EnvironmentName` is doing some path normalization?

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+    env_id: str,
+    num_trajectories: int,
+) -> Sequence[types.TrajectoryWithRew]:
+    """Generate or load expert trajectories from cache.
+
+    Args:
+        cache_path: A path to the folder to be used as cache for the expert
+            trajectories.
+        env_id: The environment to generate trajectories for.
+        num_trajectories: The number of trajectories to generate.
```
AdamGleave, line 59: Nit: in `generate_expert_trajectories` you describe it as "number of episodes" not "number of trajectories". Both seem OK but should probably be consistent.

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+    environment_cache_path = pathlib.Path(cache_path) / hfsb3.EnvironmentName(env_id)
+    environment_cache_path.mkdir(parents=True, exist_ok=True)
+
+    trajectories_path = environment_cache_path / "rollout.npz"
+
+    with FileLock(environment_cache_path / "rollout.npz.lock"):
+        try:
+            trajectories = types.load_with_rewards(trajectories_path)
+        except (FileNotFoundError, pickle.PickleError):  # pragma: no cover
+            warnings.warn(
```
AdamGleave, line 73: Nitpick, but this error message seems a bit obscure if it's just a `FileNotFoundError`, which will always happen the first time it's called. Maybe have different warning messages for FileNotFound vs pickle?

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+                "Recomputing expert trajectories due to the following error when "
+                "trying to load them:\n" + traceback.format_exc(),
+            )
+            trajectories = generate_expert_trajectories(env_id, num_trajectories)
+            types.save(trajectories_path, trajectories)
+
+    if len(trajectories) >= num_trajectories:
+        return trajectories[:num_trajectories]
+    else:
+        # If it is not enough, just throw clear the cache and generate more.
```
AdamGleave, line 83: "just clear the cache" would also be OK
```suggestion
        # If it is not enough, just throw away the cache and generate more.
```

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+        trajectories_path.unlink()
+        return lazy_generate_expert_trajectories(cache_path, env_id, num_trajectories)
+
+
+def make_expert_transition_loader(
+    cache_dir: Path,
+    batch_size,
```
AdamGleave, line 90:
```suggestion
    batch_size: int,
```

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+    expert_data_type,
```
AdamGleave, line 91:
```suggestion
    expert_data_type: str,
```

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+    env_name: str,
+    num_trajectories: int = 1,
+):
+    """Creates different kinds of pytorch data loaders for expert transitions.
```
AdamGleave, line 95:
```suggestion
    """Creates different kinds of PyTorch data loaders for expert transitions.
```

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+    transitions = rollout.flatten_trajectories(trajectories)
+
+    if len(transitions) < batch_size:
+        # If we have less transitions than the batch size, we estimate the trajectory
+        # length and generate enough trajectories to fill the batch size.
+        trajectory_length = len(transitions) // len(trajectories)
+        min_required_trajectories = math.ceil(batch_size / trajectory_length)
+        transitions = rollout.flatten_trajectories(
+            lazy_generate_expert_trajectories(
```
AdamGleave, line 124: This might return data that overlaps with `trajectories` -- is that OK? (Probably?)

src/imitation/testing/expert_trajectories.py
```diff
@@ -0,0 +1,167 @@
...
+                        acts=self.trans.acts[start:end],
+                    )
+                    d = {k: util.safe_to_tensor(v) for k, v in d.items()}
+                    yield d
+
+        return DucktypedDataset(transitions, batch_size)
+    elif expert_data_type == "transitions":
+        return transitions
+    else:  # pragma: no cover
+        raise ValueError(expert_data_type)
```
AdamGleave, line 167:
```suggestion
        raise ValueError(f"Unexpected data type '{expert_data_type}'")
```

AdamGleave (commented):
Thanks for implementing this! The tests are much easier to understand, and I like the use of hypothesis, should give us more comprehensive coverage. I've mostly just left some minor comments.

I'm a bit unsure about adding some of the additions `imitation.testing`. It seems best to reserve this for generally useful methods, that will either be used across multiple files in our test suite (and doesn't make sense to share as a fixture in `conftest`) or by users of the library. 

I think the `expert_trajectories.py` largely satisfies that, though maybe it could live in `conftest.py` instead as it seems to only be used via some fixtures right now. But `failing_data_loader.py` as far as I can tell is used in just a single test in `test_bc.py` -- why not just define it alongside the test?

AdamGleave:
Also one test is failing on Mac OS (maybe flaky? I don't see anything Mac OS specific). And type checking failing --- I think pytype may be confused by the presence of two `conftest.py` files, adding `__init__.py` to `tests/algorithms` may help with this.

src/imitation/testing/expert_trajectories.py
ernestum: Yes, I introduced it [here](https://github.com/huggingface/huggingface_sb3/pull/13). It basically just removes slashes from environment names and ensures that environment names are normalized in an uniform way.

src/imitation/testing/expert_trajectories.py
ernestum: Yes good idea!

src/imitation/testing/expert_trajectories.py
ernestum: Thanks, fixed!

src/imitation/testing/expert_trajectories.py
ernestum: Good point! Users would probably be clueless.
I tried to give better warnings. Not sure if I like the implementation. What do you think?

ernestum:
> Also one test is failing on Mac OS (maybe flaky? I don't see anything Mac OS specific)

Me neither. It definitively is one of the flaky tests. Should probably be investigated in a separate PR.

ernestum:
> And type checking failing

Yep, thanks for the hint. That fixed it.

ernestum:
> I'm a bit unsure about adding some of the additions `imitation.testing`. It seems best to reserve this for generally useful methods, that will either be used across multiple files in our test suite (and doesn't make sense to share as a fixture in `conftest`) or by users of the library.

I agree. I was unsure where to place all the helpers too. In the end I decided to move them out of the way to improve readability of the tests. Placing them into `conftest.py` was somewhat weird because of the special treatment it gets. I assumed it was reserved for fixtures? Where would you recommend to put it? Maybe just moving the failing data loader back to the test is reasonable ...

src/imitation/testing/expert_trajectories.py
AdamGleave: Yeah, I think this is probably fine as-is, the solutions I can think of would add more complexity than they're worth.

src/imitation/testing/expert_trajectories.py
AdamGleave: Seems fine. I might still include the traceback just for extra info.

AdamGleave:
> Maybe just moving the failing data loader back to the test is reasonable ...

I think this is the least bad option. I don't think it distracts from the tests that much, especially if you add a part of the file designated by a comment header for helper methods.

ernestum:
I moved the data loader back to `test_bc.py`.

The test is only flaky on the MacOS runner. So this will take some more time to investigate since I don't have a MacOS system right now.

tests/algorithms/test_bc.py
```diff
@@ -259,23 +287,41 @@ def test_that_bc_raises_error_when_data_loader_is_empty(
...
     trans = rollout.flatten_trajectories(cartpole_expert_trajectories)
     dummy_yield_value = dataclasses.asdict(trans[:batch_size])
 
-    bad_data_loader = failing_data_loader.DataLoaderThatFailsOnNthIter(
-        dummy_yield_value=dummy_yield_value,
-        # add 1 as BC uses up an iteration from getting the first element
-        # for type checking
-        no_yield_after_iter=no_yield_after_iter + 1,
-    )
+    class DataLoaderThatFailsOnNthIter:
```
AdamGleave, line 295: Our style guide favors classes at top-level except if they're closing over variables: https://google.github.io/styleguide/pyguide.html#26-nestedlocalinner-classes-and-functions

Can you move it? You can prefix it with an underscore to indicate it's private though I don't think it matters given this is in a test file.

AdamGleave (approved):
LGTM -- one minor style suggestion, but no need for me to re-review

tests/algorithms/test_bc.py
ernestum: Thanks for the hint. Turns out I was closing over the dummy yield value.

I updated the implementation to also close over `no_yield_after_iter`. This way I could simplify the  `DataLoaderThatFailsOnNthIter` even more.

I also used existing fixtures to remove some more setup code and added some code to ensure that the training loop actually executed `no_yield_after_iter` many times. This implicitly tests the `DataLoaderThatFailsOnNthIter` to address the concerns in the style guide.

AdamGleave: Ah, I missed it was closing over `dummy_yield_value`. I agree keeping it an inner class makes sense, and this is an elegant solution that simplifies the code :)

tests/algorithms/test_bc.py
```diff
@@ -268,60 +268,49 @@ def test_that_wrong_training_duration_specification_raises_error(
...
-        def __init__(self, no_yield_after_iter: int):
-            """Builds dummy data loader.
-
-            Args:
-                no_yield_after_iter: `__iter__` will raise `StopIteration` after
-                    this many calls.
-            """
-            self.iter_count = 0
-            self.no_yield_after_iter = no_yield_after_iter
+        iter_count = 0
```
AdamGleave, line 297: Nit: Using a class rather than instance variable here is a bit odd. Doesn't matter right now as we only ever have a single instance of the class. But it could be a gotcha if we change the test later, so I lean slightly towards keeping it in `__init__` even though it's more code?

AdamGleave (approved):
LGTM, one minor suggestion to make iter_count instance rather than class variable.

tests/algorithms/test_bc.py
ernestum: Fixed this.

PR #572, opened 2022-10-06
Remove awscli dependency

### LINKED ISSUE #560: `pip install -e .[dev]` installs tons of versions of `awscli`
## Bug description
When installing dependencies in master, seemingly every version of `awscli` is being installed. See [here](https://imgur.com/a/x8JkQt1) for illustration.

## Steps to reproduce
Run the following:
```
git clone git@github.com:HumanCompatibleAI/imitation.git
cd imitation/
virtualenv venv
. venv/bin/activate
pip install -e .[dev]
```

## Environment
- Operating system and version: Ubuntu 22.04.1 LTS
- Python version: 3.8.13 and 3.10.5 (tried both)
- Output of `pip freeze --all`: `pip==22.2.2 setuptools==65.0.2 wheel==0.37.1`

dfilan:
## Description

Fixes #560.

## Testing

Ran all tests. Results: 3543 passed, 595 skipped, 4234 warnings.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -211,7 +211,6 @@ setup(
         # recommended packages for development
         "dev": [
             "autopep8",
-            "awscli",
             "ipdb",
             "isort~=5.0",
             "codespell",


dfilan:
TODO get rid of scripts that download from aws

dfilan:
Having a touch of trouble testing due to #576.

dfilan:
Current status:
- have removed experiment scripts that critically rely on aws
- have retouched `experiments/README.md` accordingly
- have edited experiment config files to match with what's in HuggingFace
- experiment scripts don't currently work, because they don't like the format in which things are saved in HuggingFace

I turned the last bullet point into its own issue (#576), but think it should be addressed separately - IMO it kind of sucks that someone can't just clone this repository and run `pip install .[dev]` but instead has to comment out that bit of `setup.py`.

dfilan:
BTW I kept `imit_benchmark.sh` because I think it makes sense as its own thing, even without the context of `download_models.sh`.

experiments/README.md
```diff
@@ -8,37 +8,19 @@ macOS to install some GNU-compatible binaries before all experiments scripts wil
...
-Demonstrations are used in Phase 3 for imitation learning.
-
-### Phase 3: Train imitation learning.
+### Phase 2: Train imitation learning.
 
 Run `experiments/imit_benchmark.sh --run_name RUN_NAME`. To choose AIRL or GAIL, add the `--airl` and `--gail` flags (default is GAIL).
 
 To analyze these results, run `python -m imitation.scripts.analyze with run_name=RUN_NAME`. Analysis can be run even while training is midway (will only show completed imitation learner's results). [Example output.](https://gist.github.com/shwang/4049cd4fb5cab72f2eeb7f3d15a7ab47)
 
-### Phase 4: Transfer learning.
```
AdamGleave, line 38: Why did you delete this? If it was because `experiments/transfer_benchmark.sh` doesn't exist I think it just got renamed to `experiments/transfer_learn_benchmark.sh`.

AdamGleave (approved):
LGTM apart from deleting one section that I think shouldn't have been.

experiments/README.md
dfilan: I think maybe I deleted it because currently stuff doesn't save to `data/reward_models`. I guess I can reinstate it and leave fixing it for a general "make the experiment scripts work" PR.

dfilan: Have reinstated it.

AdamGleave: Ah yeah that is a problem but agree probably best to punt it to another issue. We should really overhaul the whole `experiments/` directory -- I expect @taufeeque9 will end up doing this as part of his benchmarking work.

PR #577, opened 2022-10-10
Fix policy format error

dfilan:
## Description

There was an error message that clearly should have been an f-string (had a variable in curly brackets inside the string), but wasn't. I made it an f-string.

## Testing

I thought about whether making that string an f-string would break anything, and decided that it wouldn't.


AdamGleave (approved):
LGTM

PR #581, opened 2022-10-18
Allow exploration in eval_policy

dfilan:
## Description

When running the script `eval_policy.py`, this PR lets you add an exploration wrapper to the policy before rolling it out. I want this so that I can train a GAN on a variety of transitions, not just good ones. Not sure if it's useful for anything else.

## Testing

- Ran CI code checks
- Tested running it with a saved policy


src/imitation/scripts/eval_policy.py
```diff
@@ -96,12 +101,19 @@ def eval_policy(
...
-        trajs = rollout.generate_trajectories(
-            expert.get_expert_policy(venv),
-            venv,
-            sample_until,
-            rng=rng,
-        )
+        policy = expert.get_expert_policy(venv)
+        if explore_kwargs is not None:
+            policy = ExplorationWrapper(
+                rollout._policy_to_callable(policy, venv),
```
AdamGleave, line 107: This is using a private method in `rollout` (prefixed with `_`). However it seems we also use it in `preference_comparisons.py` in order to use `ExplorationWrapper`.

One option is to remove the `_` in `rollout.py` -- it seems OK to expose this in the public API, and it's already documented.

I'm also thinking perhaps we should change `ExplorationWrapper` to accept `types.AnyPolicy` instead, and then call `policy_to_callable` internally? We'd still need to expose that method in public API as `ExplorationWrapper` is outside of `rollout`, but this seems better than pushing the responsibility onto the caller.

AdamGleave (changes requested):
Mostly LGTM, but there's some ugliness due to how we've designed `ExplorationWrapper` that forces us to use a private method in rollout to convert the policy into a callable. Ideally we'd fix that before adding more instances of this anti-pattern in the codebase; I've outlined a way we could do that in the inline comment.

src/imitation/scripts/eval_policy.py
dfilan: Second idea sounds good to me, am implementing that.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -189,17 +189,12 @@ def __init__(
...
-            algo_venv,
-            # By setting deterministic_policy to False, we ensure that the rollouts
-            # are collected from a deterministic policy only if self.algorithm is
-            # deterministic. If self.algorithm is stochastic, then policy_callable
-            # will also be stochastic.
-            deterministic_policy=False,
-        )
+        # This wrapper ensures that rollouts are only collected from a deterministic
+        # policy (when not in exploration mode) if self.algorithm is deterministic.
+        # If self.algorithm is stochastic, then self.exploration_wrapper will also be
```
AdamGleave, line 194: I found this a bit confusing -- isn't `self.exploration_wrapper` *always* stochastic? It's also a bit apropos of nothing now we don't have the `deterministic_policy` keyword argument to anchor this.

I think we can probably simplify this? E.g. a rough attempt:
```
# This wrapper collects rollouts from a mixture of `self.algorithm` and a policy that acts randomly.
# The samples from `self.algorithm` are themselves stochastic if `self.algorithm` is stochastic.
# Otherwise, they are determinsitic, and action selection is only stochastic when sampling from the random policy.
```

src/imitation/policies/exploration_wrapper.py
```diff
@@ -20,22 +20,26 @@ class ExplorationWrapper:
...
         Args:
-            policy_callable: The policy callable to randomize.
+            policy: The policy to randomize.
             venv: The environment to use (needed for sampling random actions).
             random_prob: The probability of picking the random policy when switching.
             switch_prob: The probability of switching away from the current policy.
             rng: The random state to use for seeding the environment and for
                 switching policies.
+            deterministic_policy: Whether to make the policy deterministic when not
+                exploring. If policy is a PolicyCallable, this must be False.
```
AdamGleave, line 40:
```suggestion
                exploring. This must be False when ``policy`` is a ``PolicyCallable``.
```

AdamGleave (approved):
LGTM -- couple of minor suggestions

src/imitation/algorithms/preference_comparisons.py
dfilan: Your version is better - will slightly modify and go with that.

PR #584, opened 2022-10-21
Detect flaky tests

Rocamonde:
## Description

Add support for detecting flaky tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -377,6 +377,30 @@ jobs:
           path: \tmp\resource-usage
           destination: resource-usage
 
+  flakiness-benchmark:
+    executor: unit-test-linux
+    parallelism: 4
+    steps:
+      - dependencies-linux
+      - restore-pytest-cache
+      - run:
+          name: run flakiness benchmark
+          command: |
+            pip install pytest-flakefinder pytest-xdist
+            TESTFILES=$(circleci tests glob tests/**/test*.py | circleci tests split --split-by=timings)
+            Xdummy-entrypoint.py pytest -n auto \
+#                   --cov=/venv/lib/python3.8/site-packages/imitation \
+#                   --cov=tests \
+                   --junitxml=/tmp/test-reports/junit.xml \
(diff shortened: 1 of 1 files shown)


AdamGleave:
Thanks for looking into this!

We probably don't want to run this benchmark on every commit given it's expense, right?

So should either find some way of gating it in imitation (maybe only run it on  tagged commits, so something that's triggered each time we make a release), or if the intention is to just use this PR to do flakiness detection on CircleCI (because you can't replicate it locally) then consider marking the PR as a draft.

Rocamonde:
The idea was to trigger this only on master or on releases, plus on this branch while it does not get merged. My guess is that there is a hook set up on the repo settings that automatically triggers all workflows for all branches by default (since there are no triggers set up on circleci). I would be happy to configure this if I am granted admin rights for the repo.

The ridiculously long runtime is because I probably ran the first batch with too big a sample size. Thinking about it more, 100 samples is probably enough. The variance of the mean of $n$ Bernoulli trials is $p(1-p)/n$, and for 100 sample sizes the standard deviation of the mean is upper bounded by 5%. I also don't expect most flaky tests to fail e.g. 1% of the time, but more like e.g. >5% of the time, for which there's >99% probability of having at least one failed test with 100 samples. Maybe we can do 150 or 200 if you think 100 is too low.

We can have a more intense checks with e.g. 1000 samples when we have e.g. major releases by parallelizing the tests over more containers.

I am also getting errors of processes crashing, I'll have to look into that in more detail.

AdamGleave:
> The idea was to trigger this only on master or on releases, plus on this branch while it does not get merged. My guess is that there is a hook set up on the repo settings that automatically triggers all workflows for all branches by default (since there are no triggers set up on circleci). I would be happy to configure this if I am granted admin rights for the repo.

You can configure this in `.circleci/config.yml` using the `filter` keyword: see https://support.circleci.com/hc/en-us/articles/115015953868-Filter-workflows-by-branch- You shouldn't need admin rights for it, but let me know if you run into any problems.

PR #586, opened 2022-10-23
Remove type ignore for `forward()` override

### LINKED ISSUE #553: Remove type ignore once DLR-RM/stable-baselines3#1027 is fixed
Remove type: ignore[override] from `forward()` overrides in BaseModel subclasses once https://github.com/DLR-RM/stable-baselines3/issues/1027 is fixed.

Rocamonde:
## Description

Closes #553. Also removes instantiation from `NotImplementedError` from some forward calls since no error message is passed.


AdamGleave (approved):
LGTM

PR #588, opened 2022-10-24
docs: add Prerequisites section in installation instructions

ianyfan:
This is useful information, especially for developing, that is currently only visible in the setup file itself.

To do:
- [x] add the same information in the documentation
- [x] any other requirements that should be added? The only thing I had to check when I installed was OpenGL, but not sure whether that is essential, or just for visualisation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,8 @@ You can find [the documentation here](https://imitation.readthedocs.io/en/latest
 ### Prerequisites
 
 - Python 3.8+
-- (Optional) OpenGL (to render gym environments)
+- (Optional) OpenGL (to render Gym environments)
+- (Optional) FFmpeg (to encode videos of renders)
 - (Optional) MuJoCo (follow instructions to install [mujoco_py v1.5 here](https://github.com/openai/mujoco-py/tree/498b451a03fb61e5bdfcb6956d8d7c881b1098b5#install-mujoco))
 
 ### Installing PyPI release
--- a/README.md
+++ b/README.md
@@ -23,6 +23,12 @@ You can find [the documentation here](https://imitation.readthedocs.io/en/latest
 
 ## Installation
 
+### Prerequisites
+
+- Python 3.8+
+- (Optional) OpenGL (to render gym environments)
+- (Optional) MuJoCo (follow instructions to install [mujoco_py v1.5 here](https://github.com/openai/mujoco-py/tree/498b451a03fb61e5bdfcb6956d8d7c881b1098b5#install-mujoco))
+
 ### Installing PyPI release
 
 Installing the PyPI release is the standard way to use `imitation`, and the recommended way for most users.
@@ -55,10 +61,6 @@ For macOS users, some packages are required to run experiments (see `./experimen
 ```
 brew install coreutils gnu-getopt parallel
 ```
-
-### Optional Mujoco Dependency
-
-Follow instructions to install [mujoco_py v1.5 here](https://github.com/openai/mujoco-py/tree/498b451a03fb61e5bdfcb6956d8d7c881b1098b5#install-mujoco).
 
 ## CLI Quickstart
 
--- a/docs/getting-started/installation.rst
+++ b/docs/getting-started/installation.rst
@@ -2,6 +2,21 @@
 Installation
 ============
 
+Prerequisites
+-------------
+
+- Python 3.8+
+
+- (Optional) OpenGL (to render gym environments)
+
+- (Optional) MuJoCo (follow instructions to install `mujoco\_py v1.5 here`_)
+
+.. _mujoco_py v1.5 here:
+    https://github.com/openai/mujoco-py/tree/498b451a03fb61e5bdfcb6956d8d7c881b1098b5#install-mujoco
+
+
+Installation from PyPI
+----------------------
 
 To install the latest PyPI release, simply run:
 
@@ -10,18 +25,13 @@ To install the latest PyPI release, simply run:
     pip install imitation
 
 
-Alternatively, you can install the latest version of ``imitation`` from source. This is useful if you wish to contribute to the development of ``imitation``, or if you need features that have not yet been made available in a stable release:
+Installation from source
+------------------------
+
+Installation from source is useful if you wish to contribute to the development of ``imitation``, or if you need features that have not yet been made available in a stable release:
 
 .. code-block:: bash
 
     git clone http://github.com/HumanCompatibleAI/imitation
     cd imitation
     pip install -e .
-
-
-**Optional Mujoco Dependency**
-
-Follow instructions to install `mujoco\_py v1.5 here`_.
-
-.. _mujoco_py v1.5 here:
-    https://github.com/openai/mujoco-py/tree/498b451a03fb61e5bdfcb6956d8d7c881b1098b5#install-mujoco
--- a/docs/getting-started/installation.rst
+++ b/docs/getting-started/installation.rst
@@ -6,9 +6,8 @@ Prerequisites
 -------------
 
 - Python 3.8+
-
 - (Optional) OpenGL (to render gym environments)
-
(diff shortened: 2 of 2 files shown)


AdamGleave:
Most of our requirements are Python dependencies, so should get pulled in automatically during installation. It wouldn't hurt to briefly document some of our key Python dependencies anyway, like Gym, seals and Stable Baselines3.

We have an optional MuJoCo binary dependency -- that's already documented under "Optional Mujoco Dependency" in README, but you could probably combine those sections.

I don't think we have much else in the way of unusual dependencies. Our [Dockerfile](https://github.com/HumanCompatibleAI/imitation/blob/master/Dockerfile) is one way to check for this (what do you need on top of default Ubuntu 20.04), but it's mostly just GL stuff + X (which is needed only for rendering anyway). Oh, and ffmpeg for recording videos of the RL agents.

README.md
```diff
@@ -23,6 +23,12 @@ You can find [the documentation here](https://imitation.readthedocs.io/en/latest
 
 ## Installation
 
+### Prerequisites
+
+- Python 3.8+
+- (Optional) GLU (to render gym environments)
```
AdamGleave, line 29: This is a bit opaque, can we add a link? I think GLU usually refers to https://en.wikipedia.org/wiki/OpenGL_Utility_Library -- do we really depend on this, I thought it got deprecated a while back?

Looking at our Dockerfile we have:
```
    ffmpeg \
    libgl1-mesa-dev \
    libgl1-mesa-glx \
    libglew-dev \
    libosmesa6-dev \
```
Which seems like a few different GL packages plus ffmpeg

docs/getting-started/installation.rst
```diff
@@ -2,6 +2,21 @@
 Installation
 ============
 
+Prerequisites
+-------------
+
+- Python 3.8+
+
+- (Optional) GLU (to render gym environments)
```
AdamGleave, line 10: Same comment as above about GLU.

AdamGleave (changes requested):

README.md
ianyfan: It was what was missing on my system when I tried to run the tests initially, but I'll check again

ianyfan: This is the issue I get (with a reproducible example using gym, for added context):
```bash
>>> import gym; gym.make('CartPole-v0').render()
Traceback (most recent call last):
  File "/home/i/venv/lib/python3.10/site-packages/gym/envs/classic_control/rendering.py", line 27, in <module>
    from pyglet.gl import *
  File "/home/i/venv/lib/python3.10/site-packages/pyglet/gl/__init__.py", line 95, in <module>
    from pyglet.gl.gl import *
  File "/home/i/venv/lib/python3.10/site-packages/pyglet/gl/gl.py", line 45, in <module>
    from pyglet.gl.lib import link_GL as _link_function
  File "/home/i/venv/lib/python3.10/site-packages/pyglet/gl/lib.py", line 149, in <module>
    from pyglet.gl.lib_glx import link_GL, link_GLU, link_GLX
  File "/home/i/venv/lib/python3.10/site-packages/pyglet/gl/lib_glx.py", line 46, in <module>
    glu_lib = pyglet.lib.load_library('GLU')
  File "/home/i/venv/lib/python3.10/site-packages/pyglet/lib.py", line 168, in load_library
    raise ImportError('Library "%s" not found.' % names[0])
ImportError: Library "GLU" not found.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/i/venv/lib/python3.10/site-packages/gym/core.py", line 295, in render
    return self.env.render(mode, **kwargs)
  File "/home/i/venv/lib/python3.10/site-packages/gym/envs/classic_control/cartpole.py", line 179, in render
    from gym.envs.classic_control import rendering
  File "/home/i/venv/lib/python3.10/site-packages/gym/envs/classic_control/rendering.py", line 29, in <module>
    raise ImportError(
ImportError: 
    Error occurred while running `from pyglet.gl import *`
    HINT: make sure you have OpenGL installed. On Ubuntu, you can run 'apt-get install python-opengl'.
    If you're running on a server, you may need a virtual frame buffer; something like this should work:
    'xvfb-run -s "-screen 0 1400x900x24" python <your_script.py>'
```
So GLU seems to be the library they are looking for, but clearly, and solutions in https://stackoverflow.com/questions/50446867/importerror-library-glu-not-found suggest installing the GLU library.

It appears that GLU itself is not deprecated, but features it depends on are deprecated in OpenGL. (a bit more discussion here: https://stackoverflow.com/questions/41580315/are-glutess-functions-deprecated )

Clearly, the docker setup doesn't seem to require installing GLU per se, but since it is quite old, I would be inclined to see whether anything requires updating, but that would take some time which I don't want to spend fiddling with Docker right now.

I'd suggest an alternative solution to change "GLU" to "OpenGL", which is more transparent, albeit not the complete picture, but leaving it open to be made more precise in a future change.

ianyfan: This is the issue I get (with a reproducible example using gym, for added context):
```bash
>>> import gym; gym.make('CartPole-v0').render()
Traceback (most recent call last):
  File ".../lib/python3.10/site-packages/gym/envs/classic_control/rendering.py", line 27, in <module>
    from pyglet.gl import *
  File ".../lib/python3.10/site-packages/pyglet/gl/__init__.py", line 95, in <module>
    from pyglet.gl.gl import *
  File ".../lib/python3.10/site-packages/pyglet/gl/gl.py", line 45, in <module>
    from pyglet.gl.lib import link_GL as _link_function
  File ".../lib/python3.10/site-packages/pyglet/gl/lib.py", line 149, in <module>
    from pyglet.gl.lib_glx import link_GL, link_GLU, link_GLX
  File ".../lib/python3.10/site-packages/pyglet/gl/lib_glx.py", line 46, in <module>
    glu_lib = pyglet.lib.load_library('GLU')
  File ".../lib/python3.10/site-packages/pyglet/lib.py", line 168, in load_library
    raise ImportError('Library "%s" not found.' % names[0])
ImportError: Library "GLU" not found.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File ".../lib/python3.10/site-packages/gym/core.py", line 295, in render
    return self.env.render(mode, **kwargs)
  File ".../lib/python3.10/site-packages/gym/envs/classic_control/cartpole.py", line 179, in render
    from gym.envs.classic_control import rendering
  File ".../lib/python3.10/site-packages/gym/envs/classic_control/rendering.py", line 29, in <module>
    raise ImportError(
ImportError: 
    Error occurred while running `from pyglet.gl import *`
    HINT: make sure you have OpenGL installed. On Ubuntu, you can run 'apt-get install python-opengl'.
    If you're running on a server, you may need a virtual frame buffer; something like this should work:
    'xvfb-run -s "-screen 0 1400x900x24" python <your_script.py>'
```
So GLU seems to be the library they are looking for, but clearly, and solutions in https://stackoverflow.com/questions/50446867/importerror-library-glu-not-found suggest installing the GLU library.

It appears that GLU itself is not deprecated, but features it depends on are deprecated in OpenGL. (a bit more discussion here: https://stackoverflow.com/questions/41580315/are-glutess-functions-deprecated )

Clearly, the docker setup doesn't seem to require installing GLU per se, but since it is quite old, I would be inclined to see whether anything requires updating, but that would take some time which I don't want to spend fiddling with Docker right now.

I'd suggest an alternative solution to change "GLU" to "OpenGL", which is more transparent, albeit not the complete picture, but leaving it open to be made more precise in a future change.

AdamGleave: Changing this to OpenGL seems reasonable. GLU depends on OpenGL after all, so just listing GLU on its own seems incomplete. The most helpful thing here would probably be to list what packages to install on common operating systems, but that'll be tricky to test/maintain, so keeping it vague for now is likely best.

AdamGleave (approved):
LGTM

PR #589, opened 2022-10-25
DAgger: remove redundant trajectory save-handling code

### LINKED ISSUE #555: Replace `_save_dagger_demo` by `types.save`
## Problem
In `algorithms/dagger.py`, there is a function called `_save_dagger_demo` that saves trajectories in `.npz` format. This is redundant with the `save` function in `data/types.py` which also saves trajectories in `.npz` format. Apparently when `_save_dagger_demo` was written, `types.save` used `pickle` rather than NPZ, and only accepted `TrajectoryWithRew`s, both of which are no longer true.

## Solution
Rewrite `_save_dagger_demo` to use `types.save`.

## Possible alternative solutions
Make it more clear why `_save_dagger_demo` is importantly different to `types.save`, if it indeed is.

ianyfan:
## Description

Fixes #555. This removes the redundant `_load_trajectory` function and delegates the save-handling in `_save_dagger_demo` to `types.save`. However, it still contains the extra logic to determine the file name, and I'm not sure whether it can be removed, since the function is called in two places, so may result in duplicate code.

## Testing

All the tests in `test_dagger.py` still pass with this change.

### DIFF
--- a/src/imitation/algorithms/dagger.py
+++ b/src/imitation/algorithms/dagger.py
@@ -7,7 +7,6 @@ policy.
 """
 
 import abc
-import dataclasses
 import logging
 import os
 import pathlib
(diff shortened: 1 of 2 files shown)


AdamGleave (approved):
LGTM. I agree keeping `_save_dagger_demo` is worth it given the filename generation logic.

I do wonder if we really want to even stick to the one file per demo structure v.s. maintain a buffer and periodically save it all, but that's out of scope for this PR.

PR #590, opened 2022-10-25
Small typing fixes

ianyfan:
## Description

This fixes some typing issues that arose on a fresh install:

- pytype fails in `preference_comparison.py`: https://github.com/HumanCompatibleAI/imitation/blob/944edce059d2278324b105df6b453a1be039440d/src/imitation/algorithms/preference_comparisons.py#L773 since `var_estimate` is defined in the previous code branches, except one, which calls `raise_uncertainty_on_not_supported` which raises an error, and this requires an explicit type annotation to indicate that
- mypy should be updated to version 0.981 since we don't target Python 3.7 (ref https://github.com/python/mypy/pull/13500)

This is a very small PR, so I can add more typing fixes where I notice them to add some more substance, if desired.

## Testing

`ci/code_checks.sh`, which failed on a fresh install, now passes

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ TESTS_REQUIRE = (
         # TODO: upgrade jupyter-client once
         #  https://github.com/jupyter/jupyter_client/issues/637 is fixed
         "jupyter-client~=6.1.12",
-        "mypy==0.971",
+        "mypy~=0.981",
         "pandas~=1.4.3",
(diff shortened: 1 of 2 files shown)


AdamGleave (approved):
LGTM.

Small PRs are fine -- preferable, in fact, easy to review. 3 lines is admittedly on the smaller end of a small PR, but feel free to just send these across as they come up.

PR #591, opened 2022-10-25
Save `.npz` files with the correct extension

### LINKED ISSUE #556: `.npz`-formatted files saved with `.pkl` extension
## Bug description
The `save` function in `data/types.py` saves sequences of trajectories in numpy's `.npz` format. That said, some scripts save these with the name `final.pkl` or similar, making it appear that the files are pickle-formatted. Fixing this basically seems to break a few other things (see my recent reverted commits in https://github.com/HumanCompatibleAI/imitation/pull/519), and patching things over would steal focus away from my main task right now.

## Steps to reproduce
Run `python -m imitation.scripts.train_rl with pendulum common.fast train.fast rl.fast fast common.log_dir=quickstart/rl/`.

ianyfan:
## Description

Fixes #556 by checking every occurrence of `.pkl` in the repo and renaming where necessary. Hopefully everything is correct. (the tests pass, at least)


AdamGleave (commented):
Largely LGTM.

One question: how did you generate the `final.npz` files in `testdata`? Since they're used in a bunch of tests it's good if we don't change the contents too much (it's fine if exact demos are different, but we probably want to have the same or higher return so the "expert" data doesn't become weaker). We have `src/imitation/scripts/convert_trajs.py` if you need (perhaps you used this).

There's a merge conflict now -- that's my fault, I think, I just merged a PR that fixed a bunch of mypy type issues before I reviewed this.

Once merge conflict addressed + question above answered tag me for a re-review.

ianyfan:
Rebased, just had to fix `src/imitation/scripts/train_preference_comparisons.py`

AdamGleave (approved):
LGTM

PR #592, opened 2022-10-25
Replace `OrderedDict` with `dict` where possible.

ianyfan:
## Description

Since Python 3.7, `dict` is guaranteed to preserve insertion order. This makes `OrderedDict` mostly redundant, except where:
- it is used as an argument to `torch.nn.Sequential` which only accepts `OrderedDict`
- equality in insertion order matters (`dict` compares equally if the items are the same, regardless of the order they are inserted)

## Testing

All tests still pass.

### DIFF
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -1,19 +1,8 @@
 """Core code for adversarial imitation learning, shared between GAIL and AIRL."""
 import abc
-import collections
 import dataclasses
 import logging
-from typing import (
-    Callable,
-    Iterable,
-    Iterator,
-    Mapping,
-    Optional,
-    Sequence,
-    Tuple,
-    Type,
-    overload,
-)
+from typing import Callable, Iterable, Iterator, Mapping, Optional, Type, overload
 
 import numpy as np
 import torch as th
@@ -81,23 +70,20 @@ def compute_train_stats(
         label_dist = th.distributions.Bernoulli(logits=disc_logits_expert_is_high)
         entropy = th.mean(label_dist.entropy())
 
-    pairs = [
-        ("disc_loss", float(th.mean(disc_loss))),
-        # accuracy, as well as accuracy on *just* expert examples and *just*
-        # generated examples
-        ("disc_acc", float(acc)),
-        ("disc_acc_expert", float(expert_acc)),
-        ("disc_acc_gen", float(generated_acc)),
+    return {
+        "disc_loss": float(th.mean(disc_loss)),
+        "disc_acc": float(acc),
+        "disc_acc_expert": float(expert_acc),  # accuracy on just expert examples
+        "disc_acc_gen": float(generated_acc),  # accuracy on just generated examples
         # entropy of the predicted label distribution, averaged equally across
         # both classes (if this drops then disc is very good or has given up)
-        ("disc_entropy", float(entropy)),
+        "disc_entropy": float(entropy),
         # true number of expert demos and predicted number of expert demos
-        ("disc_proportion_expert_true", float(pct_expert)),
-        ("disc_proportion_expert_pred", float(pct_expert_pred)),
-        ("n_expert", float(n_expert)),
-        ("n_generated", float(n_generated)),
-    ]  # type: Sequence[Tuple[str, float]]
-    return collections.OrderedDict(pairs)
+        "disc_proportion_expert_true": float(pct_expert),
+        "disc_proportion_expert_pred": float(pct_expert_pred),
+        "n_expert": float(n_expert),
+        "n_generated": float(n_generated),
+    }
 
 
 class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
--- a/src/imitation/scripts/analyze.py
+++ b/src/imitation/scripts/analyze.py
@@ -1,6 +1,5 @@
 """Commands to analyze experimental results."""
 
-import collections
 import itertools
 import json
 import logging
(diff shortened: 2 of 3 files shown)


src/imitation/algorithms/adversarial/common.py
```diff
@@ -33,7 +22,7 @@ def compute_train_stats(
     disc_logits_expert_is_high: th.Tensor,
     labels_expert_is_one: th.Tensor,
     disc_loss: th.Tensor,
-) -> Mapping[str, float]:
+) -> Dict[str, float]:
```
AdamGleave, line 25: We've usually used `Mapping` over `Dict` for return values as it is more generic so gives us greater implementation flexibility in the future. Open to arguments to change this though!

src/imitation/scripts/analyze.py
```diff
@@ -196,34 +194,26 @@ def _return_summaries(sd: sacred_util.SacredDicts) -> dict:
     )
 
 
-sd_to_table_entry_type = Mapping[str, Callable[[sacred_util.SacredDicts], Any]]
+sd_to_table_entry_type = Dict[str, Callable[[sacred_util.SacredDicts], Any]]
```
AdamGleave, line 197: As before, we've usually used `Mapping` over `Dict` where we don't need `dict`-specific features

AdamGleave (commented):
Mostly LGTM, my comments are just about whether we really want to replace more generic `Mapping` with more specific `Dict`?

Reduction in code coverage seems to be spurious so we can override.

src/imitation/algorithms/adversarial/common.py
ianyfan: My rule of thumb is to type inputs (i.e. function parameters) as generally as possible and outputs (i.e. return variables and global variables) as specifically as possible.

Two main reasons off the top of my head:
- it is more helpful for users (in my opinion, at least) when looking at documentation to see the exact type that is being returned, and to use in code, since then they know what the exact object interface is; also, it is easier for them to cast to a more general type than it is to cast to a more specific one
- it can lead to unwanted inconsistencies in the code if you change the implementation, but there are multiple exit points in the function and you forget to change one of them. For example, say you have something like:
```python
def f(x: int) -> Sequence[int]:
    if x == 0:
        return [1, 2]
    return [1, 2, 3]
```
and then you change it to
```python
def f(x: int) -> Sequence[int]:
    if x == 0:
        return [1, 2]
    return (1, 2, 3)
```
without realising you had another exit point in the function, this wouldn't be caught, but would be if you had typed the function as `-> List[int]` originally.

AdamGleave: Agreed return types should be as specific as possible so user can do more with them. However, only so far as we want to keep that promise in the future. If we really thought we might change e.g. a list to a tuple, then having returned a `List[int]` in the past is problematic -- client code might depend on features of list (e.g. mutability) that are broken when we change to a `Tuple[int]`. Whereas if we'd only made the more humble promise `Sequence[int]` we can freely change it.

So I think the key question here is do we want to commit to always returning something of Dict-type? Not crazy, dicts are pretty fundamental in Python, and it's not clear what we'd replace them with. On the other hand, there's not many features users might need in `dict` that aren't already in `Mapping`.

Overall don't feel that strongly here -- WDYT @Rocamonde, perhaps you can cast the deciding vote?

ianyfan: I would say that changing the type is bad practice anyway, and that once code is mature enough, it should become clear what the type ought to be, which would allow the specific annotation to be used.

In any case, I'm happy to remove these changes from the PR since they are not the focus and can be addressed in a separate PR.

AdamGleave: Deferring them to another PR sounds good, I am still interested to have Juan @Rocamonde weigh in though.

ianyfan:
Rebased and updated to remove changes from `Mapping -> Dict`. Forgot that it would mess up the old review comment references, sorry

src/imitation/algorithms/adversarial/common.py
Rocamonde: So IIUC you're proposing to follow the rule of thumb of "expect general types, return specific types". I agree that this seems to give greatest flexibility and safety for the users calling the function. 

However, the extent to which this should be followed in my understanding depends on where along the axis of abstract vs final the implementation falls. Since function signatures are contravariant on the argument types and covariant on the return types, an abstract signature that has very specific return types and very general argument types will be very hard to subclass, because subclasses will have to support at least as many input types and at most those many output types. 

In this case, it seems that the function declaration is a final declaration, as it's not a method part of a class, so I would be fine with making the return type as concrete as possible since users should not expect different return types when calling this function due to inheritance.

Rocamonde: In any case, if we are to improve the types in this regard, I agree with both that it's probably best to defer it to another PR since there are many other places where we use general types e.g Mapping, Sequence, where we might want to use a concrete type, or vice versa (we could and maybe should hint a more general type), following the principles above. 

I think that overall having a clear stance on this and making the code follow that would be a very positive change to the codebase's types.

Rocamonde: Another important difference particular to the Dict/Mapping discussion that has been raised by [this SO comment](https://stackoverflow.com/questions/52487663/python-type-hints-typing-mapping-vs-typing-dict#comment113661999_52487800): 

> Dict is invariant, Mapping is covariant. It's important when e.g. you have a function receiving Dict[A,B] and try to pass Dict[A,C] where C is a subtype of B. You will get typing error which can be avoided by using Mapping[A,B]

I actually did not know this. It does not matter much here because we are _returning_ a dict and the annotation is not generic, but it's worth keeping in mind.

ianyfan: Yes, it is an important point, though I've always been a bit confused about how this works from the type theory side of things.

AdamGleave (approved):
LGTM

PR #593, opened 2022-10-26
Enable mypy analysis for all files.

### LINKED ISSUE #573: Fix mypy type erorrs in skipped files
## Bug description
Some files are failing mypy type checks (but passing on pytype) and being skipped by `EXCLUDE_MYPY` in `.circleci/config.yml` introduced in https://github.com/HumanCompatibleAI/imitation/pull/534 We should fix these errors, by ignoring spurious errors and for true-positive errors refactoring the code or adding assert checks where appropriate.

## Steps to reproduce
Run `mypy` on the excluded files

ianyfan:
## Description

**EDIT: remaining issues have been resolved. The issues with `env.horizon` and `evaluate_policy` are to be taken upstream, and then depending on how quickly the turnaround is on those, either wait for them to be merged before merging this, or ignore the issues for now and un-ignore them once the fixes are done.**

ORIGINAL:

Fixes #573 by adding the remaining type annotations. Some of the issues in `mce_irl.py` is resolved by changing `env.horizon` to `int(env.horizon)` which indicates that there would be a problem if `env.horizon` was not an integer (I think it can be `np.inf`?). Does this occur in practice (since this should necessitate a runtime check)?

One other issue in `mce_irl.py` I fixed by using a `cast`, which I guarded behind `if TYPE_CHECKING:`. I don't think there is a better way of doing this, since there is no other way of restricting the type to what we want.

Remaining issues:
```
src/imitation/algorithms/preference_comparisons.py:121: error: Argument 1 to "shuffle" of "Generator" has incompatible type "List[TrajectoryWithRew]"; expected "Union[_SupportsArray[dtype[Any]], _NestedSequence[_SupportsArray[dtype[Any]]], bool, int, float, complex, str, bytes, _NestedSequence[Union[bool, int, float, complex, str, bytes]]]"  [arg-type]
src/imitation/algorithms/preference_comparisons.py:644: error: Argument 1 to "choice" of "Generator" has incompatible type "List[TrajectoryWithRew]"; expected "Union[_SupportsArray[dtype[Any]], _NestedSequence[_SupportsArray[dtype[Any]]], bool, int, float, complex, str, bytes, _NestedSequence[Union[bool, int, float, complex, str, bytes]]]"  [arg-type]
tests/algorithms/test_bc.py:133: error: Item "None" of "Optional[Cache]" has no attribute "mkdir"  [union-attr]
tests/algorithms/test_bc.py:163: error: Item "None" of "Optional[Cache]" has no attribute "mkdir"  [union-attr]
tests/algorithms/test_bc.py:186: error: Argument 1 to "evaluate_policy" has incompatible type "ActorCriticPolicy"; expected "BaseAlgorithm"  [arg-type]
tests/algorithms/test_bc.py:196: error: Argument 1 to "evaluate_policy" has incompatible type "ActorCriticPolicy"; expected "BaseAlgorithm"  [arg-type]
```

The ones in `preference_comparisons.py` appear to be due to an over-conservative annotation in `numpy`though I may have missed something. If it is due to this, then I'm not sure if there is a better way to deal with the issue than to `type: ignore` it.

The issues in `test_bc.py` are also weird and may just be worth ignoring.
`pytestconfig.cache` is typed as `Optional[Cache]` so I'm not sure whether this implies that it could actually be `None` sometimes, or whether it is just typed that way. If the latter, either an extra `assert is not None` may be sufficient; if the former, some extra logic may be required to run the tests properly.

## Testing

Ran `mypy` over relevant files, as well as `ci/code_checks.sh`.


src/imitation/algorithms/mce_irl.py
```diff
@@ -45,7 +56,7 @@ def mce_partition_fh(
         \pi is a 3d array, indexed \pi[t,s,a].
     """
     # shorthand
-    horizon = env.horizon
+    horizon = int(env.horizon)
```
AdamGleave, line 59: Can we add a TODO here so we remember to fix this before merge (if the plan is to fix it in a seals PR first), or open a GH issue to track it (if we want to merge this PR before fixing it upstream)?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1597,7 +1602,7 @@ def train(
         total_timesteps: int,
         total_comparisons: int,
         callback: Optional[Callable[[int], None]] = None,
-    ) -> Mapping[str, Any]:
+    ) -> Dict[str, Any]:
```
AdamGleave, line 1605: as above would prefer to not change these types in this PR

src/imitation/data/rollout.py
```diff
@@ -426,7 +426,7 @@ def generate_trajectories(
 
 def rollout_stats(
     trajectories: Sequence[types.TrajectoryWithRew],
-) -> Mapping[str, float]:
+) -> Dict[str, float]:
```
AdamGleave, line 429: as above would prefer to not change these types in this PR

tests/algorithms/test_bc.py
```diff
@@ -183,7 +185,7 @@ def test_that_bc_improves_rewards(
 ):
     # GIVEN
     novice_rewards, _ = evaluation.evaluate_policy(
-        cartpole_bc_trainer.policy,
+        cartpole_bc_trainer.policy,  # type: ignore[arg-type]
```
AdamGleave, line 188: Similar to with the seals upstream issue, let's add a TODO or GH issue for this.

src/imitation/algorithms/mce_irl.py
```diff
@@ -383,6 +394,7 @@ def set_demonstrations(self, demonstrations: MCEDemonstrations) -> None:
         if isinstance(demonstrations, Iterable):
             first_item, demonstrations = util.get_first_iter_element(demonstrations)
             if isinstance(first_item, types.Trajectory):
+                demonstrations = cast(Iterable[types.Trajectory], demonstrations)
```
AdamGleave, line 397: Copying from https://github.com/HumanCompatibleAI/imitation/pull/594/files (by @dan-pandori):
```suggestion
                # Safe since we know the first item is a Trajectory and `demonstrations`
                # is an Iterable
                demonstrations = cast(Iterable[types.Trajectory], demonstrations)
```

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -118,7 +118,7 @@ def __init__(
     def sample(self, steps: int) -> Sequence[TrajectoryWithRew]:
         # make a copy before shuffling
         trajectories = list(self._trajectories)
-        self.rng.shuffle(trajectories)
+        self.rng.shuffle(trajectories)  # type: ignore[arg-type]
```
AdamGleave, line 121: From https://github.com/HumanCompatibleAI/imitation/pull/594/files
```suggestion
        # numpy's typing thinks it can't accept a list here, but it can
        self.rng.shuffle(trajectories)  # type: ignore[arg-type]
```

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -641,7 +641,8 @@ def __call__(
 
         # we need two fragments for each comparison
         for _ in range(2 * num_pairs):
-            traj = self.rng.choice(trajectories, p=np.array(weights) / sum(weights))
+            p = np.array(weights) / sum(weights)
+            traj = self.rng.choice(trajectories, p=p)  # type: ignore[arg-type]
```
AdamGleave, line 645: From https://github.com/HumanCompatibleAI/imitation/pull/594/files
```suggestion
            # numpy's typing thinks it can't accept a list here, but it can
            traj = self.rng.choice(trajectories, p=p)  # type: ignore[arg-type]
```

tests/algorithms/test_bc.py
```diff
@@ -127,10 +127,11 @@ def test_smoke_bc_creation(
     rng: np.random.Generator,
     pytestconfig: pytest.Config,
 ):
+    assert pytestconfig.cache is not None
```
AdamGleave, line 130: We use `pytestconfig.cache` a bunch of times here so I like the idea of adding a temporary variable `cache = pytestconfig.cache` from https://github.com/HumanCompatibleAI/imitation/pull/594/files#diff-648af1fe52e7f75f6ddc457b062fe9ee7b456d9358c493cef1a3547f5e87785b -- consider cherry-picking that change?

AdamGleave (changes requested):
Largely looks good, only some minor changes needed.

AdamGleave:
On upstream: I don't think we're going to have time to turnaround the SB3 change before you wrap up. The seals one should be a short fix, and I can review and merge same-day, so let's push for fixing that one upstream though.

src/imitation/algorithms/mce_irl.py
ianyfan: I've opened a PR and added an extra check that should still be relevant even once the upstream has been changed, but can still add an extra comment if that would be helpful

src/imitation/algorithms/dagger.py
```diff
@@ -162,7 +162,7 @@ def __init__(
         self._last_user_actions = None
         self.rng = rng
 
-    def seed(self, seed=Optional[int]) -> List[Union[None, int]]:
+    def seed(self, seed: Optional[int] = None) -> List[Optional[int]]:
```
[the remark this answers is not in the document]
ianyfan: Ah it happens, at least this appears to be the only occurrence. Though to be honest, I thought this sort of thing usually gets picked up by linters, maybe that can be added

src/imitation/algorithms/mce_irl.py
ianyfan: I've found a better way to do this, I think

ianyfan:
There's more places that are missing type annotations, but since they don't cause errors in mypy, I will leave them be, so as not to add any more noise to this PR.

AdamGleave (approved):
LGTM.

@Rocamonde, I made some commits to this PR, can you please review the diff: https://github.com/HumanCompatibleAI/imitation/pull/593/files/f3805f8c7e2af8dd4e3803325ae586dee9e6f7ef..b89b7f067f04576b7845ead486089e46317f19a6

ianyfan:
New changes look good to me, at least

Rocamonde:
Apologies, this did not show up on my GitHub PR notifications. Will review tomorrow.

src/imitation/algorithms/mce_irl.py
```diff
@@ -503,9 +503,11 @@ def train(self, max_iter: int = 1000) -> np.ndarray:
                 predicted_r_np, visitations = self._train_step(torch_obs_mat)
 
                 # these are just for termination conditions & debug logging
-                grad_norm = util.tensor_iter_norm(
-                    p.grad for p in self.reward_net.parameters()
-                ).item()
+                grads = []
+                for p in self.reward_net.parameters():
+                    assert p.grad is not None
+                    grads.append(p)
```
Rocamonde, line 509: Before, the list comprehension added the gradient of the parameter, `p.grad`. Now you're adding the parameter, `p`. This is probably a bug?

tests/algorithms/test_bc.py
```diff
@@ -186,7 +186,7 @@ def test_that_bc_improves_rewards(
     cartpole_venv: vec_env.VecEnv,
 ):
     # GIVEN
-    # TODO: the upstream annotation for this function is overly-conservative
+    # TODO(GH#600): the upstream annotation for this function is overly-conservative
```
Rocamonde, line 189: May I suggest
```suggestion
    # TODO(https://github.com/HumanCompatibleAI/imitation/issues/600): the upstream annotation for this function is overly-conservative
```
Easier to click?

tests/algorithms/test_mce_irl.py
```diff
@@ -304,17 +341,8 @@ def test_tabular_policy_randomness(rng):
...
-        n_actions=3,
-        branch_factor=2,
-        horizon=10,
-        random_obs=False,
-        obs_dim=None,
-        generator_seed=42,
-    )
-    state_env = base_envs.ExposePOMDPStateWrapper(mdp)
+def test_mce_irl_demo_formats(rng, random_mdp):
+    state_env = base_envs.ExposePOMDPStateWrapper(random_mdp)
```
Rocamonde, line 345: Nice! Cool to see the wrapper being used in a composable, clean way in tests.

Rocamonde (changes requested):
Looks good, just one thing I noticed that might be a bug, do take a look.

Rocamonde:
Yay! Thanks for doing this @ianyfan , nice to see what we started being nicely completed. This will certainly make our codebase more robust to random bugs and errors.

src/imitation/algorithms/mce_irl.py
AdamGleave: Oh dear, yes this is a bug, thanks for catching! It's a little sad this was not caught by our tests. :/

tests/algorithms/test_bc.py
AdamGleave: Makes line too long but it's a multi-line comment so sure I'll just reflow it.

AdamGleave:
Thanks for the review @Rocamonde, issues should be resolved now

src/imitation/algorithms/density.py
```diff
@@ -343,7 +343,7 @@ def __call__(
         rew_array = np.asarray(rew_list, dtype="float32")
         return rew_array
 
-    def train_policy(self, n_timesteps: int = int(1e6), **kwargs):
+    def train_policy(self, n_timesteps: int = int(1e6), **kwargs: Any) -> None:
```
Rocamonde, line 346: We could try to be more specific than this, but ParamSpec is not (yet) a very well supported feature by most type checkers, and it might not be worth the effort of typing it.

src/imitation/algorithms/mce_irl.py
```diff
@@ -182,7 +192,7 @@ def forward(
         self,
         observation: th.Tensor,
         deterministic: bool = False,
-    ):
+    ) -> NoReturn:
```
Rocamonde, line 195: Interesting . Is this type not incompatible with the superclass? How does that work?

src/imitation/algorithms/mce_irl.py
```diff
@@ -381,9 +397,9 @@ def set_demonstrations(self, demonstrations: MCEDemonstrations) -> None:
         # Demonstrations are either trajectories or transitions;
         # we must compute occupancy measure from this.
         if isinstance(demonstrations, Iterable):
-            first_item, demonstrations = util.get_first_iter_element(demonstrations)
+            first_item, demonstrations_it = util.get_first_iter_element(demonstrations)
```
Rocamonde, line 400: I'm not sure how this type is incompatible since we are narrowing the scope of the type, we should be able to use the same variable if the generic is well constructed. (right?)

src/imitation/algorithms/mce_irl.py
```diff
@@ -487,9 +503,11 @@ def train(self, max_iter: int = 1000) -> np.ndarray:
                 predicted_r_np, visitations = self._train_step(torch_obs_mat)
 
                 # these are just for termination conditions & debug logging
-                grad_norm = util.tensor_iter_norm(
-                    p.grad for p in self.reward_net.parameters()
-                ).item()
+                grads = []
```
Rocamonde, line 506: I guess you could also just cast it and keep it was it was. But I don't really feel strongly about it, how it's now seems fine too. Just not obvious why we are doing it, so maybe add a comment saying it's for the type checker.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -641,7 +642,11 @@ def __call__(
 
         # we need two fragments for each comparison
         for _ in range(2 * num_pairs):
-            traj = self.rng.choice(trajectories, p=np.array(weights) / sum(weights))
+            # NumPy's annotation here is overly-conservative, but this works at runtime
```
Rocamonde, line 645: same comment as above

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -902,7 +907,7 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
         return np.array(rews1, dtype=np.float32), np.array(rews2, dtype=np.float32)
 
 
-class PreferenceDataset(th.utils.data.Dataset):
+class PreferenceDataset(data_th.Dataset):
```
Rocamonde, line 910: Is this the google style guide?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -960,7 +969,7 @@ def push(self, fragments: Sequence[TrajectoryWithRewPair], preferences: np.ndarr
                 self.fragments2 = self.fragments2[extra:]
                 self.preferences = self.preferences[extra:]
 
-    def __getitem__(self, i) -> Tuple[TrajectoryWithRewPair, float]:
+    def __getitem__(self, i: int) -> Tuple[TrajectoryWithRewPair, float]:
```
Rocamonde, line 972: `i` can also be a slice object in general. Is there a reason why we're not allowing this here? (also, you would have to change the return type through an overload)

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1163,7 +1172,7 @@ def __init__(
             else None
         )
 
-    def _make_data_loader(self, dataset: PreferenceDataset) -> data_th.DataLoader:
+    def _make_data_loader(self, dataset: data_th.Dataset) -> data_th.DataLoader:
```
Rocamonde, line 1175: thanks, i remember this being an issue when splitting the preference dataset (you get a subset type) but did not want to change the type as I was not sure that was fine.

src/imitation/rewards/serialize.py
```diff
@@ -174,13 +174,13 @@ def load_zero(path: str, venv: VecEnv) -> reward_function.RewardFn:
     del path, venv
 
     def f(
-        obs: np.ndarray,
-        act: np.ndarray,
-        next_obs: np.ndarray,
-        dones: np.ndarray,
+        state: np.ndarray,
```
Rocamonde, line 177: Maybe this is worth making positional only in the original type and everywhere else?

Rocamonde (commented):
This is excellent work! LGTM. I left some comments with suggestions and questions, worth looking over them before merging to make sure we're happy with the direction of things.

src/imitation/algorithms/dagger.py
```diff
@@ -44,7 +44,7 @@ def __call__(self, round_num: int) -> float:
 class LinearBetaSchedule(BetaSchedule):
     """Linearly-decreasing schedule for beta."""
 
-    def __init__(self, rampdown_rounds: int):
+    def __init__(self, rampdown_rounds: int) -> None:
```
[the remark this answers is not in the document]
ianyfan: https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-methods I think it is an error to not return None. However, omitting it would not cause an error to be thrown.

This has not been added everywhere. In fact, there are a large number of missing annotations. I started to go through them, but stopped because this PR is in principle just to fix the errors that were raised by enabling mypy on these files, and I didn't want to add lots of extra noise to the changes. However, I kept the changes in the commits I had already pushed.

src/imitation/algorithms/mce_irl.py
ianyfan: I presume mypy doesn't actually narrow the scope of the variable, even if you use it like this. Not sure though

src/imitation/algorithms/preference_comparisons.py
ianyfan: I'm not sure. It's not a very important change anyway, I think, so can just revert it

setup.py
```diff
@@ -198,7 +198,7 @@ def get_local_version(version: "ScmVersion", time_format="%Y%m%d") -> str:
         "torch>=1.4.0",
         "tqdm",
         "scikit-learn>=0.21.2",
-        "seals==0.1.4",
+        "seals==0.1.5",
```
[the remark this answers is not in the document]
AdamGleave: Yeah, I agree. We decided in another PR to use `==` for test-only dependencies that had `0.x` revisions because semantic versioning says they can make breaking API changes at any time. However, now that it's an install requirement not a test requirement we should be permissive with versions so as not to cause conflicts for our users. With seals in particular if we make a breaking change there and forget to update `imitation` I have limited sympathy for us!

I will note that we'll still need to bump the version in `setup.py` for any changes that are _needed_ in imitation: CI doesn't install new versions automatically, only when `setup.py` is changed invalidating the existing cache.

src/imitation/algorithms/dagger.py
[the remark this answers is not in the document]
AdamGleave: Adding `-> None` to `__init__` seems like it should be doable with a `sed` one-liner. But I agree keeping this PR minimal is good, if we have another PR that just does that change it'll be easy to review because all the changes will be the same...

Having `__init__` annotated with None feels kinda silly to me: what else is it going to return? But according to https://stackoverflow.com/a/73334605/14507224 `mypy` will complain if there is an `__init__` method with no type annotations (it's OK if one of the arguments is typed though), so there's some argument for adding it everywhere for consistency.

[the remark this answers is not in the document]
AdamGleave: OK I change my mind, adding `-> None` to `__init__` does seem good given https://peps.python.org/pep-0484/ says:
> (Note that the return type of __init__ ought to be annotated with -> None. The reason for this is subtle. If __init__ assumed a return annotation of -> None, would that mean that an argument-less, un-annotated __init__ method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that __init__ ought to have a return annotation; the default behavior is thus the same as for other methods.)

src/imitation/algorithms/density.py
AdamGleave: I've not used `ParamSpec` much so may be missing something, but how would it help us here? It seems designed in cases where the `P = ParamSpec("P")` variable is reused, like decorators or other higher-order functions. But if we're returning `None` I don't think it makes a difference.

src/imitation/algorithms/mce_irl.py
AdamGleave: I think it works because `NoReturn` is the bottom type: i.e. a type that is a subtype of all other types. There's a good discussion at https://github.com/python/mypy/issues/4116

src/imitation/algorithms/mce_irl.py
AdamGleave: Gradients really can be none in general, but it's true we know they're not here because we called `loss.backward()` in the `self._train_step` call, so a cast does seem fine. I've added a comment for now, but happy to change it to cast, I was pretty on the fence about it.

src/imitation/algorithms/preference_comparisons.py
AdamGleave: It looks like `numpy` is fine with lists actually, the issue is just that it expects the arrays to be of scalars:

```
(venv) adam@puffin:~/dev/imitation$ cat test.py
from numpy import random

rng : random.Generator = random.default_rng()

my_list = [1, 2, 3, 4, 5]
random.shuffle(my_list)
rng.shuffle(my_list)

class Foo:
    def __init__(self, x) -> None:
        self.x = x

my_complex_list = [Foo(42), Foo(21)]
random.shuffle(my_complex_list)
rng.shuffle(my_complex_list)
(venv) adam@puffin:~/dev/imitation$ mypy test.py
test.py:14: error: Argument 1 has incompatible type "List[Foo]"; expected "Union[_SupportsArray[dtype[Any]], _NestedSequence[_SupportsArray[dtype[Any]]], bool, int, float, complex, str, bytes, _NestedSequence[Union[bool, int, float, complex, str, bytes]]]"
test.py:15: error: Argument 1 to "shuffle" of "Generator" has incompatible type "List[Foo]"; expected "Union[_SupportsArray[dtype[Any]], _NestedSequence[_SupportsArray[dtype[Any]]], bool, int, float, complex, str, bytes, _NestedSequence[Union[bool, int, float, complex, str, bytes]]]"
Found 2 errors in 1 file (checked 1 source file)
```

I'm not sure whether they'd really want to change it given this, the one for `shuffle` is already maximally permissive within their type hierarchy taking `ArrayLike`: https://github.com/numpy/numpy/blob/266aad7478bc7fbcc55eea7f942a0d373b838396/numpy/random/_generator.pyi But `ArrayLike` is just quite restrictive: https://github.com/numpy/numpy/blob/main/numpy/_typing/_array_like.py#L85  Seems like a deliberate choice, changing that might break a lot of other annotations...

src/imitation/algorithms/preference_comparisons.py
AdamGleave: We were already importing this as `data_th` so it seems good to do this for consistency IMO.

src/imitation/algorithms/preference_comparisons.py
AdamGleave: Good idea, added!

src/imitation/algorithms/preference_comparisons.py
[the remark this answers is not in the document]
AdamGleave: I don't think it matches the signature for Sequence, `__getitem__` with a slice does not return a Sequence of tuples but a tuple of sequences. We could handle int and slice keys differently like in types, but reluctant to add that complexity until we have a use case for it.

src/imitation/rewards/serialize.py
AdamGleave: I'm always a bit scared of positional functions where all arguments have the same type, very easy to get them mixed up. I'd rather just actually stick to a convention here, the current one seems OK.

AdamGleave:
@Rocamonde I believe I've addressed all outstanding comments on this, please do re-review when you have a moment. Note we'll need to merge this one before https://github.com/HumanCompatibleAI/imitation/pull/603 because that removes the mypy excludes.

Rocamonde:
Sorry about this. I reviewed the PRs sequentially from my inbox and only saw this should have been merged before now.

src/imitation/algorithms/preference_comparisons.py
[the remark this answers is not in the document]
Rocamonde: Good catch, makes sense.

Rocamonde: Cool.

src/imitation/algorithms/mce_irl.py
Rocamonde: Yeah makes sense. I think the assert is slightly safer, and if there's a comment it should be OK.

src/imitation/algorithms/preference_comparisons.py
Rocamonde: Yeah I see, interesting, let's just deal with this internally then.

src/imitation/algorithms/density.py
Rocamonde: I meant for Any, not for None haha. The problem with this is that any subtype has to take any kwarg in the signature, but you only really want it to take the kwargs that `.learn()` can take. But not sure how you can use ParamSpec for this now I think about it. I think it was designed as a generic that is filled in implicitly at the callsite, not for extracting the type signature of an explicit method (which tbh should be a valid use case).

Rocamonde:
Actually it was fine, the merge was trivial. Will wait for tests to run and then merge.

Rocamonde (approved):
LGTM

PR #594, opened 2022-10-27
Resolve the remaining mypy issues

dan-pandori:
## Description

Address the remaining mypy issues in the repo.
https://github.com/HumanCompatibleAI/imitation/issues/573

Note that there are two instances which are really caused by type issues elsewhere, which I'd love to get a second opinion on. I just did the minimal change of suppressing the issue, but described alternative options that is more involved but also better utilizes the type system for safety.

## Testing

Ran code_checks.sh


dan-pandori:
Ah, I still need to fix the test_preference_comparisons.py test. Interesting that splitting the dataset doesn't split into more datasets of the same type, but understandable. I'll see how to best fix.

Sorry for the noise!

dan-pandori:
@AdamGleave to take a look, since you reviewed the cleanup of the other files.

AdamGleave:
Thanks for tagging, Dan! Will take a look now. I think there may be some overlap between this and concurrent https://github.com/HumanCompatibleAI/imitation/pull/593 that I just reviewed. Sorry for not noticing this duplication earlier. I'll see if we can combine them somehow.

src/imitation/algorithms/mce_irl.py
```diff
@@ -45,6 +45,10 @@ def mce_partition_fh(
         \pi is a 3d array, indexed \pi[t,s,a].
     """
     # shorthand
+    # TODO ask Adam if this isn't better solved by changing the types in
```
AdamGleave, line 48: I agree this should be fixed upstream, I think @ianyfan is planning on working on this.

src/imitation/rewards/serialize.py
```diff
@@ -173,16 +173,21 @@ def _validate_wrapper_structure(
...
-        obs: np.ndarray,
-        act: np.ndarray,
-        next_obs: np.ndarray,
-        dones: np.ndarray,
-    ) -> np.ndarray:
-        del act, next_obs, dones  # Unused.
-        return np.zeros(obs.shape[0])
-
-    return f
+    class f(reward_function.RewardFn):
```
AdamGleave, line 176: Do we need to inherit? `reward_function.RewardFn` is a protocol so I think a function with the type-signature should match too (perhaps not?)

tests/algorithms/test_bc.py
```diff
@@ -182,8 +186,15 @@ def test_that_bc_improves_rewards(
     cartpole_venv: vec_env.VecEnv,
 ):
     # GIVEN
+    # TODO ask reviewer about preference between:
```
AdamGleave, line 189: 2) long-term (I think @ianyfan might also work on this, so check in with me/him before working on this to avoid duplication), but probably 1) as hotfix since it may take a while to get it fixed upstream.

AdamGleave (commented):
OK I think I'm going to resolve this in favor of https://github.com/HumanCompatibleAI/imitation/pull/593 as that PR is overall making more changes. Where you both touched the same code, the changes were usually similar, but I've gone through some of the differences and cherry-picked relevant parts as suggestions in the review of https://github.com/HumanCompatibleAI/imitation/pull/593

I've left a couple of clarifying questions/comments on this PR.

Downside of this is GitHub won't track you as author in the commits, I'll see if I can fix this at the squash-merge stage.

src/imitation/algorithms/mce_irl.py
dan-pandori: Great, good to know for reference.

tests/algorithms/test_bc.py
dan-pandori: Will do!

src/imitation/rewards/serialize.py
dan-pandori: Ah, yes it seems like it's sufficient to rename the argument names. I would not have thought that argument names were a type enforced part of a Protocol!

dan-pandori:
And @AdamGleave, I'll just delete this PR to let https://github.com/HumanCompatibleAI/imitation/pull/593 go ahead since it's doing more of the same.

No worries on attribution, I'm just using this issue as a toy problem to get familiar with the code base a little.

ianyfan:
I've incorporated some changes from here into #594 and added @dan-pandori as a co-author.

dan-pandori:
Thanks Ian! That's very kind.

Closing this PR

Issue #506, opened 2022-07-28
[Feature request] Gradient accumulation in CrossEntropyRewardTrainer in preference_comparisons.py


dfilan:
### 🚀 Feature

When the reward function is trained by the `CrossEntropyRewardTrainer` in `preference_comparisons.py`, it currently takes a batch of trajectory fragments and preferences, calculates its reward for each pair of fragments, and then at the end of the batch calculates the cross-entropy loss and computes a gradient. Instead, when some flag like `accumulate_gradients` is passed to the reward trainer, the loss and gradient could be computed for each pair of fragments in the batch independently.

### Motivation

This would significantly increase the batch size that is compatible with any given amount of GPU memory. When training a CNN reward net on an unwrapped atari environment with a default batch size of 30, each batch uses 30 fragment pairs * 2 fragments / pair * 100 transitions / fragment * (3 observation channels + 14 action channels + 64 channels in hidden layers) * 210 height * 160 width * 4 bytes / float32 = 65.3 GB of memory per batch. If gradients are accumulated per-fragment pair, then this gets divided by 30 for a more manageable 2.2 GB per fragment pair.

### Pitch

Basically, compute losses on individual trajectories, and call `loss.backward()` inside the loss calculation loop when the reward net is being trained.

### Alternatives

Use smaller batches and adjust learning rates accordingly - should often work OK, but it's nice to support larger batches.

levmckinney:
Quick suggestion here. Rather than having a binary choice either `accumulate_grad=True` or `accumulate_grad=False`, we could have a setting like in pytorch lightning's `accumulate_grad_batches` see [here](https://pytorch-lightning.readthedocs.io/en/latest/advanced/training_tricks.html#accumulate-gradients).  Say your GPU memory can support a batch size of 16, but you want to emulate a batch size of 64. Then you set `accumulate_grad_batches = 4`. This will run loss.backward() after each step but  only call optim.step() every 4 batches. This allows you to take full advantage of your hardware even when you can't fit the full batch size into memory.

Finding the correct name for `accumulate_grad_batches` in our code base might be a bit tricky, but it seems more general than only allowing the pairs to be processed all in one batch or one at a time.

AdamGleave:
Yeah I agree we should either specify the "inner" batch size to execute on GPU, or the number of accumulation batches. Computing each pair of trajectory fragments on the GPU separately would come with a considerable performance overhead. 

[...]

PR #595, opened 2022-10-27
Implement gradient accumulation.

ianyfan:
## Description

Implements #506. ~Currently only for `preference_comparisons.py` but looking at the other algorithms too. They should follow a similar pattern to this, so any feedback would be welcome.~ Implemented for adversarial, BC, and preference comparisons (but not MCEIRL). Also haven't actually tested it on a GPU yet.

Also not finalised the parameter name for `minibatch_size` yet, if anyone has any better suggestions.

## Testing

Added a `test_gradient_accumulation` test. Seeding didn't end up being as bad an issue as I thought it would be.

### DIFF
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -356,8 +356,9 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
                     batch["labels_expert_is_one"].float(),
                 )
 
-                # Renormalise the loss to be averaged over the whole
-                # batch size instead of the minibatch size.
+                # Renormalise the loss to be averaged over the whole batch size
+                # instead of the minibatch size.
+                assert len(batch["state"]) == self.demo_minibatch_size
                 loss *= self.demo_minibatch_size / self.demo_batch_size
                 loss.backward()
 
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -356,9 +356,9 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
                     batch["labels_expert_is_one"].float(),
                 )
 
-                # Renormalise the loss to be averaged over the whole batch size
-                # instead of the minibatch size.
-                assert len(batch["state"]) == self.demo_minibatch_size
+                # Renormalise the loss to be averaged over the whole
+                # (doubled) batch size instead of the minibatch size.
+                assert len(batch["state"]) == 2 * self.demo_minibatch_size
                 loss *= self.demo_minibatch_size / self.demo_batch_size
                 loss.backward()
 
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -357,7 +357,7 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
                 )
 
                 # Renormalise the loss to be averaged over the whole
-                # (doubled) batch size instead of the minibatch size.
+                # batch size instead of the minibatch size.
                 assert len(batch["state"]) == 2 * self.demo_minibatch_size
                 loss *= self.demo_minibatch_size / self.demo_batch_size
                 loss.backward()
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -512,14 +512,14 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
         gen_samples: Optional[Mapping] = None,
         expert_samples: Optional[Mapping] = None,
     ) -> Iterator[Mapping[str, th.Tensor]]:
-        """Build and return training batch for the next discriminator update.
+        """Build and return training minibatches for the next discriminator update.
 
         Args:
             gen_samples: Same as in `train_disc`.
             expert_samples: Same as in `train_disc`.
 
-        Returns:
-            The training batch: state, action, next state, dones, labels
+        Yields:
+            The training minibatch: state, action, next state, dones, labels
             and policy log-probabilities.
 
         Raises:
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -570,15 +570,11 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
         assert batch_size == len(gen_samples["acts"])
         assert batch_size == len(gen_samples["next_obs"])
 
-        for i in range(0, batch_size, self.demo_minibatch_size):
+        for start in range(0, batch_size, self.demo_minibatch_size):
             # take minibatch slice (I think this creates views so no memory issues)
-            expert_batch = {
-                k: v[i : i + self.demo_minibatch_size]
-                for k, v in expert_samples.items()
-            }
-            gen_batch = {
-                k: v[i : i + self.demo_minibatch_size] for k, v in gen_samples.items()
-            }
+            end = start + self.demo_minibatch_size
+            expert_batch = {k: v[start:end] for k, v in expert_samples.items()}
+            gen_batch = {k: v[start:end] for k, v in gen_samples.items()}
 
             # Concatenate rollouts, and label each row as expert or generator.
             obs = np.concatenate([expert_batch["obs"], gen_batch["obs"]])
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -111,6 +111,7 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
         venv: vec_env.VecEnv,
         gen_algo: base_class.BaseAlgorithm,
         reward_net: reward_nets.RewardNet,
+        demo_minibatch_size: Optional[int] = None,
         n_disc_updates_per_round: int = 2,
         log_dir: types.AnyPath = "output/",
         disc_opt_cls: Type[th.optim.Optimizer] = th.optim.Adam,
@@ -139,6 +140,11 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
                 `venv` and `custom_logger`.
             reward_net: a Torch module that takes an observation, action and
                 next observation tensors as input and computes a reward signal.
+            demo_minibatch_size: size of minibatch to calculate gradients over.
+                The gradients are accumulated until the entire batch is
+                processed before making an optimization step.
+                Must be a factor of `demo_batch_size`.
+                Optional, defaults to `demo_batch_size`.
             n_disc_updates_per_round: The number of discriminator updates after each
                 round of generator updates in AdversarialTrainer.learn().
             log_dir: Directory to store TensorBoard logs, plots, etc. in.
@@ -169,8 +175,14 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
                 condition, and can seriously confound evaluation. Read
                 https://imitation.readthedocs.io/en/latest/guide/variable_horizon.html
                 before overriding this.
+
+        Raises:
+            ValueError: if the batch size is not a multiple of the minibatch size.
         """
         self.demo_batch_size = demo_batch_size
+        self.demo_minibatch_size = demo_minibatch_size or demo_batch_size
+        if self.demo_batch_size % self.demo_minibatch_size != 0:
+            raise ValueError("Batch size must be a multiple of minibatch size.")
         self._demo_data_loader = None
         self._endless_expert_iterator = None
         super().__init__(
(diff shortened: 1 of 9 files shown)


src/imitation/algorithms/bc.py
```diff
@@ -142,26 +142,6 @@ def __call__(
         )
 
 
-@dataclasses.dataclass(frozen=True)
```
AdamGleave, line 145: Why do we need to remove this? It seems largely unrelated to the PR. I don't have strong preferences on whether to keep it or not in general -- for reference https://github.com/HumanCompatibleAI/imitation/pull/422 is where we introduced it and tagging @ernestum who may have ideas on whether we should keep this and, if so, how to refactor it.

src/imitation/algorithms/bc.py
```diff
@@ -316,10 +301,14 @@ def __init__(
 
         Raises:
             ValueError: If `weight_decay` is specified in `optimizer_kwargs` (use the
-                parameter `l2_weight` instead.)
+                parameter `l2_weight` instead.),
```
AdamGleave, line 304:
```suggestion
                parameter `l2_weight` instead),
```

src/imitation/algorithms/bc.py
```diff
@@ -431,11 +416,17 @@ def _on_epoch_end(epoch_number: int):
             if on_epoch_end is not None:
                 on_epoch_end()
 
+        n_minibatches = (
```
AdamGleave, line 419: This is a bit long for one expression, can we split it up more? E.g. introduce a temporary variable `num_minibatches = self.batch_size // self.minibatch_size`

src/imitation/algorithms/bc.py
```diff
@@ -445,12 +436,13 @@ def _on_epoch_end(epoch_number: int):
...
                 batches_with_stats,
                 unit="batch",
-                total=n_batches,
+                total=n_minibatches,
             )
             tqdm_progress_bar = batches_with_stats
 
-        for (batch_num, batch_size, num_samples_so_far), batch in batches_with_stats:
-            loss = self.trainer(batch)
+        def process_batch():
```
AdamGleave, line 443: This could just be a (private) instance method as it only closes over instance variable `self.optimizer`

src/imitation/algorithms/bc.py
```diff
@@ -459,13 +451,30 @@ def _on_epoch_end(epoch_number: int):
...
             if on_batch_end is not None:
                 on_batch_end()
 
+        self.optimizer.zero_grad()
+        for (batch_num, batch_size, num_samples_so_far), batch in batches_with_stats:
+            obs = th.as_tensor(batch["obs"], device=self.policy.device).detach()
+            acts = th.as_tensor(batch["acts"], device=self.policy.device).detach()
+            training_metrics = self.loss_calculator(self.policy, obs, acts)
+
+            loss = training_metrics.loss * batch_size / self.batch_size
```
AdamGleave, line 467: The `batch_size` here is really `minibatch_size`?

tests/algorithms/test_bc.py
```diff
@@ -212,6 +212,61 @@ def test_that_bc_improves_rewards(
...
+        expert_data_type="transitions",
+        env_name="seals/CartPole-v0",
+        rng=None,
+        num_trajectories=17,
+    )
+
+    seed = rng.integers(2**32)
+
+    th.manual_seed(seed)
+    bc_trainer1 = bc.BC(
```
AdamGleave, line 232: A lot of code duplication between setup of `bc_trainer{1,2}` and subsequent training -- perhaps move this into a helper method?

tests/algorithms/test_bc.py
```diff
@@ -212,6 +212,61 @@ def test_that_bc_improves_rewards(
...
+    rng,
+    pytestconfig,
+):
+    demonstrations = make_expert_transition_loader(
+        cache_dir=pytestconfig.cache.makedir("experts"),
+        batch_size=6,
+        expert_data_type="transitions",
+        env_name="seals/CartPole-v0",
+        rng=None,
+        num_trajectories=17,
```
AdamGleave, line 226: Is it important this is `17`? If not, I'd suggest adding a `cartpole_expert_transitions` fixxture to `tests/algorithms/conftest.py` factoring out what's currently embedded inside `cartpole_bc_trainer`.

AdamGleave (changes requested):
Implementation looks correct -- left some comments focused primarily on readability and maintainability. Thanks for your work on this and especially spotting the naive way of implementing it doesn't work -- not something I'd realised before our call. Please tag me when ready for a re-review.

src/imitation/algorithms/bc.py
ianyfan: I'm not sure, it also needs a bunch of local variables. In theory, it can be rewritten as a method, but would require quite a few arguments

tests/algorithms/test_bc.py
ianyfan: I wanted a number that didn't divide cleanly into the minibatch size, but also didn't leave any remainders of 1, because an incorrect implementation can accidentally work for batch sizes of 1

src/imitation/algorithms/bc.py
ianyfan: It's possible to refactor, but I think the main reason I inlined the logic is to keep the `optim.step()` line close to the one for the incomplete batch, so you can see them right next to each other.

src/imitation/algorithms/bc.py
ianyfan: I don't think there's any point, since it wouldn't be used elsewhere, and the verbosity is due to checking that `n_batches` is not None.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1217,19 +1228,27 @@ def _train(
...
+                    accumulated_size = 0
+                    self.optim.zero_grad()
                     for fragment_pairs, preferences in dataloader:
-                        self.optim.zero_grad()
                         with self.logger.add_key_prefix("train"):
                             loss = self._training_inner_loop(
                                 fragment_pairs,
                                 preferences,
                             )
+                            loss *= len(fragment_pairs) / self.batch_size
```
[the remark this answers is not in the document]
ianyfan: I've added a comment, see what you think

tests/algorithms/test_bc.py
ianyfan: I'm not sure if it meaningfully can since it's just a single function call each time

tests/algorithms/test_preference_comparisons.py
```diff
@@ -415,6 +415,71 @@ def test_discount_rate_no_crash(
...
+    trajectory = agent_trainer.sample(17)
+    fragments = random_fragmenter(trajectory, 1, 17)
+    preferences = preference_gatherer(fragments)
+    dataset.push(fragments, preferences)
+
+    seed = rng.integers(2**32)
+    rng1 = np.random.default_rng(seed)
+    rng2 = np.random.default_rng(seed)
+
+    seed = rng.integers(2**32)
```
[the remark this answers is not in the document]
ianyfan: I just did it for extra generality

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -1217,19 +1228,27 @@ def _train(
...
                         train_loss += loss.item()
                         if self.regularizer:
                             self.regularizer.regularize_and_backward(loss)
                         else:
                             loss.backward()
-                        self.optim.step()
+
+                        accumulated_size += len(fragment_pairs)
+                        if accumulated_size == self.batch_size:
```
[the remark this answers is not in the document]
ianyfan: I've changed it to `>=`, hopefully that's fine

tests/algorithms/test_bc.py
```diff
@@ -212,6 +212,66 @@ def test_that_bc_improves_rewards(
...
+        th.manual_seed(seed)
+        bc_trainer2.train(n_batches=1)
+
+        # Note: due to numerical instability, the models are
+        # bound to diverge at some point, but should be stable
+        # over the short time frame we test over; however, it is
+        # theoretically possible that with very unlucky seeding,
+        # this could fail.
+        assert all(
+            th.allclose(p2, p2, atol=3e-8)
```
AdamGleave: Typo -- `p2` should always be close to `p2`! (I'll fix this; requires increasing `atol` though.)

tests/algorithms/test_bc.py
AdamGleave: It's also `th.manual_seed(seed)` and the function call is quite long, I think it can improve readability, I'll make a commit to show what I mean

tests/algorithms/test_bc.py
AdamGleave: Gotcha, this makes sense :)

AdamGleave (commented):
Largely looks good. I made some changes to simplify the test code, can you please review my changes in https://github.com/HumanCompatibleAI/imitation/pull/595/files/f54afd7818efa0a954e0b955e674fc5b89e45b42..5d6031d8cabcfd123733f95618be69a7597980d6 ? I'd usually avoid making changes directly to someone else's PR, but I want to minimize the amount of work you need to do to wrap this up.

Otherwise asked a couple of clarifying questions / made some minor suggestions.

tests/algorithms/test_bc.py
ianyfan: Makes sense, your change looks good

src/imitation/algorithms/adversarial/common.py
```diff
@@ -322,25 +337,31 @@ def train_disc(
...
+                    batch["log_policy_act_prob"],
+                )
+                loss = F.binary_cross_entropy_with_logits(
+                    disc_logits,
+                    batch["labels_expert_is_one"].float(),
+                )
+
+                # Renormalise the loss to be averaged over the whole
+                # batch size instead of the minibatch size.
+                loss *= self.demo_minibatch_size / self.demo_batch_size
```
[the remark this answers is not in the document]
ianyfan: I think it is, since there are checks to make sure `len(expert_samples) == len(gen_samples) == self.demo_batch_size` and an extra check of `self.demo_batch_size % self.demo_minibatch_size == 0`. However, I can see it being confusing, may be worth adding a comment.

src/imitation/algorithms/adversarial/common.py
```diff
@@ -543,46 +565,60 @@ def _make_disc_train_batch(
...
-        # labelled with 1 and generator samples with 0.
-        labels_expert_is_one = np.concatenate(
-            [np.ones(n_expert, dtype=int), np.zeros(n_gen, dtype=int)],
-        )
+        assert batch_size == len(expert_samples["acts"])
+        assert batch_size == len(expert_samples["next_obs"])
+        assert batch_size == len(gen_samples["acts"])
+        assert batch_size == len(gen_samples["next_obs"])
+
+        for i in range(0, batch_size, self.demo_minibatch_size):
```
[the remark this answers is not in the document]
ianyfan: Sounds reasonable, I didn't put a lot of thought into the variable name but I agree with your assessment

tests/algorithms/test_bc.py
ianyfan: But feel free to refactor it. Pytest fixtures are fiddly and I didn't want to put the work into dealing with that when I was just getting the tests to work.

tests/algorithms/test_bc.py
ianyfan: Yeah, not sure how I missed that. `atol=1e-5` seems a bit high though, did you try increasing `rtol` instead? Default values are `atol=1e-8, rtol=1e-5` (from https://pytorch.org/docs/stable/generated/torch.allclose.html) I would hope that increasing them to `atol=1e-7, rtol=1e-4` would be enough.

AdamGleave: I didn't try increasing `rtol`, thanks for the suggestion! Unfortunately fails sometimes even with `rtol=1e-3`, so I think `atol=1e-5` is our best bet. The need for something this large does make me a bit suspicious of the code, although float32 precision is only around `5e-8` so if the tensor has 20 elements then getting an error of up to `1e-6` is possible even in the best-case scenario.

src/imitation/algorithms/adversarial/common.py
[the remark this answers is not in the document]
AdamGleave: OK, I agree this is fine, difference is adversarial trainer can choose how many rollouts it samples.

tests/algorithms/test_bc.py
ianyfan: Just as a sanity check, did you increase them both at the same time?

If it still behaves the same, then just do what you think is best. But let me know if you think it warrants a bigger change in case I can help.

AdamGleave:
Thanks for the commit fixing my dumb mistake with batch size and making the allclose assert depend on `step` which is a good idea. Unfortunately I'm still seeing flakiness in at around a 1.5% with the lower thresholds in that commit. When they do fail, they fail on step 0. So, I think we're going to have to stick to `atol=1e-5` unfortunately.

AdamGleave (approved):
LGTM

ianyfan:
Yes, I noticed the flakiness too, but couldn't really work out what to do about it.

When I investigated it for bc, it also happened at step 0, and was due to a single parameter entry which was zero-initialised and had a gradient on the first step that was a lot smaller than the others so had a larger relative difference. For example, the other entries in the parameter would have a gradient on the other of 1e-7 whereas this outlier entry would have a gradient on the order of 1e-10. So if the difference in the models' gradients are ~1e-13 then the relative difference would be 1e-3 vs 1e-6, which would fail the closeness check.

Not sure if that's enough information to work out what's going on, or even open an issue about it.

Also, I wonder if there's a better way of testing closeness, such as Euclidean distance or cosine similarity.

PR #598, opened 2022-10-28
Fix flakiness in test_trajectory_dataset_shuffle

AdamGleave:
## Description

test_trajectory_dataset_shuffle samples twice from a `TrajectoryDataset` and checks they're not equal. There is a `1/number_of_trajectories` chance of sampling the same trajectory twice. In our case the number of trajectories is 60, and we run this test multiple times in different configs, so it was failing in around 10% of cases.

This PR instead samples repeatedly and checks if any of them are different, ensuring flakiness below a threshold.

Main issues I can see with this change: (a) it only really tests the sampling isn't deterministic; if it was very "sticky" the test would often pass because of the repeated sampling; (b) it feels perhaps a bit overengineered with the num_samples calculation, but anything else I felt might break if we change the number of expert trajectories later.

## Testing

I ran the test 1000x and confirmed it always passed so is no longer flaky.

### DIFF
--- a/tests/algorithms/test_preference_comparisons.py
+++ b/tests/algorithms/test_preference_comparisons.py
@@ -163,11 +163,12 @@ def test_trajectory_dataset_too_long(
         dataset.sample(100000)
 
 
-def test_trajectory_dataset_shuffle(
+def test_trajectory_dataset_not_static(
     cartpole_expert_trajectories: Sequence[TrajectoryWithRew],
     rng,
     num_steps: int = 400,
 ):
+    """Tests sample() doesn't always return the same value."""
     dataset = preference_comparisons.TrajectoryDataset(
         cartpole_expert_trajectories,
         rng,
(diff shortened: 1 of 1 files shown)


tests/algorithms/test_preference_comparisons.py
```diff
@@ -171,10 +172,15 @@ def test_trajectory_dataset_shuffle(
...
         rng,
     )
-    sample = dataset.sample(num_steps)
-    sample2 = dataset.sample(num_steps)
+    n_trajectories = len(cartpole_expert_trajectories)
+    flakiness_prob = 1 / n_trajectories
+    max_flakiness = 1e-6
+    # Choose num_samples s.t. flakiness_prob**num_samples <= max_flakiness
+    num_samples = math.ceil(math.log(max_flakiness) / math.log(flakiness_prob))
+    samples = [dataset.sample(num_steps) for _ in range(num_samples)]
```
samuelarnesen, line 180: This makes sense to me -- personally it's slightly more intuitive to resample iff the trajectories are equal rather than generating all the samples at the start but num_samples is small enough that I don't think it makes a different.

samuelarnesen (approved):

Rocamonde:
I understand why the test fails, but I'm struggling to understand what is the point of the test in the first place? You say that
> it only really tests the sampling isn't deterministic
what else would you like this test to do?

AdamGleave:
> what else would you like this test to do? I sense that the test might be conceptually ill-defined

Ideally you'd test that the probability of it returning a particular trajectory is uniform at random.

That's an awkward thing to test though.

tests/algorithms/test_preference_comparisons.py
AdamGleave: Good point, changed.

Issue #587, opened 2022-10-23
Add support to warm-start GAIL and AIRL using behavioral cloning


AdamGleave:
## Problem
GAIL and AIRL can be slow to train due to RL in the inner loop, which is both computationally expensive and can require many environment interactions. Behavioral cloning is supervised learning so is fast and does not need any environment interactions, however its peak performance in complex environments is often weaker than GAIL and AIRL.

## Solution
Add an option to use BC to train a policy from the demonstrations, and then to fine-tune that policy using GAIL/AIRL with the same set of demonstrations. This option was supported in the [Stable Baselines v2](https://stable-baselines.readthedocs.io/en/master/modules/gail.html) implementation of GAIL, for example.

This may not always help. If BC learns a bad policy, we could get stuck in a local minima. For AIRL the resulting reward function might also be more fragile as the transition distribution seen during training will be more limited (for GAIL the discriminator could be more fragile as well but it was never intended to be reused).

We already have a BC implementation, so I think this should probably be a feature added to the `train_imitation` script rather than the algorithm itself, although if it ends up being an involved implementation adding a helper method could be appropriate so that people directly using the Python API can also benefit from this.

## Possible alternative solutions
We could potentially make this supported more widely, even for algorithms that don't learn from demonstrations like the `preference_comparisons` module, in case users have both demonstrations and preference comparisons and want to learn from both (a setting that has been studied in e.g. Ibarz et al, 2018). I don't think it's worth us going out of our way to support this, but if it factors out nicely (e.g. some extra Sacred ingredient to warm-start the policy) it could be worth adding.

PR #602, opened 2022-10-30
enable warm starting adversarial algorithms w/ behavior cloning

samuelarnesen:
## Description

Addresses issue #587 

Enables out-of-the-box warm starting of a GAIL/AIRL training run w/ behavior cloning

## Testing

New unit test

### DIFF
--- a/src/imitation/scripts/config/train_adversarial.py
+++ b/src/imitation/scripts/config/train_adversarial.py
@@ -34,6 +34,9 @@ def defaults():
 
     checkpoint_interval = 0  # Num epochs between checkpoints (<0 disables)
     agent_path = None  # Path to load agent from, optional.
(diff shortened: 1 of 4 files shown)


ernestum:
@samuelarnesen @AdamGleave are we still interested in merging this PR? If so I would review it and resolve the conflicts.

AdamGleave:
It's a feature I'd like to have. But the solution ended up quite messy, I think we might need a more significant refactor of our scripts to accommodate it cleanly, so I'm inclined to icebox it.

FYI samuelarnesen worked on this during his work trial which is now over, so out of respect for his time let's try and resolve this without his involvement.

ernestum:
I rebased/squashed this onto the current master. Before properly cleaning up I would like to decide if we keep sacred or not in #79

Issue #549, opened 2022-09-03
Weird behavior training CNN policies with train_rl


PavelCz:
## Problem
Not sure I would call this a bug, but it is definitely unintuitive behavior for me.
The fundamental issue is that all CNN Policies in SB3 are just ActorCriticPolicies with a single changed default argument ([compare](https://github.com/DLR-RM/stable-baselines3/blob/fdca786f0999ea402ad5da98e3d69aa68bcbf635/stable_baselines3/common/policies.py#L674)).
The only thing that makes a CnnPolicy is the NatureCNN feature extractor.

If I want to train a CNN policy in an image-based env this is the relevant part of the config I need to modify:
```
train:
  n_episodes_eval = 50             # Num of episodes for final mean ground truth return
  policy_cls = <class 'stable_baselines3.common.policies.ActorCriticCnnPolicy'>    # Training
  policy_kwargs:
    features_extractor_class = <class 'imitation.policies.base.NormalizeFeaturesExtractor'>
    features_extractor_kwargs:
      normalize_class = <class 'imitation.util.networks.RunningNorm'>
```
This will silently give me an MlpPolicy, even though it looks like I should have a CnnPolicy from the config.

Now if I know this I might think I can circumvent this by setting `policy_kwargs=dict()` so the NatureCNN doesn't get overridden.
This again silently gives me an MlpPolicy due to [this config hook](https://github.com/HumanCompatibleAI/imitation/blob/312b091d550c11076d40d056d5c699fee965c2df/src/imitation/scripts/common/train.py#L62).

The only way to actually get a CNN policy is to explicitly set `policy_kwargs=dict(features_extractor_class=NatureCNN)` which makes the CnnPolicy obsolete, as this is equivalent to just using ActorCriticPolicy as the policy.

(I say it uses MlpPolicy silently but if you have loglevel set to INFO and remember to look it does print a summary of the net architecture.)

Here is a small script that illustrates these problems. Since there are no image-based environments in seals I use procgen here.
```
"""Thin wrapper around imitation's train_rl script"""
from imitation.scripts.config.train_rl import train_rl_ex
from imitation.scripts.train_rl import main_console
from stable_baselines3.common.torch_layers import NatureCNN
from stable_baselines3.ppo import CnnPolicy


# doesn't work
@train_rl_ex.named_config
def a():
    common = dict(env_name="procgen:procgen-coinrun-v0")
    train = dict(
        policy_cls=CnnPolicy,
    )
[...]

PR #610, opened 2022-11-07
Add default config for cnn policies

anshradh:
## Description

Adds a default configuration option for CNN policies, addressing #549 

## Testing

Tested with following command: 
`python -m imitation.scripts.train_rl with common.env_name=AsteroidsNoFrameskip-v4 train.cnn_policy`

### DIFF
--- a/src/imitation/scripts/config/train_rl.py
+++ b/src/imitation/scripts/config/train_rl.py
@@ -1,8 +1,6 @@
 """Configuration settings for train_rl, training a policy with RL."""
 
 import sacred
-from stable_baselines3.common.torch_layers import NatureCNN
-from stable_baselines3.ppo import CnnPolicy
 
 from imitation.scripts.common import common, rl, train
 
@@ -44,16 +42,6 @@ def default_end_cond(rollout_save_n_timesteps, rollout_save_n_episodes):
     # without getting an error that `rollout_save_n_timesteps is not None`.
     if rollout_save_n_timesteps is None and rollout_save_n_episodes is None:
         rollout_save_n_timesteps = 2000  # Min timesteps saved per file, optional.
-
-
-# Default config for CNN Policies
-@train_rl_ex.named_config
-def cnn_policy():
-    train = dict(
-        policy_cls=CnnPolicy,
-        policy_kwargs=dict(features_extractor_class=NatureCNN),
-    )
-    locals()  # make flake8 happy
 
 
 # Standard Gym env configs
--- a/src/imitation/scripts/config/train_rl.py
+++ b/src/imitation/scripts/config/train_rl.py
@@ -1,6 +1,8 @@
 """Configuration settings for train_rl, training a policy with RL."""
 
 import sacred
+from stable_baselines3.common.torch_layers import NatureCNN
+from stable_baselines3.ppo import CnnPolicy
 
 from imitation.scripts.common import common, rl, train
 
@@ -42,6 +44,16 @@ def default_end_cond(rollout_save_n_timesteps, rollout_save_n_episodes):
     # without getting an error that `rollout_save_n_timesteps is not None`.
     if rollout_save_n_timesteps is None and rollout_save_n_episodes is None:
         rollout_save_n_timesteps = 2000  # Min timesteps saved per file, optional.
+
+
+# Default config for CNN Policies
+@train_rl_ex.named_config
+def cnn_policy():
+    train = dict(
+        policy_cls=CnnPolicy,
+        policy_kwargs=dict(features_extractor_class=NatureCNN),
+    )
+    locals()  # make flake8 happy
 
 
 # Standard Gym env configs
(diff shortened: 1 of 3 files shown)


src/imitation/scripts/config/train_rl.py
```diff
@@ -44,6 +46,16 @@ def default_end_cond(rollout_save_n_timesteps, rollout_save_n_episodes):
         rollout_save_n_timesteps = 2000  # Min timesteps saved per file, optional.
 
 
+# Default config for CNN Policies
+@train_rl_ex.named_config
+def cnn_policy():
+    train = dict(
+        policy_cls=CnnPolicy,
+        policy_kwargs=dict(features_extractor_class=NatureCNN),
```
AdamGleave, line 54: We should probably remove the config hook in `common/train.py` and then remove `policy_kwargs`.

Should probably also move the named config itself into `common/train.py`.

anshradh: Do you have pointers to where I should update documentation since I'm removing the hook?

tests/scripts/test_scripts.py
```diff
@@ -723,6 +723,29 @@ def test_train_rl_double_normalization(tmpdir: str, rng):
...
 
 
+def test_train_rl_cnn_policy(tmpdir: str, rng):
+    venv = util.make_vec_env(
+        "AsteroidsNoFrameskip-v4",
+        n_envs=1,
+        parallel=False,
+        rng=rng,
+    )
+    print(venv)
```
AdamGleave, line 733: Do you want to leave this in?

tests/scripts/test_scripts.py
```diff
@@ -723,6 +723,29 @@ def test_train_rl_double_normalization(tmpdir: str, rng):
...
+    net = reward_nets.CnnRewardNet(venv.observation_space, venv.action_space)
+    tmppath = os.path.join(tmpdir, "reward.pt")
+    th.save(net, tmppath)
+
+    log_dir_data = os.path.join(tmpdir, "train_rl")
+    with pytest.warns(RuntimeWarning):
```
AdamGleave, line 739: What warning are we expecting to see here? We don't have any warning specifically for this case do we?

(Separately I'm worried the other tests checking for `RuntimeWarning` are too vague, they might be passing because of other warnings that always get raised, we should probably add a match to check for the relevant message.)

tests/scripts/test_scripts.py
```diff
@@ -723,6 +723,29 @@ def test_train_rl_double_normalization(tmpdir: str, rng):
...
+        train_rl.train_rl_ex.run(
+            named_configs=["train.cnn_policy"] + ALGO_FAST_CONFIGS["rl"],
+            config_updates=dict(
+                common=dict(log_dir=log_dir_data, env_name="AsteroidsNoFrameskip-v4"),
+                reward_path=tmppath,
+            ),
+        )
+
```
AdamGleave, line 747: Consider checking:
```
```

like in `test_train_rl_main`.

Right now I think this test could pass even if there were some exceptions as Sacred will just capture them (but return `run.status == "ERROR"` or the like.)

src/imitation/scripts/config/train_rl.py
AdamGleave: On reflection I think we basically don't have the script configs documented, which is sad :( So maybe it's OK to not update the docs. Pro tip: the documentation can't be out of date if you don't have it...

Should update the comment in `normalize_disable` at least though as that's now out of date (mentions the now-removed config hook).

AdamGleave (changes requested):
Config itself looks good. Some questions about the test case.

tests/scripts/test_scripts.py
anshradh: Yeah good point - removed checking for a warning here (doesn't really make sense). Should I make an issue for checking for the right messages on some of the other tests?

AdamGleave: Yes, please make an issue. If you could also make a PR to solve that issue that'd be great, but it's definitely out of scope for this PR.

AdamGleave:
> I think we can actually just remove the `normalize_disable` named_config? Seems like it's not needed after removing the config hook

Yeah, that seems right. There might be some edge case where we want it around (some policy with weird default?), but it's not hard to add back in later if we need it, so may as well simplify config code and remove it.

tests/scripts/test_scripts.py
anshradh: Done.

tests/scripts/test_scripts.py
anshradh: Haha nope, thanks.

tests/scripts/test_scripts.py
anshradh: Looks like there's actually just one test that does this, made an issue at #614. Will put up a PR in a bit to address.

AdamGleave (approved):
LGTM

PR #612, opened 2022-11-07
Copy doc string to suppress mypy type error.

dan-pandori:
## Description

mypy is a little overly worried about using abstract function calls, even to just grab the doc string. Copying the doc string makes this not a problem.

## Testing

Ran mypy on it, noticed that this was failing in circleci previously.

### DIFF
--- a/src/imitation/rewards/reward_nets.py
+++ b/src/imitation/rewards/reward_nets.py
@@ -318,7 +318,7 @@ class PredictProcessedWrapper(RewardNetWrapper):
         next_state: th.Tensor,
         done: th.Tensor,
     ) -> th.Tensor:
-        __doc__ = super().forward.__doc__  # noqa: F841
+        __doc__ = super().forward.__doc__  # type: ignore[safe-super] # noqa: F841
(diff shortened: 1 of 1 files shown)


dan-pandori:
@AdamGleave honestly, I don't totally understand the error that this is suppressing. Sending this your way on the basis that it's better to only be surfacing new errors that folks introduce in their own PRs, rather than existing issues in the code base.

You previously mentioned that this might be coming from a mypy version update.

AdamGleave (commented):
This seems like a `mypy` false-positive. If I remove the `abstractmethod` decorator from `forward` then it's fine with this. But we're not actually _calling_ `forward()`, just reading its docstring.

Your fix works OK, but I wonder if it's more effort than it's worth at this point, perhaps we should just copy the (one-sentence!) docstring over from `RewardNet.forward` and avoid the `__doc__ = ...` magic?

dan-pandori:
Works for me, updated to just copy the doc string from RewardNet, which is also sufficient to make mypy happy (at least locally).

Issue #605, opened 2022-11-02
Use `pre-commit` in `.circleci/config.yml`


AdamGleave:
## Problem
We currently have linting, type checking, and other static analysis duplicated between `.circleci/config.yml` and `pre-commit`. This makes it easy to introduce inconsistencies.

## Solution
Call `pre-commit run --all <hook-name>` in `.circleci/config.yml` instead of the command directly.

## Possible alternative solutions
We could also migrate some of our CI to something like https://pre-commit.ci/

PR #613, opened 2022-11-08
Refactor circleci to reuse commands from precommit

dan-pandori:
## Description

Run various hooks in circleci via calling them through precommit. This should reduce the likelihood that there is a divergence between the two.
https://github.com/HumanCompatibleAI/imitation/issues/605

It's somewhat difficult to pass the arguments as we'd like to, due to:
https://github.com/pre-commit/pre-commit/issues/2083

To get around this, we directly rely on NUM_CPUS in the bash call to pytype. For flake8, we instead allow the existing flake8 parallelism to work w/o specifying --j.

## Testing

Confirmed that circleci continues to pass, and ran the hooks locally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,6 +222,7 @@ jobs:
             rm -rf shellcheck-stable
             chmod a+x /usr/local/bin/shellcheck
 
+      # TODO can this be combined?
       - shellcheck/check:
           dir: .
           # Orb invokes shellcheck once per file. shellcheck complains if file
@@ -231,7 +232,7 @@ jobs:
 
       - run:
           name: ipynb-check
-          command: ./ci/clean_notebooks.py --check ./docs/tutorials
+          command: pre-commit run --all check-notebooks
 
       - run:
           name: typeignore-check
@@ -244,11 +245,19 @@ jobs:
 
       - run:
           name: black
-          command: black --version && black --check --diff ${SRC_FILES}
+          # Unsure if removing the --diff arg is OK here
+          command: |
+            pre-commit run --all black
+            pre-commit run --all black-jupyter
+
+      - run:
+          name: isort 
+          command: pre-commit run --all isort
 
       - run:
           name: codespell
-          command: codespell -I .codespell.skip --skip='*.pyc,tests/testdata/*,*.ipynb,*.csv' ${SRC_FILES}
+          # How to pass SRC files?
+          command: pre-commit run --all codespell
 
   doctest:
     executor: static-analysis-medium
@@ -270,11 +279,11 @@ jobs:
 
       - run:
           name: pytype
-          command: pytype --version && pytype -j "${NUM_CPUS}" ${SRC_FILES[@]}
+          command: pytype --version && pre-commit run --all pytype
 
       - run:
           name: mypy
-          command: mypy --version && mypy ${SRC_FILES[@]} --follow-imports=silent --show-error-codes
+          command: mypy --version && pre-commit run --all mypy
 
   unit-test-linux:
     executor: unit-test-linux
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,7 @@ version: 2.1
 
 orbs:
   codecov: codecov/codecov@3.2.3
+  shellcheck: circleci/shellcheck@3.1.2
   win: circleci/windows@4.1
 
 defaults: &defaults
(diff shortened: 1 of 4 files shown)


dan-pandori:
@AdamGleave, ready for review when you've got a chance.

.circleci/config.yml
```diff
@@ -231,23 +231,40 @@ jobs:
...
-          command: codespell -I .codespell.skip --skip='*.pyc,tests/testdata/*,*.ipynb,*.csv' ${SRC_FILES}
+          command: pre-commit run --all codespell
+
+      - run:
+          name: lint-misc
+          command: |
+            pre-commit run --all check-ast
+            pre-commit run --all trailing-whitespace
+            pre-commit run --all end-of-file-fixer
+            pre-commit run --all check-toml
```
AdamGleave, line 267: Should we have `check-added-large-files` as well?

Maybe that's OK to exclude as won't cause CI to fail on subsequent commits.

.pre-commit-config.yaml
```diff
@@ -42,7 +42,7 @@ repos:
   rev: v2.2.2
   hooks:
   - id: codespell
-    args: ["--skip=*.pyc,tests/testdata/*,*.ipynb,*.csv"]
+    args: ["--skip=*.pyc,tests/testdata/*,*.ipynb,*.csv","--ignore-words-list=reacher"]
```
AdamGleave, line 45: We used to have `iff` skipped in `.codespell.skip` -- is that not needed?

We should probably also delete `.codespell.skip` if no longer used.

.circleci/config.yml
```diff
@@ -231,23 +231,40 @@ jobs:
```
AdamGleave, line 231: Should we replace the shellcheck part of this file with the pre-commit hook?

.circleci/config.yml
```diff
@@ -231,23 +231,40 @@ jobs:
...
+
+      - run:
+          name: lint-misc
+          command: |
+            pre-commit run --all check-ast
+            pre-commit run --all trailing-whitespace
+            pre-commit run --all end-of-file-fixer
+            pre-commit run --all check-toml
 
   doctest:
```
AdamGleave, line 269: Should we replace this with `pre-commit run --all docs`?

AdamGleave (changes requested):
Left some minor comments inline that should be addressed before merging -- couple of hooks that were missed (or if not included on purpose, we should add a comment to explain it).

Other question is higher-level. `pre-commit` installs these hooks in their own virtual environment. I think our caching logic doesn't handle this as the CircleCI logs suggest it's installing black, etc from scratch each time. It's quite quick (the dominant cost is actually doing the type checking and running darglint, which is very expensive) but we should probably still update the caching logic to save these extra directories.

AdamGleave:
@anshradh would you mind finishing this PR off now that Dan's finished his trial (I just had a few minor suggestions for changes)?

anshradh:
Yup can do!

.circleci/config.yml
anshradh: Ignoring per our earlier conversation.

.circleci/config.yml
anshradh: I tried to do this, not confident I got it right though.

.pre-commit-config.yaml
anshradh: Done.

.circleci/config.yml
anshradh: Added.

.circleci/config.yml
AdamGleave: One small issue: `types: [bash]` actually skips everything (even with `pre-commit run --all shellcheck` nothing gets run). Leaving it as the default `types: [shell]` seems to work. My guess is pre-commit isn't smart enough to parse the shebang and realize we're using bash not sh, and is just going off of file extension.

I've pushed a commit to fix this.

AdamGleave:
Oh, doh, seems like the shellcheck hook needs Docker to run and we can't do Docker-in-Docker on CircleCI. So might have to just stick with the orb anyway.

AdamGleave (approved):
LGTM. Will merge once tests pass.

Rocamonde:
Seems like linting is failing after this has been merged.

AdamGleave:
Yes there's one notebook it wants to reformat. Can you open a PR to fix
this? I don't know why the tests passed before merge given this.

Issue #614, opened 2022-11-08
Improve `RuntimeWarning` check in `test_scripts:test_train_rl_double_normalization`


anshradh:
## Problem
The test checks for a RuntimeWarning when train_rl is called with normalization on an already normalized reward function, but doesn't actually check the message included in the warning.

## Solution
Test should also assert that the warning's message matches the warning we think it's hitting.

AdamGleave:
Fixed in #615

PR #615, opened 2022-11-08
Actually check RuntimeWarning message

anshradh:
## Description
Solves #614, adding a better check for the Runtime Warning in `test_scripts.py:test_train_rl_double_normalization`. 

## Testing
Tested locally.

### DIFF
--- a/tests/scripts/test_scripts.py
+++ b/tests/scripts/test_scripts.py
@@ -711,7 +711,10 @@ def test_train_rl_double_normalization(tmpdir: str, rng):
     th.save(net, tmppath)
 
     log_dir_data = os.path.join(tmpdir, "train_rl")
-    with pytest.warns(RuntimeWarning) as record:
+    with pytest.warns(
+        RuntimeWarning,
+        match=r"Applying normalization to already normalized reward function.*",
+    ):
         train_rl.train_rl_ex.run(
             named_configs=["cartpole"] + ALGO_FAST_CONFIGS["rl"],
             config_updates=dict(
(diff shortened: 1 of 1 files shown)


tests/scripts/test_scripts.py
```diff
@@ -711,7 +711,7 @@ def test_train_rl_double_normalization(tmpdir: str, rng):
     th.save(net, tmppath)
 
     log_dir_data = os.path.join(tmpdir, "train_rl")
-    with pytest.warns(RuntimeWarning):
+    with pytest.warns(RuntimeWarning) as record:
```
dan-pandori, line 714: Thoughts on just:

with pytest.warns(RuntimeWarning, match=r"Applying normalization to already normalized reward function.*"):

?

The upside is a little less code to think about and if the underlying code starts to generate other warnings as well we don't need to change the test.

The main downside seems to be that we're not longer guaranteeing that this is the only warning surfaced, but I'm not sure that maintaining that guarantee is going to be easy given the number of warnings I see elsewhere in the codebase due to usage of deprecated functions.

anshradh: I prefer that, actually, little bit easier to reason about and cleaner.

The change I’ve implemented actually has that same flaw I think, since it filters the list of warnings just to the warnings that contain this message and assets that the filtered list has length 1. Both solutions are robust to other warnings being generated I think.

anshradh: Done.

dan-pandori (approved):
LGTM assuming that tests pass

AdamGleave (approved):
LGTM

PR #617, opened 2022-11-08
Add test for reward trainers improving agent

dan-pandori:
## Description

Adds a test that BasicRewardTrainer and EnsembleRewardTrainer improve the performance of an agent when train is called.

## Testing

Ran the test with --flakefinder to see how long this needs to train before we can confidently expect reward to have improved. Training for (500, 20) was still fairly flaky with 3/80 flakes, but at (1000, 20) I didn't get any flakes when run over 100 tests.

### DIFF
--- a/tests/algorithms/test_preference_comparisons.py
+++ b/tests/algorithms/test_preference_comparisons.py
@@ -1016,7 +1016,7 @@ def ensemble_reward_trainer(venv, rng):
     "action_is_reward_trainer_func",
     [basic_reward_trainer, ensemble_reward_trainer],
 )
-def test_that_trainer_improves_rewards(
+def test_that_trainer_improves(
     action_is_reward_venv,
     action_is_reward_agent,
     action_is_reward_trainer_func,
@@ -1024,7 +1024,7 @@ def test_that_trainer_improves_rewards(
     custom_logger,
     rng,
 ):
-    """Tests that training improves performance of the reward network."""
+    """Tests that training improves performance of the reward network and agent."""
     action_is_reward_trainer = action_is_reward_trainer_func(action_is_reward_venv, rng)
     agent_trainer = preference_comparisons.AgentTrainer(
         action_is_reward_agent,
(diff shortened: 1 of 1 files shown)


dan-pandori:
@AdamGleave, what's your preference on keeping this test split up as it is now vs combining it with the test above? I was initially leaning towards keeping it split up so that what each test verifies is very clear individually.

On the other hand, it is also true that if the reward network doesn't improve we wouldn't expect the agent to improve, so it might be better to have one test where reward network improvement gates checking agent reward.

AdamGleave:
> @AdamGleave, what's your preference on keeping this test split up as it is now vs combining it with the test above? I was initially leaning towards keeping it split up so that what each test verifies is very clear individually.
> 
> On the other hand, it is also true that if the reward network doesn't improve we wouldn't expect the agent to improve, so it might be better to have one test where reward network improvement gates checking agent reward.

I lean slightly towards combining them as most of the setup code is identical, so I think it'll be easier for people to read through and understand the tests.

If it wasn't for that then having them split up seems fine. Multiple tests failing because of one root cause is annoying but fairly common scenario that we can deal with, and having extra granularity of where the test is failing can be helpful.

AdamGleave:
The flakiness is a problem though. We might need to adopt a more unit test style approach and check the reward training works, then that agent training works conditioned on a pre-initialized good reward network. Also possible RL hyperparameters just need to be tuned.

dan-pandori:
Moved it into one test then.

I misspoke on the flakiness. It was somewhat flaky when trained for shorter amounts of time, but this training length seems to be both not noticeably flaky and runs in ~10 seconds locally per test.

anshradh:
LGTM, though I’d be interested to see if it’s flaky at all when run for e.g. 500 runs (I think < 1% failure rate is totally fine though).

anshradh (approved):

dan-pandori:
This test takes ~9s to run on my laptop, so I replicated that running the test 100 times had 0 flakes because I didn't want to wait super long.

I wouldn't be shocked if the flake rate was ~1/500, but it currently seems to pass the bar of 'far from the flakiest test in the file'.

Issue #599, opened 2022-10-28
Nobody can figure out how to get rollouts of policies trained on image environments


dfilan:
## Problem
In issues #547 and #596, people try training an SB3 policy (let's call it `expert`) on some image environment (let's call it `my_img_env`) and then generating rollouts by calling `rollout.rollout(expert, my_img_env, sample_until_whatever)`.

The problem is that this doesn't work. `my_img_env` is basically always (h,w,c)-formatted, and the training of `expert` always involved internally transposing to (c,h,w) before feeding it into the policy. The code that people should write is `rollout.rollout(expert, expert.get_env(), sample_until_whatever)`.

How are people supposed to know they should do this? It's not a very natural thing to do - I don't think anybody learns about this feature of SB3 before running into it when it causes a bug, it's kind of weird that `expert` has a `get_env` method in the first place, and the documentation for `rollout.rollout` doesn't warn them that anything might go wrong.

## Possible solutions
- indicate that this can go wrong in the documentation of `rollout.rollout`
- indicate that this can go wrong in some tutorial
- change `rollout.rollout` so that it tries to transpose stuff in case of an observation space mismatch

AdamGleave:
Thanks for raising the issue, Daniel.

I think our first port of call should be to add an example to our [CNN preference comparison tutorial](https://github.com/HumanCompatibleAI/imitation/blob/master/docs/tutorials/5a_train_preference_comparisons_with_cnn.ipynb) and generally document the weirdness of their being different conventions between `imitation`/`torch` and `stable-baselines3`. We should also add more informative error messages -- so if someone tries to do this, they get a message linking to the docs, rather than just some weird shape error.

Long-term I'd like a solution that either eliminates this complexity (e.g. make SB3 consistent), or (since that might break a lot of existing SB3 users) some adaptor layer that hides most of this complexity from users. Perhaps we will get inspiration for that when writing the docs.

Assigning this to you @dan-pandori to work on next week. But it's a thorny issue, we should discuss in person.

dfilan:
Closed as solved by #618.

Disastorm:
I think I'm getting this or something similar to this when using BufferingWrapper.pop_transitions()  , is this also another expected location that will be different dimensions?

PR #618, opened 2022-11-08
Add a verbose warning for calling rollout.rollout on image environments

dan-pandori:
## Description

Add a better warning message for the common issue of calling rollout.rollout using the environment instead of the expert's get_env() method.
https://github.com/HumanCompatibleAI/imitation/issues/599

## Testing

Added a test and also used the repro script from the issue and confirmed that you get a nice error message now. Ex:
```
ValueError: Policy and environment observation shape mismatch. This is likely caused by https://github.com/HumanCompatibleAI/imitation/issues/599. If encountering this from rollout.rollout, try calling:
rollout.rollout(expert, expert.get_env(), ...) instead of
rollout.rollout(expert, env, ...)

Policy observation shape: (4, 84, 84) 
Environment observation shape: (84, 84, 4)
```


dan-pandori:
@AdamGleave, ready for you to take a look. Also of course feel free to hand this off to someone else to review, I'm still just not super sure who is the right match for these reviews.

anshradh (approved):
LGTM! I think you can add a `# pragma: no cover` to get around the code coverage check btw (not sure why it thinks the added code isn't covered in the first place though, so might be a deeper issue here I'm missing).

AdamGleave (approved):
The code coverage happens because you only test that the error message occurs, not that it _doesn't_ occur and the original exception is raised when you have some different mismatch.

I'd weakly vote in favour of adding another test to check it provides original exception on other kinds of shape mismatch, but otherwise it LGTM.

dan-pandori:
Mm, that makes sense. Added a test for that case.

I didn't add a third test for the branch where both the correct observation space and the one passed have a shape of length 3, but don't line up in a way that corresponds to this common mistake. But I can add it if you'd prefer.

dan-pandori:
OK, codecov's demands are more draconian than I anticipated, so I extended the test to cover this case as well.

AdamGleave:
Thanks! We can override codecov check when it's obviously stupid, but in this case the additional test seems simple enough anyway.

Issue #551, opened 2022-09-05
Review and fix flaky tests


Rocamonde:
## Bug description

Siince #534, the random number generator convention in the entire code has been redesigned, and all tests have been changed by default to not have fixed seeds and explore internal sources of entropy instead, unless specifically testing for seed consistency. Some of the tests have been exposed as flaky as a result, and this requires either fixing the seeds to a value that always works (not ideal), or improving the test design to reduce the flakiness to an acceptable fraction e.g. 10^-3.

A specific test that seems to be flaky is the test `test_trainer_makes_progress`, specifically the assertion
https://github.com/HumanCompatibleAI/imitation/blob/312b091d550c11076d40d056d5c699fee965c2df/tests/algorithms/test_dagger.py#L302
fails fairly frequently when developing locally. 

## Steps to reproduce
run
```bash
$ pytest -n 10 tests/algorithms/test_dagger.py::test_trainer_makes_progress --flake-finder --flake-runs=100
```
to get 38% flakiness:
```
=========== test session starts ===========
platform darwin -- Python 3.8.13, pytest-7.1.2, pluggy-1.0.0
rootdir: /Users/rocamonde/git/HumanCompatibleAI/imitation, configfile: setup.cfg
plugins: anyio-3.6.1, xdist-2.5.0, forked-1.4.0, pytest_notebook-0.8.0, hypothesis-6.54.4, flakefinder-1.0.0, cov-3.0.0
gw0 [800] / gw1 [800] / gw2 [800] / gw3 [800] / gw4 [800] / gw5 [800] / gw6 [800] / gw7 [800] / gw8 [800] / gw9 [800]
sssss.......FF.ss......ss.ss.ss.ssF......F...F....ss....ss.ss...ss...ss............ss.s.sss...........ss.ss.......ss.ss........ss........ss.ssF.sss...s....s...sF....ss........s.....F..s.Fs.......s..s...............s.s..s..s...s.s...s.s.F.s.........s...F.s.s.F..s...s.............s.........s..s....s..s... [ 38%]
```

## Environment
- Operating system and version: macOS 12.4 Monterey
- Python version: Python 3.8.18

AdamGleave:
38% is a lot, I'd expect to see errors happening frequently in CI. This happens even on the master branch?

AdamGleave:
Ah 3.9% failures is believable we wouldn't have noticed. Agreed we should fix this, and probably don't want it to be randomly skipped either...

[...]

PR #621, opened 2022-11-09
Address Flaky Tests

anshradh:
## Description
Fix issues with some flaky tests, addressing #551. Most of these were flaky when run in parallel so their flakiness wasn't showing up in CI
## Testing
Ran new tests locally with `pytest-flakefinder` for many (e.g. 500) runs

### DIFF
--- a/experiments/bc_benchmark.sh
+++ b/experiments/bc_benchmark.sh
@@ -11,7 +11,8 @@ source experiments/common.sh
 
 ENVS=(seals_cartpole)
 SEEDS=(0 1 2)
-OUTPUT_DIR="output/bc_benchmark/${TIMESTAMP}"
+# To prevent race conditions, we use a different output dir for each process id.
+OUTPUT_DIR="output/bc_benchmark/${TIMESTAMP}/$BASHPID"
 extra_configs=()
 extra_options=()
 extra_parallel_options=()
--- a/experiments/bc_benchmark.sh
+++ b/experiments/bc_benchmark.sh
@@ -12,7 +12,7 @@ source experiments/common.sh
 ENVS=(seals_cartpole)
 SEEDS=(0 1 2)
 # To prevent race conditions, we use a different output dir for each process id.
-OUTPUT_DIR="output/bc_benchmark/${TIMESTAMP}/$BASHPID"
+OUTPUT_DIR="output/bc_benchmark/${TIMESTAMP}-${BASHPID}"
 extra_configs=()
 extra_options=()
 extra_parallel_options=()
--- a/experiments/dagger_benchmark.sh
+++ b/experiments/dagger_benchmark.sh
@@ -5,7 +5,7 @@ source experiments/common.sh
 ENVS=(seals_cartpole)
 SEEDS=(0 1 2 3 4)
 # To prevent race conditions, we use a different log root for each process id.
-LOG_ROOT="output/dagger_benchmark/${TIMESTAMP}/$BASHPID"
+LOG_ROOT="output/dagger_benchmark/${TIMESTAMP}-${BASHPID}"
 extra_configs=()
 extra_options=()
 extra_parallel_options=()
--- a/experiments/dagger_benchmark.sh
+++ b/experiments/dagger_benchmark.sh
@@ -4,7 +4,8 @@ source experiments/common.sh
 
 ENVS=(seals_cartpole)
 SEEDS=(0 1 2 3 4)
-LOG_ROOT="output/dagger_benchmark/${TIMESTAMP}"
+# To prevent race conditions, we use a different log root for each process id.
+LOG_ROOT="output/dagger_benchmark/${TIMESTAMP}/$BASHPID"
 extra_configs=()
 extra_options=()
 extra_parallel_options=()
--- a/experiments/imit_benchmark.sh
+++ b/experiments/imit_benchmark.sh
@@ -11,7 +11,8 @@ source experiments/common.sh
 
 SEEDS=(0 1 2 3 4)
 CONFIG_CSV="experiments/imit_benchmark_config.csv"
-LOG_ROOT="output/imit_benchmark/${TIMESTAMP}"
+# To prevent race conditions, we use a different log root for each process id.
+LOG_ROOT="output/imit_benchmark/${TIMESTAMP}/$BASHPID"
 extra_configs=()
 extra_options=()
 extra_parallel_options=()
--- a/experiments/imit_benchmark.sh
+++ b/experiments/imit_benchmark.sh
@@ -12,7 +12,7 @@ source experiments/common.sh
 SEEDS=(0 1 2 3 4)
 CONFIG_CSV="experiments/imit_benchmark_config.csv"
 # To prevent race conditions, we use a different log root for each process id.
-LOG_ROOT="output/imit_benchmark/${TIMESTAMP}/$BASHPID"
+LOG_ROOT="output/imit_benchmark/${TIMESTAMP}-${BASHPID}"
 extra_configs=()
 extra_options=()
 extra_parallel_options=()
--- a/experiments/rollouts_from_policies.sh
+++ b/experiments/rollouts_from_policies.sh
@@ -17,7 +17,7 @@ source experiments/common.sh
 
 CONFIG_CSV=${CONFIG_CSV:-experiments/rollouts_from_policies_config.csv}
 # To prevent race conditions, we use a different output dir for each process id.
-OUTPUT_DIR="output/train_experts/${TIMESTAMP}/$BASHPID"
+OUTPUT_DIR="output/train_experts/${TIMESTAMP}-${BASHPID}"
 
 if ! TEMP=$($GNU_GETOPT -o f -l fast -- "$@"); then
   exit 1
--- a/experiments/rollouts_from_policies.sh
+++ b/experiments/rollouts_from_policies.sh
@@ -16,7 +16,8 @@ set -e  # Exit on error.
 source experiments/common.sh
 
 CONFIG_CSV=${CONFIG_CSV:-experiments/rollouts_from_policies_config.csv}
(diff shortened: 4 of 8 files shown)


anshradh:
@AdamGleave I think this PR is reasonable to merge in now, but also happy to wait in case I find some other candidate tests we want to include - these are the ones that have a > 2% flake rate when run in parallel.

tests/algorithms/test_preference_comparisons.py
```diff
@@ -71,6 +71,21 @@ def agent_trainer(agent, reward_net, venv, rng):
     return preference_comparisons.AgentTrainer(agent, reward_net, venv, rng)
 
 
+def assert_info_arrs_equal(arr1, arr2):  # pragma: no cover
```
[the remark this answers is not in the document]
anshradh: Case 1 is right here, we should be asserting that items1 and 2 are dictionaries!

tests/policies/test_exploration_wrapper.py
```diff
@@ -90,37 +90,65 @@ def test_switch_prob(rng):
...
-    assert num_constant == 1000
+
+    num_randoms_prob_one = []
+    num_constants_prob_one = []
+    num_randoms_prob_half = []
+    num_constants_prob_half = []
+    num_randoms_prob_zero = []
+    num_constants_prob_zero = []
+
+    for s in range(5):
```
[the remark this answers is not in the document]
anshradh: Yeah the specific reason I went for this route was because Windows has a different RNG and the original seed we were using caused the test to fail only on Windows, so this was a way to check that that seed was the actual issue. I think probably the better thing to do here is to assert that the mean is > 450 across the five runs with different seeds.

tests/policies/test_exploration_wrapper.py
```diff
@@ -90,37 +90,65 @@ def test_switch_prob(rng):
...
+        np.random.seed(s)
+        obs = np.random.rand(100, 2)
+        for action in wrapper(obs):
+            assert venv.action_space.contains(action)
+            assert wrapper.current_policy == policy
```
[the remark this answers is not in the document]
anshradh: Yeah moved this out, thanks!

tests/policies/test_exploration_wrapper.py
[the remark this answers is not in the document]
dan-pandori: Isn't asserting that the mean is between 450 and 550 mathematically equivalent to having one run that's 5 times as long, and asserting the total is between 2250 and 2750? Admittedly, that's ignoring the unlucky seed issue. 

I also thought that there was a push to try to remove static seed usage in these tests in general.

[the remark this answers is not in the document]
anshradh: I agree with you, removed dependence on seeds entirely and increased the number of steps.

experiments/bc_benchmark.sh
```diff
@@ -11,7 +11,8 @@ source experiments/common.sh
 
 ENVS=(seals_cartpole)
 SEEDS=(0 1 2)
-OUTPUT_DIR="output/bc_benchmark/${TIMESTAMP}"
+# To prevent race conditions, we use a different output dir for each process id.
+OUTPUT_DIR="output/bc_benchmark/${TIMESTAMP}/$BASHPID"
```
AdamGleave, line 15: Preference for this being the same directory so we don't create lots of directories with a single subdirectory.

Also suggest using `${}` for consistency with surrounding code although in this situation it doesn't matter.

```suggestion
OUTPUT_DIR="output/bc_benchmark/${TIMESTAMP}-${BASHPID}"
```

experiments/dagger_benchmark.sh
```diff
@@ -4,7 +4,8 @@ source experiments/common.sh
 
 ENVS=(seals_cartpole)
 SEEDS=(0 1 2 3 4)
-LOG_ROOT="output/dagger_benchmark/${TIMESTAMP}"
+# To prevent race conditions, we use a different log root for each process id.
+LOG_ROOT="output/dagger_benchmark/${TIMESTAMP}/$BASHPID"
```
AdamGleave, line 8: as above
```suggestion
LOG_ROOT="output/dagger_benchmark/${TIMESTAMP}-${BASHPID}"
```

experiments/imit_benchmark.sh
```diff
@@ -11,7 +11,8 @@ source experiments/common.sh
 
 SEEDS=(0 1 2 3 4)
 CONFIG_CSV="experiments/imit_benchmark_config.csv"
-LOG_ROOT="output/imit_benchmark/${TIMESTAMP}"
+# To prevent race conditions, we use a different log root for each process id.
+LOG_ROOT="output/imit_benchmark/${TIMESTAMP}/$BASHPID"
```
AdamGleave, line 15: as above
```suggestion
LOG_ROOT="output/imit_benchmark/${TIMESTAMP}-${BASHPID}"
```

experiments/rollouts_from_policies.sh
```diff
@@ -16,7 +16,8 @@ set -e  # Exit on error.
 source experiments/common.sh
 
 CONFIG_CSV=${CONFIG_CSV:-experiments/rollouts_from_policies_config.csv}
-OUTPUT_DIR="output/train_experts/${TIMESTAMP}"
+# To prevent race conditions, we use a different output dir for each process id.
+OUTPUT_DIR="output/train_experts/${TIMESTAMP}/$BASHPID"
```
AdamGleave, line 20:
```suggestion
OUTPUT_DIR="output/train_experts/${TIMESTAMP}-${BASHPID}"
```

experiments/transfer_learn_benchmark.sh
```diff
@@ -8,7 +8,8 @@ source experiments/common.sh
 SEEDS=(0 1 2)
 CONFIG_CSV="experiments/imit_benchmark_config.csv"
 REWARD_MODELS_DIR="data/reward_models"
-LOG_ROOT="output/train_experts/${TIMESTAMP}"
+# To prevent race conditions, we use a different log root for each process id.
+LOG_ROOT="output/train_experts/${TIMESTAMP}/$BASHPID"
```
AdamGleave, line 12:
```suggestion
LOG_ROOT="output/train_experts/${TIMESTAMP}-${BASHPID}"
```

experiments/transfer_learn_benchmark.sh
```diff
@@ -67,15 +68,18 @@ done
 
 if [[ $NEED_TEST_FILES == "true" ]]; then
   # Generate quick reward models for test.
-  save_dir=tests/testdata/reward_models/${ALGORITHM}
+  # To prevent race conditions, we use a different save_dir for each process id.
+  save_dir=tests/testdata/reward_models/${ALGORITHM}/${TIMESTAMP}/$BASHPID
```
AdamGleave, line 72:
```suggestion
  save_dir=tests/testdata/reward_models/${ALGORITHM}/${TIMESTAMP}-${BASHPID}
```

experiments/transfer_learn_benchmark.sh
```diff
@@ -67,15 +68,18 @@ done
...
 
   # Wipe directories for writing later.
   if [[ -d ${save_dir} ]]; then
-    rm -r ${save_dir}
+    echo "Wiping ${save_dir}"
```
AdamGleave, line 76: do we still want this or was it just for debugging?

experiments/transfer_learn_benchmark.sh
```diff
@@ -88,7 +92,7 @@ parallel -j 25% --header : --results "${LOG_ROOT}/parallel/" --colsep , --progre
   '{env_config_name}' seed='{seed}' \
   common.log_dir="${LOG_ROOT}/${ALGORITHM}/{env_config_name}_{seed}/n_expert_demos_{n_expert_demos}" \
   reward_type="RewardNet_unshaped" \
-  reward_path="${REWARD_MODELS_DIR}/${ALGORITHM}/{env_config_name}_0/n_expert_demos_{n_expert_demos}/checkpoints/final/reward_test.pt" \
+  reward_path="${REWARD_MODELS_DIR}/${ALGORITHM}/${TIMESTAMP}/$BASHPID/{env_config_name}_0/n_expert_demos_{n_expert_demos}/checkpoints/final/reward_test.pt" \
```
AdamGleave, line 95:
```suggestion
  reward_path="${REWARD_MODELS_DIR}/${ALGORITHM}/${TIMESTAMP}-${BASHPID}/{env_config_name}_0/n_expert_demos_{n_expert_demos}/checkpoints/final/reward_test.pt" \
```

tests/algorithms/test_preference_comparisons.py
```diff
@@ -72,6 +72,21 @@ def agent_trainer(agent, reward_net, venv, rng):
...
+    for item1, item2 in zip(arr1, arr2):
+        assert isinstance(item1, dict)
+        assert isinstance(item2, dict)
+        for key, val in item1.items():
+            assert key in item2
+            if isinstance(val, dict):
```
AdamGleave, line 81: Do we want to enforce that dicts are only nested one layer deep? If not we could just turn this into a recursive call (it'd also simplify the code)?

AdamGleave (commented):
Some fairly nitpicky comments, otherwise it LGTM. Thanks for making our tests reliable Ansh!

experiments/transfer_learn_benchmark.sh
anshradh: Good catch, thanks!

tests/algorithms/test_preference_comparisons.py
anshradh: I think it might make sense to leave as is, so we can enforce that dicts are either unnested or nested one layer deep, so that the test breaking will alert us that the info array contains dicts nested even further? I don't have a strong opinion here though.

anshradh: Yeah on reflection changed this to the recursive call, makes it easier to reason about.

PR #623, opened 2022-11-10
Require files to be changed before running publish to PyPi

dan-pandori:
## Description

When creating a new branch, this workflow runs but will fail. The leading theory is that this is caused by the branch having the same hash, due to no files being changed.

By only running this on pushes that actually change a file, this should skip the run on new branch creation.

## Testing

I created another branch on top of this one (dpandori_quiet_test), and I confirmed that there was no workflow running for it.

Then, to test that it still works when you actually made an edit, I made an arbitrary edit, pushed, and confirmed that the workflow ran.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -2,7 +2,12 @@
 
 name: Publish imitation distributions 📦 to PyPI and TestPyPI
 
-on: push
+on:
+  push:
+    paths:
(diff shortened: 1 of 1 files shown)


dan-pandori:
@AdamGleave, this addresses the overly eager workflow I mentioned in the Slack.

anshradh:
LGTM

anshradh (approved):

PR #624, opened 2022-11-11
Remove cause of possible flakiness of tests when executing train_rl in parallel

anshradh:
## Description

Remove modification to experiment adding observer for running experiments in parallel, reducing flakiness of tests ran in parallel. Related to #551 

## Testing

Tested locally

@AdamGleave I wasn't able to reliably repro this issue but I figured it made sense to make the modification we discussed earlier anyway. I think it's maybe reasonable to hold off on merging this until we have more solid evidence that there's an issue, here, though it does seem like the previous code could have led to issues.


AdamGleave (approved):
LGTM. Mutating a global variable in tests is definitely undesirable so even if we can't repro the problem that led us to look at this issue we should fix it

Issue #607, opened 2022-11-07
How to properly inserts sb3 `EvalCallback` for `AdversarialTraining`?


gunnxx:
Seems like there is no easy way to insert `EvalCallback` to `AdversarialTraining`, no?

`AdversarialTraining.train_gen()` populates `callback` argument of `sb3.BaseAlgorithm.learn()` with `self.gen_callback` which is either `None` or `reward_wrapper.RewardVecEnvWrapper()`.

As an additional note, `learn_kwargs` argument in `AdversarialTraining.train_gen()` is always `None` because `AdversarialTraining.train()` ignores that argument.

I think it is really crucial to monitor the progress of the imitation learning algorithm against true reward function. Using `EvalCallback` instead of specifying `eval_freq` and `create_eval_env` 1) is the recommended way in sb3 and 2) will be a "seamless plug" (because we just need to create eval environment inestad of doing fancy wrapping/unwrapping if we try to infer eval environment from the training environment).

AdamGleave:
At first glance this seems like a real issue -- I think we first wrote this code for SB2 before the callback API was finalized, and then it just got ported across. We do support our own `callback` method for `train()` but that's not serving quite the same purpose.

`learn_kwargs` seems like we could probably just delete it.

@taufeeque9 do you mind looking into this?

taufeeque9:
Sure! 

Seems like adding a `gen_callback` argument to `train()` method should resolve this. We can use the `sb3.common.callbacks.CallbackList` class to call both the callbacks -- `reward_wrapper.RewardVecEnvWrapper` and `gen_callback`.

Deleting `learn_kwargs` seems okay to me too.

I'll create a PR with the above changes.

AdamGleave:
That change sounds good to me. We may want to consider deprecating our own `callback` feature, although it may still serve some purpose, so should first check if it's being used anywhere in the codebase that can't be replaced by `gen_callback`.

gunnxx:
I checked that `callback` feature is not used anywhere. It is specific to `AdversarialTrainer.train()` method and the children (ie. `GAIL` and `AIRL`) do not overload the `train()` hence they do not use it. Can I be the one who make the PR for the changes as mentioned by @taufeeque9 ? It looks trivial to implement.

taufeeque9:
[...]

PR #626, opened 2022-11-15
Change callback for `AdversarialTrainer`

gunnxx:
Changing the callback mechanism of `AdversarialTrainer` such that we can insert `sb3.EvalCallback`. See #607.

### DIFF
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -2,13 +2,14 @@
 import abc
 import dataclasses
 import logging
-from typing import Callable, Iterable, Iterator, Mapping, Optional, Type, overload
+from typing import Iterable, Iterator, List, Mapping, Optional, Type, overload
 
 import numpy as np
 import torch as th
 import torch.utils.tensorboard as thboard
 import tqdm
 from stable_baselines3.common import base_class, on_policy_algorithm, policies, vec_env
+from stable_baselines3.common.callbacks import BaseCallback
 from stable_baselines3.sac import policies as sac_policies
 from torch.nn import functional as F
 
@@ -421,7 +422,7 @@ class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
     def train(
         self,
         total_timesteps: int,
-        callback: Optional[Callable[[int], None]] = None,
+        callback: Optional[List[BaseCallback]] = None
     ) -> None:
         """Alternates between training the generator and discriminator.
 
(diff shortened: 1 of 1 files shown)


src/imitation/algorithms/adversarial/common.py
```diff
@@ -434,10 +435,15 @@ def train(
...
-                single argument, the round number. Round numbers are in
-                `range(total_timesteps // self.gen_train_timesteps)`.
+            callback: List of stable_baslines3 callback to be passed to the policy
+                learning function.
         """
+        if callback is not None:
+            if self.gen_callback is None:
+                self.gen_callback = callback
+            else:
+                self.gen_callback = callback + [self.gen_callback]
```
mifeet, line 445: Can someone abuse the API by calling `train()` multiple times? If so, the value of `self.gen_callback` would contain nested list, which is not correct. Generally, the value of gen_callback is currently `Optional[BaseCallback]` and we shouldn't  change the type to a list at runtime.

Perhaps it would be better to add an optional callback argument to `train_gen()`, merge callbacks there, and avoid the stateful change here?

mifeet: Also, can the `learn_kwargs` argument from `train_gen()` be removed, as discussed in the original issue #607 ?

src/imitation/algorithms/adversarial/common.py
```diff
@@ -421,7 +422,7 @@ def train_gen(
     def train(
         self,
         total_timesteps: int,
-        callback: Optional[Callable[[int], None]] = None,
+        callback: Optional[List[BaseCallback]] = None
```
mifeet, line 425: Do we want to change the semantics of the argument here, or should we rather deprecate the feature (and introduce a different parameter for additional gen_callback)?

I think the suggestion in the original issue was to add a new `gen_callback` argument. (Btw, stable-baselines [supports both](https://stable-baselines3.readthedocs.io/en/master/guide/callbacks.html#callbacklist) CallbackList and list of callbacks if we wanted to be fancy)

mifeet (changes requested):

src/imitation/algorithms/adversarial/common.py
```diff
@@ -421,7 +422,7 @@ def train_gen(
     def train(
```
mifeet, line 422: One more thing - if you change the arguments, update of [training_adversarial.py](https://github.com/gunnxx/imitation/blob/7b3ed752bae1d48af7c4e5d6be06a3447b39c713/src/imitation/scripts/train_adversarial.py#L154) will also be needed

mifeet (changes requested):

PR #627, opened 2022-11-21
Add benchmarking results

Rocamonde:
## Description

Add benchmarking results and hyperparameters, previous to publishing the paper preprint.


Rocamonde:
@taufeeque9 please take a look if you can!

taufeeque9 (approved):
Looks good. Thanks for writing the util file and adding all the hyperparameter configs!

We can update the files with the argument for the demonstration data once #629 gets approved.

benchmarking/util.py
```diff
@@ -0,0 +1,114 @@
...
+    with open(write_path / f"{file.parents[3].name}.json", "w") as f:
+        json.dump(config, f, indent=4)
+
+
+if __name__ == "__main__":
+    # get two arguments from the terminal. The first positional argument contains the
+    # path to a txt file that has a list of paths to config files.
+    # The second positional argument is the path to the directory where
+    # the cleaned config files should be written.
+    import argparse
```
AdamGleave, line 92: Style guide violation: https://google.github.io/styleguide/pyguide.html#313-imports-formatting Imports should always be at top level of module to make it clear what modules file depends on, unless there's an explicit need for them to be conditional/only locally imported.

I'd also suggest moving all this to a `main()` function that's called in the `__main__` guard block to avoid polluting the global namespace.

AdamGleave (commented):
One small linting issue that I fixed for you. Please do try and make tests green before requesting review ;)

Some minor suggested changes on `util.py`, after that I think it should be good to merge.

Rocamonde:
Thanks! Just added those changes. Should be good to merge then once tests pass.

AdamGleave:
Tests are failing on Windows now? Seems like an unrelated error so perhaps best fixed in another PR.

Rocamonde:
Everything is green now but codecov is not sending the results over. Coverage should be unchanged anyways (as can be checked on https://app.codecov.io/gh/HumanCompatibleAI/imitation/pull/627) @AdamGleave can you re-trigger codecov sending the info / override?

AdamGleave (approved):
LGTM

PR #629, opened 2022-11-23
Add support to download rollouts from huggingface

taufeeque9:
## Description

Uploaded RL agents trained with imitation to huggingface along with rollout files containing demonstration data. Adding support through this PR to download the demonstration data to train imitation algorithms by specifying `demonstration.rollout_path={algo}-huggingface`.

## Testing

Tested by locally running commands with `demonstration.rollout_path=ppo-huggingface` for all the environments for which the rollout data is uploaded.


src/imitation/scripts/common/demonstrations.py
```diff
@@ -90,6 +91,10 @@ def load_expert_trajs(
     Raises:
         ValueError: There are fewer trajectories than `n_expert_demos`.
     """
+    # wrap with str() in case rollout_path is a Path object
+    if "huggingface" in str(rollout_path):
```
Rocamonde, line 95: Would it not be better to check more explicitly where within the path it is? e.g. you could do `rollout_path.name`.

src/imitation/scripts/common/expert.py
```diff
@@ -43,3 +43,19 @@ def config_hook(config, command_name, logger):
 @expert_ingredient.capture
 def get_expert_policy(venv, policy_type, loader_kwargs, common):
     return serialize.load_policy(policy_type, venv, **loader_kwargs)
+
+
+@expert_ingredient.capture
+def download_expert_rollouts(rollout_path, loader_kwargs):
+    if not rollout_path.endswith("-huggingface"):
+        raise ValueError(
+            "`rollout_path` must follow the convention `{algo}-huggingface`"
```
Rocamonde, line 52: Ironically, it seems like you're following a different convention check on `demonstrations.py`?

src/imitation/scripts/common/expert.py
```diff
@@ -43,3 +43,19 @@ def config_hook(config, command_name, logger):
...
+            "to download rollouts from huggingface of an expert trained using {algo}."
+            "Example: demonstrations.rollout_path=ppo-huggingface",
+        )
+    algo_name = rollout_path.split("-")[0]
```
Rocamonde, line 56: This will fail if the path has multiple dashes. You should use regex or pathlib instead

Rocamonde (commented):
I would add some tests to catch some of the issues I commented on.

src/imitation/policies/serialize.py
```diff
@@ -230,3 +230,14 @@ def _on_step(self) -> bool:
         output_dir = self.policy_dir / f"{self.num_timesteps:012d}"
         save_stable_model(output_dir, self.model)
         return True
+
+
+def _load_rollouts_from_huggingface(
```
AdamGleave, line 235: Why is this prefixed with `_` (private method) if it's being used from other modules?

src/imitation/policies/serialize.py
```diff
@@ -230,3 +230,14 @@ def _on_step(self) -> bool:
...
+    algo_name: str,
+    env_name: str,
+    organization: str = "HumanCompatibleAI",
+) -> str:
+    model_name = hfsb3.ModelName(algo_name, hfsb3.EnvironmentName(env_name))
+    repo_id = hfsb3.ModelRepoId(organization, model_name)
+    filename = hfsb3.load_from_hub(repo_id, "rollouts.pkl")
```
AdamGleave, line 242: I thought we were using `npz` format for rollouts now not pickle? As of https://github.com/HumanCompatibleAI/imitation/pull/448

src/imitation/scripts/common/expert.py
```diff
@@ -43,3 +43,19 @@ def config_hook(config, command_name, logger):
 @expert_ingredient.capture
 def get_expert_policy(venv, policy_type, loader_kwargs, common):
     return serialize.load_policy(policy_type, venv, **loader_kwargs)
+
+
+@expert_ingredient.capture
+def download_expert_rollouts(rollout_path, loader_kwargs):
```
AdamGleave, line 49: Does this get used anywhere other than in `common/demonstrations.py`? If not, why is it here rather than in `demonstrations.py`?

src/imitation/scripts/common/demonstrations.py
AdamGleave: Do we want to overload `rollout_path` in this way? An alternative would be to add an explicit `rollout_type` that can be e.g. `disk` or `huggingface`, similar to what we did for `policy_type`.

AdamGleave (changes requested):

src/imitation/scripts/common/expert.py
taufeeque9: I've defined it in `expert.py` so that we can use the [`config_hook`](https://github.com/HumanCompatibleAI/imitation/blob/2f7e4835a061516a394756a5874fc3b51faec48d/src/imitation/scripts/common/expert.py#L20) in expert that provides the default `loader_kwargs` with `organization_name` and `env_name`.

src/imitation/policies/serialize.py
taufeeque9: When I trained the RL agents, the rollouts were saved using the `npz` format but with the `pkl` extension, which got fixed in #591. Thanks for pointing this out. I'll rename the extensions on the hub and update the file.

src/imitation/scripts/common/expert.py
taufeeque9: Moved it to `demonstrations.py`.

src/imitation/scripts/common/demonstrations.py
```diff
@@ -33,9 +36,10 @@ def fast():
 
 @demonstrations_ingredient.capture
 def get_expert_trajectories(
+    rollout_type: str,
     rollout_path: str,
 ) -> Sequence[types.Trajectory]:
-    if rollout_path is not None:
+    if "huggingface" in rollout_type or rollout_path is not None:
```
AdamGleave, line 42:
```suggestion
    if rollout_type.endswith("-huggingface") or rollout_path is not None:
```

src/imitation/scripts/common/demonstrations.py
```diff
@@ -106,3 +113,19 @@ def load_expert_trajs(
...
+        raise ValueError(
+            "`rollout_type` must follow the convention `{algo}-huggingface`"
+            "to download rollouts from huggingface of an expert trained using {algo}."
+            "Example: rollout_type=ppo-huggingface",
+        )
+    algo_name = rollout_type.split("-")[0]
+    return serialize.load_rollouts_from_huggingface(
+        algo_name,
+        env_name=loader_kwargs["env_name"],
+        organization=loader_kwargs["organization"],
```
AdamGleave, line 130: It's a little counterintuitive that it depends on this. I can't think of anything better, it'd be worse to have code duplication, but maybe we want to document this somewhere?

src/imitation/scripts/common/demonstrations.py
```diff
@@ -91,9 +99,8 @@ def load_expert_trajs(
     Raises:
         ValueError: There are fewer trajectories than `n_expert_demos`.
     """
-    # wrap with str() in case rollout_path is a Path object
-    if "huggingface" in str(rollout_path):
-        rollout_path = expert.download_expert_rollouts(rollout_path=str(rollout_path))
+    if "huggingface" in rollout_type:
```
AdamGleave, line 102:
```suggestion
    if rollout_type.endswith("-huggingface"):
```

src/imitation/scripts/common/demonstrations.py
```diff
@@ -91,9 +99,8 @@ def load_expert_trajs(
...
+        rollout_path = _download_expert_rollouts(rollout_type)
```
AdamGleave, line 104: Should we do input validation? Right now if I specify `rollout_type="huggingfce"` (typo) or `rollout_type="foobar"` it just defaults to disk.

src/imitation/scripts/common/demonstrations.py
AdamGleave: Also consider checking that previous `rollout_path=None`, so users don't set rollout_path to something non-None and expect it to have an effect and be surprirsed?

src/imitation/scripts/common/demonstrations.py
taufeeque9: Added input validation. Changed the "disk" value of `rollout_type` to "local" since we can also generate expert trajectories by setting `rollout_path=None`. Let me know if it looks better or if we should change it.

src/imitation/scripts/common/demonstrations.py
taufeeque9: Have raised a warning for such a case. Let me know if this feels appropriate or if we should do something else like raising an error.

src/imitation/scripts/common/demonstrations.py
taufeeque9: Adding an example in the documentation should be good. Although I suspect users would rarely want to change the default `loader_kwargs` since it requires uploading the rollouts in our format to an organization other than HumanCompitableAI.

src/imitation/scripts/common/demonstrations.py
```diff
@@ -39,8 +40,36 @@ def get_expert_trajectories(
...
     rollout_path: str,
 ) -> Sequence[types.Trajectory]:
-    if "huggingface" in rollout_type or rollout_path is not None:
-        return load_expert_trajs()
+    """Loads expert demonstrations.
+
+    Args:
+        rollout_type: Can be either `local` to load rollouts from the disk or to
+            generate them locally or of the format `{algo}-huggingface` to load
+            from the huggingface hub of expert trained using `{algo}`
```
AdamGleave, line 48:
```suggestion
            from the huggingface hub of expert trained using `{algo}`.
```

AdamGleave (approved):
LGTM, one tiny suggested change.

Sorry for the slow review :sweat_smile: Do feel free to chase me if I'm this slow in the future!

PR #632, opened 2022-11-27
Update citation to point to arXiv paper

AdamGleave:
Update citation to point to arXiv rather than GitHub repo; add citation to docs as well as README.md


Rocamonde (commented):
Looks good. It would be good if you could also add a CITATION file so the GitHub UI also picks it up (https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files). I can do this myself if you're time-constrained.

AdamGleave:
Good idea @Rocamonde -- I've added a CITATION.bib file now.

I tried GitHub's suggested CFF format. It produced a nicer prompt, but an ugly BibTeX and couldn't find a way to make it include identifiers like our arXiv ePrint. So, I've just added the BibTeX for now, as I expect most of our users to be using LaTeX anyway.

Rocamonde (approved):
LGTM

Issue #619, opened 2022-11-08
Device mismatch in PreferenceModel when discounting


rk1a:
## Bug description
When training on preferences with a discounted sum of rewards the discounts are always on CPU, but the rewards from the reward model can be on GPU.

## Steps to reproduce
Run this on a machine with GPU:
`python src/imitation/scripts/train_preference_comparisons.py with preference_model_kwargs.discount_factor=0.99 fast`

## Environment
- Operating system and version: Ubuntu 20.04.5 LTS
- Python version: 3.8.13

AdamGleave:
Not sure I quite understand the problem here. Is this a performance issue, or a correctness issue? Do you see an error message?The "discounts are always on CPU" -- the discount is a scalar float, so it seems appropriate it's on CPU, I assume you're referring to the discounted returns.

rk1a:
It's a bug. Here is the error I get:
```
  File "src/imitation/scripts/train_preference_comparisons.py", line 257, in train_preference_comparisons
    results = main_trainer.train(
  File "/home/imitation/src/imitation/algorithms/preference_comparisons.py", line 1722, in train
    self.reward_trainer.train(self.dataset, epoch_multiplier=epoch_multiplier)
  File "/home/imitation/src/imitation/algorithms/preference_comparisons.py", line 1130, in train
    self._train(dataset, epoch_multiplier)
  File "/home/imitation/src/imitation/algorithms/preference_comparisons.py", line 1256, in _train
    loss = self._training_inner_loop(
  File "/home/imitation/src/imitation/algorithms/preference_comparisons.py", line 1316, in _training_inner_loop
    output = self.loss.forward(fragment_pairs, preferences, self._preference_model)
  File "/home/imitation/src/imitation/algorithms/preference_comparisons.py", line 1068, in forward
    probs, gt_probs = preference_model(fragment_pairs)
  File "/home/miniconda3/envs/imitation/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/home/imitation/src/imitation/algorithms/preference_comparisons.py", line 447, in forward
    probs[i] = self.probability(rews1, rews2)
  File "/home/imitation/src/imitation/algorithms/preference_comparisons.py", line 515, in probability
    returns_diff = (discounts * (rews2 - rews1)).sum(axis=0)
[...]

PR #636, opened 2022-11-28
#619 fix discounting device mismatch

mifeet:
Fix for https://github.com/HumanCompatibleAI/imitation/issues/619

### DIFF
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -510,6 +510,7 @@ class PreferenceModel(nn.Module):
             returns_diff = (rews2 - rews1).sum(axis=0)  # type: ignore[call-overload]
         else:
             device = rews1.device
+            assert device == rews2.device
             discounts = self.discount_factor ** th.arange(len(rews1), device=device)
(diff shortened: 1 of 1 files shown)


src/imitation/algorithms/preference_comparisons.py
```diff
@@ -509,7 +509,8 @@ def probability(self, rews1: th.Tensor, rews2: th.Tensor) -> th.Tensor:
         if self.discount_factor == 1:
             returns_diff = (rews2 - rews1).sum(axis=0)  # type: ignore[call-overload]
         else:
-            discounts = self.discount_factor ** th.arange(len(rews1))
+            device = rews1.device
```
AdamGleave, line 512: We could have problems if `rews1` and `rews2` are on different devices, I think? That shouldn't happen, but maybe worth adding an explicit assert to check they're the same?

AdamGleave (approved):
LGTM (1 small suggestion)

src/imitation/algorithms/preference_comparisons.py
mifeet: Fixed

PR #563, opened 2022-09-13
CNN scripts

dfilan:
## Description

Add config options to train CNNs on image environments. This also involves letting configs set environment wrappers, so that environments like Atari can be appropriately wrapped.

## Testing

Ran some scripts of interest locally, ran `pytest tests/scripts/test_scripts.py`.

## TODOs:

- Add nice configs for all algorithms, not just preference comparisons.
- Add tests for CNN+image environment scripts.

PR #639, opened 2022-11-30
Use procgen

dfilan:
## Description

Import `procgen` and add configs to `train_rl` and `eval_policy` for procgen environments.

This is basically a very pared-down version of #563 that just enables things that are immediately useful to me.

## Testing

Added a test that `train_rl` works on coinrun, the test passes.


dfilan:
Procgen has some issues with reward learning, so it probably doesn't make sense to merge this into imitation. (Also it's kind of interesting that tests seem to be 'failing' due to timing out as a result of adding new tests?)

PR #648, opened 2022-12-28
Fix CI errors

taufeeque9:
## Description

Fixes the `np.object` error by replacing it with `object` and installs OpenSSL on Windows.


AdamGleave (approved):
LGTM

PR #649, opened 2023-01-03
Fix non-determinism in DAgger (fixes #643)

### LINKED ISSUE #643: best way to reproduce results (w Dagger) ?
Each run with Dagger gives different results, and I am wondering what is the best/recommended way to freeze the random state when using Dagger? I freeze all possible seeds (torch, np, etc) but this module keeps giving different results on each run, most likely due to the `InteractiveTrajectoryCollector`. I even tried explicitly setting the seed in the `init` of this class, but the non-determinism persists. Any ideas would be appreciated. Thanks!

hacobe:
## Description

The issue (#643) was that DAgger demonstrations were loaded from disk for training in a different order each run. This was because the filenames for the saved demonstrations changed each run and that changed the order in which os.listdir returned the filenames. The filenames changed each run, because they included a timestamp and the first 6 characters of a UUID generated without fixing a random seed.

This PR fixes the non-determinism by making the filenames the same each run as long as the same random seed is used. It does so by removing the timestamp from the filename and fixing the seed of the UUID. Because the timestamp is removed, the PR introduces a trajectory index in the filename, so that a user can tell the order in which trajectories were created. It also includes the entire UUID instead of just the first 6 characters. Finally, it sorts the filenames returned by os.listdir. listdir returns filenames in an arbitrary order that depends on the file system implementation (https://stackoverflow.com/questions/31534583/is-os-listdir-deterministic). We sort the filenames to ensure the order is consistent across file systems.

We discuss a few other design decisions below.

Why include a UUID in the filename at all? If we removed the UUID from the filename, then the DAgger trainers would not overwrite filenames, because they take care to write to a new directory each round of training. However, if the InteractiveTrajectoryCollector is used independently of those trainers, then it can end up overwriting filenames without the UUID.

Do we need to shuffle the filenames returned by os.listdir after sorting? We could, but the demonstrations loaded from the files are passed to a DataLoader, which shuffles them. That seems like the right place to handle the shuffling rather than making it the responsibility of the utility function that returns the filenames.

## Testing

This PR also adds unit tests for the reproducibility of the InteractiveTrajectoryCollector, DAggerTrainer, and the SimpleDAggerTrainer. In particular, each unit test consists of running a block of code twice with the same random seeds and checking that the code produces the same result each time.

### DIFF
--- a/src/imitation/algorithms/dagger.py
+++ b/src/imitation/algorithms/dagger.py
@@ -382,9 +382,8 @@ class DAggerTrainer(base.BaseImitationAlgorithm):
         # https://stackoverflow.com/questions/31534583/is-os-listdir-deterministic
         # To ensure the order is consistent across file systems,
         # we sort by the filename.
-        return [
-            round_dir / p for p in sorted(os.listdir(round_dir)) if p.endswith(".npz")
-        ]
+        filenames = sorted(os.listdir(round_dir))
+        return [round_dir / f for f in filenames if f.endswith(".npz")]
 
     def _demo_dir_path_for_round(self, round_num: Optional[int] = None) -> pathlib.Path:
         if round_num is None:
--- a/src/imitation/algorithms/dagger.py
+++ b/src/imitation/algorithms/dagger.py
@@ -111,6 +111,9 @@ def _save_dagger_demo(
     random_uuid = uuid.UUID(int=randbits, version=4).hex
     filename = f"{actual_prefix}dagger-demo-{trajectory_index}-{random_uuid}.npz"
     npz_path = save_dir / filename
+    assert (
+        not npz_path.exists()
+    ), "The following DAgger demonstration path already exists: {0}".format(npz_path)
     types.save(npz_path, [trajectory])
     logging.info(f"Saved demo at '{npz_path}'")
 
--- a/src/imitation/algorithms/dagger.py
+++ b/src/imitation/algorithms/dagger.py
@@ -10,6 +10,7 @@ import abc
 import logging
 import os
 import pathlib
+import uuid
 from typing import Any, Callable, List, Mapping, Optional, Sequence, Tuple, Union
 
 import numpy as np
@@ -21,7 +22,6 @@ from torch.utils import data as th_data
 from imitation.algorithms import base, bc
 from imitation.data import rollout, types
 from imitation.util import logger as imit_logger
-from imitation.util import util
 
 
 class BetaSchedule(abc.ABC):
@@ -99,15 +99,17 @@ def reconstruct_trainer(
 
 def _save_dagger_demo(
     trajectory: types.Trajectory,
+    trajectory_index: int,
     save_dir: types.AnyPath,
+    rng: np.random.Generator,
     prefix: str = "",
 ) -> None:
     save_dir = types.parse_path(save_dir)
     assert isinstance(trajectory, types.Trajectory)
     actual_prefix = f"{prefix}-" if prefix else ""
-    timestamp = util.make_unique_timestamp()
-    filename = f"{actual_prefix}dagger-demo-{timestamp}.npz"
-
+    randbits = int.from_bytes(rng.bytes(16), "big")
+    random_uuid = uuid.UUID(int=randbits, version=4).hex
+    filename = f"{actual_prefix}dagger-demo-{trajectory_index}-{random_uuid}.npz"
     npz_path = save_dir / filename
     types.save(npz_path, [trajectory])
     logging.info(f"Saved demo at '{npz_path}'")
@@ -246,8 +248,8 @@ class InteractiveTrajectoryCollector(vec_env.VecEnvWrapper):
             infos=infos,
             dones=dones,
         )
-        for traj in fresh_demos:
-            _save_dagger_demo(traj, self.save_dir)
+        for traj_index, traj in enumerate(fresh_demos):
+            _save_dagger_demo(traj, traj_index, self.save_dir, self.rng)
 
         return next_obs, rews, dones, infos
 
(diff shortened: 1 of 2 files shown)


src/imitation/algorithms/dagger.py
```diff
@@ -372,7 +374,14 @@ def _load_all_demos(self) -> Tuple[types.Transitions, List[int]]:
         return demo_transitions, num_demos_by_round
 
     def _get_demo_paths(self, round_dir: pathlib.Path) -> List[pathlib.Path]:
-        return [round_dir / p for p in os.listdir(round_dir) if p.endswith(".npz")]
+        # listdir returns filenames in an arbitrary order that depends on the
+        # file system implementation:
+        # https://stackoverflow.com/questions/31534583/is-os-listdir-deterministic
+        # To ensure the order is consistent across file systems,
+        # we sort by the filename.
+        return [
```
AdamGleave, line 382: This list comprehension is getting a bit complicated (and now needs to be multiple lines). Consider splitting it up by using a temporary variable?

src/imitation/algorithms/dagger.py
```diff
@@ -99,15 +99,17 @@ def reconstruct_trainer(
...
     save_dir = types.parse_path(save_dir)
     assert isinstance(trajectory, types.Trajectory)
     actual_prefix = f"{prefix}-" if prefix else ""
-    timestamp = util.make_unique_timestamp()
-    filename = f"{actual_prefix}dagger-demo-{timestamp}.npz"
-
+    randbits = int.from_bytes(rng.bytes(16), "big")
+    random_uuid = uuid.UUID(int=randbits, version=4).hex
+    filename = f"{actual_prefix}dagger-demo-{trajectory_index}-{random_uuid}.npz"
     npz_path = save_dir / filename
```
AdamGleave, line 113: If the intention is to never overwrite files it might be worth adding something like `assert not npz_path.exists()`. I can imagine some edge cases where `rng` gets reseeded where overwriting ends up happening.

src/imitation/algorithms/dagger.py
```diff
@@ -99,15 +99,17 @@ def reconstruct_trainer(
...
     prefix: str = "",
 ) -> None:
     save_dir = types.parse_path(save_dir)
     assert isinstance(trajectory, types.Trajectory)
     actual_prefix = f"{prefix}-" if prefix else ""
-    timestamp = util.make_unique_timestamp()
-    filename = f"{actual_prefix}dagger-demo-{timestamp}.npz"
-
+    randbits = int.from_bytes(rng.bytes(16), "big")
+    random_uuid = uuid.UUID(int=randbits, version=4).hex
```
AdamGleave, line 111: I think we could not seed the UUID and we'd still have replicable results, as the trajectory index should take precedence over UUID in the sorting? Doesn't hurt to seed it though.

tests/algorithms/test_dagger.py
```diff
@@ -325,6 +366,69 @@ def test_trainer_makes_progress(init_trainer_fn, pendulum_venv, pendulum_expert_
...
+                run_dir,
+                pendulum_venv,
+                beta_schedule,
+                pendulum_expert_policy,
+                maybe_pendulum_expert_trajectories,
+                custom_logger,
+                rng,
+            )
+
+            # Train for 10 iterations. (6 or less causes test to fail on some configs.)
```
AdamGleave, line 406: Is this comment / the 10 iteration minimum still relevant? I think the test you copied it from was checking reward improved by some significant threshold, which needs multiple iterations. But here you're just trying to check the results line up, so can probably get away with fewer iterations.

tests/algorithms/test_dagger.py
```diff
@@ -325,6 +366,69 @@ def test_trainer_makes_progress(init_trainer_fn, pendulum_venv, pendulum_expert_
...
+                dones = [False] * pendulum_venv.num_envs
+                while not np.any(dones):
+                    expert_actions, _ = pendulum_expert_policy.predict(
+                        obs,
+                        deterministic=True,
+                    )
+                    obs, _, dones, _ = collector.step(expert_actions)
+                trainer.extend_and_update(dict(n_epochs=1))
+
+            rewards, _ = evaluation.evaluate_policy(
```
AdamGleave, line 420: Why check that rewards are equal rather than checking the entire rollouts for equality? You might find the `rollout` method in `imitation.data.rollout` useful.

tests/algorithms/test_dagger.py
```diff
@@ -105,6 +105,47 @@ def get_random_acts(obs):
...
+            for i in range(1000):
+                _, _, dones, _ = collector.step(zero_acts)
+
+            # Get the observations from all the collected trajectories.
+            file_paths = glob.glob(os.path.join(save_dir, "dagger-demo-*.npz"))
+            trajs = [types.load(p)[0] for p in file_paths]
+            trajs_obs = np.stack([trajs[i].obs for i in range(len(trajs))], axis=0)
+            trajs_obs_list.append(trajs_obs)
+
+    np.testing.assert_almost_equal(trajs_obs_list[0], trajs_obs_list[1])
```
AdamGleave, line 146: Does this fail if you do exact equality rather than almost equality?

If we're seeding everything the same I think we should be getting same results even up to floating point error. I don't particularly care if we don't, but provided exact equality holds I think we can compare the `Trajectory` objects directly rather than using the observations, which would be a stronger check (would make sure actions, rewards, etc are the same not just observations).

AdamGleave (commented):
Thanks for the PR! The fix looks good to me, I've left some minor suggestions on the tests.

Minor point but in GitHub if you write "fixes #N" it'll close issue N when the PR gets merged which is convenient, so I edited your title to include that rather than "reported by #N"

AdamGleave:
There was a lint error originally but bizarrely it seems to have fixed itself on rerun. Never seen a flaky lint error before. My guess is some upstream dependency broke and a hotfix got released in the meantime?

src/imitation/algorithms/dagger.py
hacobe: That's correct.

AdamGleave (approved):
LGTM

PR #650, opened 2023-01-04
Update benchmarking configs

taufeeque9:
## Description

Added `demonstrations.rollout_type` in all the configs to download rollouts from the huggingface hub. Also, updated config key names changed in #604.

## Testing

Tested the configs locally using: `python -m imitation.scripts.<train_script> <algo> with benchmarking/<config_name>.json
`


AdamGleave (approved):
LGTM too. Thanks for reviewing ernestum!

PR #652, opened 2023-01-06
Fix benchmarking configs + Add a minimal test suite for those configs.

hacobe:
## Description

This PR patches in the changes from PR #650. It also fixes an error in the dagger configs by replacing "py/object" with "py/type". "py/object" failed the assert in sacred.config.utils.assert_is_valid_key.

## Testing

The PR adds a minimal test suite for the benchmarking configs. It makes progress on testing, but still does not test that all the benchmarking configs work. Running all the configs with the fast configs applied is too slow. Calling sacred's print_config command for all the configs is also too slow.

### DIFF
--- a/benchmarking/example_airl_seals_ant_best_hp_eval.json
+++ b/benchmarking/example_airl_seals_ant_best_hp_eval.json
@@ -6,9 +6,9 @@
   },
   "checkpoint_interval": 0,
   "demonstrations": {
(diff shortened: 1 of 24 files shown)


hacobe:
I accidentally included the [4e9c85a](https://github.com/HumanCompatibleAI/imitation/pull/652/commits/4e9c85abcd079dbed7c92e9ab745cee1ca077325) commit in this PR.

PR #653, opened 2023-01-06
Fix benchmarking configs + Add a minimal test suite for those configs.

hacobe:
## Description

This PR patches in the changes from PR #650. It also fixes an error in the dagger configs by replacing "py/object" with "py/type". "py/object" failed the assert in sacred.config.utils.assert_is_valid_key. 

## Testing

The PR adds a minimal test suite for the benchmarking configs. Running all the configs with the fast configs applied is too slow and requires MuJoCo. Instead, we test that the print_config command works for all the configs.

### DIFF
--- a/tests/test_benchmarking.py
+++ b/tests/test_benchmarking.py
@@ -1,8 +1,10 @@
 """Tests for config files in benchmarking/ folder."""
+import os
+import subprocess
+
 import pytest
 
 from imitation.data import types
-from imitation.scripts import train_adversarial, train_imitation
 
 ALGO_FAST_CONFIGS = {
     "adversarial": [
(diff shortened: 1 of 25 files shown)


tests/test_benchmarking.py
```diff
@@ -0,0 +1,44 @@
...
+import os
+import subprocess
+
+import pytest
+
+from imitation.data import types
+
+BENCHMARKING_DIR = types.parse_path("benchmarking")
+
+if not BENCHMARKING_DIR.exists():  # pragma: no cover
```
AdamGleave, line 11: I think we can get rid of this if we make the path relative to the script like https://github.com/HumanCompatibleAI/imitation/blob/master/tests/test_examples.py#L30

The other way I can think of is adding it as `package_data` in `setup.py` then using `pkg_resources` library -- but we probably don't actually want to bake these JSON configs into the wheel.

tests/test_benchmarking.py
```diff
@@ -0,0 +1,44 @@
...
+    # We test that the configs using the print_config command
+    # only for the half_cheetah environment,
+    # because the print_config command is slow
+    # (takes about 5 seconds per execution).
+    # We do not test that the configs run even with the fast configs applied,
+    # because running the configs requires MuJoCo. Requiring MuJoCo to run
+    # the tests adds too much complexity.
+    config_file = f"example_{command_name}_seals_half_cheetah_best_hp_eval.json"
+    config_path = os.path.join(BENCHMARKING_DIR.stem, config_file)
+    completed_process = subprocess.run(
```
AdamGleave, line 37: Do we need a subprocess for this? I think you can do something like:

```
from imitation.scripts import train_imitation
train_imitation.train_imitation_ex.run(command_name='print_config')
```

This should make things faster (5 seconds I expect is mostly from spinning up a new Python interpreter) so you could maybe check all the config files.

Not sure if you'll need to load the JSON yourself and feed it in as `config_updates` or if there's an argument to `run` for that.

You might find `tests/scripts/test_scripts.py` generally instructive on how we've been testing Sacred experiments.

tests/test_benchmarking.py
```diff
@@ -0,0 +1,44 @@
...
+    "script_name, command_name",
+    [
+        ("train_imitation", "bc"),
+        ("train_imitation", "dagger"),
+        ("train_adversarial", "airl"),
+        ("train_adversarial", "gail"),
+    ],
+)
+def test_benchmarking_configs(tmpdir, script_name, command_name):
+    # We test that the configs using the print_config command
```
AdamGleave, line 28:
```suggestion
    # We test the configs using the print_config command
```

AdamGleave (changes requested):
Thanks for adding this test! Some minor suggestions.

tests/test_benchmarking.py
hacobe: Great! I can test all the configs.

tests/test_benchmarking.py
```diff
@@ -1,44 +1,33 @@
...
-    assert completed_process.returncode == 0
+def test_benchmarking_configs(tmpdir, command_name):
+    # We test the configs using the print_config command,
+    # because running the configs requires MuJoCo.
+    # Requiring MuJoCo to run the tests adds too much complexity.
+    if command_name in ("bc", "dagger"):
+        ex = train_imitation.train_imitation_ex
+    elif command_name in ("airl", "gail"):
+        ex = train_adversarial.train_adversarial_ex
+    cfg_pattern = os.path.join(BENCHMARKING_DIR.stem, f"example_{command_name}_*.json")
```
AdamGleave, line 26: Why are you using `.stem`? I think just `BENCHMARKING_DIR` should work here, whereas `BENCHMARKING_DIR.stem` will only work if we're at the root of the repository because `BENCHMARKING_DIR.stem == 'benchmark'`

tests/test_benchmarking.py
```diff
@@ -1,44 +1,33 @@
...
 
-BENCHMARKING_DIR = types.parse_path("benchmarking")
-
-if not BENCHMARKING_DIR.exists():  # pragma: no cover
-    raise RuntimeError(
-        "The benchmarking/ folder has not been found. Make sure you are "
-        "running tests relative to the base imitation project folder.",
-    )
+THIS_DIR = pathlib.Path(__file__).absolute().parent
+BENCHMARKING_DIR = THIS_DIR / ".." / "benchmarking"
```
AdamGleave, line 11:
```suggestion
BENCHMARKING_DIR = THIS_DIR.parent() / "benchmarking"
```

AdamGleave (approved):
LGTM, 2 small suggestions

tests/test_benchmarking.py
hacobe: Good catch. That was a holdover from the previous commit.

tests/test_benchmarking.py
hacobe: parent not parent() but yeah that's nicer.

PR #654, opened 2023-01-06
Add a program to generate commands to run training scripts

hacobe:
## Description

See the header file of experiments/commands.py for details. The primary motivation is to make it easy to run the training scripts using the config files in benchmarking/. It attempts to simplify (1) running local commands in parallel compared to the bash scripts in experiments/ and (2) running commands in separate containers on the Hofvarpnir cluster compared to parallel.py, which relies on Ray. AdamGleave@'s thoughts on using Ray on Hofvarpnir: "We could make Ray work I think, but it'd be janky. It'd be fairly easy to e.g. have a Docker container that just starts a Ray server, then the command runs on that Ray server, and the whole container ends when the job is complete. But then you only get parallelism up to the max machine size, no distributed training. The other option is to just launch a bunch of Ray server jobs up to the expected size, then run other jobs that connect to those Ray servers. But then you'd have to teardown the Ray servers manually to avoid bleeding resources. Overall I think it probably does make sense to find a non-Ray solution for your part of the project, though I still feel moderately excited about getting a Ray solution working longer-term. It's probably not that hard to make the Ray autoscaler handle this."

## Testing

The program is mostly tested by checking that it prints out the commands as expected. There is one test that actually runs the command that is printed out to make sure it works. Also, I've launched all the algorithms with 3 random seeds in the half cheetah environment on the Hofvarpnir cluster using the commands printed out by this program. The bc commands have completed and the others are still running. I checked that the analyze_imitation command worked on the outputs from the bc runs.

### DIFF
--- a/experiments/commands.py
+++ b/experiments/commands.py
@@ -1,0 +1,234 @@
+r"""Generate commands to run training scripts with different configs.
+
+This program takes as input a set of config files and outputs a command
+for each config file. Each command runs a training script using the
+associated config file. By default, each command uses the same random seed
+for training. However, if multiple random seeds are specified, the program
+generates multiple commands for each config file (one for each random seed).
+The commands can be executed by copying and pasting them into a command line
+or by piping them to another command line utility. This program is helpful
+for running a large number of training scripts with different config files
+using multiple random seeds.
+
+For example, we can run:
+
+python commands.py \
+    --name=run0 \
+    --cfg_pattern=../benchmarking/*ai*_seals_walker_*.json \
+    --output_dir=output
+
+And get the following commands printed out:
+
+python -m imitation.scripts.train_adversarial airl \
+    --capture=sys --name=run0 \
+    --file_storage=output/sacred/$USER-cmd-run0-airl-0-a3531726 \
+    with ../benchmarking/example_airl_seals_walker_best_hp_eval.json \
+    seed=0 logging.log_root=output
+
+python -m imitation.scripts.train_adversarial gail \
+    --capture=sys --name=run0 \
+    --file_storage=output/sacred/$USER-cmd-run0-gail-0-a1ec171b \
+    with ../benchmarking/example_gail_seals_walker_best_hp_eval.json \
+    seed=0 logging.log_root=output
+
+We can execute commands in parallel by piping them to GNU parallel:
+
+python commands.py ... | parallel -j 8
+
+If the --remote flag is enabled, then the program prints out commands
+to run training scripts in containers on the Hofvarpnir cluster.
+
+For example, we can run:
+
+python commands.py \
+    --name=run0 \
+    --cfg_pattern=../benchmarking/example_bc_seals_half_cheetah_best_hp_eval.json \
+    --output_dir=/data/output \
+    --remote
+
+And get the following command printed out:
+
+ctl job run --name $USER-cmd-run0-bc-0-72cb1df3 \
+    --command python\\ -m\\ imitation.scripts.train_imitation\\ bc\\ \
+    --capture=sys\\ --name=run0\\ \
+    --file_storage=/data/output/sacred/$USER-cmd-run0-bc-0-72cb1df3\\ \
+    with\\ \
+    /data/imitation/benchmarking/example_bc_seals_half_cheetah_best_hp_eval.json\\ \
+    seed=0\\ logging.log_root=/data/output --container hacobe/devbox:imitation \
+    --login --high-priority --force-pull --never-restart
+"""
+import argparse
+import glob
+import os
+import zlib
+
+_ALGO_NAME_TO_SCRIPT_NAME = {
+    "bc": "train_imitation",
+    "dagger": "train_imitation",
+    "airl": "train_adversarial",
+    "gail": "train_adversarial",
+}
+
+_CMD_ID_TEMPLATE = "$USER-cmd-{name}-{algo_name}-{seed}-{cfg_id}"
+
+_TRAIN_CMD_TEMPLATE = (
+    "python -m imitation.scripts.{script_name} {algo_name} "
+    "--capture=sys --name={name} --file_storage={file_storage} "
+    "with {cfg_path} seed={seed} logging.log_root={log_root}"
+)
+
+_HOFVARPNIR_CLUSTER_CMD_TEMPLATE = (
+    "ctl job run --name {name} --command {command} --container {container} "
+    "--login --high-priority --force-pull --never-restart"
+)
+
+
+def _get_algo_name(cfg_file: str) -> str:
+    algo_names = set()
+    for key in _ALGO_NAME_TO_SCRIPT_NAME:
+        if cfg_file.find("_" + key + "_") != -1:
+            algo_names.add(key)
+
+    if len(algo_names) == 0:
+        raise ValueError("Unable to find algo_name in cfg_file: " + cfg_file)
+
+    if len(algo_names) >= 2:
+        raise ValueError("algo_name is ambiguous in cfg_file: " + cfg_file)
+
+    algo_name = algo_names.pop()
+    return algo_name
+
+
+def _get_cfg_id(cfg_path: str) -> str:
+    checksum = zlib.adler32(cfg_path.encode())
+    checksum_hex = hex(checksum)
+    assert checksum_hex.startswith("0x")
+    return checksum_hex[2:]
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="Generate commands to run training scripts with different configs.",
+    )
+    parser.add_argument(
+        "--name",
+        type=str,
+        required=True,
+        help="A name that identifies multiple commands as "
+        "coming from the same 'run'. In particular, this flag is "
+        "passed to imitation training scripts directly in the "
+        "--name flag and as part of the path in the "
+        "--file_storage flag. If the --remote flag is enabled, "
+        "this flag is also used in the cluster job name.",
+    )
+    parser.add_argument(
+        "--cfg_pattern",
+        type=str,
+        default="example_bc_seals_half_cheetah_best_hp_eval.json",
+        help="Generate a command for every file that matches this glob pattern. "
+        "Each matching file should be a config file that has its algorithm name "
+        "(bc, dagger, airl or gail) bookended by underscores in the filename. "
+        "If the --remote flag is enabled, then generate a command "
+        "for every file in the --remote_cfg_dir directory "
+        "that has the same filename as a file that matches this glob pattern. "
+        "E.g., suppose the current, local working directory is 'foo' and "
+        "the subdirectory 'foo/bar' contains the config files "
+        "'example_bc_best.json' and 'example_dagger_best.json'."
+        "If the pattern 'bar/*.json' is supplied, then globbing will return "
+        "['bar/example_bc_best.json', 'bar/example_dagger_best.json']. "
+        "If the --remote flag is enabled, 'bar' will be replaced "
+        "with `remote_cfg_dir` and commands will be created for "
+        "the following configs: [`remote_cfg_dir`/example_bc_best.json, "
+        "`remote_cfg_dir`/example_dagger_best.json]"
+        "Why not just supply the pattern '`remote_cfg_dir`/*.json' directly? "
+        "Because the `remote_cfg_dir` directory may not exist on the local machine.",
+    )
+    parser.add_argument(
+        "--seeds",
+        type=str,
+        default="0",
+        help="Comma-delimited list of random seeds to use for each config.",
+    )
+    parser.add_argument(
+        "--output_dir",
+        type=str,
+        required=True,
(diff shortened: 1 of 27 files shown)


experiments/commands.py
```diff
@@ -0,0 +1,234 @@
...
+The commands can be executed by copying and pasting them into a command line
+or by piping them to another command line utility. This program is helpful
+for running a large number of training scripts with different config files
+using multiple random seeds.
+
+For example, we can run:
+
+python commands.py \
+    --name=run0 \
+    --cfg_pattern=../benchmarking/*ai*_seals_walker_*.json \
```
AdamGleave, line 17: To match `gail` and `airl`? Cute, but if `(gail|airl)` works this would be more explicit intent.

experiments/commands.py
```diff
@@ -0,0 +1,234 @@
...
+And get the following command printed out:
+
+ctl job run --name $USER-cmd-run0-bc-0-72cb1df3 \
+    --command python\\ -m\\ imitation.scripts.train_imitation\\ bc\\ \
+    --capture=sys\\ --name=run0\\ \
+    --file_storage=/data/output/sacred/$USER-cmd-run0-bc-0-72cb1df3\\ \
+    with\\ \
+    /data/imitation/benchmarking/example_bc_seals_half_cheetah_best_hp_eval.json\\ \
+    seed=0\\ logging.log_root=/data/output --container hacobe/devbox:imitation \
+    --login --high-priority --force-pull --never-restart
```
AdamGleave, line 58: `--force-pull` is a bit evil, better to just build Docker images with a unique tag (e.g. derived from Git hash), we have a workflow you might be able to adapt from https://github.com/HumanCompatibleAI/go_attack/blob/main/kubernetes/update_images.py

AdamGleave: Where is `/data` coming from? Is this just mounted by default? I was expecting to need some `-shared-host-dir` or the like.

I'd also not run with `--high-priority` by default, this could monopolize a significant % of cluster for full benchmark.

experiments/commands.py
```diff
@@ -0,0 +1,234 @@
...
+
+python commands.py \
+    --name=run0 \
+    --cfg_pattern=../benchmarking/example_bc_seals_half_cheetah_best_hp_eval.json \
+    --output_dir=/data/output \
+    --remote
+
+And get the following command printed out:
+
+ctl job run --name $USER-cmd-run0-bc-0-72cb1df3 \
```
AdamGleave, line 51: This defaults to requesting a GPU. Do you want that? Can disable by `--gpu 0` I believe

experiments/commands.py
```diff
@@ -0,0 +1,234 @@
...
+    "with {cfg_path} seed={seed} logging.log_root={log_root}"
+)
+
+_HOFVARPNIR_CLUSTER_CMD_TEMPLATE = (
+    "ctl job run --name {name} --command {command} --container {container} "
+    "--login --high-priority --force-pull --never-restart"
+)
+
+
+def _get_algo_name(cfg_file: str) -> str:
```
AdamGleave, line 86: Can we add a docstring for this (one sentence is enough)?

experiments/commands.py
```diff
@@ -0,0 +1,234 @@
...
+        raise ValueError("Unable to find algo_name in cfg_file: " + cfg_file)
+
+    if len(algo_names) >= 2:
+        raise ValueError("algo_name is ambiguous in cfg_file: " + cfg_file)
+
+    algo_name = algo_names.pop()
+    return algo_name
+
+
+def _get_cfg_id(cfg_path: str) -> str:
```
AdamGleave, line 102: Docstring?

experiments/commands.py
```diff
@@ -0,0 +1,234 @@
...
+    checksum = zlib.adler32(cfg_path.encode())
+    checksum_hex = hex(checksum)
+    assert checksum_hex.startswith("0x")
+    return checksum_hex[2:]
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
```
AdamGleave, line 110: Nit: can we make this call a main method? Anything run at module-level pollutes the global namespace. Doesn't matter too much, but does slightly raise risk of other methods accidentally using a non-local value.

AdamGleave: By the looks of it could probably also split into a parsing method, and a method that does the bulk of the logic.

experiments/commands.py
```diff
@@ -0,0 +1,234 @@
...
+        description="Generate commands to run training scripts with different configs.",
+    )
+    parser.add_argument(
+        "--name",
+        type=str,
+        required=True,
+        help="A name that identifies multiple commands as "
```
AdamGleave, line 117: This might be easier to read if you used `"""`?

experiments/commands.py
```diff
@@ -0,0 +1,234 @@
...
+        "If the pattern 'bar/*.json' is supplied, then globbing will return "
+        "['bar/example_bc_best.json', 'bar/example_dagger_best.json']. "
+        "If the --remote flag is enabled, 'bar' will be replaced "
+        "with `remote_cfg_dir` and commands will be created for "
+        "the following configs: [`remote_cfg_dir`/example_bc_best.json, "
+        "`remote_cfg_dir`/example_dagger_best.json]"
+        "Why not just supply the pattern '`remote_cfg_dir`/*.json' directly? "
+        "Because the `remote_cfg_dir` directory may not exist on the local machine.",
+    )
+    parser.add_argument(
```
AdamGleave, line 146: `type=int` with `nargs=+` might be more natural way of expressing it?

experiments/commands.py
```diff
@@ -0,0 +1,234 @@
...
+                file_storage=file_storage,
+                log_root=args.output_dir,
+            )
+
+            if local:
+                print(train_cmd)
+                continue
+
+            # Escape spaces.
+            command = train_cmd.replace(" ", "\\ ")
```
AdamGleave, line 226: This makes commands output quite hard to read, and might still fail in presence of some other special characters. How about changing format to have '' around parameters, so `--name '{name}'` etc?

AdamGleave (changes requested):

tests/test_experiments.py
```diff
@@ -14,6 +18,257 @@
     "transfer_learn_benchmark.sh",
 )
 
+BENCHMARKING_DIR = types.parse_path("benchmarking")
+
+if not BENCHMARKING_DIR.exists():  # pragma: no cover
```
AdamGleave, line 23: Similar to comment in #653:
>I think we can get rid of this if we make the path relative to the script like https://github.com/HumanCompatibleAI/imitation/blob/master/tests/test_examples.py#L30
> 
> The other way I can think of is adding it as package_data in setup.py then using pkg_resources library -- but we probably don't actually want to bake these JSON configs into the wheel.

tests/test_experiments.py
```diff
@@ -14,6 +18,257 @@
...
+    commands = formatted_stdout.split("\n")
+
+    return commands
+
+
+def test_commands_local_config():
+    commands = _run_commands_from_flags()
+    assert len(commands) == 1
+    expected = (
+        "python -m imitation.scripts.train_imitation dagger "
```
AdamGleave, line 102: `"""` might be easier here as well?

tests/test_experiments.py
```diff
@@ -14,6 +18,257 @@
...
+        "seed=0 logging.log_root=output"
+    )
+    assert commands[0] == expected
+
+
+def test_commands_local_config_runs(tmpdir):
+    output_dir = types.parse_path(tmpdir).as_posix()
+    commands = _run_commands_from_flags(output_dir=output_dir)
+    assert len(commands) == 1
+    expected = (
```
AdamGleave, line 116: same as above but with `output_dir`? Makes sense to split up the tests to more precisely localize failures, but maybe just have one `EXPECTED_LOCAL_CONFIG` or the like template that you can reuse between these two tests to cut down on duplication.

AdamGleave (commented):
Tests look comprehensive, thanks for adding those!

experiments/commands.py
hacobe: That doesn't work. Globbing has limited expressiveness. I could use regular expressions instead, but that seems like overkill since you can usually create a directory with the config files that you want. Thoughts?

experiments/commands.py
hacobe: I tried out one command escaping spaces and one command enclosed in quotes.

(1) `ctl job run --command echo\ aaa --container hacobe/devbox:imitation --login --high-priority --force-pull --never-restart --name echo-test`

(2) `ctl job run --command "echo 'aaa'" --container hacobe/devbox:imitation --login --high-priority --force-pull --never-restart --name echo-test2`

(1) completes quickly and (2) gets stuck in Pending.

Can we leave the escaping spaces as is?

experiments/commands.py
hacobe: 1) Would it be OK to skip addressing --force-pul in this PR? Relatedly, I created an issue to push a new image to humancompatibleai with MuJoCo (#658).

2) /data is the default value for the --shared-host-dir-mount flag. I added a --shared-host-dir-mount set to /data in the command template.

3) I removed the --high-priority flag

experiments/commands.py
hacobe: I agree that the escaped spaces make the command less readable. I pushed a [commit](https://github.com/HumanCompatibleAI/imitation/pull/654/commits/b0e133ccbf85c309a26577453c50d37250d032f2) that escapes double quotes in the training command and encloses it within the cluster command in double quotes.

I don't want to enclose other flag values in single quotes, because then an environment variable like $USER in the flag value won't get evaluated. I also don't want to enclose other flag values in double quotes, because that again clutters the command. In the commit message, I show an example of how you would run commands.py if you wanted to use a special character like a space in the output directory.

Let me know if this approach works for you.

hacobe:
I pushed a [commit](https://github.com/HumanCompatibleAI/imitation/pull/654/commits/66b661a4ace880a7efaee997fa70addfdb12ca5f) to skip tests for commands.py on Windows. The test failures are here: https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3280/workflows/4708172e-7633-475d-bc73-aabbce65c90e/jobs/11919. Let me know if that's reasonable.

experiments/commands.py
AdamGleave: I see, let's leave it just using globbing for now then, someone can always run `commands.py` multiple times to express things like this if needed.

experiments/commands.py
AdamGleave: To check I understand: the point is most other flags won't have spaces in them, so enclosing them all in double quotes would just clutter things? That seems like a reasonable call. I worry a bit that this is fragile (e.g. what if IFS was set to things other than just spaces), but for this application readability may trump robustness.

AdamGleave:
> I pushed a [commit](https://github.com/HumanCompatibleAI/imitation/pull/654/commits/66b661a4ace880a7efaee997fa70addfdb12ca5f) to skip tests for commands.py on Windows. The test failures are here: https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3280/workflows/4708172e-7633-475d-bc73-aabbce65c90e/jobs/11919. Let me know if that's reasonable.

Issue is paths are `\` not `/`? Skipping seems fine.

experiments/commands.py
AdamGleave: For 1): yes, OK, but can we either add a TODO in the code or add this to issue #658?

experiments/commands.py
```diff
@@ -100,90 +99,16 @@ def _get_algo_name(cfg_file: str) -> str:
...
-    )
-    parser.add_argument(
-        "--container",
-        type=str,
-        default="hacobe/devbox:imitation",
-        help="The image name to use for the containers.",
-    )
-    args = parser.parse_args()
-
+def main(args: argparse.Namespace):
```
AdamGleave, line 109:
```suggestion
def main(args: argparse.Namespace) -> None:
```

experiments/commands.py
```diff
@@ -232,3 +157,83 @@ def _get_cfg_id(cfg_path: str) -> str:
             )
 
             print(hofvarpnir_cluster_cmd)
+
+
+def parse():
```
AdamGleave, line 162:
```suggestion
def parse() -> argparse.Namespace:
```

AdamGleave (approved):
LGTM. 2 minor suggestions + would be nice to have the `--force-pull` problem documented in a TODO or issue.

experiments/commands.py
hacobe: I added it to issue #658 under the "Solution" heading.

experiments/commands.py
hacobe: Yeah that's my thinking. And a user can always enclose the flag values they pass in double quotes like this --output-dir="\"/path/with/special characters\"" if they want the flag in the generated command to be enclosed in double quotes like this --logging_dir.log_root="/path/with/special characters".

PR #657, opened 2023-01-07
Document running the entire benchmarking suite

hacobe:


benchmarking/README.md
```diff
@@ -4,16 +4,93 @@ This directory contains sacred configuration files for benchmarking imitation's
...
+```bash
+python -m imitation.scripts.analyze analyze_imitation with \
+    source_dir_str="output/sacred" table_verbosity=0  \
+    csv_output_path=results.csv \
+    run_name="run0"
+```
+
+To compute a p-value to test whether the differences from the paper are statistically significant:
+
+```python
```
AdamGleave, line 66: (we'll want to make this a separate script, not just an entry in README)

benchmarking/README.md
```diff
@@ -4,16 +4,93 @@ This directory contains sacred configuration files for benchmarking imitation's
...
+  --cfg_pattern=benchmarking/example_*.json \
+  --seeds 0 1 2 \
+  --output_dir=/data/output \
+  --remote
+```
+
+To run those commands pipe them into bash:
+
+```bash
+python experiments/commands.py ... | bash
```
AdamGleave, line 52: Some preference for commands actually being runnable (expand the ellipsis)

AdamGleave: Perhaps could just combine with above, or make it a two-step process (`| tee cmds_to_run`), verify, then `bash cmds_to_run`.

benchmarking/README.md
```diff
@@ -4,16 +4,93 @@ This directory contains sacred configuration files for benchmarking imitation's
...
+  --name=run0 \
+  --cfg_pattern=benchmarking/example_*.json \
+  --seeds 0 1 2 \
+  --output_dir=output
+```
+
+To run those commands in parallel:
+
+```bash
+python experiments/commands.py ... | parallel -j 8
```
AdamGleave, line 35: Some preference for commands actually being runnable: perhaps combine with above (see discussion below).

AdamGleave (commented):
Thanks for making a start on this @hacobe.

@taufeeque9 we should add documentation for how to generate the table of results in the paper, you might find this draft useful for this purpose, though feel free to modify as it needed.

benchmarking/README.md
timbauman: I opted to just copy the command over for simplicity

benchmarking/README.md
```diff
@@ -1,19 +1,99 @@
...
+To produce a table with all the results:
+
+```bash
+python -m imitation.scripts.analyze analyze_imitation with \
+    source_dir_str="output/sacred" table_verbosity=0  \
+    csv_output_path=results.csv \
+    run_name="<name>"
+```
+
+To compute a p-value to test whether the differences from the paper are statistically significant:
```
timbauman, line 95: FWIW, in my test run they were statistically significant so something may have changed since the paper

src/imitation/scripts/compare_to_baseline.py
```diff
@@ -0,0 +1,100 @@
...
+"""
+
+import numpy as np
+import pandas as pd
+import scipy
+
+from imitation.data import types
+
+
+def compare_results_to_baseline(results_file: types.AnyPath) -> str:
```
timbauman, line 25: todo: add a test?

src/imitation/scripts/compare_to_baseline.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    summary.columns = summary.columns.get_level_values(1)
+    summary = summary.reset_index()
+
+    # Table 2 (https://arxiv.org/pdf/2211.11972.pdf)
+    # todo: store results in this repo outside this file
+    baseline = pd.DataFrame.from_records(
+        [
+            {
+                "algo": "??exp_command=bc",
```
timbauman, line 52: For whatever reason this is how the analyze script stores the algo field

benchmarking/README.md
AdamGleave: Not too surprising, can you share the results and if they moved in a positive or negative direction since the paper? ;)

benchmarking/README.md
```diff
@@ -1,19 +1,99 @@
...
+To view the results:
+
+```bash
+python -m imitation.scripts.analyze analyze_imitation with \
+    source_dir_str="output/sacred" table_verbosity=0  \
+    csv_output_path=results.csv \
+    run_name="<name>"
+```
+
+To run a single benchmark from Python add the config to your experiment:
```
AdamGleave, line 27:
```suggestion
To run a single benchmark from Python add the config to your Sacred experiment `ex`:
```

benchmarking/README.md
```diff
@@ -1,19 +1,99 @@
...
+
+```bash
+python experiments/commands.py \
+  --name=<name> \
+  --cfg_pattern "benchmarking/example_*.json" \
+  --seeds 0 1 2 \
+  --output_dir=output | parallel -j 8
+```
+
+(You may need to `brew install parallel` to get this to work on Mac)
```
AdamGleave, line 58:
```suggestion
(You may need to `brew install parallel` to get this to work on Mac.)
```

src/imitation/scripts/compare_to_baseline.py
```diff
@@ -0,0 +1,100 @@
+"""Compare experiment results to baseline results.
+
+This script compares experiment results to the results reported in the
```
AdamGleave, line 3: I wonder if it's better to compare two CSVs, and we just ship a "reference" CSV with the GitHub? That way we can allow things to deviate from the paper over time, which seems desirable (as we add new algorithms, environments). We could still include the paper results as a CSV to allow people to check for regressions over longer periods.

src/imitation/scripts/compare_to_baseline.py
```diff
@@ -0,0 +1,100 @@
...
+    summary = (
+        data[["algo", "env_name", "imit_return"]]
+        .groupby(["algo", "env_name"])
+        .describe()
+    )
+    summary.columns = summary.columns.get_level_values(1)
+    summary = summary.reset_index()
+
+    # Table 2 (https://arxiv.org/pdf/2211.11972.pdf)
+    # todo: store results in this repo outside this file
```
AdamGleave, line 48: :+1:

src/imitation/scripts/compare_to_baseline.py
AdamGleave: Hmm, we should probably try to clean that up! @taufeeque9 any idea why it's doing this?

src/imitation/scripts/compare_to_baseline.py
```diff
@@ -0,0 +1,100 @@
...
+            {
+                "algo": "??exp_command=bc",
+                "env_name": "seals/HalfCheetah-v0",
+                "mean": 3446,
+                "margin": 130,
+            },
+        ],
+    )
+    baseline["count"] = 5
+    baseline["confidence_level"] = 0.95
```
AdamGleave, line 66: Do we want confidence level stored in the DF? Is there a reason it should vary between algo?

src/imitation/scripts/compare_to_baseline.py
```diff
@@ -0,0 +1,100 @@
...
+    # Back out the standard deviation from the margin of error.
+    baseline["std"] = (
```
AdamGleave, line 68: Consider splitting up into multiple expressions, 6 lines can be hard to read.

AdamGleave: Also I don't think this is actually the standard deviation given you're dividing by `ppf`

tests/scripts/test_scripts.py
```diff
@@ -1075,3 +1076,42 @@ def test_convert_trajs_from_current_format_is_idempotent(
     assert (
         filecmp.dircmp(converted_path, original_path).diff_files == []
     ), "convert_trajs not idempotent"
+
+
+@pytest.mark.parametrize(
```
AdamGleave, line 1081: Thanks for adding a test!

benchmarking/README.md
timbauman: I'll include this once I include the canonical results CSV

src/imitation/scripts/compare_to_baseline.py
timbauman: I was thinking I'd just commit the latest results we have, whether that's from the paper or otherwise. We could include the paper results somewhere too although that feels less essential.

It's likely that many of the behaviors have changed since the paper so we may need to rerun to get up-to-date results (probably outside of this PR)

src/imitation/scripts/compare_to_baseline.py
timbauman: Oops good catch - should have read this more thoroughly

benchmarking/README.md
timbauman: Actually, I think we should just do a bulk update of all results - I might need help with getting access to more compute for this though

timbauman: Created issue for this https://github.com/HumanCompatibleAI/imitation/issues/710

timbauman: Okay for Ant it looks like the original results were mean 1953 std dev 99 and the new results (for me) are mean 1794 std dev 244. The p-value is 0.20 so it's not a statistically significant difference though

src/imitation/scripts/compare_to_baseline.py
taufeeque9: That is the default behavior (else statement below) when the Sacred command is not recognized. The commands to train `bc` and `dagger` was changed earlier without updating the `_get_algo_name` function in `analyze.py`. A check for `preference_comparisons` is also missing from the function.

Function reproduced from `analyze.py`:
```python
def _get_algo_name(sd: sacred_util.SacredDicts) -> str:
    exp_command = _get_exp_command(sd)

    if exp_command == "gail":
        return "GAIL"
    elif exp_command == "airl":
        return "AIRL"
    elif exp_command == "train_bc":
        return "BC"
    elif exp_command == "train_dagger":
        return "DAgger"
    else:
        return f"??exp_command={exp_command}"
```

benchmarking/README.md
```diff
@@ -1,19 +1,99 @@
...
+  --seeds 0 1 2 \
+  --output_dir=/data/output \
+  --remote
+```
+
+To run those commands pipe them into bash:
+
+```bash
+python experiments/commands.py \
+  --name <name> \
```
ernestum, line 77: Should this be `--name <name>`?

Issue #655, opened 2023-01-06
Issues with gpu using SAC+GAIL


cwj22:
Hi! I'm having some trouble running SAC with GAIL if I use gpu. If I change the learner to SAC in  `docs/tutorials/3_train_gail.ipynb` it throws an error.
```

Traceback (most recent call last):
  File "/home/catherine/anaconda3/envs/py38_GAIL/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/catherine/Documents/Research/imitation/src/imitation/util/networks.py", line 28, in training_mode
    yield m
  File "/home/catherine/Documents/Research/imitation/src/imitation/algorithms/adversarial/common.py", line 458, in train
    self.train_disc()
  File "/home/catherine/Documents/Research/imitation/src/imitation/algorithms/adversarial/common.py", line 351, in train_disc
    for batch in batch_iter:
  File "/home/catherine/Documents/Research/imitation/src/imitation/algorithms/adversarial/common.py", line 608, in _make_disc_train_batches
    log_policy_act_prob = self._get_log_policy_act_prob(obs_th, acts_th)
  File "/home/catherine/Documents/Research/imitation/src/imitation/algorithms/adversarial/common.py", line 512, in _get_log_policy_act_prob
    scaled_acts_th = self.policy.scale_action(acts_th)
  File "/home/catherine/anaconda3/envs/py38_GAIL/lib/python3.8/site-packages/stable_baselines3/common/policies.py", line 364, in scale_action
    return 2.0 * ((action - low) / (high - low)) - 1.0
  File "/home/catherine/anaconda3/envs/py38_GAIL/lib/python3.8/site-packages/torch/_tensor.py", line 955, in __array__
    return self.numpy()
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

python 3.8 on Ubuntu 20.04.3 LTS

Thanks in advance!

AdamGleave:
Assigning to @ernestum to investigate & triage.

We should consider adding some tests in our CI to catch things like this, right now our CI run is CPU only making it easy to slip past, we might be able to mock a GPU just to catch device type errors.

ernestum:
In this line we call `policy.scale_action()` on a pytorch tensor:

[...]

PR #660, opened 2023-01-11
Fix GAIL with SAC learner on GPU

ernestum:
## Description

Right now GAIL with a SAC learner is broken on the GPU. Read details in #655 

## Testing

This PR includes a test to reproduce the issue.
Note: the test in the pipeline does not fail since it is executed without GPU.

### DIFF
--- a/tests/algorithms/test_adversarial.py
+++ b/tests/algorithms/test_adversarial.py
@@ -399,9 +399,9 @@ def test_logits_expert_is_high_log_policy_act_prob(
         trans.dones,
     )
     log_act_prob_non_none = np.log(0.1 + 0.9 * np.random.rand(n_timesteps))
-    log_act_prob_non_none = th.as_tensor(log_act_prob_non_none).to(obs.device)
+    log_act_prob_non_none_th = th.as_tensor(log_act_prob_non_none).to(obs.device)
 
-    for log_act_prob in [None, log_act_prob_non_none]:
+    for log_act_prob in [None, log_act_prob_non_none_th]:
         maybe_error_ctx: contextlib.AbstractContextManager
         if isinstance(trainer_diverse_env, airl.AIRL) and log_act_prob is None:
             maybe_error_ctx = pytest.raises(TypeError, match="Non-None.*required.*")
--- a/tests/algorithms/test_adversarial.py
+++ b/tests/algorithms/test_adversarial.py
@@ -433,3 +433,35 @@ def test_compute_train_stats(n_samples):
     for k, v in stats.items():
         assert isinstance(k, str)
         assert isinstance(v, float)
+
+
+def test_regression_gail_with_sac(pendulum_expert_trajectories, pendulum_venv):
+    """Right now GAIL with a SAC learner on GPU crashes when training (see #655).
+
+    This is a minimal test to reproduce it.
+
+    Args:
+        pendulum_expert_trajectories: expert trajectories for Pendulum env.
+        pendulum_venv: the Pendulum environment.
+    """
+    from stable_baselines3 import SAC
+
+    from imitation.algorithms.adversarial.gail import GAIL
+    from imitation.rewards.reward_nets import BasicRewardNet
+
+    learner = SAC(
+        env=pendulum_venv,
+        policy=stable_baselines3.sac.policies.SACPolicy,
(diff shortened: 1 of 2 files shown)


ernestum:
This now fixes the issue and makes the test pass again (also on a GPU device).
However the regression test is pointless since the CircleCI runners don't have a GPU.

I tried to use the `meta` device in pytorch [as suggested here](https://github.com/pytorch/pytorch/issues/61654#issuecomment-879989145) but then we run into issues because the tensors on the meta device contain no data.

Another option would be to switch to a cuda-enabled runner but then the [supported images](https://circleci.com/docs/configuration-reference/#available-linux-gpu-images) seem to be quite restricted and it would probably be costly/wasteful to run all of our tests in such a runner.

I would propose to introduce a GPU flag for our tests and then run only those on a dedicated GPU runner. What do you think @AdamGleave ?

tests/algorithms/test_adversarial.py
```diff
@@ -433,3 +433,35 @@ def test_compute_train_stats(n_samples):
...
+def test_regression_gail_with_sac(pendulum_expert_trajectories, pendulum_venv):
+    """Right now GAIL with a SAC learner on GPU crashes when training (see #655).
+
+    This is a minimal test to reproduce it.
+
+    Args:
+        pendulum_expert_trajectories: expert trajectories for Pendulum env.
+        pendulum_venv: the Pendulum environment.
+    """
+    from stable_baselines3 import SAC
```
AdamGleave, line 447: Is there a reason we need to import these inside the test? We generally avoid that in this codebase and https://google.github.io/styleguide/pyguide.html#313-imports-formatting requires that "Imports are always put at the top of the file"

tests/algorithms/test_adversarial.py
```diff
@@ -433,3 +433,35 @@ def test_compute_train_stats(n_samples):
...
+
+    from imitation.algorithms.adversarial.gail import GAIL
```
AdamGleave, line 449: This is already present as `gail.GAIL` -- why are we importing it again?

tests/algorithms/test_adversarial.py
```diff
@@ -433,3 +433,35 @@ def test_compute_train_stats(n_samples):
...
+    from imitation.rewards.reward_nets import BasicRewardNet
```
AdamGleave, line 450: This is already present as `reward_nets.BasicRewardNet` -- why are we importing it again?

tests/algorithms/test_adversarial.py
```diff
@@ -433,3 +433,35 @@ def test_compute_train_stats(n_samples):
     for k, v in stats.items():
         assert isinstance(k, str)
         assert isinstance(v, float)
+
+
+def test_regression_gail_with_sac(pendulum_expert_trajectories, pendulum_venv):
+    """Right now GAIL with a SAC learner on GPU crashes when training (see #655).
```
AdamGleave, line 439: "Right now" will be wrong as soon as this is merged! Rephrase?

AdamGleave:
Sorry for only just getting around to reviewing this. Overall the changes looks good, I had a few minor comments on the tests.

> I would propose to introduce a GPU flag for our tests and then run only those on a dedicated GPU runner. What do you think @AdamGleave ?

At a high-level makes sense. We don't actually have access to the CircleCI GPU executors (they require a minimum spend of $6k/year to enable it). We have self-hosted GPU test runners in the past (and that lets us use any Docker image), but it is a bit of a pain to maintain. So I might lean towards adding GPU tests with the flag but putting enabling it on CI on the back-burner, and we can always try to manually run the GPU tests before a release to sanity check?

ernestum:
Thanks for the review! I implemented all of your suggestions.

Now the test only runs when a GPU is available (otherwise it is pointless) and I opened #671 so we keep in mind to automate GPU tests someday.

ernestum:
Note: the reduction in coverage is intentional. We lost some lines of coverage since I disabled the regression test for CPU-only instances.

AdamGleave (approved):
LGTM

AdamGleave:
@ernestum can you add `# pragma: no cover` to the test that's being skipped? Otherwise codecov check will keep complaining about the test code not being executed.

Issue #600, opened 2022-10-28
Remove type: ignore once evaluate_policy type annotation generalized


AdamGleave:
Remove `# type: ignore[arg-type]` introduced in https://github.com/HumanCompatibleAI/imitation/pull/593/files#diff-648af1fe52e7f75f6ddc457b062fe9ee7b456d9358c493cef1a3547f5e87785b for `evaluate_policy` once https://github.com/DLR-RM/stable-baselines3/pull/1146 is merged

AdamGleave:
https://github.com/DLR-RM/stable-baselines3/pull/1146 got merged so now we're just waiting on the next SB3 release

ernestum:
@AdamGleave should I take care of the new SB3 release so we can get forward on this?

AdamGleave:
If you can make a new SB3 release in <1 hour then go for it. Otherwise probably not worth it: this issue is just a nit, it could wait.

ernestum:
@araffin made a new release for us so we can progress on this.

PR #661, opened 2023-01-11
Remove type: ignore in evaluate_policy calls 

ernestum:
## Description

Remove type: ignore in evaluate_policy calls that are not needed since the issue is solved in stable-baselines3
Solves #600 

Note: I had to change the way entropy is handled when the policy reports no action entropy.


AdamGleave (approved):
LGTM

PR #663, opened 2023-01-24
Establish BC as an ingredient

ernestum:
## Description

Working towards #79, this PR
- splits the `train` ingredient into a `policy_evaluation` and a `policy` ingredient.
- pulls a `bc` ingredient out of `train_imitation`
- untangles the `BehaviorCloning` and the `DAgger` execution inside of `train_imitation`

## Testing

### DIFF
--- a/benchmarking/example_dagger_seals_hopper_best_hp_eval.json
+++ b/benchmarking/example_dagger_seals_hopper_best_hp_eval.json
@@ -1,5 +1,5 @@
 {
-  "bc_kwargs": {
+  "bc": {
     "batch_size": 16,
     "l2_weight": 0.0001,
     "optimizer_cls": {
@@ -7,12 +7,12 @@
     },
     "optimizer_kwargs": {
       "lr": 0.001
+    },
+    "train_kwargs": {
+      "log_interval": 500,
+      "n_batches": null,
+      "n_epochs": 1
     }
-  },
-  "bc_train_kwargs": {
-    "log_interval": 500,
-    "n_batches": null,
-    "n_epochs": 1
   },
   "dagger": {
     "beta_schedule": {
--- a/benchmarking/example_dagger_seals_hopper_best_hp_eval.json
+++ b/benchmarking/example_dagger_seals_hopper_best_hp_eval.json
@@ -27,7 +27,7 @@
     "rollout_type": "ppo-huggingface",
     "n_expert_demos": null
   },
-  "train": {
+  "policy": {
     "policy_cls": {
       "py/type": "imitation.policies.base.FeedForward32Policy"
     },
--- a/benchmarking/example_dagger_seals_hopper_best_hp_eval.json
+++ b/benchmarking/example_dagger_seals_hopper_best_hp_eval.json
@@ -28,7 +28,6 @@
     "n_expert_demos": null
   },
   "train": {
-    "n_episodes_eval": 50,
     "policy_cls": {
       "py/type": "imitation.policies.base.FeedForward32Policy"
     },
@@ -42,6 +41,9 @@
         }
       }
     }
+  },
+  "policy_evaluation": {
+    "n_episodes_eval": 50
   },
   "environment": {
     "gym_id": "seals/Hopper-v0"
--- a/experiments/transfer_learn_benchmark.sh
+++ b/experiments/transfer_learn_benchmark.sh
@@ -29,7 +29,7 @@ while true; do
       REWARD_MODELS_DIR="tests/testdata/reward_models"
       NEED_TEST_FILES="true"
       SEEDS=(0)
-      extra_configs=("${extra_configs[@]}" environment.fast rl.fast train.fast fast)
+      extra_configs=("${extra_configs[@]}" environment.fast rl.fast policy_evaluation.fast fast)
       shift
       ;;
     -w | --wandb)
--- a/src/imitation/scripts/train_imitation.py
+++ b/src/imitation/scripts/train_imitation.py
@@ -4,11 +4,11 @@ import logging
 import os.path as osp
 import pathlib
 import warnings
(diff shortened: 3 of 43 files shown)


ernestum:
Not sure what is the matter with the pipeline right now. I hope it is just circleci having some temporary issues.

Rocamonde:
Sure! Seems like a fairly big PR so I'll try to take a look this weekend.

ernestum:
@Rocamonde it is a large PR but I made sure that the individual commits are coherent. I recommend to review commit-by-commit.

Regarding splitting up `policy` and `policy_evaluation` ingredients: yes most probably whenever we use `policy_evaluation` we need a policy too. And also in most cases where a `policy` is needed we want to we want to evaluate it. But the conceptual distinction is still important:
For example we previously had a `train.fast` named config in the `train` ingredient. This suggested that the training would be fast when actually this meant that the evaluation will be fast. Now we have a `policy_evaluation.fast` named config which makes things clear. Merging `policy` and `policy_evaluation` back into a single ingredient would mean we have a `policy.fast` ingredient which is misleading again.

AdamGleave:
I'd prefer if @Rocamonde reviewed as I have limited time right now but I can look into it or re-assigning to someone else if Rocamonde is slammed -- don't want this PR to get blocked.

experiments/transfer_learn_benchmark.sh
```diff
@@ -29,7 +29,7 @@ while true; do
       REWARD_MODELS_DIR="tests/testdata/reward_models"
       NEED_TEST_FILES="true"
       SEEDS=(0)
-      extra_configs=("${extra_configs[@]}" environment.fast rl.fast train.fast fast)
+      extra_configs=("${extra_configs[@]}" environment.fast rl.fast fast)
```
Rocamonde, line 32: Why remove? You had been replacing them elsewhere

src/imitation/scripts/train_imitation.py
```diff
@@ -3,173 +3,139 @@
...
-                # parameter mandatory for ActorCriticPolicy, but not used by BC
-                "lr_schedule": utils.get_schedule_fn(1),
-            },
-        )
-    policy: policies.BasePolicy
-    if agent_path is not None:
-        warnings.warn(
-            "When agent_path is specified, policy_cls and policy_kwargs are ignored.",
-            RuntimeWarning,
+def _add_expert_stats_to_stats_dict_if_all_are_trajectories_with_reward(
```
Rocamonde, line 22: This seems like an unreasonably large function name, but not sure I can think of a way to make it shorter.

src/imitation/scripts/train_imitation.py
```diff
@@ -3,173 +3,139 @@
...
 
-@train_imitation_ex.capture
-def train_imitation(
-    bc_kwargs: Mapping[str, Any],
-    bc_train_kwargs: Mapping[str, Any],
-    dagger: Mapping[str, Any],
-    use_dagger: bool,
-    agent_path: Optional[str],
+@train_imitation_ex.command
+def bc(
```
Rocamonde, line 40: Rationale behind splitting them? I guess it does simplify the function signature quite a lot. But I wonder what the original explanation for doing it this way was, and why this is now outweighed.

Rocamonde (commented):
Looks good. Left some comments. The overall design decision seems good and does clean up the API for the better.

benchmarking/example_dagger_seals_hopper_best_hp_eval.json
```diff
@@ -1,19 +1,19 @@
 {
-  "bc_kwargs": {
+  "bc": {
```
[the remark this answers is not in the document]
ernestum: That is my hope too :-)

experiments/transfer_learn_benchmark.sh
ernestum: Good catch! I added it back again as `policy_evaluation.fast`

src/imitation/scripts/train_imitation.py
ernestum: I refactored this a bit to hopefully make it easier to read. What do you think about it?

src/imitation/scripts/train_imitation.py
ernestum: The original decision to have them both in the same function was probably to reduce code duplication? I can only guess here. BC is needed as an ingredient to DAgger so there is some kind of hierarchical dependency between the two.

I think it is conceptually cleaner to split them. In the long term I want to have each algorithm in its own ingredient and its own script and this is a first step in that direction. For more details you can read the discussion starting [here](https://github.com/HumanCompatibleAI/imitation/issues/79#issuecomment-1401904262).

Rocamonde:
FWIW in the future avoid force pushing since that messes up the PR review interface on GitHub. Now I cannot seem to access the most recent changes only, and I am forced to re-review the whole PR.

src/imitation/scripts/train_imitation.py
```diff
@@ -19,21 +19,25 @@
 logger = logging.getLogger(__name__)
 
 
-def _add_expert_stats_to_stats_dict_if_all_are_trajectories_with_reward(
-    stats: Dict[str, Any],
+def _all_trajectories_have_reward(trajectories: Sequence[types.Trajectory]) -> bool:
```
Rocamonde, line 22: Okay this is way better, thank you for being creative

Rocamonde (approved):
LGTM

ernestum:
> FWIW in the future avoid force pushing since that seems to mess up the PR review interface on GitHub. Now I cannot see the most recent changes only, and I am forced to re-review the whole PR.

Thanks for letting me know. I like to do it for the sake of keeping a clean history and sometimes to sync up with master. I will remember to avoid it after the first reviews came in.

PR #674, opened 2023-02-08
Switch back to the upstream release of sacred

ernestum:
## Description

Issue https://github.com/IDSIA/sacred/issues/879 is resolved, therefore we can switch back to using upstream sacred.

## Testing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,7 +122,7 @@ commands:
       # Download and cache dependencies
       - restore_cache:
           keys:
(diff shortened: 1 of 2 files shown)


ernestum:
@AdamGleave when you reviewed this feel free to merge right away. I would also recommend to take down https://github.com/HumanCompatibleAI/sacred to avoid confusion. Maybe we should deprecate the `sacred-chai` package as well otherwise other people might get stuck on an old sacred version.

AdamGleave (approved):
LGTM

PR #675, opened 2023-02-09
Add scripts and configs for hyperparameter tuning

taufeeque9:
## Description

This PR adds the scripts and configs for tuning hyperparameters of imitation algorithms. The results in the first draft of the [paper](https://arxiv.org/abs/2211.11972) were computed using the configs in the PR. Note that the preference comparisons algorithm still needs to be tuned on all of the environments.

Changes:

- `imitation/scripts/parallel.py`: Added features useful for tuning hyperparameters (HPs) of the algorithms. Specifically, added options to 1) repeat trials across configs to get a better (mean) estimate of the return of HP configs, 2) evaluate the best config from an HP tune run, and 3) restart failed trials of a tune run.
- `imitation/scripts/analyze.py`: Added table verbosity level 3, which generates a CSV with all the config HPs, including RL HPs and the specific algorithm's arguments, along with the expert returns and the returns of the imitation algorithm.
- `imitation/scripts/config/parallel.py`: Added HP search space for various algorithms like BC, Dagger, GAIL, and AIRL. Search space for Preference Comparisons may not be good enough.
- `imitation/scripts/config/train_*.py`: The added configs contain the HP for the tuned base RL algorithm.

## Testing

Updated test cases for `parallel` and tested HP tuning manually.


ernestum (changes requested):
First round of review mostly with suggestions to the parallel script.

ernestum:
Not part of this PR but @taufeeque9  could you elaborate on the `benchmarking/util.py` script? It is unclear right now where in the benchmark pipeline it is used.

src/imitation/scripts/config/train_rl.py
```diff
@@ -70,8 +72,30 @@ def cartpole():
...
+            clip_range=0.4,
+            ent_coef=0.008508727919228772,
+            gae_lambda=0.9,
+            gamma=0.9999,
+            learning_rate=0.0012403278189645594,
+            max_grad_norm=0.8,
+            n_epochs=10,
+            vf_coef=0.489343896591493,
+        ),
+    )
```
ernestum, line 98: I think we should somehow document where all those new (also below and in other config files) hyperparameter values come from. Otherwise, it is unclear why we should trust them. What script generated them? What was the HP search space? How large the optimization budget?

src/imitation/scripts/train_adversarial.py
```diff
@@ -167,6 +167,7 @@ def callback(round_num: int, /) -> None:
     return {
         "imit_stats": imit_stats,
         "expert_stats": rollout.rollout_stats(expert_trajs),
+        "mean_return": imit_stats["monitor_return_mean"],
```
ernestum, line 170: Unclear why you set `mean_return` to `monitor_return_mean` here. Can we at least document this quirk of the `eval_policy` output? We also seem to do it in `train_rl.py` and `train_preference_comparisons.py`. So maybe add a function that just does this weird copying and document the reason in it's docstring to make this explicit.

ernestum (changes requested):

src/imitation/scripts/config/train_adversarial.py
```diff
@@ -242,7 +239,6 @@ def seals_swimmer():
 
 @train_adversarial_ex.named_config
 def seals_walker():
-    locals().update(**MUJOCO_SHARED_LOCALS)
```
ernestum, line 148: Careful! While the others were commented out, this one was not. Are you sure you want to remove this line?

src/imitation/scripts/parallel.py
```diff
@@ -95,15 +126,110 @@ def parallel(
...
+                print("Mean return:", row["mean_return"])
+                print("All returns:", df[df["mean_return"] == row["mean_return"]][key])
+                print("Total seeds:", (df["mean_return"] == row["mean_return"]).sum())
+                best_config["config_updates"].update(
+                    seed=ray.tune.grid_search(list(range(100, 100 + eval_trial_seeds))),
+                )
+                # update cpus per trial only if it is provided in `resources_per_trial`
+                # Uses the default values (cpu=1) if it is not provided
+                if "cpu" in resources_per_trial:
+                    resources_per_trial_eval = copy.deepcopy(resources_per_trial)
```
ernestum, line 211: What is the point of this variable? It is not used anywhere.

src/imitation/scripts/analyze.py
```diff
@@ -262,26 +265,35 @@ def analyze_imitation(
...
-        table_verbosity: Increasing levels of verbosity, from 0 to 2, increase the
-            number of columns in the table.
+        table_verbosity: Increasing levels of verbosity, from 0 to 3, increase the
+            number of columns in the table. Level 3 prints all of the columns available.
 
     Returns:
         The DataFrame generated from the Sacred logs.
     """
-    table_entry_fns_subset = _get_table_entry_fns_subset(table_verbosity)
+    if table_verbosity == 3:
```
ernestum, line 274: Could you add a note to the code to explain this edge case?

src/imitation/scripts/analyze.py
```diff
@@ -262,26 +265,35 @@ def analyze_imitation(
...
+        table_entry_fns_subset = _get_table_entry_fns_subset(2)
+    else:
+        table_entry_fns_subset = _get_table_entry_fns_subset(table_verbosity)
 
-    rows = []
+    df = pd.DataFrame()
```
ernestum, line 279: Can we have a bit more expressive variable names than just `df` and `new_df`. What does the df contain? What is the purpose of `new_df`?

src/imitation/scripts/parallel.py
```diff
@@ -95,15 +126,114 @@ def parallel(
...
+            )
+        if sacred_ex_name == "train_rl":
+            key_prefix = ""
+        elif sacred_ex_name == "train_preference_comparisons":
+            key_prefix = "rollout/"
+        else:
+            key_prefix = "imit_stats/"
+        key = key_prefix + "monitor_return_mean"
+
+        if eval_best_trial:
```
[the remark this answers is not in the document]
ernestum: After looking at this again I would have a strong preference to exclude the result evaluation from the `parallel.py` script. This scrip is already way to complex and grown over the years. Now we add another feature to it and I fear it will become entirely incomprehensible in the future. I am happy to discuss other solutions.

src/imitation/scripts/config/train_adversarial.py
taufeeque9: Yes, this line was vestigial as the `MUJOCO_SHARED_LOCALS` are anyways getting updated with the config below it.

taufeeque9:
> Not part of this PR but @taufeeque9 could you elaborate on the `benchmarking/util.py` script? It is unclear right now where in the benchmark pipeline it is used.

Running the HP tuning script using `parallel.py` generates JSON files for all the HP tuning config that contains the sacred config that can be used to re-run the experiment with the exact arguments. 

The `benchmarking/util.py` file contains the code to clean the generated JSON files by removing specific arguments like seed, rollout path, etc., that were specific for the run. The cleaned files are what we have uploaded as `benchmarking/*.json` for all the algorithms and the environments.

src/imitation/scripts/train_adversarial.py
taufeeque9: Removed the need to do all of this. Earlier, I used this hack to have a single location in the returned dict for the metric to be specified in `ray.tune.run`. Simplified this by using the original location of the metric for the different algorithms.

ernestum:
Just to make sure that I understand it correctly:

For each algorithm in {BC, AIRL, GAIL, DAgger} and each environment in {ant, half_cheetah, hopper, swimmer, walker} we tune the hyperparameters using `parallel.py`.
This means we have 4*5 = 20 runs of `parrallel.py`.

This generates some amount of files.

From those files we extract the best hyperparameter configurations using `benchmarking/utils.py` and place them in the json files in `benchmarking/`.

To reproduce a run with one of those runs with the tuned hyperparameters we run `python -m imitation.scripts.<train_script> <algo> with benchmarking/<config_name>.json`

taufeeque9:
> Just to make sure that I understand it correctly:
> 
> For each algorithm in {BC, AIRL, GAIL, DAgger} and each environment in {ant, half_cheetah, hopper, swimmer, walker} we tune the hyperparameters using `parallel.py`. This means we have 4*5 = 20 runs of `parrallel.py`.
> 
> This generates some amount of files.
> 
> From those files we extract the best hyperparameter configurations using `benchmarking/utils.py` and place them in the json files in `benchmarking/`.
> 
> To reproduce a run with one of those runs with the tuned hyperparameters we run `python -m imitation.scripts.<train_script> <algo> with benchmarking/<config_name>.json`.
> 
> Can you verify this @taufeeque9 ?

Yes, that sounds correct. The only thing to add is that we get the best hyperparameter configurations from running `parallel.py` with `eval_best_trial=True` and we use `benchmarking/utils.py` to clean the best configs that can be uploaded to the `benchmarking/` directory on GitHub.

src/imitation/scripts/parallel.py
```diff
@@ -31,14 +29,12 @@ def parallel(
...
-    repeat: int = 1,
-    eval_best_trial: bool = False,
-    eval_best_trial_resource_multiplier: int = 1,
-    eval_trial_seeds: int = 5,
-    experiment_checkpoint_path: str = "",
-    syncer=None,
-    resume: Union[str, bool] = False,
-) -> None:
+    repeat: int,
+    search_alg: Optional[str],
```
ernestum, line 32: I would prefer if you did not introduce any tuning-specific arguments to the `parallel` script. One way to solve this is to add a `tune_run_kwargs` parameter and use that to inject the optuna search algo etc.

Then it becomes the responsibility of the `tuning.py` script to set everything up correctly.

ernestum (changes requested):
Thanks for the updates @taufeeque9 !
Could you also make sure to add an overview documentation with some example calls for the `tuning.py` script?

Also some rough tests for it would be nice.

src/imitation/scripts/parallel.py
taufeeque9: I think `repeat` is a good argument to keep in the `parallel` script because it allows a sampled hyperparameter config to run multiple times, which is a good general feature to have in the `parallel` script. So if we want to keep it in the `parallel` script, we must keep `search_alg` in the `parallel` script because `repeat` requires a search algorithm to be specified first.

If we don't think `repeat` is beneficial to keep in the `parallel` script, I can introduce a `tune_run_kwargs` in `parallel` and move the repeat and search algorithm options to the `tuning` script only.

benchmarking/tuning_config.py
```diff
@@ -0,0 +1,237 @@
...
+    parallel = dict(
+        sacred_ex_name="train_imitation",
+        run_name="bc_tuning",
+        base_named_configs=["logging.wandb_logging"],
+        base_config_updates={
+            "environment": {"num_vec": 1},
+            "demonstrations": {"source": "huggingface"},
+        },
+        search_space={
+            "config_updates": {
```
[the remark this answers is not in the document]
taufeeque9: I think it is more straightforward and clear to keep the "config_updates" hierarchy as it matches the way we use config_updates in the parallel script and the tuning config.

benchmarking/tuning.py
```diff
@@ -0,0 +1,102 @@
...
+    run = parallel_ex.run(config_updates=parallel)
+    result = run.result
+
+    if eval_best_trial:
+        if parallel["sacred_ex_name"] == "train_rl":
+            return_key = "monitor_return_mean"
+        else:
+            return_key = "imit_stats/monitor_return_mean"
+        df = result.results_df
+        df = df[df["config/named_configs"].notna()]
```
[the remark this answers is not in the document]
taufeeque9: Thanks for pointing this out! This is an unnecessary line that was useful at some point when I was benchmarking the algorithms but is not required now. I have removed it.

benchmarking/tuning_config.py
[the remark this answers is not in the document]
ernestum: All right.

src/imitation/scripts/parallel.py
ernestum: I agree with the `repeat` argument. However, if we just repeat the same configuration a couple of times, why do we need a search algorithm? Maybe the Ray API needs one up just any search algorithm will do right?

src/imitation/scripts/parallel.py
```diff
@@ -72,19 +72,22 @@ def parallel(
         init_kwargs: Arguments to pass to `ray.init`.
         local_dir: `local_dir` argument to `ray.tune.run()`.
         upload_dir: `upload_dir` argument to `ray.tune.run()`.
-        search_alg: can be either "optuna" or None.
+        search_alg: can be either "optuna" or None. Setting `None` allows for
```
ernestum, line 75: Now it is a lot clearer what this parameter does. It is still confusing why it behaves that way. Again, I agree that we should keep the `repeat` argument but I think there should be a way to go without the `search_alg` argument.

src/imitation/scripts/parallel.py
```diff
@@ -72,19 +72,22 @@ def parallel(
...
             of the experiment.
         syncer: `syncer` argument to `ray.tune.syncer.SyncConfig`.
 
     Raises:
         TypeError: Named configs not string sequences or config updates not mappings.
 
     Returns:
-        The result of `ray.tune.run()`.
+        The result of running the parallel experiment with `ray.tune.run()`.
+        Useful for fetching the configs and results dataframe of all the trials.
```
ernestum, line 90: Perfect! This is the context that was missing. :+1:

tests/test_benchmarking.py
```diff
@@ -44,3 +46,28 @@ def test_benchmarks_print_config_succeeds(algorithm: str, environment: str):
...
     assert run.status == "COMPLETED"
+
+
+@pytest.mark.parametrize("algorithm", ALGORITHMS)
+def test_tuning_print_config_succeeds(algorithm: str):
+    # We test the configs using the print_config command,
+    # because running the configs requires MuJoCo.
+    # Requiring MuJoCo to run the tests adds too much complexity.
+
+    # We need to use sys.executable, not just "python", on Windows as
```
ernestum, line 57: I like how you explain the use of `sys.executable` here!

ernestum (commented):
Thanks for implementing the changes I requested in the last round! The code is already a lot more readable. This round only includes a couple not nits and we still need to come to a conclusion with the `search_alg` parameter.

benchmarking/README.md
```diff
@@ -17,3 +17,24 @@ python -m imitation.scripts.<train_script> <algo> with benchmarking/<config_name
...
+The benchmarking hyperparameter configs were generated by tuning the hyperparameters using
+the search space defined in the `tuning_config.py` script. The tuning script proceeds in two
+phases: 1) The hyperparameters are tuned using the search space provided, and 2) the best 
+hyperparameter config found in the first phase based on the maximum mean return is
+re-evaluated on a separate set of seeds, and the mean and standard deviation of these trials
+are reported.
+
+To tune the hyperparameters of an algorithm using the default search space provided:
+```bash
+python tuning.py with example_{algo} 'parallel_run_config.base_named_configs=["{env}"]'
```
ernestum, line 33: Why is there the `example_` prefix?

taufeeque9: A vestigial naming convention left from the earlier code of parallel. Have thought about removing it for quite some time. Will remove it now.

src/imitation/scripts/parallel.py
taufeeque9: > if we just repeat the same configuration a couple of times, why do we need a search algorithm? 

The Ray API for repeating takes a search algorithm and wraps a repeater on top of it to give the modified search algorithm: `search.Repeater(optuna.OptunaSearch(), repeat=repeat)`.

The repeater allows running multiple trials with the same configuration sampled by the algorithm. Hence, it needs the search algorithm as an argument. I don't think we can have the repeat argument without the search algorithm.

src/imitation/scripts/parallel.py
taufeeque9: See the comment above: https://github.com/HumanCompatibleAI/imitation/pull/675#discussion_r1266249803

src/imitation/scripts/parallel.py
ernestum: What I meant was: we could just set the search algo to any simple search algo as soon as `repeat > 0` then we do not have to expose this quirk of the Ray API to our users.

src/imitation/scripts/config/parallel.py
```diff
@@ -7,10 +7,8 @@
 Adding custom named configs is necessary because the CLI interface can't add
 search spaces to the config like `"seed": tune.choice([0, 1, 2, 3])`.
 
-For tuning hyperparameters of an algorithm on a given environment, override
-the `base_named_configs` argument with the named config of the environment.
-Ex: python -m imitation.scripts.parallel with example_gail \
-    'base_named_configs=["logging.wandb_logging", "seals_half_cheetah"]'
+For tuning hyperparameters of an algorithm on a given environment,
+check out the benchmarking/tuning.py script.
```
ernestum, line 11: Why did you remove the example here?

ernestum (commented):
Thanks for the changes. Looking better and better now!

src/imitation/scripts/config/parallel.py
taufeeque9: The example was for tuning the HPs using the parallel script which we have now moved to the tuning script along with all the named configs for tuning HPs.

src/imitation/scripts/parallel.py
taufeeque9: That makes sense. I have redacted `search_alg` from the arguments of the parallel script, along with some other rarely used arguments to introduce `tune_run_kwargs`. Let me know if this looks good.

src/imitation/scripts/config/parallel.py
```diff
@@ -31,19 +34,11 @@ def config():
...
-def seeds(n_seeds):
-    search_space = {"config_updates": {"seed": tune.grid_search(list(range(n_seeds)))}}
-
-
-@parallel_ex.named_config
-def s3():
-    upload_dir = "s3://shwang-chai/private"
+    num_samples = 1  # Number of samples per grid search configuration
+    repeat = 1  # Number of times to repeat a sampled configuration
+    search_alg = "optuna"  # search algorithm to use
```
ernestum, line 39: This config can go away now right?

benchmarking/tuning.py
```diff
@@ -0,0 +1,167 @@
...
+            allocating more resources per trial to the evaluation trials than the
+            resources for hyperparameter tuning since number of evaluation trials
+            is usually much smaller than the number of tuning trials.
+        num_eval_seeds: Number of distinct seeds to evaluate the best trial on.
+            Set to 0 to disable evaluation.
+
+    Raises:
+        ValueError: If no trials are returned by the parallel run of tuning.
+    """
+    run = parallel_ex.run(config_updates=parallel_run_config)
```
ernestum, line 35: Set the `tune_run_kwargs` here to ensure that the search algorithm etc. is set correctly? It will work when `repeat>1` but not all of your configs hae that.

ernestum (commented):
This looks a lot better. Still some small things to consider.

benchmarking/tuning.py
```diff
@@ -32,7 +33,15 @@ def tune(
     Raises:
         ValueError: If no trials are returned by the parallel run of tuning.
     """
-    run = parallel_ex.run(config_updates=parallel_run_config)
+    search_alg = optuna.OptunaSearch()
+    updated_parallel_run_config = copy.deepcopy(parallel_run_config)
```
ernestum, line 37: Took me quite some time to understand how you are updating the config here (ok maybe I should have a :coffee: ). But I think this is essentially the same as this?
```python
    if "tune_run_kwargs" in updated_parallel_run_config:
        updated_parallel_run_config["tune_run_kwargs"]["search_alg"] = optuna.OptunaSearch()
    else:
        updated_parallel_run_config["tune_run_kwargs"] = dict(search_alg=optuna.OptunaSearch())
```

taufeeque9: Done!

ernestum: thx!

tests/scripts/test_scripts.py
```diff
@@ -882,13 +883,13 @@ def test_parallel_train_adversarial_custom_env(tmpdir):
...
         sacred_ex_name="train_adversarial",
-        n_seeds=1,
+        repeat=2,
         base_named_configs=[env_named_config] + ALGO_FAST_CONFIGS["adversarial"],
         base_config_updates=dict(
             logging=dict(log_root=tmpdir),
             demonstrations=dict(path=path),
         ),
-        search_space=dict(command_name="gail"),
+        search_space=dict(command_name=tune.choice(["gail"])),
```
ernestum, line 892: Why was this needed? Can we document this inline?

ernestum (commented):
This time around I did a pass over the entire PR again as the previous reviews were just incremental reviews on the most recent changes.

benchmarking/tuning.py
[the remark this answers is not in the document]
taufeeque9: I agree. Adding the tuning script was a great suggestion, as both the tuning and the parallel scripts are now much easier to understand. Thanks for pushing for it!

src/imitation/scripts/parallel.py
```diff
@@ -95,15 +113,47 @@ def parallel(
...
+            updated_tune_run_kwargs["search_alg"] = optuna.OptunaSearch()
+        try:
+            algo = updated_tune_run_kwargs["search_alg"]
+            algo = search.Repeater(algo, repeat)
+            updated_tune_run_kwargs["search_alg"] = algo
+        except AttributeError:
+            raise ValueError(
+                "repeat > 1 but search_alg is not an instance of "
+                "ray.tune.search.SearchAlgorithm",
+            )
```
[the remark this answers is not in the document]
taufeeque9: Good suggestion! I often forget about the `get` method in dictionary.

> Also: why are we using Optuna if we just want to repeat some runs? This seems like overkill and a simple grid search might have less overhead?

The point of adding `repeat` is that we want to perform multiple runs of all the sampled trials. So let's say if `learning_rate` is sampled from `loguniform(1e-5, 1e-2)`, then if a value of $1.3 * 10^{-4}$ was sampled for a trial, then our `repeat=n` argument will perform this trial with the value of $1.3 * 10^{-4}$ `n` times. I believe this cannot be done using grid search unless I'm mistaken.

Regarding Optuna, I agree that we don't have to use Optuna for repeat some runs and can instead use any other search algorithm like `tune.search.basic_variant.BasicVariantGenerator` which supports both random and grid searches.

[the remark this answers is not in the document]
ernestum: Note from discussion with @taufeeque9:
We will stick with Optuna since the `BasicVariantGenerator` does not support being wrapped with a repeater.

Details here: https://github.com/ray-project/ray/issues/33677

experiments/commands.py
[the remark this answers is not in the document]
taufeeque9: The documentation of `commands.py` and related scripts are currently in the benchmarking docs PR #657.

src/imitation/scripts/analyze.py
```diff
@@ -166,6 +166,9 @@ def _get_algo_name(sd: sacred_util.SacredDicts) -> str:
 
 def _return_summaries(sd: sacred_util.SacredDicts) -> dict:
     imit_stats = get(sd.run, "result.imit_stats")
+    if imit_stats is None:
+        # stored in rollout key for preference comparison
+        imit_stats = get(sd.run, "result.rollout")
```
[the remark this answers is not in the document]
taufeeque9: I had already changed it in the preference comparisons script but didn't remove it from here. Thanks for pointing this out!

tests/scripts/test_scripts.py
taufeeque9: This was needed because the search space needs to contain at least one element (so that it can be sampled) when we use the optuna search algorithm (enabled by repeat=2).

src/imitation/scripts/config/train_adversarial.py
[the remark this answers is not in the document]
taufeeque9: Those are just the hyperparameters of the tuned PPO algorithm on the respective environment. We set those as the base hyperparameters of any algorithm that uses RL as a subroutine (including preference comparisons).

I have added a note for it at the top of the script.

src/imitation/scripts/config/train_preference_comparisons.py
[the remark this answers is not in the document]
taufeeque9: I've added a note at the top of the script that the RL algorithm's hyperparameters were tuned using the RL Zoo library. I don't remember specific details like the tuning budget or the number of trials. Let me know if the note looks sufficient or if there are things we should definitely try to add.

tests/scripts/test_scripts.py
ernestum: All right. Then please document this inline because somebody (probably me) will be confused by this construct when I read it again :-)

taufeeque9: Done 👍

src/imitation/scripts/parallel.py
```diff
@@ -148,7 +198,7 @@ def _ray_tune_sacred_wrapper(
         `ex.run`) and `reporter`. The function returns the run result.
     """
 
-    def inner(config: Mapping[str, Any], reporter) -> Mapping[str, Any]:
+    def inner(config: Dict[str, Any], reporter) -> Mapping[str, Any]:
```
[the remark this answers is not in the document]
taufeeque9: Good call!

benchmarking/tuning.py
```diff
@@ -0,0 +1,174 @@
...
+import copy
+import pathlib
+from typing import Any, Dict
+
+import numpy as np
+import ray
+from pandas.api import types as pd_types
+from ray.tune.search import optuna
+from sacred.observers import FileStorageObserver
+from tuning_config import parallel_ex, tuning_ex
```
AdamGleave, line 12: PEP8: standard library, third-party then first party-imports with line separating each.
```suggestion

from tuning_config import parallel_ex, tuning_ex
```

benchmarking/tuning.py
```diff
@@ -0,0 +1,174 @@
...
+            is usually much smaller than the number of tuning trials.
+        num_eval_seeds: Number of distinct seeds to evaluate the best trial on.
+            Set to 0 to disable evaluation.
+
+    Raises:
+        ValueError: If no trials are returned by the parallel run of tuning.
+    """
+    updated_parallel_run_config = copy.deepcopy(parallel_run_config)
+    search_alg = optuna.OptunaSearch()
+    if "tune_run_kwargs" in updated_parallel_run_config:
```
AdamGleave, line 38: Is this equivalent?
```suggestion
tune_run_kwargs = updated_parallel_run_config.setdefault("tune_run_kwargs", dict())
tune_run_kwargs["search_alg"] = search_alg
```

benchmarking/tuning.py
```diff
@@ -0,0 +1,174 @@
...
+from tuning_config import parallel_ex, tuning_ex
+
+
+@tuning_ex.main
+def tune(
+    parallel_run_config: Dict[str, Any],
+    eval_best_trial_resource_multiplier: int = 1,
+    num_eval_seeds: int = 5,
+) -> None:
+    """Tune hyperparameters of imitation algorithms using parallel script.
```
AdamGleave, line 21: I didn't understand the separate training and evaluation stage until I read the code, consider adding some explanation of that?

AdamGleave (commented):
Are the scripts in `benchmarking/` being type checked and linted? I don't see them in the output at https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3897/workflows/02ec5434-d883-4edf-9fa5-99faeb0645f2/jobs/15772 The directory `benchmarking/` is not specified in the `inputs` config for ptype in `setup.cfg`. 

I think should either move the scripts into `src/imitation/scripts/` and make them part of the package, or update CI config to test things in `benchmarking/` (worth checking we're not missing them for flake8 etc as well).

benchmarking/tuning_config.py
```diff
@@ -0,0 +1,232 @@
...
+        resources_per_trial=dict(cpu=1),
+    )
+    num_eval_seeds = 5
+
+
+@tuning_ex.named_config
+def gail():
+    parallel_run_config = dict(
+        sacred_ex_name="train_adversarial",
+        run_name="gail_tuning_hc",
```
AdamGleave, line 120: What does "hc" stand for? Why is this only present for gail?

src/imitation/scripts/analyze.py
```diff
@@ -262,38 +262,50 @@ def analyze_imitation(
...
         The DataFrame generated from the Sacred logs.
     """
-    table_entry_fns_subset = _get_table_entry_fns_subset(table_verbosity)
+    if table_verbosity == 3:
+        # Get column names for which we have get value using make_entry_fn
+        # These are same across Level 2 & 3. In Level 3, we additionally add remaining
+        #  config columns.
+        table_entry_fns_subset = _get_table_entry_fns_subset(2)
+    else:
+        table_entry_fns_subset = _get_table_entry_fns_subset(table_verbosity)
```
AdamGleave, line 277: Could drop if statement and write:
```suggestion
    table_entry_fns_subset = _get_table_entry_fns_subset(min(table_verbosity, 2))
```

AdamGleave (commented):
Right now it seems like there's a lot of duplication in specifying the pre-tuned hyperparameters in `*.json` files in `benchmarking/`, and in `named_configs` in each algorithm. If possible it seems preferable to make this more DRY (Don't Repeat Yourself) so that we have a single source of truth: otherwise I could imagine them drifting apart as we do future hyperparameter sweeps.

It is convenient to be able to write e.g. `with seals_ant` rather than `with benchmarking/airl_seals_ant_best_hp_eval.json`. But Sacred does support an `ex.add_named_config`: https://github.com/IDSIA/sacred/blob/17c530660d5b405af0f5c286b1a93f3d8911d026/sacred/ingredient.py#L251 So we could replace some of these with that.

It does seem like the named configs have been edited to be more minimal. That's great, but as a user, which one should I use? If the answer is "always use the named_config" then we should really consider editing the JSONs (manually or automatically). If the answer is "it depends" then it would help to document the pros and cons and justify why we need both.

src/imitation/scripts/parallel.py
```diff
@@ -17,29 +20,33 @@
...
+) -> ray.tune.ExperimentAnalysis:
     """Parallelize multiple runs of another Sacred Experiment using Ray Tune.
 
     A Sacred FileObserver is attached to the inner experiment and writes Sacred
     logs to "{RAY_LOCAL_DIR}/sacred/". These files are automatically copied over
-    to `upload_dir` if that argument is provided.
+    to `upload_dir` if that argument is provided in `tune_run_kwargs`.
 
     Args:
         sacred_ex_name: The Sacred experiment to tune. Either "train_rl" or
```
AdamGleave, line 40:
```suggestion
        sacred_ex_name: The Sacred experiment to tune. Either "train_rl",
```

src/imitation/scripts/parallel.py
```diff
@@ -17,29 +20,33 @@
...
-            "train_adversarial".
+            "train_imitation" or "train_adversarial" or "train_preference_comparisons".
```
AdamGleave, line 41:
```suggestion
            "train_imitation", "train_adversarial" or "train_preference_comparisons".
```

src/imitation/scripts/parallel.py
```diff
@@ -95,15 +113,45 @@ def parallel(
...
-            config=search_space,
-            name=run_name,
-            local_dir=local_dir,
-            resources_per_trial=resources_per_trial,
-            sync_config=ray.tune.syncer.SyncConfig(upload_dir=upload_dir),
-        )
+        if experiment_checkpoint_path:
+            # load experiment analysis results
+            result = ray.tune.ExperimentAnalysis(experiment_checkpoint_path)
+            result._load_checkpoints_from_latest(
```
AdamGleave, line 137: Unfortunate we have to use a private method to load this -- I assume there's no alternative via public methods?

tests/scripts/test_scripts.py
```diff
@@ -977,7 +981,7 @@ def test_analyze_imitation(tmpdir: str, run_names: List[str], run_sacred_fn):
             assert run.status == "COMPLETED"
 
     # Check that analyze script finds the correct number of logs.
-    def check(run_name: Optional[str], count: int) -> None:
+    def check(run_name: Optional[str], count: int, table_verbosity=1) -> None:
```
AdamGleave, line 984:
```suggestion
    def check(run_name: Optional[str], count: int, table_verbosity: int = 1) -> None:
```

benchmarking/tuning.py
taufeeque9: Yes, I wasn't aware of `setdefault`. Nice little feature!

benchmarking/tuning.py
```diff
@@ -0,0 +1,174 @@
...
+    df = experiment_analysis.results_df
+    # convert object dtype to str required by df.groupby
+    for col in df.columns:
+        if pd_types.is_object_dtype(df[col]):
+            df[col] = df[col].astype("str")
+    # group into separate HP configs
+    grp_keys = [c for c in df.columns if c.startswith("config") and "seed" not in c]
+    grps = df.groupby(grp_keys)
+    # store mean return of runs across all seeds in a group
+    df["mean_return"] = grps[return_key].transform(lambda x: x.mean())
```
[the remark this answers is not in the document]
taufeeque9: These return different number of rows. Your suggestion obtains mean return for every group whereas the transform method obtains mean return for every trial which is compatible with the number of rows in `df`. I have explained this in the comments now.

benchmarking/tuning.py
```diff
@@ -0,0 +1,174 @@
...
+    best_config_df = df[df["mean_return"] == df["mean_return"].max()]
+    row = best_config_df.iloc[0]
+    best_config_tag = row["experiment_tag"]
+    assert experiment_analysis.trials is not None  # for mypy
+    best_trial = [
+        t for t in experiment_analysis.trials if best_config_tag in t.experiment_tag
+    ][0]
+
+    if print_return:
+        all_returns = df[df["mean_return"] == row["mean_return"]][return_key]
```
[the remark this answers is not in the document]
taufeeque9: In case we get multiple distinct hyperparameter groups that get the same mean returns across seeds, we pick the first hyperparameter group from the `df`. And the ordering of hyperparameter groups in the `df` should be arbitrary, so it should be fine to do this, right?

benchmarking/tuning_config.py
taufeeque9: Fixed. It got there from another run I was doing but shouldn't be there.

src/imitation/scripts/parallel.py
taufeeque9: The `ExperimentAnalysis` class was recently rewritten in `ray.tune` and we no longer need to use the private methods. It's good that you pointed this out!

taufeeque9:
> I think should either move the scripts into src/imitation/scripts/ and make them part of the package

I have moved the benchmarking scripts into `src/imitation/scripts` and the tuned hyperparameter json files into `src/imitation/scripts/config/tuned_hps`. This seems more clean and elegant to me than keeping everything in the `benchmarking` directory. With these changes the benchmarking directory seems useless now as it just contains a `util.py` file to clean up the json files. We can move the `util.py` file somewhere else and delete the benchmarking directory.

> Right now it seems like there's a lot of duplication in specifying the pre-tuned hyperparameters in *.json files in benchmarking/, and in named_configs in each algorithm.

This was a great point! Thanks for noting it. I have added all of the json files as named configs in their respective `train_<algo>` script. So now we can directly include the named configs from the command line instead of the json files. As a result, I have also removed the old named configs for the algorithms & environments we have tuned.

So for example, in the `train_adversarial` script, I have removed `seals_half_cheetah` named config and added `airl_seals_half_cheetah` and `gail_seals_half_cheetah` configs for the respective algorithms of `airl` and `gail`. So now a user can run the following:
```
python -m imitation.scripts.train_adversarial airl with airl_seals_half_cheetah
```

Note that there's a redundancy in the command where `airl` is being mentioned twice. I've tried to think of ways to remove this redundancy but couldn't figure a way out that preserves the command name `airl` and removes `airl` from the named config. We can maybe remove the command name altogether and keep `algo_env` config names but that would be quite a big change for this PR. I think for now raising this as an issue might be best so that we can deal with this later with some other PR.

AdamGleave (approved):
LGTM, some minor suggestions.

CI is failing right now -- we should get it green before merging. 

Lint error is just whitespace should be easy to fix.

Unit test error is in test_scripts for eval_policy. Not sure why this is being triggered here as we've not changed the code. It may just be a Gymnasium upgrade triggering it? If so, happy for you to do a hotfix for this in a separate PR, we can then get that hotfix merged then finally merge this PR. Can probably be resolved by changing `gym.make` to include appropriate `render_mode="rgb_array"`. 

> With these changes the benchmarking directory seems useless now as it just contains a util.py file to clean up the json files. We can move the util.py file somewhere else and delete the benchmarking directory.

Agreed probably best to move `util.py` somewhere else. There's not an obvious alternative location for `benchmarking/README.md` and we might want to include benchmarking table output etc in that directory so I'm OK keeping it around as just a stub for now but you could also move it to e.g. `BENCHMARK.md` at the top-level.

> I've tried to think of ways to remove this redundancy but couldn't figure a way out that preserves the command name airl and removes airl from the named config.

I think this is fine, people can type the extra 5 keystrokes :)

ernestum:
`test_experiments` is now also crashing under MacOS, so I guess we need to fix that one too?

ernestum (approved):
LGTM
The windows pipeline will be fixed here: https://github.com/HumanCompatibleAI/imitation/pull/811
Can you merge this @adam?

AdamGleave:
Merged -- well done @taufeeque9 for pushing this over the finish line, and thanks @ernestum for the detailed review!

Issue #651, opened 2023-01-06
Use huggingface dataset repositories for rollouts, not model repositories


ernestum:
## Problem
We recently added the feature to load rollouts from HuggingFace hub [here](https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/policies/serialize.py#L235). The new function assumes that rollouts are stored in a file `rollouts.npz` in the repository of a model.
Right now I see the following issues with this solution:

1. [major] HuggingFace model repositories should be used for models and not for datasets. There are specific dataset repositories in the HuggingFace hub and I think we should use them as such.
2. [minor] The function is located in the package `imitation.policies.serialize` while it has nothing to do with serializing policies.
3. [minor] The function assumes a `rollout.npz` files is stored with the model. Placing it there is mostly a manual process which is very brittle. Training and uploading models is at least semi-automated by the zoo.

Related to this is #408 where we half decided to consistently name rollouts "trajectories".

## Solution

1. Move the rollouts to [HuggingFace datasets](https://huggingface.co/docs/datasets/index) and change the function to load them from there. I am not sure if `.npz` files are supported right now.
2. Move it to something like `imitation.data.serialize`.
3. Automate the generation and upload of the rollouts in a similar way to how the zoo automates training and upload.

AdamGleave:
Thanks for flagging this! Didn't know about HuggingFace datasets, I agree we should switch to that.

Support automation in the form of "there's a command we can run to do this given a set of policies", but we may want to do quality control on the policies at least so probably can't be fully automated end-to-end.

ernestum:
In the long run I would like to transition to make the best usage of the existing HuggingFace Dataset infrastructure but I realized that this would required some major changes to the `imitation` library. I broke it down into a number of smaller steps which in itself already provide an improvement over the status quo. [_Scroll down for a diagram to illustrate the steps._](#Diagram-of-the-steps)


## Current State

Right now Experts generate list of `TrajectoryWithReward`s via the `rollout()` function. Those trajectories are fed to the imitation learning algorithms. To store the trajectories we concatenate them all to one long trajectory and store that alongside with the split positions in an numpy `.npz` file which is basically a dict of numpy arrays.
This `.npz` file is manually uploaded to a HuggingFace Model Repository. There are functions to automatically retrieve it from there and load it.


## Step 1: Using the HuggingFace Dataset Repository with the .npz format

A first step would be to move the `.npz` files  HuggingFace Dataset Repositories.
[...]

PR #677, opened 2023-02-09
Add support for HuggingFace Datasets

ernestum:
## Description

As per #651 we want to start storing trajectories in HuggingFace Datasets

This PR adds support for HuggingFace Datasets and moves `imitation.data.types.load/save` to `imitation.data.seralize.load/save`. Datasets are stored in the HuggingFace format by default.

The conversion script is adapted accordingly and I re-structured the tests.

I also moved the loading/saving functionality to `imitation.data.serialize` instead of `imitation.data.types` to mimic the layout of `imitation.policies`.

## Design Choices
I introduced a new class ([`imitation.data.huggingface_datasets_conversion.TrajectoryDatasetSequence`](https://github.com/HumanCompatibleAI/imitation/blob/6e82c9a51f0245021c65f3d098e77db4e99654ba/src/imitation/data/huggingface_datasets_conversion.py#L12)) that presents a HuggingFace dataset as a sequence of `imitation.data.types.Trajectory`. In its [`__getitem__` ](https://github.com/HumanCompatibleAI/imitation/blob/6e82c9a51f0245021c65f3d098e77db4e99654ba/src/imitation/data/huggingface_datasets_conversion.py#L34) method the dataset is queried and the result is converted to a trajectory (or slice of trajectories) ad-hoc.

I found that the "info" dicts were very heterogeneous in nature and could therefore not be mapped to some dataset layout. I decided to serialize the info dicts using `jsonpickle` and then just store a list of strings. Since decoding this is far more expensive than reading the other (potentially memory-mapped) features (observations, actions, terminals) and often we don't care about the "infos", I decided for a [lazy decoding wrapper](https://github.com/HumanCompatibleAI/imitation/blob/6e82c9a51f0245021c65f3d098e77db4e99654ba/src/imitation/data/huggingface_datasets_conversion.py#L60). It will only decode the info dicts when accessed and keep decoded dicts in an LRU cache to avoid decoding them multiple times.

## Testing

Old tests have been updated and split up into more atomic unit tests.


ernestum:
@AdamGleave Right now the tests fail due to lack of git lfs support in our Docker image. Is there any reason we don't use git lfs? If not, I would add it according to this guide: https://naiyer.dev/post/2020/09/05/using-git-lfs-in-ci/

ernestum:
Not sure what is going on with the coverage reports again :shrug:

AdamGleave:
@Rocamonde can you review this please?

src/imitation/algorithms/dagger.py
```diff
@@ -19,8 +19,9 @@
 from stable_baselines3.common.vec_env.base_vec_env import VecEnvStepReturn
 from torch.utils import data as th_data
 
+import imitation.data.serialize
 from imitation.algorithms import base, bc
-from imitation.data import rollout, types
+from imitation.data import rollout, serialize, types
```
Rocamonde, line 24: Why `import imitation.data.serialize` and also `from imitation.data import serialize`? Better to keep things consistent.

tests/scripts/test_scripts.py
```diff
@@ -1010,32 +1014,67 @@ def test_analyze_gather_tb(tmpdir: str):
...
-    """Tests that convert_trajs is idempotent and does not change the data."""
-    shutil.copy(OLD_FMT_ROLLOUT_TEST_DATA_PATH, tmpdir)
-    tmp_path = os.path.join(tmpdir, os.path.basename(OLD_FMT_ROLLOUT_TEST_DATA_PATH))
-    with open(tmp_path, "rb") as f:
-        pickle.load(f)  # check it's in pickle format to start with
-    args = ["convert_trajs.py", tmp_path]
-    with mock.patch.object(sys, "argv", args):
-        convert_trajs.main()
+def test_pickle_fmt_rollout_test_data_is_pickle():
+    # WHEN
```
Rocamonde, line 1018: Are we doing GWT for all tests now?

src/imitation/data/huggingface_datasets_conversion.py
```diff
@@ -0,0 +1,135 @@
...
+    # arbitrary cache size just to put a limit on memory usage
+    @functools.lru_cache(maxsize=100000)
+    def __getitem__(self, idx):
+        if isinstance(idx, slice):
+            return [jsonpickle.decode(info) for info in self._encoded_list[idx]]
+        else:
+            return jsonpickle.decode(self._encoded_list[idx])
+
+
+def trajectories_to_dataset(
```
Rocamonde, line 83: I guess this is beyond the scope of reviewing the PR, but why is this better than pickling a dataclass? A compelling case for this would be very relevant for other current projects.

src/imitation/data/serialize.py
```diff
@@ -0,0 +1,178 @@
...
+import logging
+import os
+import pathlib
+import warnings
+from typing import Mapping, Optional, Sequence, cast
+
+import datasets
+import numpy as np
+
+from imitation.data import huggingface_datasets_conversion as hfds
```
Rocamonde, line 11: Hmmm... Maybe have a shorter module name in the first place?

src/imitation/algorithms/dagger.py
ernestum: Good point. Fixed it.

tests/scripts/test_scripts.py
ernestum: Yes.

I think it makes tests a lot easier to grasp when reading them again.
It also pushes us to only test one concept at a time in each test. This makes it a lot easier to find the root cause when a test fails.

src/imitation/data/huggingface_datasets_conversion.py
ernestum: Valid question I think! For me there are the following points:

- It is generally not a good idea to serialize classes for persistence. Whenever you change the class implementation, you might (silently) loose the ability to load older files. I had this happen when loading SB3 models trained with old gym versions (with gym changing implementations with no regard to backwards compatibility)
- The pickle format changes from time to time and sometimes I have to mess with pickle versions.
- The HuggingFace Datasets are backed by Apache Arrow. This is mostly relevant in case our datasets grow. It allows to stream datasets (from disk or from the internet) and does memory mapping for us. With the memory mapping we can load huge files that would never fit in RAM and let the OS do transparent caching for us.
- The HuggingFace datasets do not support pickled objects by default (probably due to the above reasons and to make users adhere to best practices). We could make it happen with a additional loading script next to the pickled classes but this would increase complexity and I really don't like storing code together with our datasets.

You can read more about the considerations in #651

src/imitation/data/serialize.py
ernestum: We could go with `hf_datasets_conversion`? I am happy about other suggestions but I would like the name to stay descriptive. I rather type some more letters now than wonder what a module was supposed to do later.

Rocamonde: I might call it huggingface_data_converter or something like that? I think having a very descriptive name but then renaming the import to a non-obvious abbreviation is probably just as bad for redability

ernestum: Hmm. I am not really happy with that name either. I would expect a class named `HuggingFaceDataConverter` in such a module.

When it is so hard to come up with a good name for a module that is often an indicator for bad architecture of the module. So I looked at its content again and the following came up:

When saving, we create a HF dataset just to write it to disk immediately. This involves a (shallow) copy of the data.

When loading, we have a wrapper that makes a HF dataset visible as a sequence of trajectories. Trajectory objects generated on-the-fly.

To make this symmetrical, I should have used the `Dataset.from_generator` method instead of the `Dataset.from_dict` method with a generator that constructs dicts from trajectories on-the-fly ([documentation here](https://huggingface.co/docs/datasets/v2.10.0/en/create_dataset#from-local-files)).
This is also the most memory-efficient way to create datasets and it would make it possible to stream sampled trajectories straight to disk.
I will push a version of this soon.

Concerning the naming of the module: I would propose to just call it `huggingface_utils` or even `hf_utils`. In symmetry to that we could refactor `imitation.policies.serialize` and pull out most of the HF-specific code in a `imitation.policies.huggingface_utils` module.

Rocamonde: That sounds like a good decision, thanks for taking a look at this! Huggingface_utils sounds good.

ernestum: Added it just now. Unfortunately this change made the save a lot slower (`test_types.test_save_trajectories` executes orders of magnitude slower). Maybe here @simoninithomas can give us some insight?

simoninithomas: So I asked the dataset team, that can explain the problem:
- `from_dict` writes in RAM
- `from_generator` writes on disk

ernestum: So should I better switch between `from_dict` and `from_generator` based on the size of the dataset (assuming that I know the number of elements coming out of the generator) or is that something, that you maybe implement on the datasets library end?

simoninithomas: If your dataset is a python dict, it takes up RAM and is therefore relatively small so you can load it with from_dict. But if you both load files one by one, then from_generator is better

ernestum:
I finally decided to go without the `from_generator` function and only use it when we actually need it. Right now our datasets are small enough to fit in memory.

@Rocamonde could you give this another pass and then we can merge it?

Rocamonde:
Sure! Will take a look tomorrow.

src/imitation/scripts/convert_trajs.py
```diff
@@ -1,45 +1,56 @@
...
-(i.e. "A.pkl" -> "A.npz", "A.npz" -> "A.npz", "A" -> "A.npz", "A.foo" -> "A.foo.npz").
+in the old .pkl or .npz format. It then saves each sequence in the new HuggingFace
+datasets format. The path is the same as the original but a directory without an
+extension (i.e. "A.pkl" -> "A/", "A.npz" -> "A/", "A/" -> "A/", "A.foo" -> "A/").
 """
-
+import pathlib
 import warnings
 
+import imitation.data.serialize
```
Rocamonde, line 14: In other places we're importing `from imitation.data import serialize`

src/imitation/data/serialize.py
```diff
@@ -0,0 +1,172 @@
+"""Serialization utilities for trajectories."""
```
Rocamonde, line 1: I understand you just moved the code into a separate file? Or are there any other additional changes? Hard to see based on the diff github is showing

Rocamonde: Otherwise all seems OK

tests/data/test_types.py
```diff
@@ -43,7 +43,7 @@ def trajectory(
         pytest.skip()
     obs = np.array([obs_space.sample() for _ in range(length + 1)])
     acts = np.array([act_space.sample() for _ in range(length)])
-    infos = np.array([{i: i} for i in range(length)])
+    infos = np.array([{f"key{i}": i} for i in range(length)])
```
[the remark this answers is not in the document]
ernestum: This is a test fixture for trajectories.
Before the test trajectories included info dicts whose keys were `int`s . This change ensures we only have dicts with `str` typed keys since
1. I don't think the infos dicts ever have any other key type except strings.
2. The current implementation relies on jsonpickle to serialize the info dicts. Jsonpickle would convert those int keys to strings anyways:
```python
>> jsonpickle.loads(jsonpickle.dumps({1: 1}))
{'1': 1}
```
This would make the tests fail that store a trajectory, load it again and compare the two. The original would have `int` keys while the loaded one would have `str` keys.

src/imitation/scripts/convert_trajs.py
ernestum: Good point. Even found some more places where this is the case. I made sure it is always imported as `from imitation.data import serialize` or `from imitation import data` whenever we also need the `policies.serialize` module in the same file.

src/imitation/data/serialize.py
ernestum: I moved the serialization related functions to a separate file but also changed the save/load functions as they now use the huggingface backend. Sorry I also cant find a way to generate a decent diff for this.

Rocamonde (approved):
LGTM

src/imitation/scripts/convert_trajs.py
```diff
@@ -1,45 +1,56 @@
...
+        warnings.warn(f"File {path} is already in the new format. Skipping.")
+        return path
+    else:
+        converted_path = path.with_suffix("")
+        serialize.save(converted_path, trajs)
+        return converted_path
 
 
-def main():
+def main():  # pragma: no cover
```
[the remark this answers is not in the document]
ernestum: The main is not included, `update_traj_file_in_place` however is.

AdamGleave:
Anything blocking merging now we have an LGTM? Is it just the code coverage?

ernestum:
What is missing in coverage right now (as I interpret the codecov output)

- loading old versions of the `Trajectory` dataclass which did not have the `terminal` flag yet.
- specifying the log-level with something that can not be cast to an `int`

If you think we can live without that coverage then you can merge this please.

AdamGleave:
Sorry for the slow turnaround -- yes, I think that coverage isn't essential, I have merged now.

PR #683, opened 2023-02-21
Enable git lfs

ernestum:
## Description

For some testdata we rather want to store it using git LFS. We need this for the new tests in #677 but I decided to pull this out to ensure there is no scope-creep in #677.
I also updated the cuda version along the way.
This is modeled mostly after https://naiyer.dev/post/2020/09/05/using-git-lfs-in-ci/

## Testing

I initially based this on top of `huggingface_datasets` branch because I needed it there. The pipeline there succeeded (with some LFS objects in the history).

- [In this job you can see the cache miss](https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3362/workflows/8ee9d338-8ebc-4acc-be99-9d85d025adaa/jobs/12473)
- [In this job you can see the cache hit working](https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3362/workflows/7cb191a1-8cfb-4393-9c8a-130e9f81207f/jobs/12480) (no more git lfs objects are pulled)

## Before Merging
We need to change the image from `base-alpha` to `base`!!!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,7 @@ executors:
     environment:
       # more CPUs visible but we're throttled to 8, which breaks auto-detect
       NUM_CPUS: 8
+      GIT_LFS_SKIP_SMUDGE: 1
   static-analysis-xlarge:
     <<: *defaults
     # darglint is slow enough that we benefit from xlarge even for linting.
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -20,6 +20,7 @@ executors:
     environment:
       # more CPUs visible but we're throttled to 8, which breaks auto-detect
       NUM_CPUS: 8
+      GIT_LFS_SKIP_SMUDGE: 1
```
AdamGleave, line 23: add comment explaining why this is needed? (I had to Google it)

AdamGleave (approved):
LGTM -- 1 small comment

.circleci/config.yml
ernestum: Good point. I added something to explain and a link with further details.

PR #684, opened 2023-02-21
add kwargs in reset

AdrianP-:
## Description

In those cases that it's necessary to set a seed to maintain a state, the RolloutInfoWrapper loses the seed and doesn't work the env

## Testing

Description of how you've tested your changes.

PR #695, opened 2023-03-24
Add kwargs in reset.

ernestum:
## Description

This is a duplicate of #684 which we can not solve due to technical issues with circleCI and @AdrianP- not responding.


AdamGleave (approved):
LGTM

PR #704, opened 2023-04-28
Add short tutorial about trajectories

ernestum:
## Description

I noticed that the handling trajectories is under-documented.

## Testing

Only documentation changes.

### DIFF
--- a/docs/tutorials/trajectories.rst
+++ b/docs/tutorials/trajectories.rst
@@ -36,7 +36,7 @@ which is an array of rewards of shape `(trajectory_len, )`.
 
 Usually, they are passed around as sequences of trajectories.
 
-Some algorithms do not care about entire trajectories, but only for individual
+Some algorithms do not need as much information about the ordering of states, actions and rewards. Rather than using trajectories, these algorithms can make use of individual
 :class:`Transitions <imitation.data.types.Transitions>` (:func:`flattened <imitation.data.rollout.flatten_trajectories>` trajectories).
 
 Generating Trajectories


ernestum:
This are just documentation changes. I don't get where those coverage changes are coming from. Maybe Codecov changed the way they count the coverage?

What is not covered any more (according to Codecov) in `bc.py`:
- The case where the batch size is not a multiple of the minibatch size 
- `on_epoch_end()` callbacks
- The case where we have an incomplete batch at the end of training

docs/tutorials/trajectories.rst
```diff
@@ -0,0 +1,67 @@
...
+        your_policy,
+        your_env,
+        sample_until=make_sample_until(timesteps=1000),
+        rng=np.random.default_rng(),
+    )
+
+Storing/Loading Trajectories
+----------------------------
+
+Trajectories can be stored on disk or uploaded to the HuggingFace Dataset Hub.
```
timbauman, line 50: It looks like this code shows how to save to disk not to a HF dataset?

docs/tutorials/trajectories.rst
```diff
@@ -0,0 +1,67 @@
+==========================
+Handle Expert Trajectories
+==========================
+
+For imitation learning we need expert trajectories.
```
timbauman, line 5: I think for e.g. preference comparison non-expert trajectories are stored with this same data structure right? And adversarial methods use this for the non-expert policy to train/evaluate against the discriminator?

timbauman (commented):
Hey! I'm Tim, a work trialer with FAR. @AdamGleave asked me to take a look at your PR - I'm new to the codebase so some of my comments or questions might be wrong or too detailed since I'm still learning

docs/tutorials/trajectories.rst
```diff
@@ -0,0 +1,67 @@
...
+
+    @dataclasses.dataclass(frozen=True)
+    class Trajectory:
+        obs: np.ndarray
+        """Observations, shape (trajectory_len + 1, ) + observation_shape."""
+
+        acts: np.ndarray
+        """Actions, shape (trajectory_len, ) + action_shape."""
+
+        infos: Optional[np.ndarray]
```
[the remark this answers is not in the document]
timbauman: Or document it in the code

docs/tutorials/trajectories.rst
ernestum: Good point! I will adjust the wording to reflect this.

docs/tutorials/trajectories.rst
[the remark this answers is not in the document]
ernestum: It is called `info` just like the `info` dict returned by the [`step()` method of the environments](https://gymnasium.farama.org/api/env/#gymnasium.Env.step). I totally agree that the name is not very descriptive, but I still think it is a good idea to keep the same name as the gym API dictates.
In fact, the `info` dict is used for all kinds of information that needs to be funneled through the gym API. Therefore it can not really be any more descriptive. It is a questionable part of the gym API ...

I ensured that the shape is defined in the same way as for the fields above, and in code added some more detailed explanation.

docs/tutorials/trajectories.rst
```diff
@@ -0,0 +1,67 @@
...
+:class:`TrajectoryWithRew <imitation.data.types.TrajectoryWithRew>` have another :class:`rews <imitation.data.types.TrajectoryWithRew.rews>` field, which is an array of rewards of shape `(trajectory_len, )`.
+
+Usually, they are passed around as sequences of trajectories.
+
+Some algorithms do not care about entire trajectories, but only for individual
+:class:`Transitions <imitation.data.types.Transitions>` (:func:`flattened <imitation.data.rollout.flatten_trajectories>` trajectories).
+
+Generating Trajectories
+-----------------------
+To generate trajectories from a given (expert) policy, run the following command:
```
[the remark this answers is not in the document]
ernestum: Yes. For example, when we train a policy and want to evaluate it, we generate some trajectories and compute their rewards. I took out the mention of the expert here.

docs/tutorials/trajectories.rst
ernestum: This code shows how to save to disk as a HuggingFace Dataset.
#705 contains instructions on how to upload/download to/from the HuggingFace Hub.

docs/index.rst
```diff
@@ -74,6 +74,7 @@ If you use ``imitation`` in your research project, please cite our paper to help
    tutorials/5a_train_preference_comparisons_with_cnn
    tutorials/6_train_mce
    tutorials/7_train_density
+   tutorials/trajectories
```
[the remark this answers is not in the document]
ernestum: You are right. I reviewed the overall structure of the documentation and I am not happy with it.
Let's keep the tutorial in here for now and postpone the restructuring to #715

docs/tutorials/trajectories.rst
```diff
@@ -0,0 +1,72 @@
+==========================
+Handle Expert Trajectories
+==========================
+
+For imitation learning we need trajectories.
```
Rocamonde, line 5: I guess it depends on how much you want to explain this / who the audience is, but briefly explaining what a trajectory is could be a good start.

docs/tutorials/trajectories.rst
```diff
@@ -0,0 +1,72 @@
...
+    @dataclasses.dataclass(frozen=True)
+    class Trajectory:
+        obs: np.ndarray
+        """Observations, shape (trajectory_len + 1, ) + observation_shape."""
+
+        acts: np.ndarray
+        """Actions, shape (trajectory_len, ) + action_shape."""
+
+        infos: Optional[np.ndarray]
+            """An array of info dicts, shape (trajectory_len, )."""
```
Rocamonde, line 21: Either link to what info dicts are for, or briefly explain. This is confusing to someone who doesn't fully understand gym already

src/imitation/data/types.py
```diff
@@ -52,7 +52,12 @@ class Trajectory:
     """Actions, shape (trajectory_len, ) + action_shape."""
 
     infos: Optional[np.ndarray]
-    """An array of info dicts, length trajectory_len."""
+    """An array of info dicts, shape (trajectory_len, ).
+
+    The info dict is returned by some environments `step()` and contains auxiliary
```
Rocamonde, line 57: Useful that you added it here, but this is not visible in the tutorial, so maybe tell the user to check these docs at least?

docs/tutorials/trajectories.rst
Rocamonde, line 1: I wouldn't mind seeing something discussing whether these trajectory objects can only be used for imitation algorithms or can also be used for stable baselines3 or offline RL algorithms, and how. I have been using a lot of imitation primitives, including the trajectories object, in projects where I don't actually do any imitation learning, just because the helper tools and data types are quite good.

docs/tutorials/trajectories.rst
```diff
@@ -0,0 +1,72 @@
...
+
+:class:`TrajectoryWithRew <imitation.data.types.TrajectoryWithRew>` is a subclass of
+:class:`Trajectory <imitation.data.types.Trajectory>` and has another
+:class:`rews <imitation.data.types.TrajectoryWithRew.rews>` field,
+which is an array of rewards of shape `(trajectory_len, )`.
+
+Usually, they are passed around as sequences of trajectories.
+
+Some algorithms do not care about entire trajectories, but only for individual
+:class:`Transitions <imitation.data.types.Transitions>` (:func:`flattened <imitation.data.rollout.flatten_trajectories>` trajectories).
```
levmckinney, line 34:
```suggestion
Some algorithms do not need as much information about the ordering of states, actions and rewards. Rather than using trajectories, these algorithms can make use of individual
:class:`Transitions <imitation.data.types.Transitions>` (:func:`flattened <imitation.data.rollout.flatten_trajectories>` trajectories).
```

docs/tutorials/trajectories.rst
```diff
@@ -0,0 +1,72 @@
...
+
+Generating Trajectories
+-----------------------
+To generate trajectories from a given policy, run the following command:
```
levmckinney, line 38: Might be good to explain what `sample_until` does here i.e. does it sample complete episodes or does it cut off the last episode.

docs/tutorials/trajectories.rst
ernestum: Good point. I added a short explanation.

docs/tutorials/trajectories.rst
ernestum: Good point! I linked to the reference documentation of Trajectory and to the gymnasium documentation.

docs/tutorials/trajectories.rst
ernestum: I am happy to include this here but I have never used the imitation primitives outside the library so I don't feel like I have a lot to say. If you could tell me the main points that I should include I will turn them into a coherent text.

docs/tutorials/trajectories.rst
ernestum: Good point. The kwarg should have been `min_timesteps` instead of `timesteps`. Then it is more clear.

I don't want to get lost in details in this high-level tutorial so decided to just go with `min_episodes`. Then the questions does not even come up when reading the example code.

docs/tutorials/trajectories.rst
ernestum: I moved this to a new issue so we can deal with it in a separate PR #719

Rocamonde (approved):
LGTM

PR #706, opened 2023-05-02
Fixing pipeline on master

ernestum:
## Description

The pipeline on master is failing right now. Issues, that I have identified and resolved in this PR are:

- I invalidated the pytest cache
- There was an issue with removing ad-hoc generated demonstration trajectories that only occurs in certain edge cases due to concurrency of the test suite. It was trying to delete a file using `path.unlink()` which did not work any more since the new trajectory format uses folders.
- Somehow the cartpole expert in the testdata directory could not be loaded any more. I suspect due to changes in SB3. I re-trained it and added a new version and now that works again.

## Notes

Right now the windows pipeline crashes because there are issues with setting up the venv. I think the script in `ci/build_and_activate_venv.ps1` is for not executed since the venv directory already exists (not sure about this though). When I try to run it manually using ssh it crashes anyways and leaves the venv in a half-initialized state.


AdamGleave (approved):
LGTM, @ernestum please review as I also made some minor changes to fix the CI and then feel free to merge if happy.

AdamGleave:
Any idea what's causing the coverage drop @ernestum?

ernestum:
I am not sure and the codecov output is a bit confusing. I am astonished, that there are any coverage differences at all. When opening the coverage report, it only seems to include the last commit.

I suspect we stopped covering the cases where we load old-format trajectories? 
But then again I expect this case to be covered by [this test](https://github.com/HumanCompatibleAI/imitation/blob/bbdcb2989d807b343a7da1d647358fc436458991/tests/scripts/test_scripts.py#L1049)

Then there are the sacred configuration functions which (partially) do not seem to be caught by coverage.
Maybe this has something to do with the configuration magic that sacred performs?

AdamGleave:
Strange, I agree the changes shouldn't cause these coverage differences. CodeCov does seem to be glitchy as only listing one commit. I'm going to just override this check and merge, the coverage change is not that large anyway.

Issue #413, opened 2022-02-28
Add support for stateful policies


AdamGleave:
`rollout.PolicyCallable` takes an observation and outputs an action. This only supports stateless policies. By contrast, `BasePolicy.predict` takes an observation, mask (is it terminal observation) and state (which is reset when mask is true), supporting stateful policies like LSTMs.

We should make `PolicyCallable` also task `mask` and `state` to be more flexible. This would be a fairly easy change to make, as its only used at a handful of call sites.

In particular, taking rolllouts of `TabularPolicy` always uses the timestep-0 policy because of this limitation.

PR #709, opened 2023-05-08
Support for stateful policies

timbauman:
## Description

Add support for stateful policies

https://github.com/HumanCompatibleAI/imitation/issues/413

## Testing

Unit tests pass, plus new unit test.

The notebook at docs/tutorials/6_train_mce.ipynb learns a slightly different reward function - the expert policies are found with the soft bellman backup, which has a more variable behavior in later time steps. @levmckinney said this is likely due to the fact that there's less future reward so the algorithm focuses more on maximizing entropy. Over longer time horizons he said the old and new policies should be more similar, so I set the horizon to 40. Both learned reward functions changed, but they are more similar (and the new policy is less variable)

Without state (horizon 8):
image

With state (horizon 8):

ex 1
image

ex 2
image

ex 3
image

Without state (horizon 40):
image

With state (horizon 40):

image

### DIFF
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -263,8 +263,9 @@ def make_sample_until(
     return sample_until
 
 
-# A PolicyCallable is a function that takes an array of observations
-# and returns an array of corresponding actions.
+# A PolicyCallable is a function that takes an array of observations, an optional
+# array of states, and an optional array of episode starts and returns an array of
+# corresponding actions.
 PolicyCallable = Callable[
     [np.ndarray, Optional[Tuple[np.ndarray, ...]], Optional[np.ndarray]],
     Tuple[np.ndarray, Optional[Tuple[np.ndarray, ...]]],
@@ -303,7 +304,7 @@ def policy_to_callable(
             assert isinstance(policy, (BaseAlgorithm, BasePolicy))
             # pytype doesn't seem to understand that policy is a BaseAlgorithm
             # or BasePolicy here, rather than a Callable
-            (acts, states) = policy.predict( # pytype: disable=attribute-error
+            (acts, states) = policy.predict(  # pytype: disable=attribute-error
                 observations,
                 state=states,
                 episode_start=episode_starts,
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -300,8 +300,6 @@ def policy_to_callable(
             state: Optional[Tuple[np.ndarray, ...]],
             episode_start: Optional[np.ndarray],
         ) -> Tuple[np.ndarray, Optional[Tuple[np.ndarray, ...]]]:
-            # pytype doesn't seem to understand that policy is a BaseAlgorithm
-            # or BasePolicy here, rather than a Callable
             assert isinstance(policy, (BaseAlgorithm, BasePolicy))
             (acts, state) = policy.predict(
                 observation,
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -301,7 +301,9 @@ def policy_to_callable(
             episode_start: Optional[np.ndarray],
         ) -> Tuple[np.ndarray, Optional[Tuple[np.ndarray, ...]]]:
             assert isinstance(policy, (BaseAlgorithm, BasePolicy))
-            (acts, state) = policy.predict(
+            # pytype doesn't seem to understand that policy is a BaseAlgorithm
+            # or BasePolicy here, rather than a Callable
+            (acts, state) = policy.predict( # pytype: disable=attribute-error
                 observation,
                 state=state,
                 episode_start=episode_start,
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -282,12 +282,12 @@ def policy_to_callable(
     if policy is None:
 
         def get_actions(
-            observation: np.ndarray,
-            state: Optional[Tuple[np.ndarray, ...]],
-            episode_start: Optional[np.ndarray],
+            observations: np.ndarray,
+            states: Optional[Tuple[np.ndarray, ...]],
+            episode_starts: Optional[np.ndarray],
         ) -> Tuple[np.ndarray, Optional[Tuple[np.ndarray, ...]]]:
-            acts = [venv.action_space.sample() for _ in range(len(observation))]
-            return np.stack(acts, axis=0), None  # is None the right state?
+            acts = [venv.action_space.sample() for _ in range(len(observations))]
+            return np.stack(acts, axis=0), None
 
     elif isinstance(policy, (BaseAlgorithm, BasePolicy)):
         # There's an important subtlety here: BaseAlgorithm and BasePolicy
@@ -296,20 +296,20 @@ def policy_to_callable(
         # (which would call .forward()). So this elif clause must come first!
 
         def get_actions(
-            observation: np.ndarray,
-            state: Optional[Tuple[np.ndarray, ...]],
-            episode_start: Optional[np.ndarray],
+            observations: np.ndarray,
+            states: Optional[Tuple[np.ndarray, ...]],
+            episode_starts: Optional[np.ndarray],
         ) -> Tuple[np.ndarray, Optional[Tuple[np.ndarray, ...]]]:
             assert isinstance(policy, (BaseAlgorithm, BasePolicy))
             # pytype doesn't seem to understand that policy is a BaseAlgorithm
             # or BasePolicy here, rather than a Callable
-            (acts, state) = policy.predict( # pytype: disable=attribute-error
-                observation,
-                state=state,
-                episode_start=episode_start,
+            (acts, states) = policy.predict( # pytype: disable=attribute-error
+                observations,
+                state=states,
+                episode_start=episode_starts,
                 deterministic=deterministic_policy,
             )
-            return acts, state
+            return acts, states
 
     elif callable(policy):
         # When a policy callable is passed, by default we will use it directly.
--- a/src/imitation/data/rollout.py
+++ b/src/imitation/data/rollout.py
@@ -12,6 +12,7 @@ from typing import (
     Mapping,
     Optional,
     Sequence,
+    Tuple,
     Union,
 )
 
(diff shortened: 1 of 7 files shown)


src/imitation/policies/exploration_wrapper.py
```diff
@@ -53,9 +55,14 @@ def __init__(
...
-    def _random_policy(self, obs: np.ndarray) -> np.ndarray:
+    def _random_policy(
+        self,
+        obs: np.ndarray,
+        state: Optional[Tuple[np.ndarray, ...]],
+        episode_start: Optional[np.ndarray],
+    ) -> Tuple[np.ndarray, Optional[Tuple[np.ndarray, ...]]]:
         acts = [self.venv.action_space.sample() for _ in range(len(obs))]
-        return np.stack(acts, axis=0)
+        return np.stack(acts, axis=0), None  # is this right?
```
timbauman, line 65: This feels like a weird case. If you have a stateful policy, it'll randomly lose all state every time exploration happens. On the other hand, if you just pass the state through this policy, it'll behave strangely since the world state will have changed without the hidden state having a chance to catch up. I think None is an okay solution - or, we could throw or warn if the state passed to the random policy is not None to indicate that the behavior is undefined?

timbauman: Or we could simply always pass an empty state to the wrapped policy

levmckinney: Stateful polices and exploration don't seem like they go together very well. I'm struggling to think of a single example of an off policy stateful algorithm that is used in practice. This should definitely have some kind of warning, maybe even an error. I would bump this one up to @AdamGleave.

timbauman: Updated to throw. (Talked with @AdamGleave offline and he agrees)

src/imitation/data/rollout.py
```diff
@@ -274,26 +278,38 @@ def policy_to_callable(
...
-        def get_actions(states):
-            acts = [venv.action_space.sample() for _ in range(len(states))]
-            return np.stack(acts, axis=0)
+        def get_actions(
+            observation: np.ndarray,
+            state: Optional[Tuple[np.ndarray, ...]],
+            episode_start: Optional[np.ndarray],
+        ) -> Tuple[np.ndarray, Optional[Tuple[np.ndarray, ...]]]:
+            acts = [venv.action_space.sample() for _ in range(len(observation))]
+            return np.stack(acts, axis=0), None  # is None the right state?
```
levmckinney, line 290: Yes the random agent isn't stateful.

src/imitation/data/rollout.py
```diff
@@ -274,26 +278,38 @@ def policy_to_callable(
...
 
     elif isinstance(policy, (BaseAlgorithm, BasePolicy)):
         # There's an important subtlety here: BaseAlgorithm and BasePolicy
         # are themselves Callable (which we check next). But in their case,
         # we want to use the .predict() method, rather than __call__()
         # (which would call .forward()). So this elif clause must come first!
 
-        def get_actions(states):
+        def get_actions(
```
levmckinney, line 298: nitpick: this should be `observations`, `states` and `episode_starts` if we want to be consistent.

src/imitation/policies/exploration_wrapper.py
```diff
@@ -64,8 +71,20 @@ def _switch(self) -> None:
...
+        observation: np.ndarray,
+        state: Optional[Tuple[np.ndarray, ...]],
+        _episode_start: Optional[np.ndarray],
+    ) -> Tuple[np.ndarray, Optional[Tuple[np.ndarray, ...]]]:
+        if state is not None:
+            raise ValueError("Exploration wrapper does not support stateful policies.")
+
+        acts, state = self.current_policy(observation, None, None)
+
+        if state is not None:
```
levmckinney, line 85: Do we need both of these checks?

levmckinney (approved):
Some very minor nitpicks. Overall this looks good.

src/imitation/policies/exploration_wrapper.py
timbauman: I guess technically you could only use one, but they are checking different things. Probably the check for the return value is more critical. But I'll add comments explaining the rationale for having both for now

src/imitation/data/rollout.py
```diff
@@ -264,7 +265,10 @@ def sample_until(trajs: Sequence[types.TrajectoryWithRew]) -> bool:
 
 # A PolicyCallable is a function that takes an array of observations
```
AdamGleave, line 266: Comment is inconsistent with the new definition

src/imitation/data/rollout.py
```diff
@@ -274,26 +278,38 @@ def policy_to_callable(
     deterministic_policy: bool = False,
 ) -> PolicyCallable:
     """Converts any policy-like object into a function from observations to actions."""
+    get_actions: PolicyCallable
```
AdamGleave, line 281: :+1:

src/imitation/policies/exploration_wrapper.py
```diff
@@ -53,9 +55,14 @@ def __init__(
         # Choose the initial policy at random
         self._switch()
 
-    def _random_policy(self, obs: np.ndarray) -> np.ndarray:
+    def _random_policy(
+        self,
+        obs: np.ndarray,
+        _state: Optional[Tuple[np.ndarray, ...]],
+        _episode_start: Optional[np.ndarray],
+    ) -> Tuple[np.ndarray, Optional[Tuple[np.ndarray, ...]]]:
```
AdamGleave, line 63: Nit: our style guide https://google.github.io/styleguide/pyguide.html#214-decision favors deleting variables to suppress unused variable warnings.

src/imitation/policies/exploration_wrapper.py
AdamGleave: It definitely *looks* redundant but checking both input and return value will catch errors faster and more localized (although I struggle to think of cases where an error wouldn't be caught on the next step even if you omitted one check). Adding comments to justify this sounds good. You could also consider modifying the error message a bit between them to emphasize where the non-None state is coming from.

tests/algorithms/test_mce_irl.py
```diff
@@ -314,6 +314,31 @@ def test_tabular_policy(rng):
     np.testing.assert_equal(timesteps[0], 2 - mask.astype(int))
 
 
+def test_tabular_policy_rollouts(rng):
+    """Tests tabular policy prediction, especially timestep calculation and masking."""
```
AdamGleave, line 318: What the test is doing is a bit hard to follow, consider expanding docstring, adding comments or otherwise clarifying?

My understanding (but it took a few minutes for me to work out) is:
  - `s_a` always takes the same action (deterministically) regardless of state, and alternates round-robin between the three actions.
  - `pi` is `s_a` tiled 7 times to get up to 21 timesteps, greater than the horizon of `20` of `ReasonablePOMDP`.
  - We then check that the actions taken in fact do alternate round robin as we'd expect from the construction.

tests/algorithms/test_mce_irl.py
```diff
@@ -314,6 +314,31 @@ def test_tabular_policy(rng):
...
+    state_space = gym.spaces.Discrete(5)
+    action_space = gym.spaces.Discrete(3)
+
+    s_a = np.stack([np.eye(3)] * 5, axis=1)
```
AdamGleave, line 322: Can we make it clear where `3` and `5` come from? e.g:
```suggestion
    s_a = np.stack([np.eye(action_space.n)] * state_space.n, axis=1)
```
or define `num_states`, `num_actions`

tests/algorithms/test_mce_irl.py
```diff
@@ -314,6 +314,31 @@ def test_tabular_policy(rng):
...
+    pi = np.concatenate([s_a] * 7, axis=0)
```
AdamGleave, line 323: I think this does the same and seems more canonical? (I don't particularly care either way though.)
```suggestion
    pi = np.repeat(s_a, 7, axis=0)
```

tests/algorithms/test_mce_irl.py
```diff
@@ -314,6 +314,31 @@ def test_tabular_policy(rng):
...
+    mdp = ReasonablePOMDP()
+    state_env = base_envs.ExposePOMDPStateWrapper(mdp)
+    state_venv = vec_env.DummyVecEnv([lambda: state_env])
+    trajs = rollout.generate_trajectories(
+        tabular,
+        state_venv,
+        sample_until=rollout.make_min_episodes(1),
+        rng=rng,
+    )
+    assert (trajs[0].acts == pi[:, 0, :].nonzero()[1][: len(trajs[0].acts)]).all()
```
AdamGleave, line 339: I wrote this to try and understand it, including in case it helps with clarifying test although would not recommend using this verbatim:
```suggestion
    # pi[t,s,a] is the same for every state, so drop that dimension
    expected_actions_onehot = pi[:, 0, :]
    # take the index in a (drop the indexes in t)
    expected_actions_index = expected_actions_onehot.nonzero()[1]
    assert (trajs[0].acts == expected_actions_index[: len(trajs[0].acts)]).all()
```

tests/policies/test_exploration_wrapper.py
```diff
@@ -1,14 +1,19 @@
...
 import seals  # noqa: F401
 
 from imitation.policies import exploration_wrapper
 from imitation.util import util
 
 
-def constant_policy(obs):
-    return np.zeros(len(obs), dtype=int)
+def constant_policy(obs, state, mask):
+    return np.zeros(len(obs), dtype=int), None
```
AdamGleave, line 12:
```suggestion
    del state, mask  # Unused
    return np.zeros(len(obs), dtype=int), None
```

tests/policies/test_exploration_wrapper.py
```diff
@@ -1,14 +1,19 @@
...
+
+
+def fake_stateful_policy(obs, state, mask):
+    return np.zeros(len(obs), dtype=int), (np.zeros(1),)
```
AdamGleave, line 16:
```suggestion
    del state, mask  # Unused
    return np.zeros(len(obs), dtype=int), (np.zeros(1),)
```

AdamGleave (approved):
LGTM, some minor changes requested but no need for a re-review.

AdamGleave:
You also have some lint errors (which were previously obscured by the whole CI pipeline being broken :smiling_face_with_tear: but that's fixed now in #706)

tests/algorithms/test_mce_irl.py
timbauman: Updated - hopefully it's clearer now

Issue #711, opened 2023-05-09
Support simple, synchronous, CLI or Jupyter human preference collection


timbauman:
## Problem

Today only synthetic preferences are supported. It would be great to support real human preferences.

## Solution

Requirements:
- record videos of trajectories
- ideally, extensible so we could factor out into a more elaborate, asynchronous service someday
- simple enough that it can be built in a day

MVP:
- In a Jupyter notebook or CLI
- Synchronous

Steps:
- Inject option to store videos into training code
- Build new interface/class for storing videos (so we could configure this to store in different directories or on the cloud, for example)
- Build new gatherer that requests user feedback
- Display videos to users in Jupyter or CLI
- Clean up watched/unneeded videos
- Build demo notebook
- Integrate into training script and test end to end

## Possible alternative solutions

Slightly more than MVP (out of scope for this issue):

- Refactor to support asynchronous preference gathering
- Separate requesting preferences from receiving them
- Periodically retrain with new preferences
- Would be nice to have a way to indicate that new preferences are available
- Would require changing fragmenter possibly? E.g. "just show the most recent pair to the user" rather than all pairs
- Build asynchronous

PR #712, opened 2023-05-11
Add support for a simple preference UI

timbauman:
## Description

See #711 

## Testing

TODO: add notebook and experiment config that use this feature, and screenshots of behavior. (I've tested myself but not in a clean way.)

### DIFF
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -5,6 +5,7 @@ between trajectory fragments.
 """
 import abc
 import math
+import os
 import pathlib
 import pickle
 import re
@@ -1103,9 +1104,12 @@ class SynchronousHumanGatherer(PreferenceGatherer):
 
     def _in_ipython(self) -> bool:
         try:
-            return get_ipython().__class__.__name__ == "ZMQInteractiveShell"  # type: ignore[name-defined] # noqa
+            return self.is_running_pytest_test() or get_ipython().__class__.__name__ == "ZMQInteractiveShell"  # type: ignore[name-defined] # noqa
         except NameError:
             return False
+
+    def is_running_pytest_test(self) -> bool:
+        return "PYTEST_CURRENT_TEST" in os.environ
 
 
 class PreferenceDataset(data_th.Dataset):
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -940,7 +940,6 @@ class SynchronousCLIGatherer(PreferenceGatherer):
             A numpy array of 1 if fragment 1 is preferred and 0 otherwise, with shape
             (b, ), where b is the length of the input
         """
-
         preferences = np.zeros(len(fragment_pairs), dtype=np.float32)
         for i, (frag1, frag2) in enumerate(fragment_pairs):
             if self._display_videos(frag1, frag2):
@@ -948,22 +947,34 @@ class SynchronousCLIGatherer(PreferenceGatherer):
         return preferences
 
     def _display_videos(
-        self, frag1: TrajectoryWithRew, frag2: TrajectoryWithRew
+        self, frag1: TrajectoryWithRew, frag2: TrajectoryWithRew,
     ) -> bool:
         """Displays the videos of the two fragments.
 
         Args:
             frag1: first fragment
             frag2: second fragment
+        
+        Returns:
+            True if the first fragment is preferred, False if not.
+        
+        Raises:
+            KeyboardInterrupt: if the user presses q to quit.
+            RuntimeError: if the video files cannot be opened.
+            ValueError: if the trajectory infos are not set.
         """
         # display the videos
+        if frag1.infos is None or frag2.infos is None:
+            raise ValueError(
+                "TrajectoryWithRew.infos must be set to display videos.",
+            )
         frag1_video_path = frag1.infos[0]["video_path"]
         frag2_video_path = frag2.infos[0]["video_path"]
         if self._in_ipython():
             self._display_videos_in_notebook(frag1_video_path, frag2_video_path)
 
             pref = input(
-                "Which video is preferred? (1 or 2, or q to quit, or r to replay): "
+                "Which video is preferred? (1 or 2, or q to quit, or r to replay): ",
             )
             while pref not in ["1", "2", "q"]:
                 if pref == "r":
@@ -978,7 +989,7 @@ class SynchronousCLIGatherer(PreferenceGatherer):
                 return False
 
             # should never be hit
-            raise ValueError(f"Unexpected input {pref}")
+            assert False
         else:
             print("Which video is preferred? (1 or 2, or q to quit, or r to replay):\n")
             cap1 = cv2.VideoCapture(str(frag1_video_path))
@@ -1029,7 +1040,7 @@ class SynchronousCLIGatherer(PreferenceGatherer):
             raise KeyboardInterrupt
 
     def _display_videos_in_notebook(
-        self, frag1_video_path: pathlib.Path, frag2_video_path: pathlib.Path
+        self, frag1_video_path: pathlib.Path, frag2_video_path: pathlib.Path,
     ) -> None:
         from IPython.display import HTML, Video, clear_output, display
 
@@ -1055,7 +1066,7 @@ class SynchronousCLIGatherer(PreferenceGatherer):
 
     def _in_ipython(self) -> bool:
         try:
-            return get_ipython().__class__.__name__ == "ZMQInteractiveShell"
+            return get_ipython().__class__.__name__ == "ZMQInteractiveShell" # type: ignore[attr-defined]
         except NameError:
             return False
 
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -947,17 +947,19 @@ class SynchronousCLIGatherer(PreferenceGatherer):
         return preferences
 
     def _display_videos(
-        self, frag1: TrajectoryWithRew, frag2: TrajectoryWithRew,
+        self,
+        frag1: TrajectoryWithRew,
+        frag2: TrajectoryWithRew,
     ) -> bool:
         """Displays the videos of the two fragments.
 
         Args:
             frag1: first fragment
             frag2: second fragment
-        
+
         Returns:
             True if the first fragment is preferred, False if not.
-        
+
         Raises:
             KeyboardInterrupt: if the user presses q to quit.
             RuntimeError: if the video files cannot be opened.
@@ -1040,7 +1042,9 @@ class SynchronousCLIGatherer(PreferenceGatherer):
             raise KeyboardInterrupt
 
     def _display_videos_in_notebook(
-        self, frag1_video_path: pathlib.Path, frag2_video_path: pathlib.Path,
+        self,
+        frag1_video_path: pathlib.Path,
+        frag2_video_path: pathlib.Path,
     ) -> None:
         from IPython.display import HTML, Video, clear_output, display
 
@@ -1066,7 +1070,7 @@ class SynchronousCLIGatherer(PreferenceGatherer):
 
     def _in_ipython(self) -> bool:
         try:
-            return get_ipython().__class__.__name__ == "ZMQInteractiveShell" # type: ignore[attr-defined]
+            return get_ipython().__class__.__name__ == "ZMQInteractiveShell"  # type: ignore[attr-defined]
         except NameError:
             return False
 
--- a/src/imitation/algorithms/preference_comparisons.py
+++ b/src/imitation/algorithms/preference_comparisons.py
@@ -908,12 +908,14 @@ class SyntheticGatherer(PreferenceGatherer):
         return np.array(rews1, dtype=np.float32), np.array(rews2, dtype=np.float32)
 
 
-class SynchronousCLIGatherer(PreferenceGatherer):
-    """Queries for human preferences using the command line interface."""
+class SynchronousHumanGatherer(PreferenceGatherer):
+    """Queries for human preferences using the command line or a notebook."""
 
     def __init__(
         self,
         video_dir: pathlib.Path,
+        video_width: int = 500,
+        video_height: int = 500,
         custom_logger: Optional[imit_logger.HierarchicalLogger] = None,
     ) -> None:
         """Initialize the human preference gatherer.
@@ -924,6 +926,8 @@ class SynchronousCLIGatherer(PreferenceGatherer):
         """
         super().__init__(custom_logger=custom_logger)
         self.video_dir = video_dir
+        self.video_width = video_width
+        self.video_height = video_height
 
     def __call__(self, fragment_pairs: Sequence[TrajectoryWithRewPair]) -> np.ndarray:
         """Displays each pair of fragments and asks for a preference.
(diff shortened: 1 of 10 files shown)


src/imitation/util/video_wrapper.py
```diff
@@ -61,18 +72,27 @@ def _reset_video_recorder(self) -> None:
...
-        return self.env.reset()
+        return new_obs
 
     def step(self, action):
-        res = self.env.step(action)
+        obs, rew, done, info = self.env.step(action)
         self.video_recorder.capture_frame()
-        return res
+        # is it crazy to save the video path at every step?
+        info["video_path"] = self.get_current_video_path()
```
timbauman, line 85: Probably best to only store this on the first step

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +908,158 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
...
+        command line, it will pop out a video player for each fragment. If in a
+        notebook, it will display the videos. Either way, it will request 1 or 2 to
+        indicate which is preferred.
+
+        Args:
+            fragment_pairs: sequence of pairs of trajectory fragments
+
+        Returns:
+            A numpy array of 1 if fragment 1 is preferred and 0 otherwise, with shape
+            (b, ), where b is the length of the input
```
AdamGleave, line 941: What's the input? (fragment_pairs, or user input, ...?)

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +908,158 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
...
+        """
+
```
AdamGleave, line 943:
```suggestion
```

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +908,158 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
...
+        preferences = np.zeros(len(fragment_pairs), dtype=np.float32)
+        for i, (frag1, frag2) in enumerate(fragment_pairs):
```
AdamGleave, line 945: Could also write as something like (not tested, may need to be modified):
```
preferences = [self._display_videos(**frag) for frag in fragment_pairs] 
preferences = np.astype(preferences, dtype=np.float32)
```

Not sure that's actually any better though -- more succinct but also a bit more magical. WDYT Tom & @levmckinney?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +908,158 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
...
+        return preferences
+
+    def _display_videos(
+        self, frag1: TrajectoryWithRew, frag2: TrajectoryWithRew
+    ) -> bool:
+        """Displays the videos of the two fragments.
+
+        Args:
+            frag1: first fragment
+            frag2: second fragment
```
AdamGleave, line 957: Missing Returns in docstring

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +908,158 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
...
+            elif pref == "1":
+                return True
+            elif pref == "2":
+                return False
+
+            # should never be hit
+            raise ValueError(f"Unexpected input {pref}")
+        else:
+            print("Which video is preferred? (1 or 2, or q to quit, or r to replay):\n")
+            cap1 = cv2.VideoCapture(str(frag1_video_path))
```
AdamGleave, line 984: The IPython and CV sections are largely independent except for initial path extraction -- consider moving this whole `else:` branch into a helper function analogous to `self._display_videos_in_notebook`?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +908,158 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
...
+                key = chr(cv2.waitKey(1) & 0xFF)
+                if key == "q":
+                    raise KeyboardInterrupt
+                elif key == "r":
+                    cap1.set(cv2.CAP_PROP_POS_FRAMES, 0)
+                    cap2.set(cv2.CAP_PROP_POS_FRAMES, 0)
+                    ret1, frame1 = cap1.read()
+                    ret2, frame2 = cap2.read()
+                elif key == "1" or key == "2":
+                    cap1.release()
```
AdamGleave, line 1021: Do we always want these lines to be executed? If so, might want to move them to a `finally:` block to both ensure they always execute and cut duplication of lines 1026-1028.

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +908,158 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
...
+            cap2.release()
+            cv2.destroyAllWindows()
+            raise KeyboardInterrupt
+
+    def _display_videos_in_notebook(
+        self, frag1_video_path: pathlib.Path, frag2_video_path: pathlib.Path
+    ) -> None:
+        from IPython.display import HTML, Video, clear_output, display
+
+        display(HTML("<h2>Video 1</h2>"))
```
AdamGleave, line 1036: Consider `for` loop to avoid duplicating instructions for video 1 & 2

src/imitation/util/video_wrapper.py
AdamGleave: The fragmenter will splice trajectories into multiple fragments, so storing it once per episode may not be enough (unless you make the fragmenter cleverer, and have it propagate the video path as needed). You can avoid extra memory by making sure `self.get_current_video_path()` returns the same object on successive calls if underlying path hasn't changed (some kind of memoization should work).

AdamGleave (commented):
Thanks for implementing this! Looks like implementation is moving in right direction. Agreed we should add test case & a notebook demo. Left some relatively minor comments on the implementation so far.

src/imitation/util/video_wrapper.py
timbauman: I think the issue I was worried about was that the trajectories would be stored on disk, in which case the video path would be duplicated a lot. Although maybe pickle is smart enough to make it a reference if it's the same object? IIRC `str` is immutable so it probably gets stored on disk as a value but `Path` probably won't be

src/imitation/algorithms/preference_comparisons.py
timbauman: Yeah I'll do that. Although the way I built it `_display_videos_in_notebook` only displays the videos, while the display and interaction code isn't cleanly separated in this code. (I could refactor to make it more modular and have the same shape but that feels like overkill.)

src/imitation/algorithms/preference_comparisons.py
timbauman: Maybe I'm wrong but based on Googling it looks like `finally` can only be used for try/except blocks? Were you thinking I'd add a dummy try/except here?

timbauman: Or for the KeyboardInterrupt case? TBH I think I don't need to run these functions in the KeyboardInterrupt case because it's going to kill the program anyway

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +908,173 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
...
+            elif pref == "2":
+                return False
+
+            # should never be hit
+            assert False
+        else:
+            print("Which video is preferred? (1 or 2, or q to quit, or r to replay):\n")
+            cap1 = cv2.VideoCapture(str(frag1_video_path))
+            cap2 = cv2.VideoCapture(str(frag2_video_path))
+            cv2.namedWindow("Video 1", cv2.WINDOW_NORMAL)
```
timbauman, line 999: Is "video" the right word here? "Trajectory"? "Demonstration"?

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +908,201 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
         return np.array(rews1, dtype=np.float32), np.array(rews2, dtype=np.float32)
 
 
+class SynchronousCLIGatherer(PreferenceGatherer):
```
timbauman, line 911: It occurs to me that this isn't very testable right now

timbauman: I'm going to leave this untested for now, since I have other tasks I need to attend to for now. 🤞 that I'll get to it later but I'm not sure I will

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -906,6 +909,216 @@ def _reward_sums(self, fragment_pairs) -> Tuple[np.ndarray, np.ndarray]:
...
+                    height=self.video_height,
+                    width=self.video_width,
+                    html_attributes="controls autoplay muted",
+                    embed=False,
+                ),
+            )
+
+    def _in_ipython(self) -> bool:
+        try:
+            return self.is_running_pytest_test() or get_ipython().__class__.__name__ == "ZMQInteractiveShell"  # type: ignore[name-defined] # noqa
```
AdamGleave, line 1114: Does this mean we never test `_display_in_windows`?

src/imitation/scripts/train_preference_comparisons.py
```diff
@@ -166,7 +169,18 @@ def train_preference_comparisons(
 
     custom_logger, log_dir = logging_ingredient.setup_logging()
 
-    with environment.make_venv() as venv:
+    post_wrappers = (
```
AdamGleave, line 172: Long expression is a bit hard to follow, perhaps:

```
post_wrappers = None
if video_log_dir is not None:
  post_wrappers = [...]
```

tests/algorithms/test_preference_comparisons.py
```diff
@@ -290,6 +292,57 @@ def build_preference_comparsions(gatherer, reward_trainer, fragmenter, rng):
...
+                            "video_path": pathlib.Path(
+                                "tests/algorithms/test_preference_comparisons.py",
+                            ),
+                        },
+                    ],
+                ),
+                True,
+                np.array([1.0]),
+            ),
+            types.TrajectoryWithRew(
```
AdamGleave, line 321: Is this identical to above? Can we assign it to a temporary variable e.g. `traj` and then rewrite `trajectory_pairs = [(traj,traj)]`?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -290,6 +292,57 @@ def build_preference_comparsions(gatherer, reward_trainer, fragmenter, rng):
...
+                        },
+                    ],
+                ),
+                True,
+                np.array([1.0]),
+            ),
+        ),
+    ]
+
+    # this is the actual test
```
AdamGleave, line 339: Can we test the "r" or "q" functionality somehow?

tests/algorithms/test_preference_comparisons.py
```diff
@@ -290,6 +292,57 @@ def build_preference_comparsions(gatherer, reward_trainer, fragmenter, rng):
...
+    mock_input.return_value = "1"
+    assert gatherer(trajectory_pairs) == np.array([1.0])
+    mock_input.return_value = "2"
+    assert gatherer(trajectory_pairs) == np.array([0.0])
+
+
```
AdamGleave, line 345: TODO: test the functionality using `self._display_in_windows` not just IPython? (Probably only worth doing if it's relatively easy, e.g. ~1 hour of work, otherwise we can just take the hit in code coverage.)

AdamGleave (commented):
Mostly looks good, some minor changes needed and addition of a test case.

@levmckinney reckon you could take this PR over the finish line?

RedTachyon (commented):
I did a general code review, of the library changes, left a few comments.

A few more notes:
- Right after running the tutorial, this is what I see:

image

It's probably because we're still running on `gym==0.21.0` which uses pyglet for rendering, which is a ton of pain. Gymnasium uses pygame, so it might be easier to make headless, but I'm not 100% sure right now.
If we could remove the individual rendering for the learning envs, that would be great. The one way I see to do this is an env without rendering for learning, and then a separate one for rendering -- would this be easy to do? I think if it works, it'd be a good improvement to the user experience of the tutorial.

- Maybe an option to skip a given sample? (in the 1-2-r-q choice, add `s` for skip) Sometimes it's so hard to tell that it feels better to not choose anything; not sure how well this plays with the algorithm (it should be alright to just get another sample, or remove that one from the dataset)

- If possible, maybe put the videos [side by side](https://stackoverflow.com/questions/67077437/multiple-videos-side-by-side-with-ipython-display)

This is in order of importance, and all of them are mainly just QoL improvements, depending on how much we want to spend on this tutorial. On the other hand, if we (or someone else) intend to use the UI for larger real workloads, they become somewhat more significant.

---

Also, for the sake of the tutorial, it'd be good to have some more commentary around each created object -- what's a reward net, what's a fragmenter, a preference model etc

src/imitation/util/video_wrapper.py
```diff
@@ -60,19 +76,25 @@ def _reset_video_recorder(self) -> None:
...
+        info["video_path"] = self.current_video_path
+        return obs, rew, done, info
 
     def close(self) -> None:
         if self.video_recorder is not None:
             self.video_recorder.close()
             self.video_recorder = None
+        if self.delete_on_close:
+            for path in self.directory.glob("*.mp4"):
+                path.unlink()
```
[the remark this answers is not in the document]
RedTachyon: One more thing I noticed -- I don't think this will delete the directories created along the way? But we need to be careful in case someone points at "." and then it'd get deleted upon closing

AdamGleave:
> I did a general code review, of the library changes, left a few comments.
Thanks!

> A few more notes:
> 
> * Right after running the tutorial, this is what I see:
> 
> image
> It's probably because we're still running on `gym==0.21.0` which uses pyglet for rendering, which is a ton of pain. Gymnasium uses pygame, so it might be easier to make headless, but I'm not 100% sure right now. If we could remove the individual rendering for the learning envs, that would be great. The one way I see to do this is an env without rendering for learning, and then a separate one for rendering -- would this be easy to do? I think if it works, it'd be a good improvement to the user experience of the tutorial.

Oh wow that is ugly! I don't think I saw this when I ran it, may be a version or OS difference.

I'm confused why the learning envs are rendering -- with Pendulum I thought the policy and reward took input in raw observations, and rendering was just used for the human preference comparisons?

Making it all headless obviously good if we're capturing the video anyway and showing it in the notebook. I'd be OK merging this PR without fixing this and just making an issue to track it -- it still seems better to have the tutorial than not.

> * Maybe an option to skip a given sample? (in the 1-2-r-q choice, add `s` for skip) Sometimes it's so hard to tell that it feels better to not choose anything; not sure how well this plays with the algorithm (it should be alright to just get another sample, or remove that one from the dataset)
> * If possible, maybe put the videos [side by side](https://stackoverflow.com/questions/67077437/multiple-videos-side-by-side-with-ipython-display)

Yeah, I think the original DRLHP paper had an option for contractors to skip which just ignored the sample. Not that principled but works OK in practice.

> This is in order of importance, and all of them are mainly just QoL improvements, depending on how much we want to spend on this tutorial. On the other hand, if we (or someone else) intend to use the UI for larger real workloads, they become somewhat more significant.

I think this is too toy to be used for real human experiments, #716 will be better for real workloads.

timokau:
> Oh wow that is ugly! I don't think I saw this when I ran it, may be a version or OS difference.

The windows can be avoided by using a virtual frame buffer. `xvfb` and it's [python wrapper](https://pypi.org/project/xvfbwrapper/) are useful for this.

PR #713, opened 2023-05-13
Preference comparison improvements

timbauman:
## Description

This adds some parameters for varying preference comparison to try to find hyperparameters that work better

## Testing

Ran some training runs with these parameters


src/imitation/scripts/analyze.py
```diff
@@ -212,6 +219,7 @@ def _return_summaries(sd: sacred_util.SacredDicts) -> dict:
     "expert_return_summary": lambda sd: _return_summaries(sd)["expert_return_summary"],
     "imit_return_summary": lambda sd: _return_summaries(sd)["imit_return_summary"],
     "imit_expert_ratio": lambda sd: _return_summaries(sd)["imit_expert_ratio"],
+    "rollout_summary": lambda sd: _return_summaries(sd)["rollout_summary"],
 }
 
 # If `verbosity` is at least the length of this list, then we use all table_entry_fns
```
timbauman, line 225: Should we include this in the verbosity=0 list?

Issue #696, opened 2023-03-29
Support human preferences in “Deep RL from human preferences” (RLHP) implementation


mschweizer:
@rk1a and I participated in last year's BASALT competition and we noticed that RLHP currently lacks support for human preferences.

## Problem: 
Only synchronous, synthetic preferences gathering is supported by RLHP. 

There is only a synthetic gatherer. More fundamentally, [PreferenceComparisons.train](https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/algorithms/preference_comparisons.py#L1656) collects preferences synchronously meaning that [PreferenceGatherer.\_\_call\_\_](https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/algorithms/preference_comparisons.py#L803) blocks until all preferences for a set of queries are gathered. This is okay for synthetic preferences that are immediately returned by an oracle but problematic for human preferences that take time to answer. (There is also no user interface to interact with the human in order to present queries and collect preferences.) 

### Making the training procedure asynchronous 
We propose to extract the querying functionality from the `PreferenceGatherer` class into a new `PreferenceQuerent` class and refine the `PreferenceGatherer` interface accordingly. The `PreferenceQuerent` sends the queries to the user and returns a dictionary of all sent queries, each augmented by an ID. The `PreferenceGatherer` maintains a list of pending queries (with their IDs) and gathers available preferences for all pending queries from the user (or an oracle).

The following code block shows how this would be used:

```python
queries = self.fragmenter(trajectories, self.fragment_length, num_queries)
identified_queries = self.preference_querent(queries)
self.preference_gatherer.add(identified_queries)
answered_queries, preferences = self.preference_gatherer()
self.dataset.push(answered_queries, preferences)
```

### Human Preferences
The `HumanPreferenceQuerent` renders the videos for each query, uploads them to a cloud storage and communicates query ID and video URLs to the UI service. The UI allows the user to state their preferences for pending queries. The `HumanPreferenceGatherer` retrieves available preferences (identified by their ID) from the UI service and adds these to the dataset.

We would follow the approach of [rl-teacher](https://github.com/nottombrown/rl-teacher) which communicates via REST requests with the external, Django-based UI service and stores the trajectory fragment videos, that are rendered by the agent and displayed by the UI, in an S3-like cloud storage. 

### Synthetic Preferences 
The existing implementation for synthetic preferences needs to be adjusted to comply with the API change described above.
[...]

PR #716, opened 2023-05-15
Support asynchronous human preference gathering in RLHP implementation 

mschweizer:
## Description
- We implemented asynchronous human preference gathering for RLHP, which includes an API change. See #696 for more details. 
- Our PR implements a REST-interface for general Web-UIs and we provide a simple [Django-web-service](https://github.com/mschweizer/pref_collect) that already works with this. 
- This PR also integrates synchronous human preference gathering via CLI/Notebook from PR #712, which is fully compatible with our implementation. 
  - Another difference is that we only generate videos for the requested fragments instead of for the entire environment interaction. 

## Testing
We added extensive unit tests in test_preference_comparisons.py 
Tests for 2 utility methods are currently missing and some existing tests still need fixing.


rk1a:
@ernestum just lettting you know that this is ready for first review. Also we think that this might be a breaking change due to the changed signature of the "gatherer" and introduction of the "querent".

ernestum:
@rk1a thanks for this huge chuck of work! We are excited to see how we can upstream it.

Right now we are focused on releasing a v1.0 of `imitation` by the end of July. We plan to introduce the asynchronous human preference gathering after that release. So please bear with us because we do not have the capacity to review in depth and integrate such a big change right now but be assured that we have it on our radar.

I had a quick glance over the changes. I think changing the signature of the gatherer is no breaking change since it is just an internal class anyways. Requiring an explicit querent to construct the `PreferenceComparisons` class, however, is a breaking change. Maybe we can adapt it so some default querent is created when it is not provided?

To me it seem to split up the gatherer into gatherer and querent. I am not 100% sure that I understand the bounds of the responsibilities of the two classes. Especially when I was just reading the docstring:
```
preference_querent: queries preferences between trajectory fragments.
preference_gatherer: gathers preferences between trajectory fragments.
```
it is not really clear to me.

@Rocamonde will have a deeper look at your PR to give you some still preliminary but more in-depth feedback in the coming days.

Rocamonde (commented):
Thanks for the PR, this seems like valuable functionality to add to imitation. I have some comments below on the HLA. There's some high-level changes that I would consider making before moving forward with a more low-level review.

1. 

I find that the `querent` name is not ideal. It is a very uncommon word to use in English, which seems to be used only in two contexts:

- legal context: archaic term for plaintiff 
- astrology: one who consults an astrologer

This makes the meaning of the word in this context unclear to both native and non-native English speakers, and therefore makes the API confusing and hard to follow. I looked up ["querent reinforcement learning"](https://www.google.com/search?rls=en&q=%22querent%22+reinforcement+learning&ie=UTF-8&oe=UTF-8) to make sure I was not missing some niche use that I was unaware of, but it does not seem like the word is established jargon either. I had to look up the meaning to confirm that my guess was correct. I think a better word to use would be inquirer (which is x10 more common according to Google Ngram), or just drop this class altogether (see below).

Then there's also other confusing names like `PrefCollectGatherer` vs `PreferenceGatherer`. Gathering and collecting are basically synonymous in English AFAICT, so this difference it not very semantically clarifying.

2. 

In your issue you say

> We propose to extract the querying functionality from the PreferenceGatherer class into a new PreferenceQuerent class and refine the PreferenceGatherer interface accordingly. The PreferenceQuerent sends the queries to the user and returns a dictionary of all sent queries, each augmented by an ID. The PreferenceGatherer maintains a list of pending queries (with their IDs) and gathers available preferences for all pending queries from the user (or an oracle).

I do not understand the rationale behind your design decision. From an API PoV, the first most natural way to do this is to have a single class to do this, or alternatively split *creating*, *storing*, and *resolving* queries into separate entities. You have the querent for creating queries, but then you still have to manually add these queries to the gatherer. It's unclear what entity is responsible from handling the queue storage, or why you'd need to call the querent separately and not just wrap this functionality for the user API directly in the gatherer.

I understand that, from a back-end PoV, it seems that the querent does the heavy lifting of preparing a query, which is conceptually distinct from handling the queries and awaiting them. But this sounds like a problem that we as developers should deal with nicely, e.g. by accepting a callable to prepare query calls, instead of expecting the user to understand this. 

3.

I don't actually understand how your PR solves the issue of gatherer calls being blocking. You say that the current design is problematic because

> This is okay for synthetic preferences that are immediately returned by an oracle but problematic for human preferences that take time to answer.

Either the preference collection is synchronous or asynchronous:
- If synchronous (as is the case in your `SynchronousHumanGatherer` class), then the high-level software design is the same. you're just adding an IO service to collect the preferences rather than generate them synthetically. No change here.
- If asynchronous (as is the case with `PrefCollectGatherer`), then you're quietly ignoring all the queries that have not resolved, which means that the speed at which humans give feedback affects the training dynamics. Correct me if this is wrong, but that does not sound like a controllable learning environment for running experiments, etc. I'm not sure I can fully understand this use case, but maybe someone with more experience on this particular matter can chime in here. 

On a lower-level note, the method to resolve the queries is a `__call__` method to the gatherer, but now you're adding a `.add()` method to add the queries. This is kind of inconsistent, since there's no reason why the user should think in terms of methods for adding queries but in terms of callables for resolving them. I'd replace this with a `.resolve()` instead. You're making breaking changes in the type signature
 of `__call__` anyways. 


4.

As a general comment, we should consider refactoring our `preference_comparisons.py` module, it now has a staggering 2200 lines of code and has too many distinct class types, e.g. querent, fragmenter, trajectory_generator, preference_model,

mschweizer:
Hi Rocamonde,

Thanks for your feedback.

1. Naming
    - We are happy to change the name “Querent” to something more common, like inquirer. 
    - PrefCollect is the name of the Service that we have used as Web GUI for preference collection. We plan to make this part more generic and remove the reference to the name of that particular service accordingly. 
2. Split into Gatherer and Querent
  
   - We will change our design and integrate the querent functionality into the gatherer. 
   - We will follow up with details as soon as we implement this change.
4. Asynchronous Preference Collection and Controllable Learning
   - We acknowledge that this asynchronous design results in a less controlled learning process.
   - We believe that there is no solution that allows a completely controlled learning process and fast training.
   - In our opinion, the trade-off between these two conflicting objectives should be made by the user. 
    For this, we would like to allow the user to choose the degree of control based, for example, on a percentage of queries that are unresolved. The range would go from completely controlled, i.e. training is blocked if any unresolved queries exist, to completely uncontrolled, i.e. training proceeds irrespective of the number of unresolved queries.
   - Our approach can be extended to include this functionality.
   - As a side note, the implementation of Christiano et al. was asynchronous with parallel, non-blocking processes for each, RL, preference collection and reward model training.
5. Agreed :-)

src/imitation/algorithms/preference_comparisons.py
```diff
@@ -778,6 +783,160 @@ def variance_estimate(self, rews1: th.Tensor, rews2: th.Tensor) -> float:
...
+        """
+        self.logger = custom_logger or imit_logger.configure()
+
+    def __call__(
+        self,
+        queries: Sequence[TrajectoryWithRewPair],
+    ) -> Dict[str, TrajectoryWithRewPair]:
+        """Queries the user for their preferences.
+
+        This dummy implementation does nothing because by default the queries are
```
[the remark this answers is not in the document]
mschweizer: You're right, we will change this accordingly

src/imitation/data/wrappers.py
```diff
@@ -1,15 +1,86 @@
...
+            use_file_cache: whether to save rendered images to disk.
+        """
+        super().__init__(env)
+        self.scale_factor = scale_factor
+        self.use_file_cache = use_file_cache
+        if self.use_file_cache:
+            self.file_cache = tempfile.mkdtemp("imitation_RenderImageInfoWrapper")
+
+    def step(self, action):
+        obs, rew, done, info = self.env.step(action)
```
timokau, line 48: This wrapper hasn't been adjusted for the new gymnasium api yet right?

rk1a: that's right, thanks for the heads up! We also still need to add tests for the wrapper.

mschweizer:
Hi @Rocamonde   
we have changed our design and integrated the querent functionality into the gatherer.
We added the two methods gather() and query() to the gatherer and removed its __call()__ method. gather() now fulfills the role of the __call()__ method.
We made the querent a component of the gatherer and query() calls its querent component’s query method.
Currently, the implemented gatherer classes use a hardcoded querent component. We are aware that there are alternative ways of designing this:
inject an object or callable for query(); injecting an object is similar to our new design because our design could be changed to accept the querent component as a constructor argument (with default value).
use multiple inheritance for query() and possibly gather()
implement gather() and query() directly in each gatherer class (and use inheritance where it is needed)Side note: we waited with the renaming of the

We would like to get your input on our new design and these alternative options.
What would be the advantages/disadvantages given a future refactoring of the algorithm?

src/imitation/data/wrappers.py
mschweizer: Hey @timokau, the wrapper is adjusted to the gymnasium api and we tested it briefly. It should work. hope this helps

timokau: Thanks Marvin!

PR #723, opened 2023-05-30
Transition from Models Hub to Datasets Hub for expert trajectories

ernestum:
## Description

HuggingFace offers a [**Models** Hub](https://huggingface.co/models) and a [**Datasets** Hub](https://huggingface.co/datasets).

We used to store expert trajectories along with expert models on the HuggingFace **Model** Hub and load them from there using the `imitation.data.serialize.load_rollouts_from_huggingface` function.

Since #677 we have support for [HuggingFace Datasets](https://huggingface.co/docs/datasets/) for storing and loading expert trajectories so it makes more sense to store the expert trajectories on the **Datasets** Hub instead of the **Models** Hub.

As part of this transition, this PR removes the `imitation.data.serialize.load_rollouts_from_huggingface` function, which used to load rollouts from the **Model** Hub.
That function was only used in the `demonstrations` ingredient for the sacred scripts.
This ingredient was changed so it downloads the demonstrations from the **Datasets** hub using `datasets.load_dataset()`.
For this to work, I uploaded the already existing expert trajectories that were stored in a (old-format) `rollouts.npz` along the expert models to their own **Dataset** repositories.
You will find the six new **Dataset** repositories on our [HuggingFace Organization page](https://huggingface.co/HumanCompatibleAI) at the bottom.

## Testing

The existing tests should cover this change.

### DIFF
--- a/src/imitation/scripts/ingredients/demonstrations.py
+++ b/src/imitation/scripts/ingredients/demonstrations.py
@@ -150,11 +150,12 @@ def _generate_expert_trajs(
 
 @demonstrations_ingredient.capture
 def _download_expert_rollouts(
-        environment: Dict[str, Any],
-        path: Optional[str],
-        organization: Optional[str],
-        algo_name: Optional[str],
-        loader_kwargs: Dict[str, Any]):
+    environment: Dict[str, Any],
+    path: Optional[str],
+    organization: Optional[str],
+    algo_name: Optional[str],
+    loader_kwargs: Dict[str, Any],
+):
     if path is not None:
         repo_id = path
     else:
--- a/src/imitation/scripts/ingredients/demonstrations.py
+++ b/src/imitation/scripts/ingredients/demonstrations.py
@@ -1,7 +1,6 @@
 """Ingredient for scripts learning from demonstrations."""
 
 import logging
-import warnings
 from typing import Any, Dict, Optional, Sequence
 
 import datasets
@@ -26,11 +25,26 @@ logger = logging.getLogger(__name__)
 
 @demonstrations_ingredient.config
 def config():
+    # Either "local" or "huggingface" or "generated".
     rollout_type = "local"
-    # path to file containing rollouts. If rollout_path is None
-    # and rollout_type is local, they are sampled from the expert.
+
+    # local path or huggingface repo id to load rollouts from.
     rollout_path = None
-    n_expert_demos = None  # Num demos used or sampled. None loads every demo possible.
+
+    # passed to `datasets.load_dataset` if `rollout_type` is "huggingface"
+    loader_kwargs: Dict[str, Any] = dict(
+        organization="HumanCompatibleAI",
+        split="train",
+    )
+
+    # Used to deduce HuggingFace repo id if rollout_path is None
+    organization = "HumanCompatibleAI"
+
+    # Used to deduce HuggingFace repo id if rollout_path is None
+    algo_name = "ppo"
+
+    # Num demos used or sampled. None loads every demo possible.
+    n_expert_demos = None
     locals()  # quieten flake8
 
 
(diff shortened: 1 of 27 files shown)


ernestum:
Hi @levmckinney, this PR is ready for review. Would you have a look at it?

src/imitation/scripts/ingredients/demonstrations.py
```diff
@@ -109,43 +141,15 @@ def generate_expert_trajs(
...
     assert rollout_type.endswith("-huggingface")
     algo_name = rollout_type.split("-")[0]
-    return serialize.load_rollouts_from_huggingface(
+    model_name = hfsb3.ModelName(
         algo_name,
-        env_name=loader_kwargs["env_name"],
-        organization=loader_kwargs["organization"],
+        hfsb3.EnvironmentName(loader_kwargs["env_name"]),
     )
+    repo_id = hfsb3.ModelRepoId(loader_kwargs["organization"], model_name)
```
levmckinney, line 152: While we are making this change we may want to add a bit of future proofing. First we might want to be able to have datasets that are more fine-grained than just algorithm + environment. For example, we might want a dataset created from a PPO agent trained on ant for 100,000 time steps and another dataset for the same agent trained after 1,000,000 time steps. In addition, we might want to update the dataset lets say we find a bug in the PPO code and upload a new dataset to the hub. For experiment repeatability people should be able to select the old `revision` of the dataset.

src/imitation/scripts/ingredients/demonstrations.py
```diff
@@ -109,43 +141,15 @@ def generate_expert_trajs(
...
+    logger.info(f"Loading expert trajectories from {repo_id}")
+    dataset = datasets.load_dataset(repo_id)["train"]
```
levmckinney, line 154: Having canonical splits for the datasets might also be useful, but that might be beyond the scope of this PR.

levmckinney (approved):
This ports the current functionality over to datasets nicely. Since we are already opening up this code, I have a couple of comments though on how the interface could be made more future-proof.

src/imitation/scripts/ingredients/demonstrations.py
ernestum: You are totally right. I looked again into the API design of this and cleaned it up a bit:

Now you can either specify `demonstrations.organization` (default [`HumanCompatibleAI`](https://huggingface.co/HumanCompatibleAI)) and `demonstrations.algo_name` (default `ppo`). Then the HuggingFace Hub `repo_id` is deduced according to `<organization>/<algo_name>-<normalized_env_name>` (e.g. [`HumanCompatibleAI/ppo-seals-Walker2d-v0`](https://huggingface.co/HumanCompatibleAI/ppo-seals-Walker2d-v0)) or you specify `demonstrations.path`. Then this is used as the `repo_id`.

The `loader_kwargs` are now part of the `demonstrations` ingredient and not pulled from the `expert` ingredient. They are passed straight to the `datasets.load_dataset()` function so you have full control over how the dataset is loaded. You can even select your own split here (the default value for the `loader_kwargs` is `{"split": "train"}`).

I also changed the design so you have to explicitly set the `demonstrations.type` to either `local`, `huggingface` or `generated`. This fixes two issues:
1. before we awkwardly extracted the `algo_name` from the `type` which was set to `<algo_name>-huggingface`.
2. We implicitly started to generate demonstrations if the `path` was not set. This could lead to unexpected behavior and was a bit harder to debug.

src/imitation/scripts/ingredients/demonstrations.py
ernestum: Turns out I could add this feature easily as a side-effect (see comment above)

levmckinney (approved):
These changes look good to me!

ernestum:
@AdamGleave would you merge this? I can't due to the coverage changes. There are some edge-cases in the new new demonstration ingredient, that we do not check right now, but I think it is not worth the effort to add tests for them (because the code is simple, they rarely occur and we probably won't keep this code for a long time anyways).

PR #735, opened 2023-06-29
Gymnasium Compatibility

ernestum:
## Description

Make `imitation` compatible with [`gymnasium`](https://gymnasium.farama.org/) because the old `gym` is outdated.

### Main changes:
- `import gym` -> `import gymnasium as gym`
- Adoption of the new gymnasium API (`.seed(x)` -> `.reset(seed=x)`, `done` -> `terminated`/`truncated`, `.reset()` not returns a tuple)
- Added some extra checks because the new API does not guarantee that spaces have a shape
- We don't import `seals` any more and rather use the `seals:` prefix on environment names.
- `make_vec_env` does not mess with `TimeLimit` wrappers any more (`gymnasium` does that for us)

### To be changed before merging
- [x] Update the [models on HuggingFace model hub](https://huggingface.co/HumanCompatibleAI) (they are already trained on `rnn`, I just need to push them)
- [x] Stop installing `gym` as part of the pipeline (that was just so we can load the current models)
- [x] Depend on upstream seals and not on a branch of it.

### Notes on the failing pipeline

- Windows pipeline is addressed in #779 

## Testing

No new tests have been added but all tests pass now.

---
Thanks to all the people that helped coding and reviewing! 
@AdamGleave, @araffin, @EdoardoPona, @Rocamonde, @simoninithomas,

This PR has been possible due to lots of work in our dependencies:

- https://github.com/huggingface/huggingface_sb3/pull/35
- https://github.com/HumanCompatibleAI/seals/pull/77
- https://github.com/HumanCompatibleAI/seals/pull/73
- https://github.com/DLR-RM/rl-baselines3-zoo/pull/403


Rocamonde:
Linking https://github.com/HumanCompatibleAI/seals/pull/72 to this

Rocamonde:
I am getting some weird pytype error that it cannot import packages. I remember facing this at some point in the past. @ernestum do you know what the cause of this is?


```
$ pytype -j auto
Computing dependencies
Analyzing 107 sources with 0 local dependencies
ninja: Entering directory `.pytype'
[1/64] check test_benchmarking
FAILED: /home/rocamonde/HumanCompatibleAI/imitation/.pytype/pyi/test_benchmarking.pyi 
/home/rocamonde/HumanCompatibleAI/imitation/venv/bin/python -m pytype.single --imports_info /home/rocamonde/HumanCompatibleAI/imitation/.pytype/imports/test_benchmarking.imports --module-name test_benchmarking --platform linux -V 3.10 -o /home/rocamonde/HumanCompatibleAI/imitation/.pytype/pyi/test_benchmarking.pyi --analyze-annotated --nofail --quick /home/rocamonde/HumanCompatibleAI/imitation/tests/test_benchmarking.py
File "/home/rocamonde/HumanCompatibleAI/imitation/tests/test_benchmarking.py", line 6, in <module>: Can't find module 'imitation.scripts.train_adversarial'. [import-error]
File "/home/rocamonde/HumanCompatibleAI/imitation/tests/test_benchmarking.py", line 6, in <module>: Can't find module 'imitation.scripts.train_imitation'. [import-error]
```

ernestum:
> @ernestum do you know what the cause of this is?

No I have never seen this error before :shrug:

ernestum:
Note @ernestum : ensure to treat all open comments in #765 here.

AdamGleave:
@dan-pandori could you review this since Juan is busy with an ICLR submission right now? Please do tag me in any areas of the code you're uncertain about.

docs/getting-started/installation.rst
```diff
@@ -23,11 +17,6 @@ Prerequisites
 
 - (Optional) OpenGL (to render gym environments)
 - (Optional) FFmpeg (to encode videos of renders)
-- (Optional) MuJoCo (follow instructions to install `mujoco\_py v1.5 here`_)
```
dan-pandori, line 26: I don't understand why MuJoCo has been removed here. Is MuJoCo incompatible with Gymnasium, and that's why it was removed?

I see MuJoCo still in some other dependencies, which makes me confused.

src/imitation/algorithms/adversarial/common.py
```diff
@@ -493,6 +499,10 @@ def _get_log_policy_act_prob(
             assert gen_algo_actor is not None
             # generate log_policy_act_prob from SAC actor.
             mean_actions, log_std, _ = gen_algo_actor.get_action_dist_params(obs_th)
+            assert isinstance(
+                gen_algo_actor.action_dist,
+                distributions.SquashedDiagGaussianDistribution,
+            )  # Note: this is just a hint to mypy
```
dan-pandori, line 505: IMO unnecessary comment, but no harm in keeping.

src/imitation/algorithms/density.py
```diff
@@ -259,22 +259,51 @@ def _preprocess_transition(
     ) -> np.ndarray:
         """Compute flattened transition on subset specified by `self.density_type`."""
         if self.density_type == DensityType.STATE_DENSITY:
-            return flatten(self.venv.observation_space, obs)
+            flat_observations = space_utils.flatten(self.venv.observation_space, obs)
+            if not isinstance(flat_observations, np.ndarray):
```
dan-pandori, line 263: Thanks for including a helpful error message here and elsewhere!

src/imitation/data/buffer.py
```diff
@@ -273,14 +273,36 @@ def __init__(
...
-            act_shape = tuple(venv.action_space.shape)
-            obs_dtype = venv.observation_space.dtype
-            act_dtype = venv.action_space.dtype
+            if venv.observation_space.shape is not None:
+                if obs_shape is not None:
+                    raise ValueError(
+                        "Cannot specify both observation shape and also environment "
+                        "with an observation space that has a shape.",
+                    )
+                obs_shape = tuple(venv.observation_space.shape)
```
dan-pandori, line 282: Minor preference for consistently adding whitespace between these `if` statements. IE:

```python
    obs_shape = tuple(venv.observation_space.shape)

if venv.observation_space.dtype is not None:
```

src/imitation/data/rollout.py
```diff
@@ -419,7 +431,9 @@ def generate_trajectories(
     #
     # To start with, all environments are active.
     active = np.ones(venv.num_envs, dtype=bool)
-    assert isinstance(obs, np.ndarray), "Dict/tuple observations are not supported."
+    if not isinstance(obs, np.ndarray):
+        raise ValueError("Dict/tuple observations are not supported.")
```
dan-pandori, line 435: Prefer:
```
"Dict/tuple observations are not supported. Currently only np.ndarray observations are supported."
```

Only mentioning the first kind leaves open the option that we might be currently supporting something other than np.ndarray. I think that the only options are dict/tuple/np.ndarray so this may be slightly pedantic, but 🤷

tests/algorithms/test_mce_irl.py
```diff
@@ -26,16 +26,17 @@ def rollouts(env, n=10, seed=None):
...
-            # give same trajectory each time)
-            env.seed(seed)
             env.action_space.seed(seed)
-        obs = env.reset()
+
         traj = [obs]
         while not done:
             act = env.action_space.sample()
-            obs, rew, done, info = env.step(act)
+            obs, rew, terminated, truncated, info = env.step(act)
```
dan-pandori, line 38: Only `obs` and `rew` seem to be used here. Thoughts on updating to:
```
obs, rew, _, _, _ = env.step(act)
traj.append((obs, rew))
```
?

tests/algorithms/test_mce_irl.py
```diff
@@ -441,7 +449,8 @@ def test_mce_irl_reasonable_mdp(
 
         # test MCE IRL on the MDP
         mdp = ReasonablePOMDP()
-        mdp.seed(715298)
+        mdp.reset(seed=715298)
+        # mdp.seed(715298)    # NOTE: not sure this is the correct approach in this case
```
dan-pandori, line 453: I'm not either! But if mdp.reset is behaving normally it seems fine to go with that for this unit test and delete this.

dan-pandori (approved):
LGTM! Thanks for cleaning up types in several places and the detailed PR description.

I just added a few minor nits about comments etc. I don't deeply understand the behavior of resetting the seed vs using `seed()`, so you should ask Adam if you want a better opinion on that.

docs/getting-started/installation.rst
ernestum: MuJoCo is still a dependency. We used to depend on the `mujoco-py` package, which is requires manual installation steps. Now we depend on the `mujoco` package, which does not need manual installation steps. Therefore it has been removed from the installation instructions.

src/imitation/data/buffer.py
ernestum: You are right, it reads a lot easier this way!

src/imitation/data/rollout.py
ernestum: You are right, I am also a fan of actionable error messages!

tests/algorithms/test_mce_irl.py
ernestum: `terminated` and `truncated` are merged to `done`, which again is used as the termination condition for the loop. I however made sure the info is ignored.

tests/algorithms/test_mce_irl.py
ernestum: You are right. Removed it.

src/imitation/algorithms/adversarial/common.py
dan-pandori: Still seems redundant to me (ie, its status as a hint to mypy is obvious), but I don't have that strong of a preference for removing. If you do remove it here, I think you've got 3 equivalent comments in another file that you should remove too.

docs/getting-started/installation.rst
dan-pandori: Makes sense, thanks!

Issue #734, opened 2023-06-29
gymnasium compatibility of imitation


ernestum:
`gym` is deprecated and we eventually want to switch to `gymnasium`. `stable-baselines3` in the version 2.0.0 already supports gymnasium, but `seals` and `huggingface_sb3` are not ported yet.

PR #738, opened 2023-07-03
Pin SB3 version to 1.7.0

RedTachyon:
## Description

Currently the whole CI is on fire due to the default SB3 version being 2.0.0, which is incompatible with gym. 

This PR pins the version to 1.7.0 (or more precisely, to anything below 2.0.0, in case there's bugfix releases) while the Gymnasium switch is underway (see #735, #734, HumanCompatibleAI/seals#72)


AdamGleave (approved):
LGTM -- thanks for fixing the CI!

Issue #727, opened 2023-06-02
Tutorial to compare a new algorithms performance to the imitation baseline


ernestum:
It should be clear how to test if a new algorithm (or one of our own algorithms) has a significant improvement in rewards.

To do a meaningful statistical test (permutation test), the `imitation.testing.reward_improvement` can be used.

PR #737, opened 2023-07-03
Custom environment tutorial

RedTachyon:
Tutorial about using a custom environment with `imitation`, see #718

PR #739, opened 2023-07-03
Tutorial on comparing algorithm performance

RedTachyon:
See #727

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -74,6 +74,7 @@ If you use ``imitation`` in your research project, please cite our paper to help
    tutorials/5a_train_preference_comparisons_with_cnn
    tutorials/6_train_mce
    tutorials/7_train_density
+   tutorials/9_compare_baselines
    tutorials/trajectories
 
 .. toctree::
--- a/docs/tutorials/9_compare_baselines.ipynb
+++ b/docs/tutorials/9_compare_baselines.ipynb
@@ -9,7 +9,7 @@
     "\n",
     "Did we actually match the expert performance or was it just luck? Did this hyperparameter change actually improve the performance of our algorithm? These are questions that we need to answer when we want to compare the performance of different algorithms or hyperparameters.\n",
     "\n",
-    "`imitation` provides some tools to help you answer these questions. For demonstration purposes, we will use Behavior Cloning on the CartPole-v1 environment. We will compare the performance of the algorithm before and after training and compare it to the performance of an expert."
+    "`imitation` provides some tools to help you answer these questions. For demonstration purposes, we will use Behavior Cloning on the CartPole-v1 environment. We will compare different variants of the trained algorithm, and also compare it with a more sophisticated algorithm, DAgger."
    ]
   },
   {
@@ -40,7 +40,7 @@
     "    n_epochs=10,\n",
     "    n_steps=64,\n",
     ")\n",
-    "expert.learn(100_000)  # Note: set to 100000 to train a proficient expert"
+    "expert.learn(10_000)  # set to 100_000 for better performance"
    ]
   },
   {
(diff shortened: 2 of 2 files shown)


docs/index.rst
```diff
@@ -74,6 +74,7 @@ If you use ``imitation`` in your research project, please cite our paper to help
    tutorials/5a_train_preference_comparisons_with_cnn
    tutorials/6_train_mce
    tutorials/7_train_density
+   tutorials/9_compare_baselines
```
AdamGleave, line 77: Note: Merge after #737 which adds tutorial 8

docs/tutorials/9_compare_baselines.ipynb
```diff
@@ -0,0 +1,429 @@
...
+    ")\n",
+    "\n",
+    "print(f\"Is the expert significantly better than the not-expert? {significant}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Here we go! We can now be 99.9% confident that the expert is better than the not-expert. It might still be an extraordinary stroke of luck, or a conspiracy to make us choose the wrong algorithm, but outside of that, we can be pretty sure that the expert is better."
```
AdamGleave, line 180: This is true in terms of these two specific policies. However, it might be worth noting that this does *not* show that training with 100,000 timesteps is better than 10,000 timesteps with 99.9% confidence -- we only have a single sample of the *training run*, even if we have many samples of the *rollouts* from those policies (holding the training run fixed).

I'm sure you understand this, but it's a misinterpretation I think someone could easily make, so it may be worth explicitly flagging.

docs/tutorials/9_compare_baselines.ipynb
```diff
@@ -0,0 +1,429 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Let's try training a behavior cloning algorithm on this dataset.\n",
+    "\n",
+    "Note that for DAgger, we have to cheat a little bit -- it's allowed to use the expert policy to generate additional data.\n",
+    "For the purposes of this tutorial, we'll stick with this to avoid spending hours training an expert for a more complex environment.\n",
+    "\n",
+    "So while this little experiment isn't definitive proof that DAgger is better than BC, you can use the same method to compare any two algorithms."
```
AdamGleave, line 329: In practice we'd also want to at least run the algorithms with several random seeds. And probably do significance tests on the mean returns across seeds rather than mean returns across rollouts.

I'm not sure the tutorial should demonstrate that exactly -- it might be too slow, although we could always just write the code and only do ~n=3 seeds and have a comment saying more are needed to reach significance. But I'd want to at least flag that this this test alone isn't enough to actually conclude one algo is better than another.

AdamGleave (commented):
Thanks for adding this tutorial! Overall it seemed well written and a good addition to the documentation.

My main comment is that there's a couple of points in the text where I think the interpretation of the hypothesis test is a bit misleading: it really tests different learned policies but it could be misread as being about testing different algorithms (that produce those policies). I'd suggest clarifying that -- and possibly demonstrating hypothesis testing for comparing algorithms, which is probably what people actually _want_ to do more often (but is slower, so may not be doable to actually run at sufficient scale in a tutorial).

docs/tutorials/9_compare_baselines.ipynb
RedTachyon: I added a bunch of commentary across the notebook to address this, hopefully it's enough

docs/tutorials/9_compare_baselines.ipynb
RedTachyon: Also addressed in the commentary, at least for now

AdamGleave (approved):
LGTM modulo a couple of typos (will fix in another PR)

AdamGleave:
Closing in favor of #747

Issue #741, opened 2023-07-04
At least some tests are accidentally stochastic


RedTachyon:
## Bug description
In a [recent CI run](https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3572/workflows/03bf9cfe-431a-47da-9c1e-a1543a2891e5/jobs/13800) for an unrelated change, it's failing on one of the tests. (`tests.algorithms.test_tabular_policy_randomness`)

From what I can tell, it's because we're not actually properly seeding the RNG:
https://github.com/HumanCompatibleAI/imitation/blob/09ca7b6b19b2770f0b0d35246edd76b3aa40c7de/tests/conftest.py#L46-L48

This means there's a low (but non-zero) chance of each CI run failing for no good reason, which is generally not a good thing to have.

## Steps to reproduce
Go through `test_tabular_policy_randomness` in a debugger twice, compare the value of `np.mean(actions)` -- it will (probably be different)

AdamGleave:
Fixed in #742

PR #742, opened 2023-07-04
Fix the seed of the RNG in tests

RedTachyon:
Addresses #741

Someone more familiar with the code should double-check this in case there are some rough edges, or the randomness is intended (though imo it shouldn't be)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -45,4 +45,4 @@ def custom_logger(tmpdir: str) -> logger.HierarchicalLogger:
 
 @pytest.fixture()
 def rng() -> np.random.Generator:
-    return np.random.default_rng()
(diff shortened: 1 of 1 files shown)


AdamGleave (approved):
Approved.

I do have somewhat mixed feelings here. Test flakiness is bad. But tests that pass only on specific seeds are also problematic. In general for stochastic tests we should aim to have the sample size high enough to make both false positives and false negative rates extremely low.

However, in a test suite with 100s of tests, run 1000s of times, even a very small flakiness rate can be problematic. So, on balance, I lean towards thinking fixing the RNG is the least-bad thing to do here. But, it does blind us slightly to certain false-positives in tests, that we should be vigilant to.

PR #744, opened 2023-07-04
Implementation of the SQIL algorithm

### LINKED ISSUE #740: Implement Soft Q imitation learning (SQIL)
## Problem
[Soft Q imitation learning (SQIL)](https://arxiv.org/pdf/1905.11108.pdf) is a simple imitation learning algorithm that claims to achieve similar performance to more complicated algorithms like GAIL and AIRL. Adding it to `imitation` would enable an independent replication of these results and facilitate others using it as a baseline in environments not studied in the original paper.

## Solution
This issues proposes implementing SQIL. The algorithm is pleasingly simple: just Q-learning with a reward of +1 when the action matches the demonstrated action in a demonstrated state, and 0 for all other behavior. It can therefore be implemented on top of existing Q-learning implementations such as SB3's DQN.

Ideally the solution would be modular: a wrapper around key components of underlying Q-learning algorithms. However, some copy-pasting and modification may be difficult to avoid.

The first author of SQIL's [implementation](https://github.com/rddy/sqil-atari/tree/master) may be a useful reference, although the code is a modified version of a course homework assignment implementing DQN so is a little difficult to disentangle.

RedTachyon:
## Description

Fixes #740 

Right now it's a basic implementation based on SB3. 

It seems to work on a basic level (i.e. I trained it on CartPole and it converged), but still needs a bunch of cleanup and testing.

One note is that there's quite a few `#type: ignore[...]` annotations. I tried to minimize them, but a good chunk of the code is either modifying SB3 code, or closely interfacing with it, and SB3 seems to have more lax type checking.

## Testing

WiP


src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,321 @@
...
+replacing half the buffer with expert demonstrations and adjusting the rewards.
+"""
+from typing import Any, Dict, Iterable, Optional, Tuple, Type, Union
+
+import numpy as np
+import torch as th
+import torch.nn.functional as F
+from stable_baselines3 import dqn
+from stable_baselines3.common import policies, vec_env
+from stable_baselines3.common.buffers import ReplayBuffer
```
AdamGleave, line 13: Nit: this codebase follows the Google Python Style Guide (except where it conflicts with PEP8, which takes precedence) and https://google.github.io/styleguide/pyguide.html#22-imports states:

(We've tended to ignore this for tutorials as users may be less familiar with this style, but are more strict about adhering to it in the rest of the code.)

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,321 @@
+"""Soft Q Imitation Learning (SQIL).
```
AdamGleave, line 1: Add link to paper?

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,321 @@
...
+
+        """
+        self.venv = venv
+
+        super().__init__(demonstrations=demonstrations, custom_logger=custom_logger)
+
+        self.orig_train_freq = train_freq
+
+        self.dqn = dqn.DQN(
+            policy=policy,
```
AdamGleave, line 125: Could do `*args, **kwargs`. Although then function would be less self-documenting, so I'm unsure if that's better. More maintainable though if DQN changes underneath us?

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,321 @@
...
+            _init_setup_model: Whether or not to build the network
+                at the creation of the instance.
+
+
+        """
+        self.venv = venv
+
+        super().__init__(demonstrations=demonstrations, custom_logger=custom_logger)
+
+        self.orig_train_freq = train_freq
```
AdamGleave, line 122: This is the same as what you pass in as `train_freq` to DQN?

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def train(self, *, total_timesteps: int):
+        self.learn_dqn(total_timesteps=total_timesteps)
+
+    @property
+    def policy(self) -> policies.BasePolicy:
+        assert isinstance(self.dqn.policy, policies.BasePolicy)
+        return self.dqn.policy
+
+    def train_dqn(self, gradient_steps: int, batch_size: int = 100) -> None:
```
AdamGleave, line 189: Bit of a shame we need to duplicate so much from DQN implementation (I assume?), but probably unavoidable

AdamGleave (commented):
High-level review, can take a closer look once PR is further along!

src/imitation/algorithms/sqil.py
RedTachyon: I was also thinking about sort of a middle ground `dqn_kwargs`, but not sure which options is the best

tests/algorithms/test_sqil.py
```diff
@@ -0,0 +1,101 @@
...
+        next_obs = model.expert_buffer.next_observations[i]
+        done = model.expert_buffer.dones[i]
+
+        np.testing.assert_array_equal(obs[0], demonstrations.obs[i])
+        np.testing.assert_array_equal(act[0], demonstrations.acts[i])
+        np.testing.assert_array_equal(next_obs[0], demonstrations.next_obs[i])
+        np.testing.assert_array_equal(done, demonstrations.dones[i])
+
+
+def test_sqil_demonstration_without_flatten(rng):
```
AdamGleave, line 50: Could parameterize previous test with `flatten: bool` and merge them

tests/algorithms/test_sqil.py
```diff
@@ -0,0 +1,101 @@
...
+        rng=rng,
+    )
+    demonstrations = rollout.flatten_trajectories(rollouts)
+    model = sqil.SQIL(
+        venv=venv,
+        demonstrations=demonstrations,
+        policy=policy,
+        learning_starts=1000,
+    )
+    model.train(total_timesteps=10_000)
```
AdamGleave, line 101: Can probably strengthen this test a bit. Might be able to check that reward actually goes up if we're in a simple enough environment?

docs/tutorials/10_train_sqil.ipynb
```diff
@@ -0,0 +1,209 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[download this notebook here](https://github.com/HumanCompatibleAI/imitation/blob/master/docs/tutorials/10_train_sqil.ipynb)\n",
+    "# Train an Agent using Soft Q Imitation Learning\n",
+    "\n",
+    "Soft Q Imitation Learning ([SQIL](https://arxiv.org/abs/1905.11108)) is a simple algorithm that can be used to clone expert behavior.\n",
+    "It's fundamentally a modification of the DQN algorithm. At each training step, whenever we sample a batch of data from the replay buffer,\n",
+    "we also sample a batch of expert data. Expert demonstrations are assigned a reward of 1, while the agent's own transitions are assigned a reward of 0.\n",
```
AdamGleave, line 12: This is an accurate description but does highlight the algorithm is a bit bizarre. If the agent perfectly mimicked expert demos, they'd still get assigned a reward of 0. Whereas at least with stuff like AIRL/GAIL, they'd get the same reward (as discriminator could no longer distinguish them).

docs/tutorials/10_train_sqil.ipynb
```diff
@@ -0,0 +1,209 @@
+{
```
AdamGleave, line 1: Tutorial looks good! :+1:

AdamGleave (commented):
Thanks for the implementation! Algorithm looks correct. SQIL is a strange beast but at least it's quite simple conceptually.

I *think* the implementation could be simplified / code duplication could be reduced by moving most of the logic into a ReplayBuffer wrapper, then setting DQN to use that replay buffer (probably with `replay_buffer_class`). I think you can still do this in composition (which I agree seems right approach -- SQIL is not an RL algorithm, so probably shouldn't inherit from DQN, and multiple inheritance gets messy). I may be missing some subtlety though, but if right this would let us eliminate many lines of code making it much easier to read. Could also pave the way to having this support any `OffPolicyAlgorithm` rather than just DQN which would be neat.

Other main area is it'd be nice to have slightly more comprehensive tests. Although to be fair the algorithm is so trivial there's not that much to actually test (beyond correct functioning of DQN). Checking it makes some progress on a simple environment might add something. If returns improving is too flaky (at least without too expensive a # of timesteps), could also check the Q-network is moving in right direction (e.g. assigns higher Q-value to expert demos than randomly chosen observation/action pairs)?

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,314 @@
...
+                obs=np.array(transition["obs"]),  # type: ignore[index]
+                next_obs=np.array(transition["next_obs"]),  # type: ignore[index]
+                action=np.array(transition["acts"]),  # type: ignore[index]
+                done=np.array(transition["dones"]),  # type: ignore[index]
+                reward=np.array(1),
+                infos=[{}],
+            )
+
+    def train(self, *, total_timesteps: int):
+        self.learn_dqn(total_timesteps=total_timesteps)
```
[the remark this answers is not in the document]
RedTachyon: Outdated now, but my idea here was keeping some level of consistency in the API. It's a bit awkward since in SB3, `train` performs one step and `learn` does multiple steps; and in imitation, `train` performs multiple steps. But imo the important thing is the consistency of the API

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,314 @@
...
+        # If demonstrations is a list of trajectories,
+        # flatten it into a list of transitions
+        if isinstance(demonstrations, Iterable):
+            (
+                item,
+                demonstrations,
+            ) = util.get_first_iter_element(  # type: ignore[assignment]
+                demonstrations,  # type: ignore[assignment]
+            )
+            if isinstance(item, types.Trajectory):
```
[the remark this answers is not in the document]
RedTachyon: You're right, I assumed at first that a `TransitionMapping` would be an iterable over individual transition, but that's not the case. Added an additional conversion function

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,314 @@
...
+        self.expert_buffer = buffers.ReplayBuffer(
+            n_samples,
+            self.venv.observation_space,
+            self.venv.action_space,
+            handle_timeout_termination=False,
+        )
+
+        for transition in demonstrations:
+            self.expert_buffer.add(
+                obs=np.array(transition["obs"]),  # type: ignore[index]
```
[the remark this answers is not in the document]
RedTachyon: `Transitions` doesn't directly implement `__iter__`, but it inherits from `TransitionsMinimal`, which in turn inherits from/implements `typing.Sequence`.

`typing.Sequence` is essentially the same as `collections.abc.Sequence` (in fact it might be deprecated in favor of the latter in newer versions)

If we have a valid `Sequence`, i.e. something that implements `__getitem__` and `__len__`, then the ABC automatically fills in `__iter__` and a few other methods, see: https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

If we look at the `__getitem__` of `TransitionsMinimal`, it indexes each individual array and reassembles it as a dictionary or a `TransitionsMinimal`, so we get the right output.

src/imitation/algorithms/sqil.py
RedTachyon: I changed to `dqn_kwargs`, not sure about having the caller construct DQN, it could lead to some subtle bugs with the new structure.

Following the other suggestion, I changed it so that we only overwrite `ReplayBuffer`, which can be passed directly into the `DQN` constructor. But this means that if someone just creates a DQN with default settings, then everything could theoretically work, except it's using a normal replay buffer, and therefore running normal DQN. 

To avoid that, we could explicitly check if it's using the right replay buffer, but at that point it feels a bit overcomplicated, so for now I'm leaning towards creating it inside of the algorithm. If you're certain the other approach would be better, I can change it though.

I would probably consider a bring-your-own-algo approach for a potential future implementation/refactoring which would automatically work with all `OffPolicyAlgorithm`s, but this is probably out of scope for now, and has a few extra nuances and edge cases we need to think about carefully. For example, HER uses its own replay buffer implementation, and dictionary obs spaces use a `DictReplayBuffer`. We'd need to be sure that these edge cases at the very least fail loudly.

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,314 @@
...
+        assert isinstance(self.dqn.policy, policies.BasePolicy)
+        return self.dqn.policy
+
+    def train_dqn(self, gradient_steps: int, batch_size: int = 100) -> None:
+
+        # Needed to make mypy happy, because SB3 typing is shoddy
+        assert isinstance(self.dqn.policy, policies.BasePolicy)
+
+        # Switch to train mode (this affects batch norm / dropout)
+        self.dqn.policy.set_training_mode(True)
```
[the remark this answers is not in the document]
RedTachyon: This is now hidden inside of the DQN implementation, but *I think* running a `train` without breaking in a weird spot in the middle will always result in the training mode being `False` due to how the SB3 DQN loop determines the breaking condition

src/imitation/algorithms/sqil.py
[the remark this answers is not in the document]
AdamGleave: Ah, thanks for the explanation, that makes sense!

AdamGleave (commented):
New version looks significantly improved.

AdamGleave:
@ernestum can you review this at some point? I've made enough changes I shouldn't be the only reviewer on it.

ernestum:
Yes I will have a look at it after the v0.4.0 release.

ernestum (commented):
Overall this looks really neat. I really like the throughout and well-written documentation. I will do a second pass of review tomorrow (did not review all the tests and all implementation details yet)

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,215 @@
...
+        self.dqn.learn(total_timesteps=total_timesteps)
+
+    @property
+    def policy(self) -> policies.BasePolicy:
+        assert isinstance(self.dqn.policy, policies.BasePolicy)
+        return self.dqn.policy
+
+
+class SQILReplayBuffer(buffers.ReplayBuffer):
+    """Replay buffer used in off-policy algorithms like SAC/TD3, modified for SQIL."""
```
ernestum, line 94: I think this docstring should describe exactly how this special implementation of the ReplayBuffer differs form a normal one. This is too vague. Basically the docstring of the __init__ should go here.

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,215 @@
...
+        Returns:
+            A batch of samples for DQN
+        """
+        new_sample_size, expert_sample_size = util.split_in_half(batch_size)
+
+        new_sample = super().sample(new_sample_size, env)
+        new_sample.rewards.fill_(0)
+
+        expert_sample = self.expert_buffer.sample(expert_sample_size, env)
+        expert_sample.rewards.fill_(1)
```
ernestum, line 206: I think we already do this in `set_demonstrations()` so either remove it there (my preference) or here.

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,215 @@
...
+
+        n_samples = len(demonstrations)
+        self.expert_buffer = buffers.ReplayBuffer(
+            n_samples,
+            self.observation_space,
+            self.action_space,
+            handle_timeout_termination=False,
+        )
+
+        for transition in demonstrations:
```
ernestum, line 172: Why can't this loop be replaced by a simple `self.expert_buffer.set_demonstrations()`?

ernestum (changes requested):
Some more small suggestions.

src/imitation/data/rollout.py
```diff
@@ -564,6 +566,34 @@ def flatten_trajectories(
...
+            parts["obs"].append(util.safe_to_numpy(data["obs"][i]))
+            parts["next_obs"].append(util.safe_to_numpy(data["next_obs"][i]))
+            parts["acts"].append(util.safe_to_numpy(data["acts"][i]))
+            parts["dones"].append(util.safe_to_numpy(data["dones"][i]))
+            parts["infos"].append(data["infos"][i])  # type: ignore[arg-type]
+
+    cat_parts = {key: np.stack(part_list, axis=0) for key, part_list in parts.items()}
+    lengths = set(map(len, cat_parts.values()))
+    assert len(lengths) == 1, f"expected one length, got {lengths}"
+    return types.Transitions(**cat_parts)
```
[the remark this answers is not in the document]
ernestum: @AdamGleave here I need your opinion.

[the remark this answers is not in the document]
AdamGleave: `DemonstrationAlgorithm.set_demonstrations` takes a parameter of type `AnyTransitions` which includes `Iterable[types.TransitionMapping]` as an option. So if we want to implement that interface we do need this. However, the fact this interface is difficult to implement perhaps suggests we should remove this from `AnyTransitions`.

The main use case where we'd need `TransitionMapping` is if something is a PyTorch data loader. However, the key point of using that was if the data was too large to fit into memory. Converting it into a list (and then storing it in a replay buffer) eliminates that advantage.

So on balance I think it's OK to drop support for transition mappings. We should either check for that case explicitly and raise a ValueError, or amend the type signature to forbid this use.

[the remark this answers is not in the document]
ernestum: Ok as mentioned in (4.) [here](https://github.com/HumanCompatibleAI/imitation/issues/651#issuecomment-1415831640), I wanted to refactor the whole demonstrations/trajectory handling in the medium term anyways.

So I will remove support for this and throw a `NotImplementedError`. Are you fine with this?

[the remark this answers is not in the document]
AdamGleave: Yes, I'm fine with that approach.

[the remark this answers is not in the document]
AdamGleave: Sounds good re: overhauling demo handling in long term.

AdamGleave:
Asking @jas-ho to look at fixing the failing/flaky tests

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,229 @@
...
+        )
+
+        super().__init__(demonstrations=demonstrations, custom_logger=custom_logger)
+
+    def set_demonstrations(self, demonstrations: algo_base.AnyTransitions) -> None:
+        assert isinstance(self.dqn.replay_buffer, SQILReplayBuffer)
+        self.dqn.replay_buffer.set_demonstrations(demonstrations)
+
+    def train(self, *, total_timesteps: int):
+        self.dqn.learn(total_timesteps=total_timesteps)
```
jas-ho, line 85: `DQN.learn` has a bunch of optional parameters which would be good to expose imo:
```python
        callback: MaybeCallback = None, # this one seems good to have
        log_interval: int = 4,  # ...same here
        tb_log_name: str = "DQN",  # this one too; maybe override the default to sth else than "DQN"?  e.g. "SQIL-DQN" ?
        reset_num_timesteps: bool = True,  # not sure about this one
        progress_bar: bool = False,  # I definitely want to have this one :D
```
maybe just pass through the kwargs?  possibly with a different default value for "tb_log_name"

AdamGleave: Good point, passing through `**kwargs` seems like a good solution. Changing `tb_log_name` to default to `SQIL` seems reasonable.

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,229 @@
...
+        policy: Union[str, Type[DQNPolicy]],
+        custom_logger: Optional[logger.HierarchicalLogger] = None,
+        dqn_kwargs: Optional[Dict[str, Any]] = None,
+    ):
+        """Builds SQIL.
+
+        Args:
+            venv: The vectorized environment to train on.
+            demonstrations: Demonstrations to use for training.
+            policy: The policy model to use (SB3).
```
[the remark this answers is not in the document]
AdamGleave: I think it's just pointing that it needs to be a policy registered/supported by Stable Baselines3.

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,229 @@
...
+            observation_space: Observation space
+            action_space: Action space
+            demonstrations: Expert demonstrations.
+            device: PyTorch device.
+            n_envs: Number of parallel environments. Defaults to 1.
+            optimize_memory_usage: Enable a memory efficient variant
+                of the replay buffer which reduces by almost a factor two
+                the memory used, at a cost of more complexity.
+        """
+        super().__init__(
```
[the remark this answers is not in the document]
AdamGleave: Agreed, happy for you to switch it to passing as kwargs.

src/imitation/algorithms/sqil.py
```diff
@@ -0,0 +1,229 @@
...
+    def add(
+        self,
+        obs: np.ndarray,
+        next_obs: np.ndarray,
+        action: np.ndarray,
+        reward: np.ndarray,
+        done: np.ndarray,
+        infos: List[Dict[str, Any]],
+    ) -> None:
+        super().add(obs, next_obs, action, np.array(0.0), done, infos)
```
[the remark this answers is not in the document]
AdamGleave: Happy for you to make that change.

tests/algorithms/test_sqil.py
```diff
@@ -0,0 +1,131 @@
...
+):
+    demonstrations = get_demos(rng, pytestconfig, "transitions")
+    model = sqil.SQIL(
+        venv=cartpole_venv,
+        demonstrations=demonstrations,
+        policy="MlpPolicy",
+        dqn_kwargs=dict(
+            learning_starts=500,
+            learning_rate=0.002,
+            batch_size=220,
```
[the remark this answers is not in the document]
AdamGleave: Thanks for doing the hyperparam sweep! Yeah, let's fix the seed. Given it tests for a *significant* improvement in reward it should be a non-trivial test even with only a single seed (if we were just checking for any improvement then it'd be 50/50 whether it passed even if the algorithm was no better than random). We could also `@pytest.mark.parametrize` the seed for extra robustness if the test runs quick enough.

tests/algorithms/test_sqil.py
```diff
@@ -1,4 +1,6 @@
 """Tests `imitation.algorithms.sqil`."""
+from unittest.mock import MagicMock
```
AdamGleave, line 2: Nit: violates our style guide https://google.github.io/styleguide/pyguide.html#22-imports
```suggestion
from unittest import mock
```

AdamGleave: (good use of mocking though!)

tests/algorithms/test_sqil.py
```diff
@@ -129,3 +135,14 @@ def test_sqil_performance(
         rewards_before,  # type:ignore[arg-type]
         rewards_after,  # type:ignore[arg-type]
     )
+
+
+@pytest.mark.parametrize("illegal_kw", ["replay_buffer_class", "replay_buffer_kwargs"])
+def test_sqil_constructor_raises(illegal_kw: str):
+    with pytest.raises(ValueError):
```
AdamGleave, line 142: Consider using `match` to check for specific error message?

AdamGleave (commented):
Thanks for tuning hyperparams, agree seems unlikely to be a bug given high success rate after tuning.

Happy for you to make the other changes you suggested; please request a re-review from me & ernestum once done.

tests/algorithms/test_sqil.py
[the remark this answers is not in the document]
jas-ho: test_sqil_performance already takes ~20 sec on my machine. Given that I did not cherry-pick the seed I don't think additional  parametrization improves robustness enough to trade off favorably against decreased dev velocity

tests/algorithms/test_sqil.py
jas-ho: thanks for pointing it out. I hadn't realized that convention.

jas-ho:
> Thanks for tuning hyperparams, agree seems unlikely to be a bug given high success rate after tuning.
> 
> Happy for you to make the other changes you suggested; please request a re-review from me & ernestum once done.

I addressed your comments @AdamGleave . From my side it's ready for final review.

AdamGleave (approved):
LGTM

src/imitation/algorithms/sqil.py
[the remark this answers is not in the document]
AdamGleave: I don't think you ever made this change -- was that intentional omission?

AdamGleave (commented):
Please decide if you want to make this change or not prior to merging @jas-ho , but I don't need to re-review just for that.

src/imitation/algorithms/sqil.py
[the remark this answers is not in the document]
jas-ho: hmm, might be an issue with how gh displays it?  I actually changed this yesterday already. anyway, it's resolved now :)

jas-ho:
> Please decide if you want to make this change or not prior to merging @jas-ho , but I don't need to re-review just for that.

That change was implemented already so it might have just been a gh display issue. -> LGTM :)

ernestum (approved):
LGTM

Issue #718, opened 2023-05-23
Example of using custom environments


UnnamedMoose:
## Problem
I have a custom environment derived from the Gym base class that works fine with stable baselines 3, including SubprocVecEnv, but I am struggling to integrate it with this framework because of the different wrapper layers.

## Solution
Having an example of a simple custom env (even a copy of on of the gym environments) would be very helpful to someone new to imitation like myself.

## Possible alternative solutions
A simple paragraph or two explaining how to use different environments instead of relying on the util::make_vec_env function would also be appreciated.

ernestum:
Good point! To (maybe) unblock you for now I recommend to 
1. reigster you env using `gym.register()` with some unique ID
2. use `util.make_vec_env()` to create your environment with the given unique ID

UnnamedMoose:
Thanks, I found a workaround by writing my own routine for creating the initial rollouts as that seemed to be the major issue when interfacing with my env as well as a PID controller that emulates a stable baselines model (apparently not well enough).  I'm pasting my code below as it may prove useful to others.

```
import matplotlib
import matplotlib.pyplot as plt
import numpy as np
import pandas
import datetime
import os

import stable_baselines3
from stable_baselines3.common.vec_env import VecMonitor, SubprocVecEnv
from stable_baselines3.common.noise import NormalActionNoise, VectorizedActionNoise
import torch

from imitation.algorithms.adversarial.gail import GAIL
from imitation.algorithms.adversarial.airl import AIRL
from imitation.data.types import TrajectoryWithRew
from imitation.rewards.reward_nets import BasicRewardNet, BasicShapedRewardNet
from imitation.util.networks import RunningNorm

import verySimpleAuv as auv
import resources

font = {"family": "serif",
        "weight": "normal",
        "size": 16}

matplotlib.rc("font", **font)
matplotlib.rcParams["figure.figsize"] = (9, 6)

# %% Set up.
if __name__ == "__main__":
    # An ugly fix for OpenMP conflicts in my installation.
    os.environ["KMP_DUPLICATE_LIB_OK"] = "TRUE"

    agentName = "SAC_sblPretrain_try0_fromPID"

    # --- Pretraining parameters ---
    nPretrainEpisodes = 100
    nPretrainSteps = 500

    # Note that the bounds check is disabled to ensure episodes are of equal length.
    env_kwargs_pretrain = {
        "currentVelScale": 1.,
        "currentTurbScale": 2.,
[...]

PR #746, opened 2023-07-05
Custom environment tutorial

AdamGleave:
Tutorial about using a custom environment with imitation, see https://github.com/HumanCompatibleAI/imitation/issues/718


AdamGleave:
Note: reviewed in #737.

PR #761, opened 2023-07-25
Additional examples of CLI usage

### LINKED ISSUE #748: Add tutorials for the common use cases of the CLI
Those include:

- training with new demonstrations,
- comparing to a baseline

EdoardoPona:
## Description

Added some examples of cli use cases, specifically 

- training and expert saving demonstrations explicitly, then training a policy reading those from disk. 
- visualising and rendering a saved policy 
- comparing performances of two policies (even though this requires a small bit of python) 

Fixes #748 

### Testing
Documentation only changes

### DIFF
--- a/docs/getting-started/cli.rst
+++ b/docs/getting-started/cli.rst
@@ -32,8 +32,8 @@ You can always find out all the configurable values by running:
 
     python -m imitation.scripts.<script> print_config
 
-Run BC on the ``CartPole-v1`` environment with a pre-trained PPO policy as expert:
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Run BC on the ``CartPole-v1`` environment with a pre-trained PPO policy as expert
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. note:: Here the cartpole environment is specified via a named configuration.
 
@@ -49,8 +49,8 @@ Run BC on the ``CartPole-v1`` environment with a pre-trained PPO policy as exper
 50 expert demonstrations are sampled from the PPO policy that is included in the testdata folder.
 2000 batches are enough to train a good policy.
 
-Run DAgger on the ``CartPole-v0`` environment with a random policy as expert:
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Run DAgger on the ``CartPole-v0`` environment with a random policy as expert
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: bash
 
@@ -63,8 +63,8 @@ Run DAgger on the ``CartPole-v0`` environment with a random policy as expert:
 This will not produce any meaningful results, since a random policy is not a good expert.
 
 
-Run AIRL on the ``MountainCar-v0`` environment with a expert from the HuggingFace model hub:
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Run AIRL on the ``MountainCar-v0`` environment with a expert from the HuggingFace model hub
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/cli.rst
```diff
@@ -93,6 +93,164 @@ The ``seals:`` prefix ensures that the seals package is imported and the environ
...
+First, train an expert and save the demonstrations. 
+Note that by default these are saved in ``<log_dir>/rollouts/final.npz``.
+Where for this script by default ``<log_dir>`` is ``output/train_rl/<environment>/<timestamp>`` .  
+However, we can pass an explicit path as logging directory. 
+By default, this will use ``ppo``. 
+
+.. code-block:: bash
+
+        python -m imitation.scripts.train_rl with pendulum \
+                logging.log_dir=output/train_rl/Pendulum-v1/my_run  \
```
AdamGleave, line 108: May want to illustrate the `rollout_save_n_{timesteps,episodes}` flags here. Could also potentially mention `eval_policy` if you want to get rollouts from an already trained expert.

docs/getting-started/cli.rst
```diff
@@ -93,6 +93,164 @@ The ``seals:`` prefix ensures that the seals package is imported and the environ
...
+    python -m imitation.scripts.eval_policy with \
+            expert.policy_type=ppo \
+            expert.loader_kwargs.path=output/train_rl/Pendulum-v1/my_run/policies/final/model.zip \
+            environment.num_vec=1 \
+            render=True \
+            environment.gym_id='Pendulum-v1' 
+
+
+
+Comparing algorithm performances
```
AdamGleave, line 137:
```suggestion
Comparing algorithms' performance
```

docs/getting-started/cli.rst
```diff
@@ -93,6 +93,164 @@ The ``seals:`` prefix ensures that the seals package is imported and the environ
...
+            expert.loader_kwargs.path=output/train_rl/Pendulum-v1/my_run/policies/final/model.zip \
+            environment.num_vec=1 \
+            render=True \
+            environment.gym_id='Pendulum-v1' 
+
+
+
+Comparing algorithm performances 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^    
+Let's use the cli to compare the performances of different algorithms.
```
AdamGleave, line 139:
```suggestion
Let's use the CLI to compare the performance of different algorithms. 
```

docs/getting-started/cli.rst
```diff
@@ -93,6 +93,164 @@ The ``seals:`` prefix ensures that the seals package is imported and the environ
...
+            'len_std': 15.686012049373561, 
+            'len_max': 104, 
+            'monitor_return_min': 8.0, 
+            'monitor_return_mean': 28.92676056338028, 
+            'monitor_return_std': 15.686012049373561, 
+            'monitor_return_max': 104.0
+    }
+    INFO - eval_policy - Completed after 0:00:17
+
+This will save the monitor csvs (one for each vectorised env, controlled by environment.num_vec).
```
AdamGleave, line 227:
```suggestion
This will save the monitor CSVs (one for each vectorised env, controlled by environment.num_vec). 
```

docs/getting-started/cli.rst
```diff
@@ -93,6 +93,164 @@ The ``seals:`` prefix ensures that the seals package is imported and the environ
...
+            'monitor_return_std': 15.686012049373561, 
+            'monitor_return_max': 104.0
+    }
+    INFO - eval_policy - Completed after 0:00:17
+
+This will save the monitor csvs (one for each vectorised env, controlled by environment.num_vec). 
+We can load these with ``pandas`` and use the ``imitation.test.reward_improvement``
+module to compare the performances of the two policies.
+
+.. code-block:: python
```
AdamGleave, line 231: Is this the way we want to recommend? What about if multiple environments (this will only read the first monitor one)? It already prints out monitor return mean and SD so I think we should be able to do a statistical test directly from that.

AdamGleave:
@jas-ho could you take a quick look at this? You can accept my suggestions where appropriate and make any edits you deem fit, then I can re-review.

docs/getting-started/cli.rst
jas-ho: Would be great if we could have a CLI for evaluating reward improvements as well (instead of falling back to Python). The CLI could take care of iterating over contained csv files if you specify a folder instead of a single file.

jas-ho (changes requested):
Looks good. The only change I'd insist on is fixing in some way the comparison of rewards between expert and non-expert around line 233.  Ideally with a CLI for comparing rewards; but just making sure all CSVs are included would be fine as well.

@AdamGleave I left some comments.

docs/getting-started/cli.rst
jas-ho: both fixed in 4807a15920953cb21f1ff593d84cc1ae9871cf04 and ccfc99d18aa5a4243b9712d1066f7321c1d87f8c , respectively.

docs/getting-started/cli.rst
jas-ho: ok, I checked with @ernestum who apparently has a tool for that already. Therefore, it does not make sense to implement sth independently right now. 
Instead I vote for just adapting the current code-block to consume all generated monitor CSVs and leaving a todo comment to fix this when a CLI tool for handling significance testing becomes available.

jas-ho: (implemented in 7f39df3fb384fa542f9c1a7746372ab3ef281f99)

jas-ho (approved):
@ernestum can you have a look?

docs/getting-started/cli.rst
ernestum: I like it a lot better that we promote doing random permutation tests here. The code surfaces the weirdness in which we store the output of the experiments but that is nothing we can fix in this PR.
Maybe add a bit more context in the print statement like:
```python
if is_signifcant_reward_improvement(...):
    print("The expert improved over the non-expert with 95% probability")
else:
    print("No significant (p=0.05) reward improvement of expert over non-expert")

**Review by ernestum — COMMENTED** [review]
LGTM except for the two points I made.

File: docs/getting-started/cli.rst (hunk 5 above)
Comment (jas-ho) [reply docs/getting-started/cli.rst:253]: thanks for the suggestions, I took it.

File: docs/getting-started/cli.rst [hunk 6]
```diff
@@ -93,6 +93,196 @@ The ``seals:`` prefix ensures that the seals package is imported and the environ
...
+
+Note that the rollout_save_path is relative to the ``log_dir`` of the imitation script.
+
+.. code-block:: bash
+
+        python -m imitation.scripts.eval_policy with seals_cartpole \
+                expert.policy_type=ppo-huggingface \
+                eval_n_episodes=50 \
+                logging.log_dir=output/ppo/seals_cartpole/loaded \
+                rollout_save_path=rollouts/final.npz
```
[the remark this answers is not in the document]
jas-ho: good point. fixed

Issue #764, opened 2023-07-28
Don't train experts in the tutorials, download from HF instead


ernestum:
## Problem
We waste quite some time and LOC in the tutorials on training experts.

## Solution
Replace by code to download experts from the HuggingFace Model hub, where we already have them.

## Possible alternative solutions
None

AdamGleave:
Assigning to @jas-ho

Issue #574, opened 2022-10-07
Replace Callable with Protocol + ParamSpec


Rocamonde:
Type definitions like
https://github.com/HumanCompatibleAI/imitation/blob/4232b454b61181e322a666b7ef4c684afa587516/src/imitation/policies/serialize.py#L23-L23

can (and should be, for better type safety) replaced with the usage of Protocol + ParamSpec.

Example:

```python
from typing import ParamSpec, Protocol

P = ParamSpec("P")

class PolicyLoaderFn(Protocol[P]):
    def __call__(self, venv: vec_env.VecEnv, *args: P.args, **kwargs: P.kwargs) -> policies.BasePolicy:
        ...
```

PR #766, opened 2023-08-04
Download experts from hf inside tutorials and docs

jas-ho:
## Description

This updates the following files to download pretrained experts from the HuggingFace Model hub (instead of training them from scratch):
- most notebooks in `docs/tutorials`
- some algorithm doc files in `docs/algorithms`
- the example script `examples/quickstart.py`

This saves computation time and allows users to experiment with competent, fully trained experts.

Note that this required changing the default example environment from "CartPole-v1"  to "seals/CartPole-v0" in some places (since https://huggingface.co/HumanCompatibleAI only contains pretrained experts for the latter).

The only remaining calls to `expert.train` are now in  
- `docs/tutorials/8_train_custom_env.ipynb` where we train on a custom environment (hence, no pretrained expert available)
- `tests/data/test_rollout.py` where experts are trained for a single timestep (hence, negligible overhead)

Solves #764 

## Testing

I have run all notebooks and the example script and checked that they still go through.


AdamGleave (commented):
Thanks, these changes mostly looks good but there's a type error as the code is currently written.

The tutorials are typically passing in `venv=env` where `env` is a `gym.Env` created by `gym.make`. But `load_policy` expects a `vec_env.VecEnv`.

Clearly the notebooks work with this as they're passing tests, so one option to fix this would be to loosen up the type annotation on `load_policy` to a `stable_baselines3.common.type_aliases.GymEnv = Union[gym.Env, vec_env.VecEnv]` since this is what `BaseAlgorithm.load` accepts, although there's a chance this will cause type checking errors elsewhere (e.g. do any of our other policy loaders strictly require a `vec_env.VecEnv`)?

Alternatively we could stick to the `vec_env.VecEnv` convention we've mostly used in `imitation` (reasoning that `VecEnv` is strictly more general than a `gym.Env`, you can always have a `VecEnv` with a single `gym.Env`) and change your `env`'s to a `venv` by e.g. using the `util.make_vec_env` function.

jas-ho:
Thanks for spotting that @AdamGleave .

jas-ho:
> one option to fix this would be to loosen up the type annotation on load_policy to a stable_baselines3.common.type_aliases.GymEnv = Union[gym.Env, vec_env.VecEnv] since this is what BaseAlgorithm.load accepts, although there's a chance this will cause type checking errors elsewhere (e.g. do any of our other policy loaders strictly require a vec_env.VecEnv)?

Neither `mypy` nor `pytype` complain if the annotation is changed to `GymEnv`. Looking over the loader factories in `registry.py` and `serialize.py` however shows that this is actually inconsistent with the type annotations in those loaders (which also expect `VecEnv` acc. to the type annotation). 

This is because the typing of the policy registry is insufficient
```python
agent_loader = policy_registry.get(policy_type)
reveal_type(agent_loader)
```
gives `Revealed type is "def (*Any, **Any) -> stable_baselines3.common.policies.BasePolicy"`  where actually it should be "def (*Any, **Any) -> stable_baselines3.common.policies.BasePolicy"`

There is an open issue for that already: https://github.com/HumanCompatibleAI/imitation/issues/574

jas-ho:
Given how widespread the assumption of working with `vec_env.VecEnv` is atm the simpler solution you suggested seems safer to me:
> stick to the vec_env.VecEnv convention we've mostly used in imitation (reasoning that VecEnv is strictly more general than a gym.Env, you can always have a VecEnv with a single gym.Env) and change your env's to a venv by e.g. using the util.make_vec_env function.

docs/algorithms/dagger.rst
```diff
@@ -26,19 +26,24 @@ Detailed example notebook: :doc:`../tutorials/2_train_dagger`
...
-    expert.learn(1000)
-    venv = DummyVecEnv([lambda: gym.make("CartPole-v1")])
+    env = gym.make("seals/CartPole-v0")
+    expert = load_policy(
+        "ppo-huggingface",
+        organization="HumanCompatibleAI",
+        env_name="seals-CartPole-v0",
+        venv=env,
+    )
+    venv = DummyVecEnv([lambda: gym.make("seals/CartPole-v0")])
```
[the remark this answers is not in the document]
jas-ho: I don't think so

docs/tutorials/3_train_gail.ipynb
jas-ho, line 1: @AdamGleave when testing it seemed like docs/tutorials/3_train_gail.ipynb and docs/tutorials/4_train_airl.ipynb might be broken. The trained policies achieve lower rewards than the untrained ones. I've actually checked that the switch to the VecEnv is not the cause of this behavior. I'll have to check on master to be sure

docs/algorithms/dagger.rst
```diff
@@ -24,21 +24,28 @@ Detailed example notebook: :doc:`../tutorials/2_train_dagger`
...
 
     rng = np.random.default_rng(0)
-    env = gym.make("CartPole-v1")
-    expert = PPO(policy=MlpPolicy, env=env)
-    expert.learn(1000)
-    venv = DummyVecEnv([lambda: gym.make("CartPole-v1")])
+    env = make_vec_env(
+        "seals/CartPole-v0",
+        rng=rng,
+        n_envs=1,
```
jas-ho, line 41: @AdamGleave just wanted to flag that I had to set `n_envs=1` to achieve the maximum reward. For the default settings, rewards fluctuated in the 350 to 450 range.

jas-ho: nevermind. increasing the number of time_steps in dagger_trainer.train fixes that

jas-ho (commented):
@AdamGleave it's ready for a final review.

docs/tutorials/3_train_gail.ipynb
```diff
@@ -93,53 +89,58 @@
...
+    "    learner, env, 100, return_episode_rewards=True\n",
     ")"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "When we look at the histograms of rewards before and after learning, we can see that the learner is not perfect yet, but it made some progress at least.\n",
-    "If not, just re-run the above cell."
```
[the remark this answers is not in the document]
jas-ho: (note that the "When we look at the histograms ..." part is still there, the diff is just a bit unfortunate

docs/tutorials/3_train_gail.ipynb
```diff
@@ -93,53 +89,58 @@
...
     "    reward_net=reward_net,\n",
     ")\n",
     "\n",
+    "# evaluate the learner before training\n",
+    "env.seed(SEED)\n",
     "learner_rewards_before_training, _ = evaluate_policy(\n",
-    "    learner, venv, 100, return_episode_rewards=True\n",
+    "    learner, env, 100, return_episode_rewards=True\n",
     ")\n",
-    "gail_trainer.train(20000)  # Note: set to 300000 for better results\n",
```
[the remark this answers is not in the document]
AdamGleave: Training performance should not be non-monotonic, points to something else wrong with the code/hyperparams, but I guess let's not have a misleading comment.

docs/algorithms/gail.rst
```diff
@@ -29,35 +29,47 @@ Detailed example notebook: :doc:`../tutorials/3_train_gail`
...
-    rng = np.random.default_rng(0)
+    SEED = 42
 
     env = make_vec_env(
         "seals/CartPole-v0",
-        rng=rng,
-        n_envs=8,
-        post_wrappers=[lambda env, _: RolloutInfoWrapper(env)],  # for computing rollouts
+        rng=np.random.default_rng(SEED),
+        n_envs=5,
```
AdamGleave, line 41: Why 5? Why do we make `env` twice here?

docs/algorithms/airl.rst
```diff
@@ -52,14 +53,22 @@ Detailed example notebook: :doc:`../tutorials/4_train_airl`
...
         expert,
         env,
-        rollout.make_sample_until(min_timesteps=None, min_episodes=60),
-        rng=rng,
+        rollout.make_sample_until(min_episodes=60),
+        rng=np.random.default_rng(SEED),
     )
 
-    learner = PPO(env=env, policy=MlpPolicy)
+    env = make_vec_env("seals/CartPole-v0", rng=np.random.default_rng(SEED))
```
AdamGleave, line 60: Why do we re-create `env` rather than reusing the previous `env` object (perhaps reseeding as needed to make deterministic)?

docs/tutorials/3_train_gail.ipynb
```diff
@@ -93,53 +89,58 @@
...
     "from stable_baselines3 import PPO\n",
+    "from stable_baselines3.ppo import MlpPolicy\n",
     "from stable_baselines3.common.evaluation import evaluate_policy\n",
-    "from stable_baselines3.common.vec_env import DummyVecEnv\n",
-    "\n",
-    "import gym\n",
     "\n",
     "\n",
-    "venv = make_vec_env(\"seals/CartPole-v0\", n_envs=8, rng=rng)\n",
+    "env = make_vec_env(\"seals/CartPole-v0\", n_envs=8, rng=np.random.default_rng(SEED))\n",
```
AdamGleave, line 100: Why are we re-creating the environment rather than using existing `env`?

docs/tutorials/4_train_airl.ipynb
```diff
@@ -92,51 +88,52 @@
...
     "from stable_baselines3 import PPO\n",
+    "from stable_baselines3.ppo import MlpPolicy\n",
     "from stable_baselines3.common.evaluation import evaluate_policy\n",
     "\n",
-    "import gym\n",
-    "import seals\n",
-    "\n",
     "\n",
-    "venv = make_vec_env(\"seals/CartPole-v0\", n_envs=8, rng=rng)\n",
+    "env = make_vec_env(\"seals/CartPole-v0\", n_envs=8, rng=np.random.default_rng(SEED))\n",
```
AdamGleave, line 96: Why are we re-creating the environment rather than using existing `env`?

AdamGleave (commented):
Thanks for the changes! Looks like we might be able to reduce code duplication by re-using some VecEnv objects; WDYT?

docs/algorithms/gail.rst
jas-ho: not sure where this came from :sweat_smile:   but it's not needed. fixed and reuses the same env now

docs/algorithms/airl.rst
jas-ho: fixed

docs/tutorials/3_train_gail.ipynb
jas-ho: it's indeed unnecessary. fixed now

docs/tutorials/4_train_airl.ipynb
jas-ho: fixed

jas-ho:
> Thanks for the changes! Looks like we might be able to reduce code duplication by re-using some VecEnv objects; WDYT?

Good catch! I grepped and did not find other instances of redundant env creations in tutorials or docs. So should be good for final review and merge @AdamGleave

AdamGleave (approved):
LGTM

PR #768, opened 2023-08-10
Use data acquired by users

jas-ho:
## Description

This PR introduces a new abstract base policy `InteractivePolicy` to support interactive data collection from human experts.

There are two implementations for demo purposes:

- `TextInteractivePolicy` which supports simple Toy Text environments like FrozenLake
- `AtariInteractivePolicy` which is meant to support Atari games like Pong. This one is work in progress with many missing pieces. The idea is to rely on some code from retro-gym which is added in this PR with very slight modifications at src/imitation/policies/retro_gym_.py

### Limitations

- The current interface design will not be sufficient to support continuous time applications like the Pong game.
- It's unclear whether we want / can support `retro-gym` environments. Maybe there are other options for demonstrating continuous-time data gathering?
- due to limited time there are still a _lot_ of unsolved todo comments
- ...and only very basic testing

## Testing

To explore the behavior of `TextInteractivePolicy` you can run 

- ` python src/imitation/policies/interactive_text.py` to observe a single episode rollout of `TextInteractivePolicy` on FrozenLake
- `python examples/train_dagger_with_human_demos.py` to run a script collecting human demonstration data on FrozenLake and using it to train DAgger
- play with the very rudimentary tests at tests/policies/test_interactive.py

To test that the interactive code taken from retro-gym is working you can `pip install gym-retro`  and then run

- `python src/imitation/policies/retro_gym_.py`

There are no demo scripts or tests for our wrapper `AtariInteractivePolicy` yet since that is still WIP.


jas-ho:
PS: also a lot of clean-up still needed to pass linters..

PR #731, opened 2023-06-17
Change adversarial algorithms to collect rollouts first

taufeeque9:
## Description

This PR changes the adversarial algorithm such that the at any iteration, the rollouts are collected first, and then the discriminator is trained, followed by training the generator. This modification matches Algorithm 1 given in the [AIRL paper](https://arxiv.org/pdf/1710.11248.pdf).

## Testing

The proposed change improves the returns obtained on many environments. The table below shows the imitation-to-expert return ratio of the algorithms on several environments. The results were obtained by tuning the hyperparameters for each environment separately. The return ratio was obtained by evaluating the tuned hyperparameters on five distinct seeds and calculating the average return ratio to the expert's return.

Algo \ Env | Ant   | Half Cheetah | Hopper | Swimmer | Walker |
-----------|-------|--------------|--------|---------|--------|
GAIL-PR        | 0.883 | 0.868        | 1.01   | 0.986   | 0.989  |
AIRL-PR        | -0.04 | 0.993        | 1.01   | 0.926   | 0.270  |
GAIL-Master | 0.864 | 0.981        | 1.004  | 0.945   | 0.893  |
AIRL-Master | 0.259 | 0.447        | 1.008  | 0.663   | 0.176  |

PR #770, opened 2023-08-10
Adversarial algorithm matching original paper's implementation

taufeeque9:
## Description

This PR updates the adversarial algorithm by training the discriminator between collecting the rollouts of the generator and training the generator. This matches the reference implementation provided in Algorithm 1 of the [AIRL paper](https://arxiv.org/pdf/1710.11248.pdf).

The modification is done by implementing the `TrainDiscriminatorCallback`, which is called to train the discriminator after collecting rollouts through the `callback.on_rollout_end()`. The callback first stores the latest rollout in the replay buffer, which is then used to train the discriminator. Once the discriminator is trained, the callback updates the generator's rollout/replay buffer by updating the rewards using the latest discriminator. 

Note that we must also update the advantages and returns in the rollout buffer of the on-policy algorithms upon updating the rewards. This is tricky to do since information like `value` and `done` on the last observations of the rollouts is not stored in the rollout buffer. These are obtained in this PR by using the original advantages and rewards. A test of whether it produces correct values needs to be added.

## Testing

All the tests for adversarial algorithms run successfully.

The performance of this implementation remains to be compared to the [other implementation](#731) and the master branch's algorithm.

### DIFF
--- a/src/imitation/algorithms/adversarial/common.py
+++ b/src/imitation/algorithms/adversarial/common.py
@@ -114,6 +114,8 @@ class TrainDiscriminatorCallback(callbacks.BaseCallback):
         return True
 
     def _on_rollout_end(self) -> None:
+        if self.gen_ctx_manager is not None:
+            self.exit_gen_ctx_manager()
         gen_trajs, ep_lens = self.adversarial_trainer.venv_buffering.pop_trajectories()
         self.adversarial_trainer._check_fixed_horizon(ep_lens)
         gen_samples = rollout.flatten_trajectories_with_rew(gen_trajs)
@@ -132,9 +134,13 @@ class TrainDiscriminatorCallback(callbacks.BaseCallback):
         self.gen_ctx_manager = self.adversarial_trainer.logger.accumulate_means("gen")
         self.gen_ctx_manager.__enter__()
 
-    def _on_training_end(self) -> None:
+    def exit_gen_ctx_manager(self) -> None:
         assert self.gen_ctx_manager is not None
         self.gen_ctx_manager.__exit__(None, None, None)
+        self.gen_ctx_manager = None
+
+    def _on_training_end(self) -> None:
+        self.exit_gen_ctx_manager()
 
 
 class AdversarialTrainer(base.DemonstrationAlgorithm[types.Transitions]):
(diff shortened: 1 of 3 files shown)


src/imitation/policies/replay_buffer_wrapper.py
```diff
@@ -23,6 +23,48 @@ def _samples_to_reward_fn_input(
...
 
+def _rollout_buffer_to_reward_fn_input(
+    buffer: RolloutBuffer,
+) -> Mapping[str, np.ndarray]:
+    """Convert a sample from a rollout buffer to a numpy array."""
+    assert buffer.observations is not None
+    assert buffer.actions is not None
+    obs = buffer.observations
+    next_obs = obs[1:]
+    next_obs = np.concatenate([next_obs, obs[-1:]], axis=0)  # last obs not available
```
ernestum, line 34: Easier to read if you do:

```python
ext_obs = np.concatenate([obs[1:], obs[-1:]], axis=0)
```

src/imitation/policies/replay_buffer_wrapper.py
```diff
@@ -23,6 +23,48 @@ def _samples_to_reward_fn_input(
...
+    actions = buffer.actions
+    dones = buffer.episode_starts
+    dones = np.roll(dones, -1, axis=0)
```
ernestum, line 37: same as above: pull `buffer.episode_starts` in this line.

src/imitation/algorithms/adversarial/common.py
```diff
@@ -222,16 +279,22 @@ def __init__(
 
         self.venv_buffering = wrappers.BufferingWrapper(self.venv)
 
+        self.disc_trainer_callback = TrainDiscriminatorCallback(self)
```
ernestum, line 282: Why not define the `gen_callback` here like this:

```python
self.gen_callback: List[callbacks.BaseCallback] = [self.disc_trainer_callback]
```

and then just append it down in the else block?

And while you are at it rename it to use a plural because it is actually more than one callback. E.g. `self.gen_callbacks`.

ernestum (commented):
Just some nits. Throughout review will follow.

PR #772, opened 2023-09-04
Tune hyperparameters in tutorials for GAIL and AIRL

zajaczajac:
## Description

This PR tunes hyperparameters for the GAIL and AIRL tutorials. 
For GAIL, the expert performance is reached (~500 on CartPole) with 800K PPO steps (~2 min run time on MacBook Air M1). For AIRL, the default is the "fast" version which improves over random but does not reach the expert performance (800K steps, ~2 min run time); if we switch off "fast" then the expert performance is reached (2M steps, ~5 min run time).

The hyperparameters were inspired by configs for half-cheetah from the `benchmarking` directory + a bit of manual tuning. Also, for GAIL I needed to change from `BasicShapedRewardNet` to `BasicRewardNet` to make it work (not exactly sure why but it affected performance a lot!).

## Testing

Just ran the notebooks, and also tested with a few different seeds to make sure results are stable.

### DIFF
--- a/docs/algorithms/airl.rst
+++ b/docs/algorithms/airl.rst
@@ -60,9 +60,13 @@ Detailed example notebook: :doc:`../tutorials/4_train_airl`
     learner = PPO(
         env=env,
         policy=MlpPolicy,
-        batch_size=16,
-        learning_rate=0.0001,
-        n_epochs=2,
+        batch_size=64,
+        ent_coef=0.0,
+        learning_rate=0.0005,
+        gamma=0.95,
+        clip_range=0.1,
+        vf_coef=0.1,
+        n_epochs=5,
         seed=SEED,
     )
     reward_net = BasicShapedRewardNet(
@@ -72,9 +76,9 @@ Detailed example notebook: :doc:`../tutorials/4_train_airl`
     )
     airl_trainer = AIRL(
         demonstrations=rollouts,
-        demo_batch_size=1024,
-        gen_replay_buffer_capacity=2048,
-        n_disc_updates_per_round=4,
+        demo_batch_size=2048,
+        gen_replay_buffer_capacity=512,
+        n_disc_updates_per_round=16,
         venv=env,
         gen_algo=learner,
         reward_net=reward_net,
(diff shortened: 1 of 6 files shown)


ernestum (approved):
Thanks for the contribution! I meant to do this for quite some time! 
The changes itsel LGTM.
I think the pipeline fails because we get the newest seals version (0.2) which is made for gymnasium. If we change our seals version specifier in `setup.py` to `seals~=0.1.5`, this should be fixed.

docs/algorithms/airl.rst
```diff
@@ -84,7 +88,7 @@ Detailed example notebook: :doc:`../tutorials/4_train_airl`
     learner_rewards_before_training, _ = evaluate_policy(
         learner, env, 100, return_episode_rewards=True,
     )
-    airl_trainer.train(20000)
+    airl_trainer.train(20000)  # Train for 2_000_000 steps to match expert.
```
AdamGleave, line 91: 2 million timesteps is a lot of timesteps for something as simple as CartPole, I expect we can do better but this seems fine for the purpose of this PR, at least the environment runs quickly.

docs/tutorials/3_train_gail.ipynb
```diff
@@ -146,17 +147,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import matplotlib.pyplot as plt\n",
-    "\n",
-    "print(\"mean reward after training:\", np.mean(learner_rewards_after_training))\n",
-    "print(\"mean reward before training:\", np.mean(learner_rewards_before_training))\n",
-    "\n",
-    "plt.hist(\n",
```
AdamGleave, line 154: Why are you removing histogram (here and in AIRL)? Fine to remove if it's not informative. But perhaps we should report the SD as well as the means?

zajaczajac: Yeah, the reason was I thought it was not super informative (especially in case we reach expert perf). Good suggestion with SD though, will add!

docs/algorithms/airl.rst
zajaczajac: I'd keep it for now (it's already an improvement) and possibly revisit in another PR.

AdamGleave (approved):
LGTM

docs/tutorials/3_train_gail.ipynb
zajaczajac: @ernestum , thanks for this, the lib looks quite cool! I'll remember about it in the future. For this PR I decided to not introduce additional dependency though.

Issue #701, opened 2023-04-25
Use data acquired by users


AdrianPrados:
I am testing 'imitation' to work with proprietary environments for the robot in our Mujoco-based lab. For testing I am generating Pygame environments based on Gym. I am creating user generated data. When working with BC it is clear that those demos have to be in Trajectory format (act,obs,infos,terminal) and inside bc.BC in demonstrations (something like demosntration = exppert_trajs). When I work with Dagger, it is not very clear in the documentation how to do this process. The bc.BC is the same as a normal BC process, but in SimpleDAggerTrainer, it is not clear which expert_policy should be added. I have tried using data generated by the PPO and there it is clear that the expert_policy = expert (where expert = PPO(...), expert.learn(...)), but if I want to use human data I am not sure what I should do. Do I need to use an expert that is for example a PPO? Do I need to do a bc_trainer.learn(...) and use that result as expert_policy inside SimpleDAggerTrainer, do I need to get the expert_policy from the acquired human data, if so, how can I generate the policy from this human data?

![DaggerQuestion](https://user-images.githubusercontent.com/95371499/234261297-b92cda0c-6198-44b7-9729-dd7c6ba204d5.png)

I have uploaded an example of a code where Arm2d is my own environment (a robotic arm in 2 dimensions), expert_trajs are 30 action state pairs data taken using PyGame and expert_policy in SimpleDAggerTrainer in this case has been tested with expert (a trained PPO with 10000 iterations). I don't know if this configuration is correct, or if as I have commented previously, it is necessary to do a BC training and use its policy as expert_policy or if I need to get the Policy from the user data (and if this is the option, how it would be done).

Neo-manchester:
Yeah I have the same question, wondering how to apply this imitation library on human demonstrations or real robot demonstration.

AdamGleave:
@jas-ho can you take a look at this?

My understanding is that `imitation` should work with human-generated demonstrations just fine, so long as they're of type `imitation.algorithms.base.AnyTransitions` (e.g. a sequence of trajectories).

However, we don't currently have any out-of-the-box demo. Collecting arbitrary human demos is out of scope (think of how different e.g. demos from motion capture of a person performing a task, demos from a person teleoperating a robot or demos of humans writing text solving tasks are). But being able to do this in some simple RL environments seems valuable.

[...]

PR #776, opened 2023-09-06
Introduce interactive policies to gather data from a user

zajaczajac:
## Description

This PR introduces interactive policies that query the user for actions, as requested in #701.
Such policies can be used e.g. in Behavioral cloning or DAgger.
An example showing the use for Atari is included.
Acknowledgement: tests were heavily based on the ones from https://github.com/HumanCompatibleAI/imitation/pull/768 by @jas-ho. 

## Testing

`pytest tests/policies/test_interactive.py` to run unit tests.
`python examples/train_dagger_atari_interactive_policy.py` to run the interactive demo.


tests/policies/test_interactive.py
```diff
@@ -0,0 +1,53 @@
+"""Tests interactive policies."""
```
zajaczajac, line 1: This has been based very heavily on https://github.com/HumanCompatibleAI/imitation/pull/768 . How to acknowledge it properly?

examples/train_dagger_atari_interactive_policy.py
```diff
@@ -0,0 +1,49 @@
...
+from imitation.policies import interactive
+
+if __name__ == "__main__":
+    rng = np.random.default_rng(0)
+
+    env = vec_env.DummyVecEnv([lambda: gym.wrappers.TimeLimit(gym.make("Pong-v4"), 10)])
+    env.seed(0)
+
+    action_names = env.envs[0].get_action_meanings()
+    names_to_keys = {
```
AdamGleave, line 17: There's only a small finite number of legal actions in Atari, so we could define a more comprehensive version of these in a constant somewhere (or even subclass ImageObsDiscreteInteractivePolicy to handle this directly) rather than it having to live in an example.

examples/train_dagger_atari_interactive_policy.py
```diff
@@ -0,0 +1,49 @@
+import tempfile
```
AdamGleave, line 1: Missing module docstring

src/imitation/policies/interactive.py
```diff
@@ -0,0 +1,84 @@
...
+    ):
+        super().__init__(
+            observation_space=observation_space,
+            action_space=action_space,
+        )
+
+        assert isinstance(action_space, gym.spaces.Discrete)
+        assert len(action_names) == len(action_keys) == action_space.n
+        # Names and keys should be unique.
+        assert len(set(action_names)) == len(set(action_keys)) == action_space.n
```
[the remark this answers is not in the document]
zajaczajac: Yeah, I did this because of ordering, and had doubts about enforcing `OrderedDict` type, but now I think I agree it's more elegant so changing into `OrderedDict`.

tests/policies/test_interactive.py
```diff
@@ -0,0 +1,53 @@
...
+]
+
+
+class NoRenderingDiscreteInteractivePolicy(interactive.DiscreteInteractivePolicy):
+    def _render(self, obs: np.ndarray) -> None:
+        pass
+
+
+@pytest.mark.parametrize("env_name", ENVS)
+def test_interactive_policy(env_name: str):
```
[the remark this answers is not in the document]
zajaczajac: Done, thanks for suggestions!

src/imitation/policies/interactive.py
```diff
@@ -0,0 +1,148 @@
...
+
+    For each query, the observation is rendered and then the action is provided
+    as a keyboard input.
+    """
+
+    def __init__(
+        self,
+        observation_space: gym.Space,
+        action_space: gym.Space,
+        action_keys_names: collections.OrderedDict,
```
zajaczajac, line 27: Ideally, we'd have `OrderedDict[str, str]` as a type here. Note however that depending on the Python version, this would need to be imported either from `collections` or `typing`, so omitting it to avoid depending on a concrete version. Reference: https://stackoverflow.com/questions/41207128/how-do-i-specify-ordereddict-k-v-types-for-mypy-type-annotation#comment115837552_43583996

tests/test_examples.py
```diff
@@ -30,9 +30,12 @@ def _paths_to_strs(x: Iterable[pathlib.Path]) -> Sequence[str]:
 EXAMPLES_DIR = THIS_DIR / ".." / "examples"
 TUTORIALS_DIR = THIS_DIR / ".." / "docs" / "tutorials"
 
-SH_PATHS = _paths_to_strs(EXAMPLES_DIR.glob("*.sh"))
+EXCLUDED_EXAMPLE_FILES = ["train_dagger_atari_interactive_policy.py"]
```
zajaczajac, line 33: Probably one could think about an alternative where we mock parts of the example script etc. However, it does not seem to be super useful, since we have unit tests that check analogous things that this mocked version would check.

tests/test_examples.py
```diff
@@ -30,9 +30,14 @@ def _paths_to_strs(x: Iterable[pathlib.Path]) -> Sequence[str]:
...
+EXCLUDED_EXAMPLE_PATHS = [
+    THIS_DIR / ".." / "examples" / f for f in EXCLUDED_EXAMPLE_FILES
```
AdamGleave, line 35:
```suggestion
    EXAMPLES_DIR / f for f in EXCLUDED_EXAMPLE_FILES
```

examples/train_dagger_atari_interactive_policy.py
```diff
@@ -0,0 +1,41 @@
+"""Training DAgger with an interactive policy that queries the user for actions.
```
AdamGleave, line 1: If a matplotlib GUI backend isn't installed, it'll fail with a somewhat cryptic error:

```/home/adam/dev/imitation/src/imitation/policies/interactive.py:99: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
  fig.show()
```

and indeed no figure displays.

Installing the relevant backend seems out-of-scope for this project. But might want to check if the backend is interactive (I think [plt.isinteractive()](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.isinteractive.html#matplotlib.pyplot.isinteractive) checks for this) and warn if not with link to relevant docs e.g. https://matplotlib.org/stable/users/explain/backends.html

src/imitation/policies/interactive.py
AdamGleave: We require Python >=3.8 for `imitation` -- does the problem arise in that (or later) version?

src/imitation/policies/interactive.py
```diff
@@ -1,36 +1,58 @@
...
         clear_screen_on_query: bool = True,
     ):
+        """Builds DiscreteInteractivePolicy.
+
+        Args:
+            observation_space: Observation space.
+            action_space: Action space.
+            action_keys_names: `OrderedDict` containing pairs (key, name) for every
+                action, where key will be used in the console interface, and name
+                is a semantic action name.
```
AdamGleave, line 37:
```suggestion
                is a semantic action name. The index of the pair in the dictionary will be used as the discrete, integer action.
```

src/imitation/policies/interactive.py
```diff
@@ -1,36 +1,58 @@
...
+            == len(set(action_keys_names.values()))
+            == action_space.n
+        )
 
-        self.action_names = action_names
-        self.action_keys = action_keys
-        self.action_key_to_index = {k: i for i, k in enumerate(action_keys)}
+        self.action_keys_names = action_keys_names
+        self.action_key_to_index = {
+            k: i for i, k in enumerate(action_keys_names.keys())
```
AdamGleave, line 54: Iterating over a dict gives you the keys by default (you can leave as-is if you want to be explicit about it)
```suggestion
            k: i for i, k in enumerate(action_keys_names)
```

src/imitation/policies/interactive.py
```diff
@@ -44,34 +66,35 @@ def _choose_action(self, obs: np.ndarray) -> np.ndarray:
...
             "Please select an action. Possible choices in [ACTION_NAME:KEY] format:",
-            ", ".join(
-                [f"{n}:{k}" for n, k in zip(self.action_names, self.action_keys)]
-            ),
+            ", ".join([f"{n}:{k}" for k, n in self.action_keys_names.items()]),
         )
 
         key = input("Your choice (enter key):")
-        while key not in self.action_keys:
+        while key not in self.action_keys_names.keys():
```
AdamGleave, line 76: similar to above (OK to leave as-is)
```suggestion
        while key not in self.action_keys_names:
```

src/imitation/policies/interactive.py
```diff
@@ -1,36 +1,58 @@
+"""Interactive policies that query the user for actions."""
+
 import abc
-from typing import Optional, List
+import collections
+import typing
```
AdamGleave, line 5: Our style guide allows importing types directly from typing (i.e. `from typing import Optional` is permissible) although it's not obligatory -- fine to use this style if you prefer. https://google.github.io/styleguide/pyguide.html#2241-exemptions

src/imitation/policies/interactive.py
```diff
@@ -44,34 +66,35 @@ def _choose_action(self, obs: np.ndarray) -> np.ndarray:
...
 
     @abc.abstractmethod
-    def _render(self, obs: np.ndarray) -> Optional[object]:
-        """Renders an observation, optionally returns a context object for later cleanup."""
+    def _render(self, obs: np.ndarray) -> typing.Optional[object]:
+        """Renders an observation, optionally returns a context for later cleanup."""
 
     def _clean_up(self, context: object) -> None:
         """Cleans up after the input has been captured, e.g. stops showing the image."""
         pass
```
AdamGleave, line 87:
```suggestion
```
Not needed given there's a docstring. Either remove `pass` or add `pass` to `_render` for consistency.

tests/policies/test_interactive.py
```diff
@@ -10,44 +11,107 @@
 from imitation.policies import interactive
 
 ENVS = [
-    "Pong-v4",
+    "CartPole-v0",
```
AdamGleave, line 14: Consider using seals version instead which has a fixed termination condition (500 timesteps). CartPole-v0 in Gym may terminate quickly (as soon as pole drops below threshold) which might cause test to be more vacuous than intended.
```suggestion
    "seals/CartPole-v0",
```

tests/policies/test_interactive.py
```diff
@@ -10,44 +11,107 @@
...
+            print(prompt)
+            if self.return_valid:
+                return action_keys[0]
+            self.return_valid = True
+            return "invalid"
+
+    with mock.patch("builtins.input", mock_input_invalid_then_valid()):
+        interactive_policy.predict(obs)
+        stdout = capsys.readouterr().out
+        assert "Your choice" in stdout and "Invalid" in stdout
```
AdamGleave, line 117: Tests for `DiscreteInteractivePolicy` looks great. We're not testing `AtariInteractivePolicy` at all though. It's pretty simple granted but might still be worth testing, even if just a simple smoke test (it runs, if we feed in a key corresponding to "FIRE" we get the correct action back, etc).

AdamGleave (changes requested):
Thanks for this PR! Looks nearly ready -- only major sugestion is to add some more tests (covering AtariInteractivePolicy), others are pretty minor suggestions.

src/imitation/policies/interactive.py
zajaczajac: Thanks, good to know!

src/imitation/policies/interactive.py
zajaczajac: Yes, the latest change was at 3.9 so 3.8 is not sufficient.

src/imitation/policies/interactive.py
zajaczajac: Yeah, in this case I slightly prefer to keep it.

examples/train_dagger_atari_interactive_policy.py
zajaczajac: Could you please check again what message you are getting and if this is an error or a warning? In my case, when I set a non-GUI backend like `Agg`, I get a warning like this (and the execution continues):
```
UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.
  fig.show()
```
Actually, `plt.isinteractive()` checks for the interactive mode; GUI backends like MacOsX can be in both modes (and we do not need to turn on interactive for it to work). What we actually would like is to check if the backend is "GUI" or "non-GUI" but from a simple search, it does not seem like there is a nice way to do it (rather than check with some white-list of backends). Given that, and the fact that the message I listed above is not that bad, I'd keep this as-is for now. Alternatively, we could opt for throwing an error/assert instead of warning, but again this would require a white-list of backends. WDYT?

zajaczajac: For additional context: on my laptop, the example runs nicely although the mode is not interactive by default.

AdamGleave (approved):
LGTM

Issue #767, opened 2023-08-08
Generalize SQIL to work with other off-policy algos


jas-ho:
## Problem
The current implementation of SQIL (PR: #744 , issue: #740 ) only works with DQN while the original paper also used soft-actor-critic and soft Q-learning. 

It would be great to have support in SQIL for, at least, SAC and, ideally, any off-policy algorithm supported in SB3.

## Solution
Generalize the SQIL constructor to allow instantiation of other off-policy algos in place of DQN.

AdamGleave:
Yes, let's add this. We should also do some basic benchmarking of SQIL with these new algorithms, and ideally try to replicate the results of the SQIL paper.

AdamGleave:
Fixed in #778

PR #778, opened 2023-09-07
Add an option to run SQIL with various off-policy algorithms

zajaczajac:
## Description

This PR adds a possibility to combine SQIL with off-policy algorithms other than DQN, such as SAC, TD3, DDPG, as requested in #767.
A tutorial with SQIL+SAC training on HalfCheetah env is also provided. Random policy gets < 0, expert demonstrations are at ~3400. SQIL+SAC reaches 1400.7 +/- 254.1 after 300K steps (mean +/- std from 5 runs).

## Testing

`pytest tests/algorithms/test_sqil.py` -- adapted relevant tests to work with new base algorithms.
Also one can run the provided tutorial.

### DIFF
--- a/docs/tutorials/8a_train_sqil_sac.ipynb
+++ b/docs/tutorials/8a_train_sqil_sac.ipynb
@@ -7,7 +7,7 @@
     "[download this notebook here](https://github.com/HumanCompatibleAI/imitation/blob/master/docs/tutorials/8a_train_sqil_sac.ipynb)\n",
     "# Train an Agent using Soft Q Imitation Learning with SAC\n",
     "\n",
-    "In the previous tutorial, we used Soft Q Imitation Learning ([SQIL](https://arxiv.org/abs/1905.11108)) on top of DQN base algorithm. In fact, SQIL can be combined with any off-policy algorithm from `stable_baselines3`. Here, we train a HalfCheetah agent using SQIL + SAC."
+    "In the previous tutorial, we used Soft Q Imitation Learning ([SQIL](https://arxiv.org/abs/1905.11108)) on top of the DQN base algorithm. In fact, SQIL can be combined with any off-policy algorithm from `stable_baselines3`. Here, we train a HalfCheetah agent using SQIL + SAC."
    ]
   },
   {
--- a/docs/tutorials/8a_train_sqil_sac.ipynb
+++ b/docs/tutorials/8a_train_sqil_sac.ipynb
@@ -62,7 +62,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "After we collected our expert trajectories, it's time to set up our behavior cloning algorithm."
+    "After we collected our expert trajectories, it's time to set up our imitation algorithm."
    ]
   },
   {
--- a/docs/tutorials/8a_train_sqil_sac.ipynb
+++ b/docs/tutorials/8a_train_sqil_sac.ipynb
@@ -1,0 +1,160 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[download this notebook here](https://github.com/HumanCompatibleAI/imitation/blob/master/docs/tutorials/8a_train_sqil_sac.ipynb)\n",
+    "# Train an Agent using Soft Q Imitation Learning with SAC\n",
+    "\n",
+    "In the previous tutorial, we used Soft Q Imitation Learning ([SQIL](https://arxiv.org/abs/1905.11108)) on top of DQN base algorithm. In fact, SQIL can be combined with any off-policy algorithm from `stable_baselines3`. Here, we train a HalfCheetah agent using SQIL + SAC."
+   ]
+  },
+  {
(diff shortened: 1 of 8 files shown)


tests/algorithms/test_sqil.py
```diff
@@ -1,25 +1,33 @@
...
+    env_name: str,
+    rng: np.random.Generator,
+    pytestconfig: pytest.Config,
+    data_type: str,
+):
     cache = pytestconfig.cache
     assert cache is not None
     return expert_trajectories.make_expert_transition_loader(
-        cache_dir=cache.mkdir("experts"),
+        cache_dir=cache.mkdir(env_name.replace("/", "_")),
```
AdamGleave, line 27: Why do we need environment name in the cache directory? Should already be included in the environment path in https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/testing/expert_trajectories.py#L74

tests/algorithms/test_sqil.py
```diff
@@ -137,12 +219,53 @@ def test_sqil_performance(
...
+        rl_kwargs=dict(
+            learning_starts=500,
+            learning_rate=0.002,
+            batch_size=220,
+            seed=42,
+        ),
+    )
+
+
+# TODO: make it work reliably for other algorithms besides SAC.
```
AdamGleave, line 242: That would be nice ;) Perfectly possible some of the other algorithms are just less sample efficient though and might be hard to get significant improvement in a short unit test.

docs/tutorials/8a_train_sqil_sac.ipynb
```diff
@@ -0,0 +1,160 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[download this notebook here](https://github.com/HumanCompatibleAI/imitation/blob/master/docs/tutorials/8a_train_sqil_sac.ipynb)\n",
+    "# Train an Agent using Soft Q Imitation Learning with SAC\n",
+    "\n",
+    "In the previous tutorial, we used Soft Q Imitation Learning ([SQIL](https://arxiv.org/abs/1905.11108)) on top of DQN base algorithm. In fact, SQIL can be combined with any off-policy algorithm from `stable_baselines3`. Here, we train a HalfCheetah agent using SQIL + SAC."
```
AdamGleave, line 10:
```suggestion
    "In the previous tutorial, we used Soft Q Imitation Learning ([SQIL](https://arxiv.org/abs/1905.11108)) on top of the DQN base algorithm. In fact, SQIL can be combined with any off-policy algorithm from `stable_baselines3`. Here, we train a HalfCheetah agent using SQIL + SAC."
```

docs/tutorials/8a_train_sqil_sac.ipynb
```diff
@@ -0,0 +1,160 @@
...
+    "    f\"The average length of each trajectory is {trajectory_stats['len_mean']}. \"\n",
+    "    f\"The average return of each trajectory is {trajectory_stats['return_mean']}.\"\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "After we collected our expert trajectories, it's time to set up our behavior cloning algorithm."
```
AdamGleave, line 65: I know this was just copied from the original tutorial but I find the reference to behavior cloning potentially ambiguous: it usually refers to supervised learning on expert trajectories (and we have a `BC` class that does exactly), SQIL is doing something conceptually similar but quite different in the details (RL rather than supervised learning).

Would suggest rephrasing this (and the original tutorial), could just call it an imitation algorithm rather than supervised learning algorithm.

docs/tutorials/8a_train_sqil_sac.ipynb
```diff
@@ -0,0 +1,160 @@
...
+    "\n",
+    "reward_before_training, _ = evaluate_policy(sqil_trainer.policy, venv, 100)\n",
+    "print(f\"Reward before training: {reward_before_training}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "After training, we can observe that agent is quite improved (> 1000), although it does not reach the expert performance in this case."
```
AdamGleave, line 119: If you have time to do more tuning, great, but not a priority; this is enough to illustrate the algorithm.

AdamGleave (changes requested):
Thanks for the implementation! Overall looks strong, just a few relatively minor changes.

tests/algorithms/test_sqil.py
zajaczajac: Yeah, indeed, I was confused about the implementation of this function that uses cache and was not sure if I need to make it unique or not. Now I see that this root cache dir can be shared.

tests/algorithms/test_sqil.py
zajaczajac: Ok, with a bit of tuning (lr / gamma) I think I got it to work.

AdamGleave (approved):
LGTM

PR #779, opened 2023-09-08
Add a test timeout

ernestum:
## Description

Adds timeout of 590 seconds to the tests. This should abort tests before CircleCI interrupts the pipeline thereby allowing us to see which test took so long.

ALSO:

Fix notebook running error.  #790 and #779.
1. docs/tutorials/3_train_gail.ipynb: 800k can bring running time exceed error
2. docs/tutorials/4_train_airl.ipynb: env.reset doesn't require seed in gymnasium setting.
3. docs/tutorials/5a_train_preference_comparisons_with_cnn.ipynb: migrate to gymnasium
4. docs/tutorials/8a_train_sqil_sac.ipynb: v0 is not supported.

## Testing

Only test parameter changes.


AdamGleave:
Thanks for this!

Looks like one of the Ray tests is failing. That's not too surprising, I think Ray can be buggy on Windows, we may want to consider just skipping that test on Windows

ernestum:
Good point. Added a skipif on the test with the timeout.

AdamGleave:
Looks like its timing out in `test_train_preference_comparisons_main`. Odd since stdout/stderr suggests that training completes so I think it must be hanging on something after that. Maybe in writing TensorBoard logs as some threads seem to be stuck there?

Skipping that test as well would be OK as a hotfix, but if can replicate that test failure via running as SSH can hopefully determine a clean resolution.

ernestum:
Possibly related: https://discuss.pytorch.org/t/got-stuck-at-tensorboardx-event-file-writer-py/112209

I updated to the newest `tensorboard`/`tensorboardX` version in the SSH session and re-ran the tests. Now there is no timeout any more but the tests fail due to a different reason:
<details>
<summary>
Some file is not found:
<code>FileNotFoundError: [Errno 2] No such file or directory: b'C:\\Users\\circleci.PACKER-64C83029\\AppData\\Local\\Temp\\pytest-of-circleci\\pytest-3\\test_train_preference_comparis0\\train_preference_comparisons\\CartPole-v1\\20230913_094320_90ea37\\log\\events.o
ut.tfevents.1694598202.packer-64c83029-cd75-bf3a-2e04-0f23beb7d9d9.7076.0'</code>
(click to expand)
</summary>
```
___________ test_train_preference_comparisons_main[plain] ___________ 

tmpdir = local('C:\\Users\\circleci.PACKER-64C83029\\AppData\\Local\\Temp\\pytest-of-circleci\\pytest-3\\test_train_preference_comparis0'), preference_comparison_config = {}

    def test_train_preference_comparisons_main(tmpdir, preference_comparison_config): 
        config_updates = dict(logging=dict(log_root=tmpdir))
        sacred.utils.recursive_update(config_updates, preference_comparison_config)
>       run = train_preference_comparisons.train_preference_comparisons_ex.run(
            # Note: we have to use the cartpole and not the seals_cartpole config because
            #  the seals_cartpole config needs rollouts with a fixed horizon,
            #  and the saved trajectory rollouts are variable horizon.
            named_configs=["cartpole"] + ALGO_FAST_CONFIGS["preference_comparison"],
            config_updates=config_updates,
        )

tests\scripts\test_scripts.py:160:
_ _ _ _ _ _ _ _ _ _ _ _ _ 
venv\lib\site-packages\sacred\experiment.py:276: in run
    run()
venv\lib\site-packages\sacred\run.py:238: in __call__
    self.result = self.main_function(*args)
venv\lib\site-packages\sacred\config\captured_function.py:42: in captured_function
    result = wrapped(*args, **kwargs)
venv\lib\site-packages\imitation\scripts\train_preference_comparisons.py:167: in train_preference_comparisons
    custom_logger, log_dir = logging_ingredient.setup_logging()
venv\lib\site-packages\sacred\config\captured_function.py:42: in captured_function
    result = wrapped(*args, **kwargs)
venv\lib\site-packages\imitation\scripts\ingredients\logging.py:114: in setup_logging
    custom_logger = imit_logger.configure(
venv\lib\site-packages\imitation\util\logger.py:413: in configure
    output_formats = _build_output_formats(folder, format_strs) 
venv\lib\site-packages\imitation\util\logger.py:67: in _build_output_formats
    output_formats.append(make_output_format(f, str(folder)))
venv\lib\site-packages\imitation\util\logger.py:44: in make_output_format
    return sb_logger.make_output_format(_format, log_dir, log_suffix)
venv\lib\site-packages\stable_baselines3\common\logger.py:462: in make_output_format
    return TensorBoardOutputFormat(log_dir)
venv\lib\site-packages\stable_baselines3\common\logger.py:398: in __init__
    self.writer = SummaryWriter(log_dir=folder)
venv\lib\site-packages\torch\utils\tensorboard\writer.py:247: in __init__
    self._get_file_writer()
venv\lib\site-packages\torch\utils\tensorboard\writer.py:277: in _get_file_writer
    self.file_writer = FileWriter(
venv\lib\site-packages\torch\utils\tensorboard\writer.py:76: in __init__
    self.event_writer = EventFileWriter(
venv\lib\site-packages\tensorboard\summary\writer\event_file_writer.py:100: in __init__
    self.flush()
venv\lib\site-packages\tensorboard\summary\writer\event_file_writer.py:125: in flush
    self._async_writer.flush()
venv\lib\site-packages\tensorboard\summary\writer\event_file_writer.py:194: in flush
    self._check_worker_status()
venv\lib\site-packages\tensorboard\summary\writer\event_file_writer.py:212: in _check_worker_status
    raise exception
C:\Python38\lib\threading.py:932: in _bootstrap_inner
    self.run()
venv\lib\site-packages\tensorboard\summary\writer\event_file_writer.py:244: in run
    self._run()
venv\lib\site-packages\tensorboard\summary\writer\event_file_writer.py:275: in _run
    self._record_writer.write(data)
venv\lib\site-packages\tensorboard\summary\writer\record_writer.py:40: in write
    self._writer.write(header + header_crc + data + footer_crc)
venv\lib\site-packages\tensorboard\compat\tensorflow_stub\io\gfile.py:768: in write
    self.fs.write(self.filename, file_content, self.binary_mode)
venv\lib\site-packages\tensorboard\compat\tensorflow_stub\io\gfile.py:157: in write
    self._write(filename, file_content, "wb" if binary_mode else "w")
_ _ _ _ _ _ _ _ _ _ _ 

self = <tensorboard.compat.tensorflow_stub.io.gfile.LocalFileSystem object at 0x000002B784756550>
filename = b'C:\\Users\\circleci.PACKER-64C83029\\AppData\\Local\\Temp\\pytest-of-circleci\\pytest-3\\test_train_preference_compa...ole-v1\\20230913_094320_90ea37\\log\\events.out.tfevents.1694598202.packer-64c83029-cd75-bf3a-2e04-0f23beb7d9d9.7076.0'
file_content = b'H\x00\x00\x00\x00\x00\x00\x00\xfe\x92H\x9d\te\xfd\x80\x0ea@\xd9A\x1a\rbrain.Event:2R.\n,tensorboard.summary.writer.event_file_writer\xc8p\xa0\xa0', mode = 'wb'

    def _write(self, filename, file_content, mode):
        encoding = None if "b" in mode else "utf8"
>       with io.open(filename, mode, encoding=encoding) as f:
E       FileNotFoundError: [Errno 2] No such file or directory: b'C:\\Users\\circleci.PACKER-64C83029\\AppData\\Local\\Temp\\pytest-of-circleci\\pytest-3\\test_train_preference_comparis0\\train_preference_comparisons\\CartPole-v1\\20230913_094320_90ea37\\log\\events.o
ut.tfevents.1694598202.packer-64c83029-cd75-bf3a-2e04-0f23beb7d9d9.7076.0'

venv\lib\site-packages\tensorboard\compat\tensorflow_stub\io\gfile.py:171: FileNotFoundError
---- Captured log call ----
WARNING  rl:initialize.py:217 Added new config entry: "rl_kwargs.batch_size"
WARNING  rl:initialize.py:217 Added new config entry: "rl_kwargs.n_epochs"
WARNING  train_preference_comparisons:run.py:458 No observers have been added to this run
ERROR    train_preference_comparisons:run.py:392 Failed after 0:00:00!
```
</details>
A manual check revealed, that the path DOES exist :shrug:

ZiyueWang25:
I double checked. The file indeed doesn't exists, the directory exists. So I experiment a bit and found the issue:

linux file path limit [4096](https://www.ibm.com/docs/en/spectrum-protect/8.1.9?topic=parameters-file-specification-syntax)
windows file path limit [256](https://it.cornell.edu/shared-file/windows-file-name-or-destination-path-you-specified-not-valid-or-too-long#:~:text=Microsoft%20Windows%20has%20a%20MAX_PATH,%2Frename%20these%20paths%2Ffiles.)

It turns out the file length is right on the limit.
```python
>>> len("C:\\Users\\circleci.PACKER-64C83029\\AppData\\Local\\Temp\\pytest-of-circleci\\pytest-4\\test_train_preference_comparis0\\train_preference_comparisons\\CartPole-v1\\20231001_144824_a538c8\\log\\events.out.tfevents.1696171705.packer-64c83029-cd75-bf3a-2e04-0f23beb7d9d9.6640.0")
262
```

So I am checking whether removing the path limit [this way](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=powershell#enable-long-paths-in-windows-10-version-1607-and-later) can work or not.

ZiyueWang25:
- The `ModuleNotFoundError: No module named 'gymnasium'` error pops up weirdly. Not sure why it happened after this change.
- If I do `pip install gymnasium==0.28.1` and then run the test `pytest tests/algorithms/test_dagger.py::test_simple_dagger_space_mismatch_error[None-False]` , it shows error: 

```
    def check_for_correct_spaces(env: GymEnv, observation_space: spaces.Space, action_space: spaces.Space) -> None:
        """
        Checks that the environment has same spaces as provided ones. Used by BaseAlgorithm to check if
        spaces match after loading the model with given env.
        Checked parameters:
        - observation_space
        - action_space

        :param env: Environment to check for valid spaces
        :param observation_space: Observation space to check against
        :param action_space: Action space to check against
        """
        if observation_space != env.observation_space:
>           raise ValueError(f"Observation spaces do not match: {observation_space} != {env.observation_space}")
E           ValueError: Observation spaces do not match: Box([-1. -1. -8.], [1. 1. 8.], (3,), float32) != Box([-1. -1. -8.], [1. 1. 8.], (3,), float32)
```

ZiyueWang25:
so the original error is gone. I noticed two new issues here:

 #791 & #790 

Note: #790 can represent the issues for:
```
FAILED tests/test_examples.py::test_run_tutorial_notebooks[C:\\Users\\circleci.PACKER-64C83029\\project\\tests\\..\\docs\\tutorials\\5a_train_preference_comparisons_with_cnn.ipynb]
FAILED tests/test_examples.py::test_run_tutorial_notebooks[C:\\Users\\circleci.PACKER-64C83029\\project\\tests\\..\\docs\\tutorials\\4_train_airl.ipynb]
FAILED tests/test_examples.py::test_run_tutorial_notebooks[C:\\Users\\circleci.PACKER-64C83029\\project\\tests\\..\\docs\\tutorials\\8a_train_sqil_sac.ipynb]
FAILED tests/test_examples.py::test_run_tutorial_notebooks[C:\\Users\\circleci.PACKER-64C83029\\project\\tests\\..\\docs\\tutorials\\3_train_gail.ipynb
```
And #791 is for:
```
FAILED tests/algorithms/test_sqil.py::test_sqil_performance_continuous[DDPG]
```

ZiyueWang25:
The numpy issue can be replicated
image

Hopefully it can be resolved by specifying the seed.

ZiyueWang25:
So it shows a timeout issue saying that `tests/test_examples.py::test_run_tutorial_notebooks[C:\\Users\\circleci.PACKER-64C83029\\project\\tests\\..\\docs\\tutorials\\4_train_airl.ipynb]` cannot pass within 5 minutes. It can pass on my computer within 3 minutes. I think this is due to win32 has less resources. The current solution is to make the timeout larger.

Created feature request for this: #793

ZiyueWang25:
> The numpy issue can be replicated image
> 
> Hopefully it can be resolved by specifying the seed.

Actually this is solved in stable_baselines3 [here](https://github.com/DLR-RM/stable-baselines3/pull/1655), I think we need to set `stable-baselines3~=2.0` to `stable-baselines3>=2.0` to keep the package updated.

.circleci/config.yml
```diff
@@ -176,13 +176,28 @@ commands:
             .\venv\Scripts\activate
             pip install --upgrade --force-reinstall --no-deps .
           shell: powershell.exe
+
+      - run:
+          name: install gymansium[mujoco]
```
AdamGleave, line 181: Why do we need to explicitly install Gymnasium? Shouldn't this dependency be picked up by `pip install .` above?

Also confused why we need the MuJoCo dependency -- do any of our tests use MuJoCo?

.circleci/config.yml
```diff
@@ -176,13 +176,28 @@ commands:
...
+          shell: powershell.exe
+
       - run:
           name: print installed packages
           command: |
             .\venv\Scripts\activate
             pip freeze --all
           shell: powershell.exe
 
+      - run:
```
AdamGleave, line 194: Well done tracking down this issue and the workaround! (I had no idea about this Windows limitation.)

docs/tutorials/4_train_airl.ipynb
```diff
@@ -127,7 +128,7 @@
     "    reward_net=reward_net,\n",
     ")\n",
     "\n",
-    "env.reset(seed=SEED)\n",
+    "env.reset()\n",
```
AdamGleave, line 131: Why are we removing the seed here? I'd be nervous this might make the test flaky.

tests/algorithms/test_sqil.py
```diff
@@ -246,18 +246,22 @@ def test_sqil_performance_continuous(
...
     rl_algo_class: Type[off_policy_algorithm.OffPolicyAlgorithm],
 ):
+    rl_kwargs = dict(
+        learning_starts=500,
+        learning_rate=0.001,
+        gamma=0.95,
+        seed=42,
+    )
+    if rl_algo_class == ddpg.DDPG:
+        rl_kwargs["gamma"] = 0.99
```
AdamGleave, line 256: Thanks for tuning this!

tests/test_examples.py
```diff
@@ -52,7 +52,8 @@ def test_run_tutorial_notebooks(nb_path) -> None:  # pragma: no cover
         nb_path: Path to the notebook to test.
     """
     nb = ptnb.notebook.load_notebook(nb_path)
-    result = ptnb.execution.execute_notebook(nb, cwd=TUTORIALS_DIR, timeout=120)
+    # TODO: Shorten timeout and ensure the notebook can still show desired improvement.
```
AdamGleave, line 55: Following https://google.github.io/styleguide/pyguide.html#312-todo-comments we suggest including username or GH issue in TODO comments so that future developers can find more information about it as needed. I think you already opened issue #793  to track this so can reference that

AdamGleave (commented):
Thanks for these changes! Overall looks almost ready, a few clarifying questions & suggestions, please tag me for a re-review once addressed.

docs/tutorials/4_train_airl.ipynb
ZiyueWang25: Because the venv created from [`make_vec_env`](https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/util/util.py#L80) doesn't accept seed.

.circleci/config.yml
ZiyueWang25: The main thing is about the mujoco part. It gets involved in the windows environment from `import seals` in the notebook and mainly [here](https://github.com/HumanCompatibleAI/seals/blob/master/src/seals/__init__.py#L35). If I don't do this, the notebook will [fail](https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3907/workflows/58f56c5c-7035-464e-a9d6-335832f75284/jobs/15842?invite=true#step-113-944164_97).

tests/test_examples.py
ZiyueWang25: Done.

.circleci/config.yml
ZiyueWang25: It is directly [mentioned](https://github.com/HumanCompatibleAI/seals/blob/master/setup.py#L147) in Seals as well. Not quite sure why it shows up only in the Windows environment and only in notebook.

docs/tutorials/4_train_airl.ipynb
ZiyueWang25: I updated the naming from `env` to `venv` to make things clearer. Also, I added `venv.seed` before `evaluate_policy` to make sure the experimentation can be the same. I ran the notebook twice and they both shown:

```
Rewards before training: 102.6 +/- 24.11514047232568 
Rewards after training: 319.54 +/- 159.8500184548003
```

.circleci/config.yml
AdamGleave: I did some digging here and the TL;DR is that the notebooks were only getting tested on Windows. The [SQIL notebook](https://github.com/HumanCompatibleAI/imitation/blob/8db6bfb0cfd533f6418e662890499670bd6b010b/docs/tutorials/8a_train_sqil_sac.ipynb#L4) depends on HalfCheetah which is a MuJoCo environment, so it makes sense we'd need thiis dependency. We made a decision a while back to have MuJoCo dependency be optional for the codebase since it's large and can be somewhat tricky to install, so I'm loathe to reverse that especially just special-casing it for the Windows CI run. I'd suggest we instead switch out HalfCheetah for something like `MountainCarContinuous` that does not require extra dependencies -- tracking in https://github.com/HumanCompatibleAI/imitation/issues/798

Note the `gym.register` line [here](https://github.com/HumanCompatibleAI/seals/blob/master/src/seals/__init__.py#L35) should succeed even if MuJoCo is not present -- it will only fail when attempting to instantiate those environments. Indeed,

This doesn't error on `unit-test-linux` since we're skipping all notebooks: https://github.com/HumanCompatibleAI/imitation/blob/master/tests/test_examples.py#L43 This was because they were meant to be run in readthedocs after https://github.com/HumanCompatibleAI/imitation/pull/565 This run is happening but the notebooks failing is just a *warning*:
```
/home/docs/checkouts/readthedocs.org/user_builds/imitation/checkouts/latest/docs/tutorials/8a_train_sqil_sac.ipynb: WARNING: Executing notebook failed: CellExecutionError [mystnb.exec]
/home/docs/checkouts/readthedocs.org/user_builds/imitation/checkouts/latest/docs/tutorials/8a_train_sqil_sac.ipynb: WARNING: Notebook exception traceback saved in: /home/docs/checkouts/readthedocs.org/user_builds/imitation/checkouts/latest/_readthedocs/html/reports/tutorials/8a_train_sqil_sac.err.log [mystnb.exec]
```
I've opened https://github.com/HumanCompatibleAI/imitation/issues/799 to track this

Mac we are not running `test_examples.py` either due to a separate bug, I'll address in #797

docs/tutorials/10_train_custom_env.ipynb
```diff
@@ -5,18 +5,17 @@
    "metadata": {},
```
AdamGleave, line 5: Newlines seem to have changed throughout this and many other notebooks making the diff messy. Can you clean this up? Most of the notebooks shouldn't be changing in this PR, and I can't tell right now which notebooks had actual semantic changes.

AdamGleave (changes requested):
The notebooks have a bunch of spurious newlines introduced -- can you please clean up that diff as it will otherwise make commit history/`git blame` messy in future.

MuJoCo dependency I think should not be necessary although I now understand (see comment) why test was failing without it -- this has unrooted a deeper problem in our CI that I'll get the team to work on addressing in other PRs/issues.

docs/tutorials/4_train_airl.ipynb
AdamGleave: Oh, I see, that change makes sense. Given we were accidentally not testing any of the notebooks before this the bug must have managed to slip through CI (it's certainly not Windows specific).

AdamGleave (approved):
On a closer look I see you made some non-trivial changes to the notebooks (not just newlines) and the newlines seem to have been introduced by black reformatting (perhaps a newer version is more opinionated?) so I won't insist on that change.

I'm removing the MuJoCo dependency as that shouldn't be in our CI. This will make it red again but I'll go ahead and merge if the only test that's failing is for `8a_train_sqil_sac`; we should fix that separately in https://github.com/HumanCompatibleAI/imitation/pull/800 by changing the notebook

.circleci/config.yml
ZiyueWang25: Thank you for the detailed explanation! That makes a lot sense!

docs/tutorials/10_train_custom_env.ipynb
ZiyueWang25: Sorry about this. Yes, I did some semantic changes and then the lint keeps failing on `docs/tutorials/4_train_airl.ipynb`. I did something wrong and formatted all notebook. I should have just formatted that only one. 

I have reverted the commit and now the notebook change should be easy to read.

AdamGleave (approved):
LGTM

PR #771, opened 2023-08-21
Tune preference comparison example hyperparameters

timokau:
## Description

This PR changes the hyperparameters of the preference comparisons example to values that result in much more reliable training. The main point of discussion is how we should handle the examples in the notebooks. I have left them unchanged for now.

See the commit message (included here for convenience) for details.

### Commit Message

The preference comparison example previously did not show significant learning. It usually ended with a reward < -1000, which can be considered "failed" in the Pendulum environment. This commit updates the parameters to avoid this. It could be argued that hyperparameter optimization for the examples is bad, since it gives a skewed impression of the library. I think as long as we acknowledge that the parameters were optimized this is okay though, and it is much nicer if we have a working example as a starting point.

I have tuned the hyperparameters with a mix of syne_tune [1] and manual tuning. Since the training can have very high variance, I repeated each training run multiple (up to 100) times and used multi-fidelity optimization (PASHA and ASHA) to find a good configuration. I set the objective to the 90% upper-confidence-bound of the mean final-evaluation reward over all the training runs.

Unfortunately the optimization process was a bit messy since I was just getting started with syne_tune, so it is difficult to provide a full script to cleanly reproduce the results. I used something akin to this configuration space:

```py
import syne_tune.config_space as cs
config_space = {
    "reward_epochs": cs.randint(1, 20),
    "ppo_clip_range": cs.uniform(0.0, 0.3),
    "ppo_ent_coef": cs.uniform(0.0, 0.01),
    "ppo_gae_lambda": cs.uniform(0.9, 0.99),
    "ppo_n_epochs": cs.randint(5, 25),
    "discount_factor": cs.uniform(0.9, 1.0),
    "use_sde": cs.choice(["true", "false"]),
    "sde_sample_freq": cs.randint(1, 5),
    "ppo_lr": cs.loguniform(1e-4, 5e-3),
    "exploration_frac": cs.uniform(0, 0.1),
    "num_iterations": cs.randint(5, 100),
    "initial_comparison_frac": cs.uniform(0.05, 0.25),
    "initial_epoch_multiplier": cs.randint(1, 4),
    "query_schedule": cs.choice(["constant", "hyperbolic", "inverse_quadratic"]),
    "total_timesteps": 50_000,
    "total_comparisons": 200,
    "max_evals": 100,
}
```

and the configuration I selected in the end is this one

```py
{
    "reward_epochs": 10,
    "ppo_clip_range": 0.1,
    "ppo_ent_coef": 0.01,
[...]

PR #782, opened 2023-09-11
Complete PR #771 (Tune preference comparison example hyperparameters)

lukasberglund:
## Description

This PR is meant to complete the changes made in this [separate PR](https://github.com/HumanCompatibleAI/imitation/pull/771). [I'm not sure this is the best way to do this. I've also considered pushing onto timokau's branch, but I think that would require his approval.]

In addition to the changes made in the previous PR (adding optimized hyperparameters to `preference_comparisons.rst`) this PR also adds these hyperparameters to `5_train_pereference_comparisons.ipynb`, as well as making slight changes `5_train_preference_comparisons.ipynb`. 

## Testing

I compared performance with the old hyperparameters to performance with new hyperparameters by running the notebooks separately. I found that when increasing the `train_timesteps` to 100_000 as advised by the comment in the notebook, we get the following performances:

* ~(-500) for original hyperparams
* ~(-170) for new hyperparams

Additionally, I found that increasing the increasing `num_iterations` to 60 only resulted in a modest increase in performance of ~30 and that increasing `total_timesteps` and `total_comparison` in `pref_comparisons.train` didn't increase performance much after that (can't give a precise number here since I only tested it once). Given that increasing `total_timesteps` and `total_comparison` increases training time by a lot. I removed the comments that recommend increasing them.


AdamGleave:
Hi @lukasberglund, the changes to the notebook look good, however note that our style is to not include any output from the notebook (to keep them small and diffs minimal), so you'll need to delete that per the linter error: https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3786/workflows/003fdb1e-1cdf-4ebc-a96f-90e378c73c39/jobs/15106 The script ci/clean_notebooks.py will be useful for this.

Separately all the CI tests are failing. Although there's a known issue with unit-test-windows (failing on master), the others should all be green and we cannot merge without them passing. It looks like your branch is 5 commits behind master -- please try merging master into it as it may fix many of these issues.

In future I'd recommend making branches directly on `imitation` (you have push access), so that others in the project can also make commits -- although we usually only have one main developer per author, it can be useful in cases like the above for others to be able to help out by doing small merges.

AdamGleave (changes requested):
Clean notebook output, and get CI tests (except unit-test-windows) passing

lukasberglund:
@AdamGleave thanks for the review. I've cleaned the notebook and gotten all tests except unit-test-windows to pass. I should have probably looked at the failing tests before requesting a review. Sorry about that!

AdamGleave (approved):
LGTM

Issue #780, opened 2023-09-09
Add CLI for SQIL


AdamGleave:
## Problem
There is currently no script in `src/imitation/scripts` that supports training SQIL implemented in `src/imitation/algorithms/sqil.py`. This both makes the algorithm harder to use, and is inconsistent (all our other algorithms have CLI support).

## Solution
Add support for SQIL to the existing script `train_imitation`. This would be preferable if practical, to maintain a consistent CLI interface simplifying benchmarking of different algorithms.

## Possible alternative solutions
Implement a new script e.g. `train_sqil`. This would be preferable if SQIL has a substantially different enough API that unifying it with an existing script would lead to an overall increase in complexity.

lukasberglund:
How I'm thinking about this right now: There are some non-trivial differences between BC/DAgger and SQIL (e.g. SQIL requires specifying an off-policy RL algorithm), which means that the two algorithms will use different parameters. I don't think it would be too bad though since we could just have separate configs for both. 

One drawback of using the same file is that it would sometimes introduce unused parameters (e.g. SQIL doesn't use an expert). I'm not sure how big a deal that is, but doesn't seem too bad.

lukasberglund:
Another issue would be that the current default for the policy is an `ActorCriticPolicy`, whereas sqil requires a `DQNPolicy`. One way to fix this would be:
 1) Change the name of the `imitation.scripts.ingredients.policy` to `ActorCriticPolicy` 
 2) Create another ingredient called `DQNPolicy`
 3) Import them as subingredients to `bc_ingredient` and `sqil_ingredient` respectively.

PR #784, opened 2023-09-12
Add CLI for SQIL

lukasberglund:
## Description

This PR adds an option to `src/imitation/scripts` to train using SQIL as addressed in #780. This is still a work in progress, but I would welcome feedback. Some uncertainties I have include:

* **What arguments to explicitly add to the sacred config for SQIL** -- Right now I'm only including the keyword-args for SQIL as well as `log_interval`, `total_timesteps`, and `progress_bar`, similarly to how it's done for `BC` and `DAgger`. The user set can set all the other arguments too, but including them would make them more salient. 
* **Adding warmstart capabilities** -- BC and DAgger both allow to continue training a previously trained model, but the SQIL API doesn't. Consequently I didn't include in the SQIL CLI. I could add this feature pretty easily though.
* **Logging** -- Currently, I don't see any logs to stdout when training with SQIL, except at the end. I'm not sure if I'm using the wrong hyperparameters or this is a bug. I will look into this more tomorrow.

## Testing
I added three tests that mirror those for DAgger and BC. Right now I have more tests for `SQIL` than for `DAgger`, which might be overkill. Let me know if I should remove some.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -202,7 +202,7 @@ setup(
         "numpy>=1.15",
         "torch>=1.4.0",
         "tqdm",
-        "shine",
+        "rich",
         "scikit-learn>=0.21.2",
         "seals~=0.1.5",
         STABLE_BASELINES3,
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ ATARI_REQUIRE = [
     "autorom[accept-rom-license]~=0.6.0",
 ]
 PYTYPE = ["pytype==2022.7.26"] if IS_NOT_WINDOWS else []
-STABLE_BASELINES3 = "stable-baselines3[extra]>=1.7.0,<2.0.0"
+STABLE_BASELINES3 = "stable-baselines3>=1.7.0,<2.0.0"
 # pinned to 0.21 until https://github.com/DLR-RM/stable-baselines3/pull/780 goes
 # upstream.
 GYM_VERSION_SPECIFIER = "==0.21.0"
@@ -202,6 +202,7 @@ setup(
         "numpy>=1.15",
         "torch>=1.4.0",
         "tqdm",
+        "shine",
         "scikit-learn>=0.21.2",
         "seals~=0.1.5",
         STABLE_BASELINES3,
--- a/setup.py
+++ b/setup.py
@@ -209,6 +209,7 @@ setup(
         "tensorboard>=1.14",
         "huggingface_sb3~=2.3",
         "datasets>=2.8.0",
+        "shine",
     ],
     tests_require=TESTS_REQUIRE,
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ ATARI_REQUIRE = [
     "autorom[accept-rom-license]~=0.6.0",
 ]
 PYTYPE = ["pytype==2022.7.26"] if IS_NOT_WINDOWS else []
-STABLE_BASELINES3 = "stable-baselines3>=1.7.0,<2.0.0"
+STABLE_BASELINES3 = "stable-baselines3[extra]>=1.7.0,<2.0.0"
 # pinned to 0.21 until https://github.com/DLR-RM/stable-baselines3/pull/780 goes
 # upstream.
 GYM_VERSION_SPECIFIER = "==0.21.0"
@@ -209,7 +209,6 @@ setup(
         "tensorboard>=1.14",
         "huggingface_sb3~=2.3",
         "datasets>=2.8.0",
-        "shine",
     ],
     tests_require=TESTS_REQUIRE,
     extras_require={
--- a/src/imitation/algorithms/base.py
+++ b/src/imitation/algorithms/base.py
@@ -13,6 +13,7 @@ from typing import (
     cast,
 )
 
+import torch as th
 import torch.utils.data as th_data
 from stable_baselines3.common import policies
 
@@ -180,6 +181,14 @@ class DemonstrationAlgorithm(BaseImitationAlgorithm, Generic[TransitionKind]):
     @abc.abstractmethod
     def policy(self) -> policies.BasePolicy:
         """Returns a policy imitating the demonstration data."""
+
+    def save_policy(self, policy_path: types.AnyPath) -> None:
+        """Save policy to a path. Can be reloaded by `.reconstruct_policy()`.
+
+        Args:
+            policy_path: path to save policy to.
+        """
+        th.save(self.policy, util.parse_path(policy_path))
 
 
 class _WrappedDataLoader:
--- a/src/imitation/algorithms/base.py
+++ b/src/imitation/algorithms/base.py
@@ -183,7 +183,7 @@ class DemonstrationAlgorithm(BaseImitationAlgorithm, Generic[TransitionKind]):
         """Returns a policy imitating the demonstration data."""
 
     def save_policy(self, policy_path: types.AnyPath) -> None:
-        """Save policy to a path. Can be reloaded by `.reconstruct_policy()`.
+        """Save policy to a path.
 
         Args:
             policy_path: path to save policy to.
--- a/src/imitation/algorithms/base.py
+++ b/src/imitation/algorithms/base.py
@@ -13,7 +13,6 @@ from typing import (
     cast,
 )
 
-import torch as th
 import torch.utils.data as th_data
 from stable_baselines3.common import policies
 
@@ -181,14 +180,6 @@ class DemonstrationAlgorithm(BaseImitationAlgorithm, Generic[TransitionKind]):
     @abc.abstractmethod
     def policy(self) -> policies.BasePolicy:
         """Returns a policy imitating the demonstration data."""
-
-    def save_policy(self, policy_path: types.AnyPath) -> None:
-        """Save policy to a path.
-
-        Args:
-            policy_path: path to save policy to.
-        """
-        th.save(self.policy, util.parse_path(policy_path))
 
 
 class _WrappedDataLoader:
--- a/src/imitation/scripts/config/train_imitation.py
+++ b/src/imitation/scripts/config/train_imitation.py
@@ -1,19 +1,19 @@
 """Configuration settings for train_dagger, training DAgger from synthetic demos."""
 
 import sacred
+from stable_baselines3 import dqn
 
 from imitation.scripts.ingredients import bc
 from imitation.scripts.ingredients import demonstrations as demos_common
 from imitation.scripts.ingredients import environment, expert
 from imitation.scripts.ingredients import logging as logging_ingredient
-from imitation.scripts.ingredients import policy, policy_evaluation
+from imitation.scripts.ingredients import policy_evaluation
 
 train_imitation_ex = sacred.Experiment(
     "train_imitation",
     ingredients=[
         logging_ingredient.logging_ingredient,
         demos_common.demonstrations_ingredient,
-        policy.policy_ingredient,
         expert.expert_ingredient,
         environment.environment_ingredient,
         policy_evaluation.policy_evaluation_ingredient,
@@ -28,6 +28,17 @@ def config():
         use_offline_rollouts=False,  # warm-start policy with BC from offline demos
         total_timesteps=1e5,
         beta_schedule=None,
+    )
+
+    sqil = dict(
+        rl_algo_class=dqn.DQN,
+        rl_kwargs=dict(),
+        total_timesteps=3e5, 
+        policy_model="MlpPolicy", # TODO: this is a placeholder, change it to something that makes sense
+        train_kwargs = dict( 
+            log_interval=4,  # Number of updates between Tensorboard/stdout logs
+            progress_bar=True,
+        ),
     )
 
 
@@ -100,3 +111,4 @@ def seals_humanoid():
 def fast():
     dagger = dict(total_timesteps=50)
     bc = dict(train_kwargs=dict(n_batches=50))
+    sqil = dict(total_timesteps=50)
--- a/src/imitation/scripts/config/train_imitation.py
+++ b/src/imitation/scripts/config/train_imitation.py
@@ -34,7 +34,7 @@ def config():
         rl_algo_class=dqn.DQN,
         rl_kwargs=dict(),
         total_timesteps=3e5,
-        policy_model="MlpPolicy",  # TODO: this is a placeholder, change it to something that makes sense
+        policy_model="MlpPolicy",
         train_kwargs=dict(
             log_interval=4,  # Number of updates between Tensorboard/stdout logs
             progress_bar=True,
--- a/src/imitation/scripts/config/train_imitation.py
+++ b/src/imitation/scripts/config/train_imitation.py
@@ -1,7 +1,6 @@
 """Configuration settings for train_dagger, training DAgger from synthetic demos."""
 
 import sacred
-from stable_baselines3 import dqn as dqn_algorithm
 
 from imitation.scripts.ingredients import bc
 from imitation.scripts.ingredients import demonstrations as demos_common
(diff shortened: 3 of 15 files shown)


setup.py
```diff
@@ -209,6 +209,7 @@ def get_local_version(version: "ScmVersion", time_format="%Y%m%d") -> str:
         "tensorboard>=1.14",
         "huggingface_sb3~=2.3",
         "datasets>=2.8.0",
+        "shine",
```
AdamGleave, line 212: What is this for? What's the minimum version required?

src/imitation/algorithms/base.py
```diff
@@ -181,6 +182,14 @@ def set_demonstrations(self, demonstrations: AnyTransitions) -> None:
     def policy(self) -> policies.BasePolicy:
         """Returns a policy imitating the demonstration data."""
 
+    def save_policy(self, policy_path: types.AnyPath) -> None:
+        """Save policy to a path. Can be reloaded by `.reconstruct_policy()`.
```
AdamGleave, line 186: Where is reconstruct_policy defined? All I see in our codebase is a module-level method in the `bc` package.

src/imitation/scripts/config/train_imitation.py
```diff
@@ -30,6 +30,17 @@ def config():
         beta_schedule=None,
     )
 
+    sqil = dict(
+        rl_algo_class=dqn.DQN,
+        rl_kwargs=dict(),
+        total_timesteps=3e5,
+        policy_model="MlpPolicy",
+        train_kwargs=dict(
+            log_interval=4,  # Number of updates between Tensorboard/stdout logs
```
AdamGleave, line 39: nit
```suggestion
            log_interval=4,  # Number of updates between TensorBoard/stdout logs
```

src/imitation/scripts/train_imitation.py
```diff
@@ -140,6 +141,41 @@ def dagger(
...
+    _rnd: np.random.Generator,
+) -> Mapping[str, Mapping[str, float]]:
+    custom_logger, log_dir = logging_ingredient.setup_logging()
+    expert_trajs = demonstrations.get_expert_trajectories()
+
+    with environment.make_venv() as venv:
+        sqil_trainer = SQIL(
+            venv=venv,
+            demonstrations=expert_trajs,
+            policy=sqil["policy_model"],
```
AdamGleave, line 157: This duplicates the policy ingredient: https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/scripts/ingredients/policy.py

I think can replace with `policy.make_policy(venv)` and remove the `policy_model` config parameter.

src/imitation/scripts/train_imitation.py
```diff
@@ -140,6 +141,41 @@ def dagger(
...
+            custom_logger=custom_logger,
+            rl_algo_class=sqil["rl_algo_class"],
```
AdamGleave, line 159: It'd be good to use the `rl` ingredient. Although this may be a bit tricky since that ingredient is designed to *make* the RL algorithm rather than pass class and keyword arguments in (which we need given how SQIL is currently designed, in order to ensure a custom replay buffer gets specified).

src/imitation/scripts/train_imitation.py
```diff
@@ -140,6 +141,41 @@ def dagger(
...
+
+        sqil_trainer.train(
+            total_timesteps=int(sqil["total_timesteps"]),
+            **sqil["train_kwargs"],
+        )
+        sqil_trainer.save_policy(policy_path=osp.join(log_dir, "final.th"))
+
+        imit_stats = policy_evaluation.eval_policy(sqil_trainer.policy, venv)
+
+    stats = {"imit_stats": imit_stats}
```
AdamGleave, line 171: A lot of boilerplate code here (almost identical to the end of dagger and bc functions) -- can we reduce code duplication some way?

src/imitation/scripts/train_imitation.py
```diff
@@ -140,6 +141,41 @@ def dagger(
...
+    expert_stats = _try_computing_expert_stats(expert_trajs)
+    if expert_stats is not None:
+        stats["expert_stats"] = expert_stats
+
+    return stats
+
+
 def main_console():
     observer_path = pathlib.Path.cwd() / "output" / "sacred" / "train_dagger"
```
AdamGleave, line 180: This script does a lot more than just dagger training, rename from `"train_dagger"` to `"train_imitation"`? (This is an existing problem not introduced in this PR but may as well fix.)

tests/scripts/test_scripts.py
```diff
@@ -409,6 +425,40 @@ def test_train_bc_warmstart(tmpdir):
...
 
+def test_train_sqil_main(sqil_config):
+    # NOTE: Having four different expert types as in bc might be overkill for sqil
+    run = train_imitation.train_imitation_ex.run(**sqil_config)
+    assert run.status == "COMPLETED"
+    assert isinstance(run.result, dict)
+
+
+def test_train_sqil_with_no_demonstrations_raises_value_error(tmpdir):
+    # NOTE: This test might be overkill
```
AdamGleave, line 436: Agreed, I don't think we need this, it's testing `_generate_expert_trajs` in the `demonstrations.py` ingredient, which is a common code path between BC and SQIL

tests/scripts/test_scripts.py
```diff
@@ -409,6 +425,40 @@ def test_train_bc_warmstart(tmpdir):
     assert isinstance(run_warmstart.result, dict)
 
 
+def test_train_sqil_main(sqil_config):
+    # NOTE: Having four different expert types as in bc might be overkill for sqil
```
AdamGleave, line 429: Probably not necessary but seems fine to leave as-is.

AdamGleave (changes requested):
Tests are currently failing (looks like a missing dependency for tqdm), please address.

Some small comments on the code itself.

setup.py
lukasberglund: I've removed this and have instead changed `STABLE_BASELINES3 = "stable-baselines3>=1.7.0,<2.0.0"` to `STABLE_BASELINES3 = "stable-baselines3[extra]>=1.7.0,<2.0.0"`. The bug I got related when running `train_imitation.py` with `sqil` told me to make this change.

src/imitation/algorithms/base.py
lukasberglund: Oh, good point. save_policy was originally there too, but I wanted to move it to `BaseImitationAlgorithm` so that `sqil` could use it. But since they reference each other it makes sense to move `reconstruct_policy` there too, which I've done now.

src/imitation/scripts/train_imitation.py
lukasberglund: > I think can replace with policy.make_policy(venv) and remove the policy_model config parameter.
This wouldn't work because `policy.make_policy(venv)` returns `policies.BasePolicy`, whereas SQIL requires `type[policies.BasePolicy]` (i.e. a constructor, not an instance). 

You are right though that I could use `policy["policy_cls"]`. However, SQIL uses DQN by default which requires a DQNPolicy and the default for `policy_cls` is `base.FeedForward32Policy`. If the default argument doesn't work for SQIL, his would mean that we couldn't run `train_imitation` with the `sqil` command without modifying the policy in the command line. 

One fix would be to change the name of `policy` to `imitation_policy` and `sqil["policy_model"]` to `sqil["sqil_policy_cls"]`. I'll go with this for now, but if you think we should do it differently (e.g. just use policy_model["policy_cls"] for both) I can roll back the changes.

src/imitation/scripts/train_imitation.py
lukasberglund: I've added the RL ingredient without using any of the functions provided which make the RL algorithm, instead only using the configs. Let me know if that works.

src/imitation/scripts/train_imitation.py
lukasberglund: Made a shared function called `_collect_stats`!

src/imitation/scripts/train_imitation.py
lukasberglund: > This script does a lot more than just dagger training

How so? It seems like everything it does (including creating a bc_trainer) is necessary for training dagger.

tests/scripts/test_scripts.py
lukasberglund: Removed.

src/imitation/algorithms/base.py
lukasberglund: I've also now moved save_policy again fyi.

setup.py
```diff
@@ -21,7 +21,7 @@
     "autorom[accept-rom-license]~=0.6.0",
 ]
 PYTYPE = ["pytype==2022.7.26"] if IS_NOT_WINDOWS else []
-STABLE_BASELINES3 = "stable-baselines3>=1.7.0,<2.0.0"
+STABLE_BASELINES3 = "stable-baselines3[extra]>=1.7.0,<2.0.0"
```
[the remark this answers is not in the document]
AdamGleave: I saw you mentioned the error you had previously told you to install `shine` but I don't even see this as a dependency of SB3: https://github.com/DLR-RM/stable-baselines3/blob/master/setup.py Perhaps it's an indirect dependency?

src/imitation/scripts/ingredients/sqil.py
```diff
@@ -0,0 +1,20 @@
+"""This ingredient provides a sqil algorithm instance."""
```
AdamGleave, line 1:
```suggestion
"""This ingredient provides a SQIL algorithm instance."""
```

src/imitation/scripts/config/train_imitation.py
```diff
@@ -96,7 +97,14 @@ def seals_humanoid():
     environment = dict(gym_id="seals/Humanoid-v0")
 
 
+@train_imitation_ex.named_config
+def dqn():
```
AdamGleave, line 101: Is this used for anything other than sqil? If not, does it need to live in this file? (I'm particularly confused that it is changing something in `rl` but `train_imitation` does not directly depend on the `rl` ingredient.)

src/imitation/scripts/train_imitation.py
```diff
@@ -9,12 +9,14 @@
...
 
 from imitation.algorithms.dagger import SimpleDAggerTrainer
+from imitation.algorithms.sqil import SQIL
 from imitation.data import rollout, types
 from imitation.scripts.config.train_imitation import train_imitation_ex
 from imitation.scripts.ingredients import bc as bc_ingredient
 from imitation.scripts.ingredients import demonstrations, environment, expert
 from imitation.scripts.ingredients import logging as logging_ingredient
 from imitation.scripts.ingredients import policy_evaluation
+from imitation.util.util import save_policy
```
AdamGleave, line 19: Our style guide https://google.github.io/styleguide/pyguide.html#22-imports says:

(It seems there are some existing violations of this in the file, so easy to have missed this, but still let's not add to them and fix existing issues where possible.)

src/imitation/scripts/train_imitation.py
```diff
@@ -40,6 +42,18 @@ def _try_computing_expert_stats(
         return None
 
 
+def _collect_stats(
+    imit_stats: Mapping[str, float],
+    expert_trajs: Sequence[types.Trajectory],
+):
```
AdamGleave, line 48: Missing return type annotation

AdamGleave (changes requested):

src/imitation/scripts/train_imitation.py
AdamGleave: That refactor looks good, thanks!

src/imitation/scripts/train_imitation.py
AdamGleave: Yes, this seems like a good solution!

src/imitation/scripts/train_imitation.py
AdamGleave: > This wouldn't work because policy.make_policy(venv) returns policies.BasePolicy, whereas SQIL requires type[policies.BasePolicy] (i.e. a constructor, not an instance).

Ah, good point, yes the difference between SQIL expecting classes v.s. the rest of our code expecting objects bites again.

> But you are right, I could use policy["policy_cls"]. One issue is that SQIL uses a DQN by, default which requires a DQNPolicy and the default for policy_cls is base.FeedForward32Policy, which is incompatible. This is bit unfortunate, but I can't think of a way around it currently. To make it easy I've made a named_config called sqil.dqn which lets users set a dqn policy.

> I currently can't think of a way to make sqil work by default. We would somehow want to override a config conditional on the sqil command being used. Not sure if that's possible/desirable.

Mm indeed this is messy and highlights a design flaw in Sacred. It is possible to have a different default depending on the context using a config hook, https://github.com/HumanCompatibleAI/imitation/blob/master/src/imitation/scripts/ingredients/rl.py#L41 is an example of this. You can add a hook in SQIL that checks if `policy_cls == base.FeedForward32Policy` and if so changes it to a `DQNPolicy`. This is a bit nasty since if for some reason the user *manually* sets the policy to a FeedForward32Policy we'll just override that silently, but it seems OK.

src/imitation/algorithms/base.py
AdamGleave: OK to leave `reconstruct_policy` where it is for now, it's not needed anywhere else right now anyway.

src/imitation/scripts/train_imitation.py
lukasberglund: This seems good. I've made the change.

src/imitation/scripts/train_imitation.py
lukasberglund: Sorry, I thought you were commenting on the train_dagger function. This makes sense. Changed.

setup.py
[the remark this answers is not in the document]
lukasberglund: Good point. I think I meant to write rich, but wrote shine instead. I think having those installed would be enough, so we can avoid install `stable-baselines3[extra]`.

src/imitation/scripts/train_imitation.py
lukasberglund: Fixed this and other imports in `train_imitation.py`

setup.py
```diff
@@ -202,6 +202,7 @@ def get_local_version(version: "ScmVersion", time_format="%Y%m%d") -> str:
         "numpy>=1.15",
         "torch>=1.4.0",
         "tqdm",
+        "rich",
```
lukasberglund, line 205: sb3 doesn't require a specific version of rich

tests/test_benchmarking.py
```diff
@@ -3,7 +3,8 @@
 
 import pytest
 
-from imitation.scripts import train_adversarial, train_imitation
```
AdamGleave, line 6: Why is this change necessary?

AdamGleave (changes requested):
Thanks for making these changes! I think this is almost ready, a few minor comments

src/imitation/scripts/ingredients/sqil.py
```diff
@@ -0,0 +1,48 @@
...
+    total_timesteps = 3e5
+    train_kwargs = dict(
+        log_interval=4,  # Number of updates between Tensorboard/stdout logs
+        progress_bar=True,
+    )
+
+    locals()  # quieten flake8 unused variable warning
+
+
+@rl.rl_ingredient.config_hook
```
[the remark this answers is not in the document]
lukasberglund: I tried doing this originally, but it didn't work. For some reason it would set the variables inside of sqil. E.g. it would set `config["sqil"]["rl"]["rl_cls"]` instead of `config["rl"]["rl_cls"]` as intended.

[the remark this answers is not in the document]
AdamGleave: Can confirm having it in SQIL ingredient will set variables inside of SQIL. More problematically moving hook to train_imitation experiment seems to then have no effect on sub-ingredients. So, ugly though it is, I think we probably do need to keep it here. Good news is these hooks are no-op when command name is not sqil.

AdamGleave (approved):
LGTM

Issue #681, opened 2023-02-14
Support for Dictionary Observation Space


damaruga:
## Problem
Robotic Env such as [SurRoL]( https://github.com/med-air/SurRoL) and [Fetch](https://robotics.farama.org/envs/fetch/) uses Dictionary Observation Space, with 
1. observation
2. desired_goal
3. achieved_goal
as keys.   

## Query
Is there a quick fix to train an agent using BC in the current release for these kinds of environments?

tudorjnu:
Hello! I have a similar issue as I need to use a dictionary input method. If one was to implement this, where would be a good starting point?

ernestum:
As long as you don't plan on treating different parts of the observation space (e.g. applying a CNN to image data but taking in accelerator data directly) you should be good with a `gym.wrappers.FlattenObservation` wrapper around your environment.

tudorjnu:
Personally that is roughly what I need as I have a custom environment generating different types of observations (i.e. image, joint pos, etc.).

On a tangent note, is it possible to generate transtitions using an environment (i.e. dict based env) and store only the image for example and then use imitation to learn those transitions (image -> action)? 

I have been thinking to create a custom rollout wrapper or something along those lines so that when the expert generates the trajectories, they can be filtered. What do you think?


I asked here as it seems to be related to the issue. Thank you!

NixGD:
PR #785 partially addresses this. Once it's merged dictionary observations should be supported in:

Core functionality
- [x] Collecting rollouts
- [ ] Saving / writing trajectories to disk 
- [ ] Buffers

Algorithms:
- [x] Behavioral Cloning
- [ ] DAgger
- [x] Density based reward modelling
- [ ] MCEIRL
- [ ] Adversarial AIRL / GAIL
- [ ] SQIL
- [ ] Preference Comparisons

benediktkreis:
I've seen that the PR #785 was merged. Are there any updates regarding the other algorithms? In particular, I'm interested in dict support for DAgger.

benediktkreis:
Are there any updates regarding dictionary support for DAgger?

PR #785, opened 2023-09-13
Add partial support for dictionary observation spaces (bc, density)

NixGD:
## Description

Partially addresses #681, by adding support for dictionary observation spaces in:
 - Core trajectory gathering and processing code (`types.py`, `rollout.py`, etc.)
 - Behavioral Cloning
 - Density based algorithms

Does not add support for any other algorithms, or trajectory saving / loading.

## Testing

- [x] Add Dict space to observation space parameterization over trajectories in `test_types.py`
- [x] Add explicit tests of `ObsDict` in `test_types.py`
- [x] Add tests of dict observation spaces to `test_rollout.py`
- [x] Add tests of dict observation spaces to `test_bc.py`
- [x] Add tests of dict observation spaces to `density.py`


src/imitation/data/types.py
```diff
@@ -21,7 +26,167 @@
...
 T = TypeVar("T")
 
+TensorVar = TypeVar("TensorVar", np.ndarray, th.Tensor)
+
+
 AnyPath = Union[str, bytes, os.PathLike]
+
+
+@dataclasses.dataclass(frozen=True)
+class DictObs:
```
NixGD, line 36: This is the most important choice in the PR: having a separate class to store observations in a dictionary space, instead of just keeping them in a `dict[str, np.ndarray]` format. I'm honestly not sure if it's the correct call.

The main advantage is that it allows code like `len(obs)` to mean a consistent thing regardless of if the observations are in a dictionary space or not. The alternative would probably be to have a collection of utilities that perform a similar role, such as a `obs_len` function that checks if `obs` is a dictionary.

Imo the main disadvantage is complexity bloat of having an additional class. It's sometimes also occasionally to shuffle observations in and out of DictObs objects.

src/imitation/data/types.py
```diff
@@ -75,7 +240,8 @@ def __eq__(self, other) -> bool:
         if not isinstance(other, Trajectory):
             return False
 
-        dict_self, dict_other = dataclasses.asdict(self), dataclasses.asdict(other)
+        dict_self = dataclass_quick_asdict(self)
```
NixGD, line 243: `dataclasses.asdict` is recursive which unwraps dictobs, so changing to `dataclass_quick_asdict(self)` in several places.

src/imitation/data/types.py
```diff
@@ -339,3 +515,22 @@ def __post_init__(self):
         """Performs input validation, including for rews."""
         super().__post_init__()
         _rews_validation(self.rews, self.acts)
+
+
+ObsType = TypeVar("ObsType", np.ndarray, DictObs)
+
+
+def concatenate_maybe_dictobs(arrs: List[ObsType]) -> ObsType:
```
NixGD, line 523: Maybe these should move into the DictObjs class.

src/imitation/data/huggingface_utils.py
```diff
@@ -124,6 +124,8 @@ def trajectories_to_dict(
         ],
         terminal=[traj.terminal for traj in trajectories],
     )
+    if any(isinstance(traj.obs, types.DictObs) for traj in trajectories):
```
NixGD, line 127: I do intend to support this.

src/imitation/data/types.py
```diff
@@ -22,6 +27,177 @@
...
+    array.
+
+    There are also utility functions for mapping / stacking / concatenating
+    lists of dictobs.
+    """
+
+    d: Dict[str, np.ndarray]
+
+    @classmethod
+    def from_obs_list(cls, obs_list: List[Dict[str, np.ndarray]]):
```
AdamGleave, line 48: Nit: check that all elements have the same keys? (This will error if `obs_list[i]` contains a key not present in `obs_list[0]` but not vice-versa.)

src/imitation/data/types.py
```diff
@@ -22,6 +27,177 @@
...
+        return cls(
+            {k: np.stack([obs[k] for obs in obs_list]) for k in obs_list[0].keys()},
+        )
+
+    def __post_init__(self):
+        if not all((isinstance(v, np.ndarray) for v in self.d.values())):
+            raise ValueError("keys must by numpy arrays")
+
+    def __len__(self):
+        """Returns the first dimension of constiuent arrays.
```
AdamGleave, line 58:
```suggestion
        """Returns the first dimension of constituent arrays.
```

src/imitation/data/types.py
```diff
@@ -22,6 +27,177 @@
...
+        return all(np.array_equal(self.d[k], other.d[k]) for k in self.d.keys())
+
+    @property
+    def shape(self) -> Dict[str, Tuple[int, ...]]:
+        """Returns a dictionary with shape-tuples in place of the arrays."""
+        return {k: v.shape for k, v in self.d.items()}
+
+    @property
+    def dtype(self) -> Dict[str, np.dtype]:
+        """Returns a dictionary with shape-tuples in place of the arrays."""
```
AdamGleave, line 131:
```suggestion
        """Returns a dictionary with dtype-tuples in place of the arrays."""
```

src/imitation/data/types.py
```diff
@@ -22,6 +27,177 @@
...
+        return {k: v.dtype for k, v in self.d.items()}
+
+    def unwrap(self) -> Dict[str, np.ndarray]:
+        return self.d
```
AdamGleave, line 135: Do we want to allow callers to mutate our internal dict? (This is a genuine question, I'm not sure.)

src/imitation/data/types.py
```diff
@@ -22,6 +27,177 @@
...
+            {
+                k: np.concatenate(arr_list)
+                for k, arr_list in cls._unravel(dictobs_list).items()
+            },
+        )
+
+    # TODO: add keys, values, items?
+
+
+def assert_not_dictobs(x: Union[np.ndarray, DictObs]) -> np.ndarray:
```
AdamGleave, line 203: Nit: `assert` sounds like it should raise an AssertionError but this raises a ValueError. Consider changing name of method or the exception raised.

src/imitation/data/types.py
```diff
@@ -168,12 +351,16 @@ def transitions_collate_fn(
         info dict into a single dict, which is incorrect.)
     """
     batch_no_infos = [
```
AdamGleave, line 362: This variable name is now a bit of a misnomer (it now also doesn't include observations).

src/imitation/data/types.py
```diff
@@ -304,16 +491,19 @@ class Transitions(TransitionsMinimal):
...
-            raise ValueError(
-                "obs and next_obs must have same shape: "
-                f"{self.obs.shape} != {self.next_obs.shape}",
-            )
-        if self.obs.dtype != self.next_obs.dtype:
-            raise ValueError(
-                "obs and next_obs must have the same dtype: "
-                f"{self.obs.dtype} != {self.next_obs.dtype}",
-            )
+        # TODO: could add support for checking dictobs shape/dtype of each array
```
AdamGleave, line 503: I think you added this now :)

src/imitation/data/types.py
AdamGleave: I'm generally fine with module-level helper methods in Python, we can always make them private if not used outside of this module.

src/imitation/policies/base.py
```diff
@@ -23,18 +24,31 @@ def __init__(self, observation_space: gym.Space, action_space: gym.Space):
             action_space=action_space,
         )
 
-    def _predict(self, obs: th.Tensor, deterministic: bool = False):
+    def _predict(
+        self,
+        obs: Union[th.Tensor, Dict[str, th.Tensor]],
```
AdamGleave, line 29: Do we need to change this? `predict` already supports dict-typed observations: https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/policies.py#L323

``` 
observation: Union[np.ndarray, Dict[str, np.ndarray]],
```

I think the way this works in SB3 is that the feature extractor extracts all the values from the dict and concatenate them together for feeding into neural network. We don't want that but I wonder if we can just operate at the level of abstraction of `predict` and be OK?

src/imitation/policies/exploration_wrapper.py
```diff
@@ -57,7 +57,7 @@ def __init__(
 
     def _random_policy(
         self,
-        obs: np.ndarray,
+        obs: Union[np.ndarray, types.DictObs],
```
AdamGleave, line 60: Might be worth defining `Union[np.ndarray, types.DictObs]` as a type alias given it occurs so much.

src/imitation/data/rollout.py
```diff
@@ -107,17 +108,22 @@ def finish_trajectory(
...
-            k: np.stack(arr_list, axis=0) for k, arr_list in out_dict_unstacked.items()
-        }
-        traj = types.TrajectoryWithRew(**out_dict_stacked, terminal=terminal)
-        assert traj.rews.shape[0] == traj.acts.shape[0] == traj.obs.shape[0] - 1
+
+        # TODO: what about infos? Does this actually handle them well?
+        traj = types.TrajectoryWithRew(
+            obs=types.stack_maybe_dictobs(out_dict_unstacked["obs"]),
+            acts=np.stack(out_dict_unstacked["acts"], axis=0),
+            infos=np.stack(out_dict_unstacked["infos"], axis=0),  # TODO: confused
```
AdamGleave, line 116: I think it forms a numpy object-dtype array. It's a bit silly/unnecessary that this is a NumPy array rather than just a list but AFAIK it works OK.

src/imitation/data/rollout.py
```diff
@@ -142,20 +148,26 @@ def add_steps_and_auto_finish(
             each `True` in the `dones` argument.
         """
         trajs: List[types.TrajectoryWithRew] = []
-        for env_idx in range(len(obs)):
+        wrapped_obs = types.DictObs.maybe_wrap(obs)
```
AdamGleave, line 151: Nit: does this belong in `DictObs`? A factory method that always returns `DictObs` would. Something that *sometimes* returns `DictObs` feels borderline.

src/imitation/data/rollout.py
```diff
@@ -142,20 +148,26 @@ def add_steps_and_auto_finish(
...
+        zip_iter = enumerate(zip(acts, wrapped_obs, rews, dones, infos))
         for env_idx, (act, ob, rew, done, info) in zip_iter:
             if done:
                 # When dones[i] from VecEnv.step() is True, obs[i] is the first
                 # observation following reset() of the ith VecEnv, and
                 # infos[i]["terminal_observation"] is the actual final observation.
                 real_ob = info["terminal_observation"]
+                if isinstance(real_ob, dict):
+                    # TODO: does this need to be unsqueezed or something?
+                    real_ob = types.DictObs(real_ob)
```
AdamGleave, line 170: I'm confused why this is being applied here (only to the terminal observation) rather than always (i.e. after the else statement on line 173, unindented).

src/imitation/data/rollout.py
```diff
@@ -460,9 +493,10 @@ def generate_trajectories(
     for trajectory in trajectories:
         n_steps = len(trajectory.acts)
         # extra 1 for the end
-        exp_obs = (n_steps + 1,) + venv.observation_space.shape
-        real_obs = trajectory.obs.shape
-        assert real_obs == exp_obs, f"expected shape {exp_obs}, got {real_obs}"
+        if not isinstance(venv.observation_space, spaces.dict.Dict):
```
AdamGleave, line 496: probably want some analogous check for dict observation spaces? Dict is a composite space so should be able to do some kind of recursive check I think: https://gymnasium.farama.org/api/spaces/composite/

src/imitation/data/rollout.py
```diff
@@ -558,11 +593,19 @@ def flatten_trajectories(
...
 
     cat_parts = {
-        key: np.concatenate(part_list, axis=0) for key, part_list in parts.items()
+        key: types.concatenate_maybe_dictobs(part_list)
+        for key, part_list in parts.items()
     }
     lengths = set(map(len, cat_parts.values()))
     assert len(lengths) == 1, f"expected one length, got {lengths}"
     return types.Transitions(**cat_parts)
+    # TODO: clean
```
AdamGleave, line 602: can this just be deleted now?

src/imitation/algorithms/bc.py
```diff
@@ -114,9 +120,23 @@ def __call__(
...
+        if isinstance(obs, types.DictObs):
+            tensor_obs = {k: util.safe_to_tensor(v) for k, v in obs.unwrap().items()}
+        elif isinstance(obs, dict):
+            tensor_obs = {k: util.safe_to_tensor(v) for k, v in obs.items()}
+        else:
+            tensor_obs = util.safe_to_tensor(obs)
         acts = util.safe_to_tensor(acts)
-        _, log_prob, entropy = policy.evaluate_actions(obs, acts)
+
+        # policy.evaluate_actions's type signature seems wrong to me.
```
AdamGleave, line 132: I think `evaluate_action` is assuming that `extract_features` is `th.Tensor -> th.Tensor` like the default implementation (although I agree you could override it).

`obs_to_tensor` seems to be resopnsible for dict flattening: https://github.com/DLR-RM/stable-baselines3/blob/master/stable_baselines3/common/policies.py#L231 (this may contradict information I told you in the past, I wrongly thought feature extractors did this, I think they maybe used to in older versions of SB3?)

src/imitation/algorithms/bc.py
```diff
@@ -114,9 +120,23 @@ def __call__(
             A BCTrainingMetrics object with the loss and all the components it
             consists of.
         """
-        obs = util.safe_to_tensor(obs)
+        tensor_obs: Union[th.Tensor, Dict[str, th.Tensor]]
+        if isinstance(obs, types.DictObs):
```
AdamGleave, line 124: This should maybe be a utility method (feels useful outside of BC).

AdamGleave (commented):
Overall this design looks good to me. As you mentioned lots of things that needed to be cleaned up but I think it's going in the right direction. I did a fairly detailed review of data/types.py, data/rollout.py and algorithms/bc.py but just skimmed the rest so do highlight if there are any other important areas.

src/imitation/algorithms/density.py
```diff
@@ -168,9 +168,11 @@ def set_demonstrations(self, demonstrations: base.AnyTransitions) -> None:
 
         if isinstance(demonstrations, types.TransitionsMinimal):
             next_obs_b = getattr(demonstrations, "next_obs", None)
+            if next_obs_b is not None:
+                next_obs_b = types.assert_not_dictobs(next_obs_b)
             transitions.update(
                 self._get_demo_from_batch(
-                    demonstrations.obs,
+                    types.assert_not_dictobs(demonstrations.obs),
```
AdamGleave, line 175: It's fine for density not to support dict observations, although we could probably also fairly easily just flatten them...

src/imitation/algorithms/mce_irl.py
```diff
@@ -411,18 +422,30 @@ def set_demonstrations(self, demonstrations: MCEDemonstrations) -> None:
...
+                None,
+                None,
+            )
         elif isinstance(demonstrations, Iterable):
             # Demonstrations are a Torch DataLoader or other Mapping iterable
             # Collect them together into one big NumPy array. This is inefficient,
             # we could compute the running statistics instead, but in practice do
             # not expect large dataset sizes together with MCE IRL.
-            collated_list = collections.defaultdict(list)
+            collated_list: Dict[
```
AdamGleave, line 440: `black` has done a terrible job formatting this! Not sure if there's much we can do unfortunately other than potentially define a type alias for part of this.

src/imitation/data/huggingface_utils.py
AdamGleave: Yes, we should support this eventually, although given time constraints (and size of this PR) I'd be fine leaving it undone and just opening an issue to track adding support for this eventually.

tests/algorithms/test_bc.py
```diff
@@ -371,3 +375,59 @@ def inc_batch_cnt():
 
     # THEN
     assert batch_cnt == no_yield_after_iter
+
+
+class FloatReward(gym.RewardWrapper):
```
AdamGleave, line 380: why do we need this? shouldn't reward be a float already?

AdamGleave: Ah, I see SimpleMultiObsEnvs sometimes returns 1 rather than 1.0. This is a bug we should probably fix upstream. I'm a maintainer of SB3 so if you make a PR I can review it.

tests/algorithms/test_bc.py
```diff
@@ -371,3 +375,59 @@ def inc_batch_cnt():
...
+        observation_space=env.observation_space,
+        policy=policy,
+        action_space=env.action_space,
+        rng=rng,
+        demonstrations=transitions,
+    )
+
+    bc_trainer.train(n_epochs=1)
+
+    reward, _ = evaluation.evaluate_policy(
```
AdamGleave, line 428: Not clear what this is checking for -- we're not asserting anything with reward (and the "expert" is random so probably won't get better than random reward anyway?)

AdamGleave (commented):
One thing to think about is how much we want to push dict observations through the code. In many cases for the imitation algorithms we could basically just flatten the dict to a NumPy array and "add" dict support with minimal additional code changes.

But that kind of defeats the purpose of adding dict support -- may as well just flatten the observations at the environment level. On the other hand, we have to flatten them at some point: the neural network will take a tensor as input not a dict. So, a lot of the design decision is deciding where to do the flattening.

It seems nice to be able to preserve the dict up until calling the policy. This gives flexibility to the user. In our case, InteractivePolicy can ignore the non-rendering component. In general, a user might want to preprocess different components of the observation differently.

NixGD:
One sad thing to note is we don't support arbitrarily nested dictionaries.

src/imitation/algorithms/bc.py
NixGD: `obs_to_tensor` claims to flatten dicts, but it doesn't

```
from gym import spaces
from stable_baselines3.common import policies

obs_space = spaces.Dict({"a": spaces.Discrete(2,), "b": spaces.Box(0,1,(2,))})
policy = policies.MultiInputActorCriticPolicy(
    obs_space,
    spaces.Discrete(4,),
    lambda _: 0.001,
)
obs = obs_space.sample()
policy.obs_to_tensor(obs)
```
gives `({'a': tensor([1]), 'b': tensor([[0.7694, 0.0154]])}, False)`

NixGD: Probably worth making an issue in SB3, I didn't see any previous discussion when I looked earlier

src/imitation/data/rollout.py
NixGD: moved outside into a cluster of utility methods in types.py

tests/data/test_types.py
```diff
@@ -179,20 +194,30 @@ def test_traj_unequal_to_perturbations(
             assert trajectory != types.Trajectory(
                 obs=trajectory.obs[: new_length + 1],
                 acts=trajectory.acts[:new_length],
-                infos=trajectory.obs[:new_length],
```
NixGD, line 182: unrelated bug fix (I think?)

tests/data/test_types.py
```diff
@@ -179,20 +194,24 @@ def test_traj_unequal_to_perturbations(
...
             for k in as_dict.keys():
                 perturbed = dict(as_dict)
                 if k == "infos":
                     perturbed["infos"] = [{"foo": 42}] * len(as_dict["infos"])
+                elif isinstance(as_dict[k], types.DictObs):
+                    perturbed[k] = as_dict[k].map_arrays(lambda x: x + 1)
                 else:
                     perturbed[k] = as_dict[k] + 1
-                assert trajectory != type(t)(**perturbed)
+                assert t != type(t)(**perturbed)
```
NixGD, line 214: Unrelated minor bug fix. Previously would always return false for `t = trajectory_rew` even without perturbations

tests/algorithms/test_bc.py
NixGD: PR [here](https://github.com/DLR-RM/stable-baselines3/pull/1676). Though the environment is more sketchy the more I look at it, e.g. they hardcode the possible transitions from each state in a way that doesn't depend on gridworld size. So maybe best to not depend on the environment much (or submit more fixes upstream).

NixGD:
> One thing to think about is how much we want to push dict observations through the code. In many cases for the imitation algorithms we could basically just flatten the dict to a NumPy array and "add" dict support with minimal additional code changes.
> 
> But that kind of defeats the purpose of adding dict support -- may as well just flatten the observations at the environment level. On the other hand, we have to flatten them at some point: the neural network will take a tensor as input not a dict. So, a lot of the design decision is deciding where to do the flattening.
> 
> It seems nice to be able to preserve the dict up until calling the policy. This gives flexibility to the user. In our case, InteractivePolicy can ignore the non-rendering component. In general, a user might want to preprocess different components of the observation differently.

I agree, I think the Policy is the place to handle the dict -> network input transition. This is consistent with SB3 (see [here](https://stable-baselines3.readthedocs.io/en/master/guide/custom_policy.html#multiple-inputs-and-dictionary-observations)) although the type signatures in SB3 obfuscate this fact.

src/imitation/data/rollout.py
NixGD: It's applied to all observations above (before we can iterate through the environments).

src/imitation/data/huggingface_utils.py
NixGD: Will not support in this PR.

src/imitation/data/types.py
NixGD: Now returns a copy of the dict. It doesn't copy arrays, so still possible to mutate, but that seems like the right compromise.

src/imitation/data/types.py
```diff
@@ -22,7 +29,267 @@
...
+    @classmethod
+    def from_obs_list(cls, obs_list: List[Dict[str, np.ndarray]]) -> "DictObs":
+        """Stacks the observation list into a single DictObs."""
+        return cls.stack(map(cls, obs_list))
+
+    def __post_init__(self):
+        if not all(
+            isinstance(v, (np.ndarray, numbers.Number)) for v in self._d.values()
+        ):
+            raise TypeError("keys must by numpy arrays")
```
AdamGleave, line 60:
```suggestion
            raise TypeError("Values must be NumPy arrays")
```

src/imitation/data/types.py
```diff
@@ -22,7 +29,267 @@
...
+        Len of a DictObs usually represents number of timesteps, or number of
+        environments in a VecEnv.
+
+        Use `dict_len` to get the number of entries in the dictionary.
+
+        Raises:
+            RuntimeError: if the arrays have different lengths or there are no arrays.
+
+        Returns:
+            The length (first dimension) of the constiuent arrays
```
AdamGleave, line 77:
```suggestion
            The length (first dimension) of the constituent arrays
```

src/imitation/data/types.py
```diff
@@ -22,7 +29,267 @@
...
+    def __contains__(self, key):
+        return key in self._d
+
+    def get(self, key: str) -> np.ndarray:
+        """Returns the array for the given key, or raises KeyError."""
+        return self._d[key]
+
+    def unwrap(self) -> Dict[str, np.ndarray]:
+        """Returns a copy of the underlying dictionary (arrays are not copied)."""
+        return {k: v for k, v in self._d.items()}
```
AdamGleave, line 160: I think this is equivalent to:
```suggestion
        return dict(self._d)
```

src/imitation/data/types.py
```diff
@@ -22,7 +29,267 @@
...
+        it1, it2 = itertools.tee(dictobs_list)
+        # assert all have same keys
+        key_set = set(frozenset(obs.keys()) for obs in it1)
+        if len(key_set) == 0:
+            raise ValueError("Empty list of DictObs")
+        if not len(key_set) == 1:
+            raise ValueError(f"Inconsistent keys: {key_set}")
+
+        unraveled: Dict[str, List[np.ndarray]] = collections.defaultdict(list)
+        for do in it2:
```
AdamGleave, line 178: Nit: `do` not obviously an abbreviation for dict obs rather than to "do" something.

src/imitation/data/types.py
```diff
@@ -22,7 +29,267 @@
...
+        """Returns a single dictobs concatenating the arrays by key."""
+        return cls(
+            {
+                k: np.concatenate(arr_list, axis=axis)
+                for k, arr_list in cls._unravel(dictobs_list).items()
+            },
+        )
+
+
+# DicObs utilities
```
AdamGleave, line 204:
```suggestion
# DictObs utilities
```

src/imitation/data/types.py
```diff
@@ -22,7 +29,267 @@
...
+
+
+Observation = Union[np.ndarray, DictObs]
+ObsVar = TypeVar("ObsVar", np.ndarray, DictObs)
+
+
+def assert_not_dictobs(x: Observation) -> np.ndarray:
+    """Typeguard to assert `x` is an array, not a DictObs."""
+    if isinstance(x, DictObs):
```
AdamGleave, line 213: Perhaps more idiomatic: `assert not isinstance(...)`?

src/imitation/data/types.py
```diff
@@ -22,7 +29,267 @@
...
+def stack_maybe_dictobs(arrs: List[ObsVar]) -> ObsVar:
+    """Stacks a list of observations appropriately (depending on type)."""
+    assert len(arrs) > 0
+    if isinstance(arrs[0], DictObs):
+        return DictObs.stack(arrs)
+    else:
+        return np.stack(arrs)
+
+
+# the following overloads have a type error as a DictObs matches both definitions, but
```
AdamGleave, line 236: Sad. Could maybe have overloads for each expected value of T (`np.ndarray`, `th.Tensor`, `int`) and just treat the others as type errors? Unsure how strict we want to be for `maybe_unwrap_dictobs`.

src/imitation/data/types.py
```diff
@@ -22,7 +29,267 @@
...
+    if isinstance(obs, dict):
+        return DictObs(obs)
+    else:
+        if not isinstance(obs, (np.ndarray, DictObs, float, int)):
+            warnings.warn(f"tried to wrap {type(obs)} as an observation")
+        return obs
+
+
+def map_maybe_dict(fn, maybe_dict):
+    """Applies fn to all values a dictionary, or to the value itself if not a dict."""
```
AdamGleave, line 284:
```suggestion
    """Applies fn to all values in a dictionary, or to the value itself if not a dict."""
```

src/imitation/data/types.py
```diff
@@ -22,7 +29,267 @@
...
+    if isinstance(maybe_dict, dict):
+        return {k: fn(v) for k, v in maybe_dict.items()}
+    else:
+        return fn(maybe_dict)
+
+
+# TODO: maybe should support DictObs?
```
AdamGleave, line 291: Nit: https://google.github.io/styleguide/pyguide.html#312-todo-comments encouraged to leave username or issue in TODO, e.g. `TODO(nix): `, so future developers know where to find more info.

src/imitation/data/rollout.py
```diff
@@ -419,13 +437,17 @@ def generate_trajectories(
...
     active = np.ones(venv.num_envs, dtype=bool)
-    assert isinstance(obs, np.ndarray), "Dict/tuple observations are not supported."
     state = None
     dones = np.zeros(venv.num_envs, dtype=bool)
     while np.any(active):
+        # policy gets unwrapped observations (eg as dict, not dictobs)
         acts, state = get_actions(obs, state, dones)
         obs, rews, dones, infos = venv.step(acts)
-        assert isinstance(obs, np.ndarray)
+        assert isinstance(
```
AdamGleave, line 446: Observation space shouldn't change type so might be overkill to assert this here as well as in line 417, although is harmless to do so and would catch a space-changing environment bug.

src/imitation/data/rollout.py
```diff
@@ -538,8 +565,19 @@ def flatten_trajectories(
...
+    def all_of_type(key, desired_type):
+        return all(
+            isinstance(getattr(traj, key), desired_type) for traj in trajectories
+        )
+
+    assert all_of_type("obs", types.DictObs) or all_of_type("obs", np.ndarray)
+    assert all_of_type("acts", np.ndarray)
+
+    # mypy struggles without Any annotation here.
+    # The necessary constraints are enforced above.
```
AdamGleave, line 578: 👍 this seems like right choice (trying to type check this all probably not worth the pain)

src/imitation/algorithms/bc.py
AdamGleave: Oh, huh, well spotted -- yes, we should open an issue in SB3 and have them fix their type annotations, this is very confusing right now.

AdamGleave: Suggest making that issue and converting this to a TODO linking to that issue.

src/imitation/algorithms/bc.py
```diff
@@ -325,12 +342,18 @@ def __init__(
         self.rng = rng
 
         if policy is None:
+            extractor = (
+                torch_layers.CombinedExtractor
```
AdamGleave, line 346: This is pleasingly straightforward.

AdamGleave (commented):
Halfway through, hitting submit so you can start responding to comments

src/imitation/algorithms/density.py
NixGD: now supported :)

tests/data/test_types.py
```diff
@@ -434,3 +456,65 @@ def test_parse_path():
...
+    with pytest.raises(IndexError):
+        abc[:, 0]
+
+    # iter
+    for i, a_row in enumerate(A):
+        np.testing.assert_equal(a_row, ab[i].get("a"))
+    assert ab[0] == next(iter(ab))
+
+    # eq
+    assert abc == types.DictObs({"a": A, "b": B, "c": C})
```
AdamGleave, line 492:
```suggestion
    assert abc == types.DictObs({"a": A, "b": B, "c": C})
    assert abc == types.DictObs({"a": np.array(A), "b": np.array(B), "c": np.array(C)})
```
In case just testing NumPy array identity rather than equality

src/imitation/data/types.py
```diff
@@ -288,8 +289,20 @@ def map_maybe_dict(fn, maybe_dict):
...
+
+    obs: Union[Observation, th.Tensor]
+    acts: AnyTensor
+
+
+# inheritance with total=False so these are not required
+class TransitionMapping(TransitionMappingNoNextObs, total=False):
+    """Dictionary with `obs` and `acts`, maybe also `next_obs`, `dones`, `rew`."""
+
+    next_obs: AnyTensor
```
AdamGleave, line 303:
```suggestion
    next_obs: Union[Observation, th.Tensor]
```
for consistency with `obs` above?

tests/algorithms/test_base.py
```diff
@@ -41,7 +41,10 @@ def test_check_fixed_horizon_flag(custom_logger):
 
 
 def _make_and_iterate_loader(*args, **kwargs):
-    loader = base.make_data_loader(*args, **kwargs)
+    # our pytype version doesn't understand optional arguments in TypedDict
```
AdamGleave, line 44: Probably fine to upgrade pytype -- we had the habit of specifying test-only dependencies as `==` so that CI wouldn't spuriously break (whereas for install dependencies we *want* it to break if there's some upstream incompatibility, since it'll hurt our users)

src/imitation/algorithms/bc.py
NixGD: https://github.com/DLR-RM/stable-baselines3/issues/1679

src/imitation/data/rollout.py
NixGD: agree. It's largely here to make the typechecker happy

tests/data/test_types.py
```diff
@@ -208,6 +227,9 @@ def test_save_trajectories(
         use_rewards,
         type_safe,
     ):
+        if isinstance(trajectory.obs, types.DictObs):
```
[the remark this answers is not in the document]
NixGD: Is this list good enough or do you want a separate issue? https://github.com/HumanCompatibleAI/imitation/issues/681#issuecomment-1722073345

NixGD:
Current state: this PR is very close to being mergable but my work trial is over and someone else will need to get it over the finish line.

The only outstanding TODO before merging (other than outstanding comments above, which are pretty minor) would be to add tests for using dictionary observations in density.py. This should be simple, I just haven't done it. Alternatively, one could add back the assert statements removed in [this commit](https://github.com/HumanCompatibleAI/imitation/pull/785/commits/8d1e0d66dba180c9fddc7a860ca1184f7f521332#diff-ec9cbb666923b5775e14a0acdaa5ee592140733e7484734fd1961f0cdc0d07b2) so that using dict observation assertion errors until it is tested.

NixGD:
There are other future tasks required to get imitation to fully support dictionary observations, outside the scope of this PR. I summarized the current status in [this comment](https://github.com/HumanCompatibleAI/imitation/issues/681#issuecomment-1722073345). Both saving/loading trajectories and buffer support seem doable but non trivial.

tests/algorithms/test_bc.py
araffin: SimpleMultiObsEnvs is just a test env, it is not meant to be used except in tests.

NixGD: Yep, we'd only be using it in tests as well.

tests/algorithms/test_bc.py
ZiyueWang25, line 1: Disregard this commit, we still need FloatReward because the newest package isn't released for us to use yet.

src/imitation/algorithms/density.py
```diff
@@ -147,7 +147,8 @@ def _get_demo_from_batch(
         assert act_b.shape[1:] == self.venv.action_space.shape
 
         if isinstance(obs_b, types.DictObs):
-            exp_shape = {k: v.shape for k, v in self.venv.observation_space.items()}
+            exp_shape = {k: v.shape for k, v in self.venv.observation_space.items()}  # type: ignore[attr-defined] # noqa: E501
```
AdamGleave, line 150: We try to minimize number of errors we suppress -- can we reformat the code rather than suppressing the line length error?

src/imitation/data/rollout.py
```diff
@@ -397,25 +408,22 @@ def generate_trajectories(
         Sequence of trajectories, satisfying `sample_until`. Additional trajectories
         may be collected to avoid biasing process towards short episodes; the user
         should truncate if required.
-
```
AdamGleave, line 400: Why are we removing this?

tests/algorithms/test_density_baselines.py
```diff
@@ -168,3 +170,54 @@ def test_density_trainer_raises(
...
+# merged and released.
+class FloatReward(gym.RewardWrapper):
+    """Typecasts reward to a float."""
+
+    def reward(self, reward):
+        return float(reward)
+
+
+def test_dict_space():
+    def make_env():
```
AdamGleave, line 185: Code duplication between this and `test_bc.py`. One potential resolution is to make this a pytest test fixture defined in e.g. `tests/algorithms/conftest.py`

tests/data/test_wrappers.py
```diff
@@ -1,14 +1,18 @@
...
 import gymnasium as gym
 import numpy as np
 import pytest
 from stable_baselines3.common.vec_env import DummyVecEnv
 
 from imitation.data import types
-from imitation.data.wrappers import BufferingWrapper
+from imitation.data.wrappers import (
+    BufferingWrapper,
+    HumanReadableWrapper,
```
AdamGleave, line 13: Should be split into separate PR.

AdamGleave (changes requested):
Mostly looks good. Left some minor comments. The main high-level issues:
  - Linting and type checking are failing. Can you fix these? We need CI green (except the known issues in the pipeline that we're in process of fixing) before merging.
  - There's some changes here that seem work-in-progress and related to adding support for human-readable observations. I think these are best handled in a separate PR since this PR is already rather large.

tests/data/test_wrappers.py
ZiyueWang25: Yes. Sorry for the confusion. I didn't do well on branching. Just fixed it.

src/imitation/data/wrappers.py
```diff
@@ -203,3 +210,56 @@ def step(self, action):
                 "rews": np.stack(self._rews),
             }
         return obs, rew, terminated, truncated, info
+
+
+class HumanReadableWrapper(gym.Wrapper):
```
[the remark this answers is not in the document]
ZiyueWang25: Yes, indeed. This is from my wrong branching practice. Fixed.

src/imitation/data/rollout.py
ZiyueWang25: Because gym.spaces.Dict doesn't have shape and I don't find this check helpful or provide valuable information for later code. Maybe I missed something here, given we support Dict obs now, do you still think it is necessary to check here?

image

ZiyueWang25:
Finished addressing your comments. Please take another look.

I moved pytype upgrade related issue to #801 because it turns out to be more complicated than simply fixing some types.

src/imitation/data/rollout.py
AdamGleave: OK, makes sense!

src/imitation/algorithms/dagger.py
```diff
@@ -508,7 +508,7 @@ def create_trajectory_collector(self) -> InteractiveTrajectoryCollector:
         beta = self.beta_schedule(self.round_num)
         collector = InteractiveTrajectoryCollector(
             venv=self.venv,
-            get_robot_acts=lambda obs: self.bc_trainer.policy.predict(obs)[0],
+            get_robot_acts=lambda acts: self.bc_trainer.policy.predict(acts)[0],
```
AdamGleave, line 511: `predict` takes an **obs**ervation and returns **act**ion**s** (so I think code was right before).

AdamGleave (approved):
LGTM -- one minor typo to fix but no need for a re-review. Can merge once we get CI green.

src/imitation/algorithms/dagger.py
ZiyueWang25: Yes I agree. Actually, the `acts` is from master branch. I planned to change `acts` to `obs` and in the end put it under #802 . You see `acts` here because I revert that change and put it under #802. So I would prefer still keep it here and fix it in #802.

saeed349:
Appreciate adding this support. 
Would be really good to see dict support extended to other algorithms as well.

benediktkreis:
Are there any updates on this? In particular, I'm interested in dict support for DAgger.

Issue #790, opened 2023-09-30
Mismatch between Tutorial doc and the actual jupyter notebook


ZiyueWang25:
## Bug description
The [tutorial](https://imitation.readthedocs.io/en/latest/tutorials/3_train_gail.html) "Train an Agent using Generative Adversarial Imitation Learning" doesn't match [the actual jupyter notebook](https://github.com/HumanCompatibleAI/imitation/blob/master/docs/tutorials/3_train_gail.ipynb) and it seems like it should be updated.

This also happens for other tutorials as well.

ernestum:
This is probably because some jupyter notebook jobs silently failed as described in #799. Should be fixed by #803

PR #792, opened 2023-10-02
Notebook fix

ZiyueWang25:
## Description

Fix notebook running error.  #790 and #779.
1. docs/tutorials/3_train_gail.ipynb: 800k can bring running time exceed error
2. docs/tutorials/4_train_airl.ipynb: env.reset doesn't require seed in gymnasium setting.
3. docs/tutorials/5a_train_preference_comparisons_with_cnn.ipynb: migrate to gymnasium
4. docs/tutorials/8a_train_sqil_sac.ipynb: v0 is not supported.

These issues only appear after #779 get half-fixed.

## Testing

`pytest tests/test_examples.py::test_run_tutorial_notebooks` success


ZiyueWang25:
This should fix #790 .

ZiyueWang25:
I think a easier case would be putting these fix all under #779 so that we can see there is no failure. I will work on that.

ZiyueWang25:
I add the fix here to #779

PR #797, opened 2023-10-04
Fix MacOS Pipeline: Include tests not in subdirectories

AdamGleave:
## Description

Previously the MacOS pipeline was only running in tests that were in subdirectories of `tests/`. This is because the glob was not quoted, so was being expanded by the shell (which did not properly parse `**`) rather than the CircleCI script

## Testing

Before: https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3888/workflows/a8c06177-af9f-4ec2-9bba-2e384f3d35c3/jobs/15718/parallel-runs/0/steps/0-109 -- observe all tests are in subdirectories
After: https://app.circleci.com/pipelines/github/HumanCompatibleAI/imitation/3925/workflows/88a81adc-aa48-4f1f-86cb-d8833171316a/jobs/15952/parallel-runs/0/steps/0-109 -- observe tests include those directly in tests such as `tests/test_examples.py`


AdamGleave:
The Mac CI is red but this is because notebook tests that were previously being skipped are in fact failing.

At least some of the notebook tests seem cross-platform, so I suggest we review this PR first and consider merging it to get the CI functional, then address fixing the issues the now-functional CI is picking up in separate PRs.

AdamGleave:
Only tests failing are related to 8a_train_sqil_sac, so merging -- these should get fixed in #800

PR #800, opened 2023-10-04
Remove MuJoCo dependency from SQIL notebook

### LINKED ISSUE #798: Update 8a_train_sqil_sac.ipynb to remove MuJoCo dependency
## Bug description
We decided against depending on MuJoCo in CI, but https://github.com/HumanCompatibleAI/imitation/blob/8db6bfb0cfd533f6418e662890499670bd6b010b/docs/tutorials/8a_train_sqil_sac.ipynb has introduced this dependency into our test suite.

## Steps to reproduce
Observe tutorial uses HalfCheetah which is a MuJoCo environment.

Suggested resolution: replace with an environment that is native Gymnasium, like [MountainCarContinuous](https://gymnasium.farama.org/environments/classic_control/mountain_car_continuous/] or [Pendulum](https://gymnasium.farama.org/environments/classic_control/pendulum/)

AdamGleave:
## Description

Fixes #798.

WIP: @ernestum can you make a HuggingFace dataset `ppo-Pendulum-v1` out of the same-name model? I think that will be enough to make this notebook work. I'm not sure how you generated datasets so deferring to you on this.


ernestum:
Pushed it [here](https://huggingface.co/datasets/HumanCompatibleAI/ppo-Pendulum-v1). Let's see what it does ...

ernestum:
Seems like the SQIL notebook tests pass (even though other tests still fail due to issues addressed in other PRs.

Rocamonde (commented):
this looks fine to me, even though I'm missing context so I cannot speak for the high-level intent of whether replacing it with the Pendulum is the right choice. Only thing seems to be that you're no longer using seals? Is this intentional?

ernestum:
The context is in the linked issue #798. Basically we just wanted to get rid of the MuJoCo dependency in the test pipeline. So we just replaced the environment with anything basic that is not MuJoCO.

PR #801, opened 2023-10-04
Upgrade pytype

ZiyueWang25:
## Description

Originally from this [commit](https://github.com/HumanCompatibleAI/imitation/pull/785/commits/194ec1ac336d5ecdb36a5df92d53e056458291be) to remove workaround for old versions. But it turns out we need to address many more type issues than expected. So I moved pytype related fix from #785 to here.

## Testing

`pytype --version && pre-commit run --all pytype` shows no error locally. Not sure why it makes CI red for now.


ZiyueWang25:
I am not sure about the solution inside [context manager related fix](https://github.com/HumanCompatibleAI/imitation/pull/801/commits/ec0061428f0b7aca4eff46e5fde4e362a2eed847), they are all related with ` environment.make_venv` or `environment.make_rollout_venv()`. Seems like pytype detects some false positive in the new version.

.pre-commit-config.yaml
```diff
@@ -78,7 +78,7 @@ repos:
     name: pytype
     language: system
     types: [python]
-    entry: "bash -c 'pytype -j ${NUM_CPUS:-auto}'"
+    entry: "bash -c 'pytype ./ --keep-going -j ${NUM_CPUS:-auto}'"
```
AdamGleave, line 81: I wouldn't be surprised if using `./` causes problems relative to specifying the partcular source directories (I see pytype is erroring out with `multiple rules generate /imitation/.pytype/pyi/imitation/scripts/parallel.pyi` now)

ZiyueWang25: Yes, it is weird. I can take a further look into it later. But before that, does the other changes look good to you? If not, I think we can probably keep the original pytype version first...

ZiyueWang25: The new pytype version requires specifying the file or directory explicitly, so `./` part is required. The `--keep-going` part means: "Keep going past errors to analyze as many files as possible.". So reverting that part wouldn't help resolve this issue. 

okay, I will look into SSH and check its detail.

AdamGleave: > The new pytype version requires specifying the file or directory explicitly

New pytype version looks in `pyproject.toml` rather than `setup.cfg` for config, so our existing config including list of source directories was being ignored. I've ported it now, and this resolved the problem. Running in `.` is not equivalent to running separately for `src/`, `tests/`, etc so it makes sense there'd be a difference but I'm not sure quite what the origin of the error was. There are multiple scripts called parallel.py (e.g. there's a joblib/parallel.py in my venv), so it's possible `pytype` pulled in one of those and got confused.

PR #802, opened 2023-10-04
Add rgb observation to dagger

ZiyueWang25:
## Description

1. Add an environment wrapper to keep the original observation and rgb version together for interactive policy
2. Remove the rgb observation and its space in the bc algo.
3. Test the interactive policy can work with the wrapped environments.

## Testing

Unit test + test by examples.


ZiyueWang25:
In the files changed part, some of them are from master branch. This is because this branch is initialized from a certain commit of support-dict-obs-space branch rather than the latest condition. This would be more clear once support-dict-obs-space get merged and I can change its base to master.

src/imitation/algorithms/bc.py
```diff
@@ -334,18 +334,19 @@ def __init__(
         self._bc_logger = BCLogger(self.logger)
 
         self.action_space = action_space
-        self.observation_space = observation_space
+        obs_space_without_rgb = wrappers.remove_rgb_obs_space(observation_space)
```
AdamGleave, line 337: Why is this the responsibility of the algorithm? This will mean we have to modify every algorithm where interactive data collection is supported. More generally the algorithm is going to have to guess what the desired behavior is from the observation space.

src/imitation/algorithms/dagger.py
```diff
@@ -496,6 +531,30 @@ def extend_and_update(
         logging.info(f"New round number is {self.round_num}")
         return self.round_num
 
+    def _get_trainable_predict_fn(
```
AdamGleave, line 534: Similar to with BC -- I'm concerned about special-casing this feature into the algorithm.

What about if a year later we want to add some other component of the observation that is to be presented to humans but ignored by the policy? Do we need to go through every algorithm and update them to include this as well?

src/imitation/data/wrappers.py
```diff
@@ -206,3 +211,124 @@ def step(self, action):
...
+        # need to reset before render.
+        self.env.reset()
+        example_rgb_obs = self.env.render()
+        new_rgb_space = gym.spaces.Box(
+            low=0,
+            high=255,
+            shape=example_rgb_obs.shape,
+            dtype=np.uint8,
+        )
+        curr_sapce = self.observation_space
```
AdamGleave, line 250:
```suggestion
        curr_space = self.observation_space
```

src/imitation/data/wrappers.py
```diff
@@ -206,3 +211,124 @@ def step(self, action):
                 "rews": np.stack(self._rews),
             }
         return obs, rew, terminated, truncated, info
+
+
+class HumanReadableWrapper(gym.Wrapper):
```
AdamGleave, line 216: May be able to simplify it by subclassing the more specialized [ObservationWrapper](https://github.com/Farama-Foundation/Gymnasium/blob/aabf65cdc31431cb0b3737f8afb27d4af0fc50e7/gymnasium/core.py#L480) since it seems this wrapper only changes the observation (and not action/reward).

src/imitation/data/wrappers.py
```diff
@@ -206,3 +211,124 @@ def step(self, action):
...
+    def reset(self, **kwargs):
+        obs, info = super().reset(**kwargs)
+        return self._add_hr_obs(obs), info
+
+    def step(self, action):
+        obs, rew, terminated, truncated, info = self.env.step(action)
+        return self._add_hr_obs(obs), rew, terminated, truncated, info
+
+
+def remove_rgb_obs_space(obs_space: gym.Space) -> gym.Space:
```
AdamGleave, line 297: I question if this is the right primitive to expose to users.

An alternative would be to define a policy wrapper or just a policy feature extractor that pulls out the relevant keys from the observation.

Then there's no need to change the observation space or observations at all -- policies are just responsible for filtering what part of the observation stream they attend to. Interactive policy attends to one part, learnable policies attend to the other.

Developer using imitation would have to be responsible for configuring the policies correctly which is a downside of this approach. But developer will need to be thinking about the data collection interface and a bunch of other things anyway so it's not like they'll be able to use an algorithm off-the-shelf anyway, and we can make it somewhat "batteries included" by having some rudimentary support for this in some of our CLI scripts.

WDYT?

tests/data/test_wrappers.py
```diff
@@ -278,3 +318,193 @@ def test_n_transitions_and_empty_error(Env: Type[gym.Env]):
...
+        assert len(got.keys()) == len(expected.keys())
+        for k, v in got.items():
+            assert v == expected[k]
+    else:
+        assert got == expected
+
+
+@pytest.mark.parametrize("Env", Envs)
+@pytest.mark.parametrize("original_obs_key", ["k1", "k2"])
+def test_human_readable_wrapper(Env, original_obs_key):
```
AdamGleave, line 335: This is a great test case!

AdamGleave (commented):
Thanks for this PR! I like the implementation of `HumanReadableWrapper` and the test cases, very clean.

However I don't understand the design decision of how this was integrated with DAgger. As I see it right now it seems like we're special-casing some logic into BC and DAgger to catch (and strip out) RGB observations from the learnable policy. Generally it seems preferable to have each part of the code responsible for a small, discrete chunk. One natural way of splitting it is the algorithms are responsible for the *training* process, the policies as responsible for *how* the observations are processed and converted to actions, and other parts of the code as responsible for data collection etc. Putting logic to strip RGB observations out into the algorithms is breaking this abstraction hierarchy.

Sometimes it is necessary to break abstractions but I'd like to better understand what we're gaining from this and what the alternatives look like before committing to this. To sketch one alternative (which I suspect will need some modification): what if we made policies (or the policies' feature extractors) responsible for removing RGB observations instead? We might be able to do this with a wrapper for a policy, or just specifying a custom feature extractor (we might need some small changes to the algorithms to let callee specify feature extractor, but it'd be a more generic change). There may well be pitfalls to this approach as well -- would love to hear your thoughts.

ZiyueWang25:
Thank you for the detailed comments! I agree with your idea and I think the original design is flawed. I update the design to use a policy wrapper. I choose it instead of feature extractor because feature_extractor works on the tensor level and the wrapper can work on the np.ndarray and dict[str, np.ndarray] level, which is closer to the input. I don't have a strong preference about it though. 

In this new design, like you said, we don't need to modify any algorithm level code to fix the data level issue. Please take another look.

src/imitation/policies/obs_update_wrapper.py
```diff
@@ -0,0 +1,121 @@
...
+
+import numpy as np
+import torch as th
+from stable_baselines3.common.policies import ActorCriticPolicy
+from stable_baselines3.common.type_aliases import Schedule
+
+from imitation.data import wrappers as data_wrappers
+
+
+class Base(ActorCriticPolicy, abc.ABC):
```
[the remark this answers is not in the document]
AdamGleave: Nit: `Base` is a bit vague, base of what? Base policy wraper? Base actor critic policy wrapper?

src/imitation/policies/obs_update_wrapper.py
```diff
@@ -0,0 +1,121 @@
...
+    """Removes the human readable observation if any."""
+    if not isinstance(obs, dict):
+        return obs
+    if data_wrappers.HR_OBS_KEY not in obs:
+        return obs
+    if len(obs) == 1:
+        raise ValueError(
+            "Only human readable observation exists, can't remove it",
+        )
+    # keeps the original observation unchanged in case it is used elsewhere.
```
AdamGleave, line 115: 👍 yes good to avoid side effects where possible, and copying a dict should be cheap as it's just a shallow-copy

tests/data/test_wrappers.py
```diff
@@ -320,191 +313,28 @@ def test_n_transitions_and_empty_error(Env):
...
-        expected_obs_space = gym.spaces.Dict(
-            {
-                "2t": gym.spaces.Box(low=0, high=np.inf, shape=()),
-                "t": gym.spaces.Box(low=0, high=np.inf, shape=()),
-                HR_OBS_KEY: expected_hr_space,
-            },
-        )
-    assert isinstance(env.observation_space, gym.spaces.Dict)
-    _check_obs_or_space_equal(env.observation_space, expected_obs_space)
+    assert hr_env.observation_space == ori_env.observation_space
```
AdamGleave, line 326: How can this be true? Does the observation returned by `step()` not belong to the observation space (that'd violate the Gymnasium API I think)? Or are we encoding the human readable information somewhere other than the observation?

src/imitation/data/wrappers.py
```diff
@@ -235,30 +234,8 @@ def __init__(self, env: Env, original_obs_key: str = "ORI_OBS"):
             )
         self._original_obs_key = original_obs_key
         super().__init__(env)
-        self._update_obs_space()
```
AdamGleave: I think we still need the observation space update. From the [Gynmasium docs](https://gymnasium.farama.org/api/wrappers/observation_wrappers/):
> The transformation defined in that method must be reflected by the env observation space. Otherwise, you need to specify the new observation space of the wrapper by setting self.observation_space in the __init__() method of your wrapper.

AdamGleave (changes requested):
@ZiyueWang25 please don't make any further changes to this PR now your period with us is up -- safe travels back to Seattle and have a good weekend :) We'll find someone else to finish off the PR. Including this for informational value for you and for the next developer to pick up on.

It looks like `HumanReadableWrapper` ends up with an observation space that is inconsistent with observations returned (does not include the human-readable component that is added). This violates the Gynmasium API and is likely to cause problems somewhere down the road: e.g. if algorithms allocate buffers to store observations based on the declared observation space. This should be fixed.

I suspect having the observation space be unchanged made the policy wrapper easier. It's workable to do it without, but might need to mangle the observation space being fed into the underlying policy. This along with the need for policy wrapper to be specialized to particular kinds (ActorCriticPolicy, OffPolicy, maybe even SACPolicy) makes me think feature extractors are likely the cleaner solution here.

PR #810, opened 2023-10-09
Reduce training time and improve expert loading code in the tutorials

### LINKED ISSUE #790: Mismatch between Tutorial doc and the actual jupyter notebook

### LINKED ISSUE #793: Shorten timeout and ensure the notebook can still show desired improvement.
The `tests/test_examples.py::test_run_tutorial_notebooks[C:\\Users\\circleci.PACKER-64C83029\\project\\tests\\..\\docs\\tutorials\\4_train_airl.ipynb]` can run more than 5 minutes inside windows platform, which makes the timeout setting high to ensure the test can pass: `result = ptnb.execution.execute_notebook(nb, cwd=TUTORIALS_DIR, timeout=540)` 

To make the timeout requirement shorter, we need to update the notebook to reach a decent performance in a short time frame. Or maybe we can adjust the windows testing environment to have more resources. 

In my own Linux environment, the notebook can finish within 3 minutes.

ernestum:
## Description

The readthedocs pipeline failed due to timeouts when executing the cells of tutorial notebooks.
This was fixed by reducing the number of steps to train in the FAST mode of the AIRL tutorial.

Along the way I fixed the way experts are loaded in the tutorials:
Before: they were loaded based on normalized** environment names. However, the loader normalizes the names itself so it should be less confusing that we just load the experts based on the original environment name.

Also in one tutorial we were not even using the `load_expert` util function. I made sure it is used in all tutorials that load an expert in the same way.

** normalized means slashes replaced by hyphens

Will fix #793 and fix #790 

## Testing

Changes are covered by the existing tests.


AdamGleave (approved):
LGTM, thanks for fixing this promptly Max!

PR #812, opened 2023-10-16
Running benchmarks

ernestum:
## Description

Add scripts to run the entire benchmark suite.

## Testing


benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+    Returns:
+        A tuple of:
+         - probability of improvement
+         - confidence interval
+         - number of samples per env
+         - number of baseline samples per env
+    """
+    envs = runs_by_env.keys()
+    baseline_envs = baseline_runs_by_env.keys()
+    comparison_envs = sorted(set(envs).intersection(set(baseline_envs)))
```
qxcv, line 81:
```suggestion
    comparison_envs = sorted(set(envs) & set(baseline_envs))
```

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+
+    run_scores = sample_matrix_from_runs_by_env(runs_by_env, comparison_envs)
+    baseline_run_scores = sample_matrix_from_runs_by_env(
+        baseline_runs_by_env,
+        comparison_envs,
+    )
```
qxcv, line 87: Suggestion: add comment explaining dimensions & an assertion that the shape makes sense (in particular `run_scores.shape` should match the pattern `(s, len(comparison_envs))`, where `s>=1`).

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+
+    probabs, error_intervals = rly.get_interval_estimates(
+        {"baseline_vs_new": (baseline_run_scores, run_scores)},
+        metrics.probability_of_improvement,
+        reps=reps,
+    )
+    probability_of_improvement = probabs["baseline_vs_new"]
+    confidence_interval = np.squeeze(error_intervals["baseline_vs_new"])
+
+    return (
```
qxcv, line 99: I'd be inclined to return a dataclass/namedtuple/typed dict here, three scalars & one tuple of scalars seems very easy to confuse.

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+        if ignoring_some_envs or ignoring_some_baseline_envs:
+            warnings.warn(
+                f"The baseline runs are for the environments "
+                f"[{', '.join(baseline_envs)}], "
+                f"while the runs are for the environments [{', '.join(envs)}]. "
+                f"The comparison will only be made for the environments "
+                f"[{', '.join(comparison_envs)}].",
+            )
+
+    except ValueError as e:
```
qxcv, line 213: Why us this handler instead of letting it dump a normal traceback? Normal traceback seems more useful.

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+        if len(algos) == 0:
+            raise ValueError(f"The run directory [{args.runs_dir}] contains no runs.")
+
+        if len(baseline_algos) == 0:
+            raise ValueError(
+                f"The baseline run directory [{args.baseline_runs_dir}] "
+                f"contains no runs.",
+            )
+
+        if "algo" not in args is None:
```
qxcv, line 142: What does this do?

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+                    f"The run directory [{args.runs_dir}] contains runs for the "
+                    f"algorithms [{', '.join(algos)}]. Please use the --algo option "
+                    f" to specify which algorithms runs to compare.",
+                )
+
+        if args.baseline_algo is None:
+            if len(baseline_algos) == 1:
+                args.baseline_algo = baseline_algos[0]
+            elif args.algo in baseline_algos:
+                args.baseline_algo = args.algo
```
qxcv, line 156: What's the intent here?

qxcv: (it seems like we're comparing an algorithm to itself in this branch; is this ever a useful operation?)

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+        args.runs_dir,
+        only_completed_runs=True,
+    )
+    baseline_runs_by_algo_and_env = group_runs_by_algo_and_env(
+        args.baseline_runs_dir,
+        only_completed_runs=True,
+    )
+
+    algos = sorted(runs_by_algo_and_env.keys())
+    baseline_algos = sorted(baseline_runs_by_algo_and_env.keys())
```
qxcv, line 130: I think it would be more clear to call these `algos_in_runs_dir` and `algos_in_baselines_dir` to avoid clashes with the attributes of `args`.

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+        baseline_samples_per_env,
+    )
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("runs_dir", type=pathlib.Path)
+    parser.add_argument("baseline_runs_dir", nargs="?", default=None, type=pathlib.Path)
+    parser.add_argument("--baseline-algo", type=str)
+    parser.add_argument("--algo", type=str)
```
qxcv, line 112: Could you please add a `help=` string to these two `--*algo` args?

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+        ):
+            warnings.warn(
+                "You are comparing two equal sets of runs. "
+                "This is probably not what you want.",
+            )
+
+        envs = runs_by_algo_and_env[args.algo].keys()
+        baseline_envs = baseline_runs_by_algo_and_env[args.baseline_algo].keys()
+
+        comparison_envs = set(envs).intersection(set(baseline_envs))
```
qxcv, line 191: Similar comment here about `&` (I don't feel strongly about this but IMO it's a bit cleaner.

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+                f"The baseline runs are for the environments "
+                f"[{', '.join(baseline_envs)}], while the runs are for the "
+                f"environments [{', '.join(envs)}]. "
+                f"There is no overlap in the environments of the two run sets, so no "
+                f"comparison can be made",
+            )
+
+        ignoring_some_envs = len(comparison_envs) < len(envs)
+        ignoring_some_baseline_envs = len(comparison_envs) < len(baseline_envs)
+        if ignoring_some_envs or ignoring_some_baseline_envs:
```
qxcv, line 204: Why not just compare lists with `sorted(a) != sorted(b)` or something?

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,251 @@
...
+        f"Comparison based on {n_samples} samples per environment for {algo_str} and"
+        f" {n_baseline_samples} samples per environment for {baseline_algo_str}.",
+    )
+    print(f"Samples taken in {', '.join(comparison_envs)}")
+    print()
+    print(f"Probability of improvement of {algo_str} over {baseline_algo_str}:")
+    print(
+        f"{probability_of_improvement:.3f} "
+        f"({error_interval[0]:.3f}, {error_interval[1]:.3f}, "
+        f"reps={args.bootstrap_reps})",
```
qxcv, line 246:
```suggestion
        f"reps={args.bootstrap_reps:,})",
```

Comma-separated b/c this seems to be a big-ish number.

qxcv (commented):
Added some requests for clarification and small changes.

src/imitation/util/sacred_file_parsing.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Yields:
+        Tuples of (config, run) dicts.
+    """
+    for config_path in run_path.rglob("config.json"):
+        run_path = config_path.parent / "run.json"
+
+        if run_path.exists():
+            run = json.loads(run_path.read_text())
+            if only_completed_runs and run["status"] != "COMPLETED":
```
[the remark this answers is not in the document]
ernestum: I decided against the warning here since IMO the `only_completed_runs==True` makes it clear that we are not interested in failed/pending runs.
The summary will contain a special section listing the number of failed/pending/running runs if there are any.

benchmarking/compute_probability_of_improvement.py
ernestum: You are right. This comparison can be a lot simpler.

benchmarking/compute_probability_of_improvement.py
ernestum: I found it easier to read but you are probably right about it. I will revert it.

benchmarking/compute_probability_of_improvement.py
ernestum: Good point changed it.

benchmarking/compute_probability_of_improvement.py
ernestum: Yes, this part of the code is a bit confusing. Added some asserts and clarification.

benchmarking/compute_probability_of_improvement.py
ernestum: Very good point. Changed that.

benchmarking/compute_probability_of_improvement.py
ernestum: Thanks for the catch. Fixed it.

benchmarking/compute_probability_of_improvement.py
ernestum: Added one for all of them.

benchmarking/compute_probability_of_improvement.py
ernestum: Fixed, good point!

ernestum:
To give you an idea of how the output CSV and Markdown look:
[summary.csv](https://github.com/HumanCompatibleAI/imitation/files/12938101/summary.csv)
[summary.md](https://github.com/HumanCompatibleAI/imitation/files/12938102/summary.md)

benchmarking/compute_probability_of_improvement.py
ernestum: Yes, probably for the better. Added a dataclass

benchmarking/run_all_benchmarks_on_slurm.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+sbatch --job-name=bc_seals_ant run_benchmark_on_slurm.sh train_imitation bc seals_ant
```
taufeeque9, line 2: Can replace the whole file here also in a similar way as above:

```bash
#!/bin/bash

envs=(
    "seals_ant"
    "seals_half_cheetah"
    "seals_hopper"
    "seals_swimmer"
    "seals_walker"
)

script_algos=(
    "imitation" "bc"
    "imitation" "dagger"
    "adversarial" "airl"
    "adversarial" "gail"
)

for env in "${envs[@]}"; do
    for ((i=0; i<${#script_algos[@]}; i+=2)); do
        script=${script_algos[$i]}
        algo=${script_algos[$((i+1))]}
        job_name="${algo}_${env}"
        echo sbatch --job-name="${job_name}" run_benchmark_on_slurm.sh "train_${script}" "${algo}" "${env}"
    done
done

File: benchmarking/compute_probability_of_improvement.py [hunk 14]
```diff
@@ -0,0 +1,303 @@
...
+from typing import Dict, List, Optional
+
+import numpy as np
+from rliable import library as rly
+from rliable import metrics
+
+from imitation.util.sacred_file_parsing import SacredRun, group_runs_by_algo_and_env
+
+
+def make_sample_matrix_from_runs_by_env(
```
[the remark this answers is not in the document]
ernestum: Yes that name is much clearer!

benchmarking/run_all_benchmarks.sh
```diff
@@ -0,0 +1,201 @@
+#!/bin/bash
+python -m imitation.scripts.train_imitation bc with bc_seals_ant seed=1
```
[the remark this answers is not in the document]
ernestum: I know that this can be easily be generated. I used:
```python
envs = [
    "seals_ant",
    "seals_half_cheetah",
    "seals_hopper",
    "seals_swimmer",
    "seals_walker",
]

for env in envs:
    for algo in ["bc", "dagger"]:
        for seed in range(1,11):
            print(f"python -m imitation.scripts.train_imitation {algo} with {algo}_{env} seed={seed}")

    for algo in ["airl", "gail"]:
        for seed in range(1, 11):
            print(f"python -m imitation.scripts.train_adversarial {algo} with {algo}_{env} seed={seed}")
```

However, I opted for solution with less moving parts. Whenever we add new algorithms or change something else I can just re-write this script in 3min. As soon as we put the code in the repo we have to maintain it, test it, document it. I think that is a lot of effort just to safe us those 3mins to re-write this on-off script.

benchmarking/run_all_benchmarks_on_slurm.sh
ernestum: See above comment.

benchmarking/run_all_benchmarks.sh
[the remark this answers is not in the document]
taufeeque9: I suppose that's true.

Having the `run_all_benchmarks.sh` file in the repo feels slightly weird. Do you think we can remove the file entirely and just keep the above python/bash script that generates those commands? There doesn't seem any need to have the `run_all_benchmarks.sh` script at all.

benchmarking/compute_probability_of_improvement.py
```diff
@@ -0,0 +1,303 @@
...
+            f"[{', '.join(comparison_envs)}].",
+        )
+
+    probability_of_improvement_result = compute_probability_of_improvement(
+        runs_by_env=runs_by_algo_and_env[args.algo],
+        baseline_runs_by_env=baseline_runs_by_algo_and_env[args.baseline_algo],
+        reps=args.bootstrap_reps,
+    )
+
+    show_path = args.algo == args.baseline_algo
```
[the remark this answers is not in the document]
ernestum: Did so!

benchmarking/run_all_benchmarks.sh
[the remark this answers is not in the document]
qxcv: I don't think this matters much either way, but I have a slight preference for the current approach because I think it makes it easier to see exactly what was executed, and also to run just a subset of the experiments by copy-pasting.

qxcv (approved):
I re-reviewed the diff from my last review. Looks good to merge!

On the debate about including the script that generates the script that runs experiments or just including the script that runs experiments: it seems like we're 2v1 in favor if the current approach, so I'm inclined to just merge as-is.

ernestum:
Great. This are the final results. Raw values as well as CSV and Markdown Summary. This is to be added as a release artifact
[benchmark_runs.zip](https://github.com/HumanCompatibleAI/imitation/files/13209245/benchmark_runs.zip)
!

PR #823, opened 2023-11-30
Fix warning in quickstart.py

ernestum:
Fix warning in `quickstart.py` caused by not properly setting the render mode in the evaluation environment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -365,6 +365,10 @@ jobs:
             TESTFILES=$(circleci tests glob "tests/**/test*.py" | circleci tests split --split-by=timings)
             echo "This shard testing: ${TESTFILES}"
             pytest -n auto --junitxml=/tmp/test-reports/junit.xml -vv $TESTFILES
+          environment:
+            # This is needed in newer versions of pygam. See
+            # https://github.com/pygame/pygame/issues/3835#issuecomment-1703717368
+            SDL_VIDEODRIVER: dummy
 
       - save-pytest-cache
       - store-test-output-unix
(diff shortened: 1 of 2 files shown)


ernestum:
@AdamGleave can you merge this? The coverage changes seem spurious.

.circleci/config.yml
```diff
@@ -365,6 +365,10 @@ jobs:
             TESTFILES=$(circleci tests glob "tests/**/test*.py" | circleci tests split --split-by=timings)
             echo "This shard testing: ${TESTFILES}"
             pytest -n auto --junitxml=/tmp/test-reports/junit.xml -vv $TESTFILES
+          environment:
+            # This is needed in newer versions of pygam. See
```
AdamGleave, line 369:
```suggestion
            # This is needed in newer versions of pygame. See
```

AdamGleave (commented):
One minor change, will wait @taufeeque9's review & then merge once outstanding changes addressed

taufeeque9:
The changes look good to me. Should we also make the same changes in the tutorial notebooks by adding `evaluation_env`? I'm not sure if rendering will work with the jupyter notebooks.

ernestum:
What do you mean @taufeeque9 ? We do not render anything in the tutorials as far as I know.

taufeeque9:
> What do you mean @taufeeque9 ? We do not render anything in the tutorials as far as I know.

Makes sense. I was also not sure so was just checking whether we need to make the same changes in `docs/tutorials/1_train_bc.ipynb` for example.

taufeeque9 (approved):
LGTM

PR #826, opened 2023-12-06
Switch to lualatex to generate the documentation PDF

ernestum:
The documentation pipeline was failing only on the master branch.
This is due to the fact that only for commits on master readthedocs tries to generate a PDF.

PDF generateion failed since latex ran out of memory.

Let's see if luatex can do better.

Note: to test if this PR works we unfortunately need to merge it into master since the PDF output is only generated for 
master builds.


AdamGleave (approved):
LGTM -- let's give LuA a try!

PR #827, opened 2023-12-07
Fix documentation pipeline

ernestum:
Next attempt at fixing the documentation pipeline.

Right now we had two issues:
1. we occasionally ran into timeouts when executing the tutorial notebooks.
2. we are missing dependencies when automatically generating API docs
3. The readthedocs.yaml file was not hidden.

I fixed 1 by increasing the timeout and splitting up the offending tutorial (GAIL) into more cells so we are less likely to run into the per-cell timeout.

Otherwise the config file was renamed and the dependencies were added.


ernestum:
@AdamGleave can you merge this one again? Looks like the first try of fixing the documentation pipeline fixed due to other issues which should have been fixed in this PR. Merging this one means another test run for LuaTex

AdamGleave (approved):
LGTM

PR #828, opened 2023-12-11
Make the default device for policies "cpu"

### LINKED ISSUE #825: Run time Error when run quickstart.py
When run examples/quickstart.py, i'm getting error with

RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument index in method wrapper_CUDA_gather)

it seems that when running imitation code (Behavior Cloning in this example), default mode is in several GPUs. If not it seems data conversion problem in tensor.

==>I solved it by changing bc.py's __init__ function's  device setting from auto to cpu. but still getting error (warning sign);

python3.10/site-packages/stable_baselines3/common/vec_env/base_vec_env.py:243: UserWarning: You tried to call render() but no `render_mode` was passed to the env constructor.
  warnings.warn("You tried to call render() but no `render_mode` was passed to the env constructor."

Does any have same issues? or know solution?

Thank you in advanced

## Environment
- OS:Ubuntu 22.04
- Python version: 3.10
- Cuda : 11.8 - 12.0

ernestum:
When there is a GPU available, using "auto" as the default device does not seem to be desirable (see #825).

This PR sets the default device to "cpu" which hopefully just works in most cases.

Fixes #825

### DIFF
--- a/src/imitation/algorithms/bc.py
+++ b/src/imitation/algorithms/bc.py
@@ -249,7 +249,7 @@ class BCLogger:
 
 def reconstruct_policy(
     policy_path: str,
-    device: Union[th.device, str] = "auto",
+    device: Union[th.device, str] = "cpu",
 ) -> policies.ActorCriticPolicy:
     """Reconstruct a saved policy.
 
@@ -285,7 +285,7 @@ class BC(algo_base.DemonstrationAlgorithm):
         optimizer_kwargs: Optional[Mapping[str, Any]] = None,
         ent_weight: float = 1e-3,
         l2_weight: float = 0.0,
-        device: Union[str, th.device] = "auto",
+        device: Union[str, th.device] = "cpu",
         custom_logger: Optional[imit_logger.HierarchicalLogger] = None,
     ):
         """Builds BC.
--- a/src/imitation/algorithms/dagger.py
+++ b/src/imitation/algorithms/dagger.py
@@ -100,7 +100,7 @@ def reconstruct_trainer(
     scratch_dir: types.AnyPath,
     venv: vec_env.VecEnv,
     custom_logger: Optional[imit_logger.HierarchicalLogger] = None,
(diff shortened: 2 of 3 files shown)


tomtseng:
Hmm so the error is this backtrace when running `examples/quickstart.py`:
```
  File "examples/quickstart.py", line 83, in <module>
    bc_trainer.train(n_epochs=1)
  File "/home/ttseng/imitation/src/imitation/algorithms/bc.py", line 495, in train
    training_metrics = self.loss_calculator(self.policy, obs_tensor, acts)
  File "/home/ttseng/imitation/src/imitation/algorithms/bc.py", line 130, in __call__
    (_, log_prob, entropy) = policy.evaluate_actions(
```

On `bc.py` line 130, `tensor_obs` is on device `auto` which assigns it to a GPU, but `acts` is on the CPU. Could the solution be to move `acts` onto the same device as `tensor_obs`? The proposed fix in this PR is to make the `device` argument default to `cpu`, but that means that this code still breaks if the user sets the argument to something else

tomtseng:
On [line 494 of `bc.py`](https://github.com/HumanCompatibleAI/imitation/pull/828/files#diff-1081b59972be59d143d90b881addc63c250e57e93a46aa2cc741579d55627316L494), right before the failing line 495, we have `acts = util.safe_to_tensor(batch["acts"], device=self.policy.device)`. Despite the `device` argument seemingly suggesting that `acts` should be moved to the same device as the policy & observation, but looking at the implementation of `util.safe_to_tensor`, `acts` doesn't actually get moved if it's already a tensor. My instinct would be to either change `util.safe_to_tensor`'s behavior, and/or to add a line `acts = acts.to(self.policy.device)`

ernestum:
Good point and thanks for the suggestions! I will start a new PR with your proposal.

ernestum:
Closing this in favor of #831

PR #829, opened 2023-12-11
Remove FloatReward

### LINKED ISSUE #794: Remove FloatReward when next SB3 is released
FloatReward that is defined inside tests/algorithms/test_bc.py should be removed once the next SB3 is released.

ernestum:
A quick fix for #794.

After the newest release of SB3 we don't need the FloatReward wrapper any more.

Fixes #794

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -203,7 +203,7 @@ setup(
         "rich",
         "scikit-learn>=0.21.2",
         "seals~=0.2.1",
-        "stable-baselines3~=2.0",
+        "stable-baselines3~=2.2.1",
         "sacred>=0.8.4",
         "tensorboard>=1.14",
         "huggingface_sb3~=3.0",
--- a/tests/algorithms/conftest.py
+++ b/tests/algorithms/conftest.py
@@ -113,20 +113,10 @@ def pendulum_single_venv(rng) -> VecEnv:
     )
 
 
-# TODO(GH#794): Remove after https://github.com/DLR-RM/stable-baselines3/pull/1676
-# merged and released.
-class FloatReward(gym.RewardWrapper):
-    """Typecasts reward to a float."""
-
-    def reward(self, reward):
-        return float(reward)
-
-
 @pytest.fixture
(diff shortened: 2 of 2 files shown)


taufeeque9 (approved):
LGTM.

For the failing tests, it seems like the issue is with the incompatibility of Ray with Pydantic 2.5+. According to this [issue](https://github.com/ray-project/ray/issues/37019), Ray 2.9 which will be released end of December should fix the error. For now we can add pydantic<2 in the dependencies as a temporary fix.

ernestum:
I had to remove some outdated test code. Would you re-review @taufeeque9 ?

taufeeque9:
The pytorch warning check in `test_train_dagger_main` was introduced in #332. That PR added the `np.array(v)` copy line to not cause the pytorch warning. I checked locally that changing `np.array(v)` -> `v` in line 465 of `imitation/data/types.py` doesn't result in pytorch raising the warning. So should we make this change as well? Or will it break anything else?

ernestum:
Hmm the two changes in #332 make no sense to me. As I understand it, `np.array(v)` should suppress the warning but the the added test code should ensure that the warning pops up. So why did the test not fail back then?
This seems to be somehow related to the `save_to_tensor` utility, that I am changing right now in #831.
Maybe @AdamGleave or @qxcv can give us some insight here?

AdamGleave:
> Hmm the two changes in #332 make no sense to me. As I understand it, `np.array(v)` should suppress the warning but the the added test code should ensure that the warning pops up. So why did the test not fail back then? This seems to be somehow related to the `save_to_tensor` utility, that I am changing right now in #831. Maybe @AdamGleave or @qxcv can give us some insight here?

IIUC the test is checking that the warning does *not* occur:
```python
        assert not (
            warning.category == UserWarning
            and "NumPy array is not writeable" in warning.message.args[0]
        )
```
so I think we can keep that test code in. Only looked at this quickly, quite possibly missing something obvious, please do LMK if so!

PR #830, opened 2023-12-11
Fix coverage issue in BC tests

ernestum:
Some missing coverage in BC testing was surfacing in a couple of the most recent PRs. This PR fixes it so we don't have to ask @AdamGleave to merge all the time.

### DIFF
--- a/tests/algorithms/test_bc.py
+++ b/tests/algorithms/test_bc.py
@@ -94,17 +94,17 @@ bc_train_args = st.builds(
     log_rollouts_venv=st.one_of(rollout_envs, st.just(None)),
 )
 bc_args = st.builds(
-    lambda env, batch_size, custom_logger, rng: dict(
+    lambda env, minibatch_size, rng, minibatch_fraction: dict(
         observation_space=env.observation_space,
         action_space=env.action_space,
-        batch_size=batch_size,
-        custom_logger=custom_logger,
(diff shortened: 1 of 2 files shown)


tests/algorithms/test_bc.py
```diff
@@ -152,7 +151,12 @@ def test_smoke_bc_creation(
     expert_data_type=expert_data_types,
     rng=rngs,
 )
-@hypothesis.settings(deadline=20000, max_examples=15)
+@hypothesis.settings(
+    deadline=20000,
+    max_examples=15,
+    # TODO: one day consider removing this. For now we are good.
```
tomtseng, line 157: maybe this is obvious for someone who's more familiar with `hypothesis` or this codebase but what does this check do? would it be helpful to clarify when it would be possible to remove this check or what's going that causes this check to fail?

tomtseng (approved):

tests/algorithms/test_bc.py
ernestum: Thanks that is a good point! I added a better description. Does it make sense now @tomtseng ?

AdamGleave:
I'll merge this @ernestum but the issue with test coverage appears to be in `test_sqil.py`, lines 251 and 257-260, as visible at https://app.codecov.io/gh/HumanCompatibleAI/imitation/pull/830/blob/tests/algorithms/test_sqil.py

Issue #852, opened 2024-05-31
TypeError: argument of type 'PosixPath' is not iterable


den-schmidt:
## Bug description
Hi everyone,

I am trying to play with the DAgger example from `docs/tutorials/2_train_dagger.ipynb` but get the following error when executing the line `dagger_trainer.train(2000)`: 

```bash
Exception has occurred: TypeError       (note: full exception trace is shown but execution is paused at: _run_module_as_main)
argument of type 'PosixPath' is not iterable
  File "/home/denni/msc/imitation/.venv/lib/python3.8/site-packages/fsspec/core.py", line 323, in _un_chain
    if "::" in path
  File "/home/denni/msc/imitation/.venv/lib/python3.8/site-packages/fsspec/core.py", line 383, in url_to_fs
    chain = _un_chain(url, kwargs)
  File "/home/denni/msc/imitation/.venv/lib/python3.8/site-packages/datasets/arrow_dataset.py", line 1515, in save_to_disk
    fs, _ = url_to_fs(dataset_path, **(storage_options or {}))
  File "/home/denni/msc/imitation/.venv/lib/python3.8/site-packages/imitation/data/serialize.py", line 23, in save
    huggingface_utils.trajectories_to_dataset(trajectories).save_to_disk(p)
  File "/home/denni/msc/imitation/.venv/lib/python3.8/site-packages/imitation/algorithms/dagger.py", line 147, in _save_dagger_demo
    serialize.save(npz_path, [trajectory])
  File "/home/denni/msc/imitation/.venv/lib/python3.8/site-packages/imitation/algorithms/dagger.py", line 285, in step_wait
    _save_dagger_demo(traj, traj_index, self.save_dir, self.rng)
  File "/home/denni/msc/imitation/.venv/lib/python3.8/site-packages/stable_baselines3/common/vec_env/base_vec_env.py", line 206, in step
    return self.step_wait()
  File "/home/denni/msc/imitation/.venv/lib/python3.8/site-packages/imitation/data/rollout.py", line 447, in generate_trajectories
    obs, rews, dones, infos = venv.step(acts)
  File "/home/denni/msc/imitation/.venv/lib/python3.8/site-packages/imitation/algorithms/dagger.py", line 669, in train
    trajectories = rollout.generate_trajectories(
  File "/home/denni/msc/imitation/test.py", line 41, in <module>
    dagger_trainer.train(2000)
[...]

PR #831, opened 2023-12-15
Ensure safe_to_tensor moves tensors to the specified device.

ernestum:
This PR fixes a bug in the `safe_to_tensor` utility: previously it did not move tensors to a new device according to the `device` kwarg which caused issues when there is more than one device available.
The bug went unnoticed for a long while since our circleCI runners do not have GPUs enabled.

Thanks to @tomtseng for drafting a fix for this in #828

PR #855, opened 2024-07-13
Fix PosixPath being passed to HuggingFace `save_to_disk` (which does not accept it)

iwishiwasaneagle:
## Description

HuggingFace save_to_disk takes `PathLike` type which is defined as str, bytes or os.PathLike. imitation.util.parse_path always returned pathlib.Path which is not one of these types. This commit converts pathlib.Path to str before calling HuggingFace save_to_disk. 

Related issue #852 

## Testing

Passed a path in both `str` and `Path` forms to `save` for `Trajectory` and `TrajectoryWithRew.` Expanded testing of serialize by adding additional `save` then `load` tests.

### DIFF
--- a/src/imitation/data/serialize.py
+++ b/src/imitation/data/serialize.py
@@ -19,7 +19,7 @@ def save(path: AnyPath, trajectories: Sequence[Trajectory]) -> None:
         path: Trajectories are saved to this path.
         trajectories: The trajectories to save.
     """
-    p = util.parse_path(path)
+    p = str(util.parse_path(path))
     huggingface_utils.trajectories_to_dataset(trajectories).save_to_disk(p)
     logging.info(f"Dumped demonstrations to {p}.")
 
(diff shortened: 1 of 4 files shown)


iwishiwasaneagle:
Needs more work.

src/imitation/data/serialize.py
```diff
@@ -19,7 +19,7 @@ def save(path: AnyPath, trajectories: Sequence[Trajectory]) -> None:
         path: Trajectories are saved to this path.
         trajectories: The trajectories to save.
     """
-    p = util.parse_path(path)
+    p = str(util.parse_path(path))
```
Jaramyy, line 22: I tested with this solution. It works.

cp2phi: It works! Cheers!

tomtseng:
Fixed in https://github.com/HumanCompatibleAI/imitation/pull/831#issuecomment-2574448401 before I saw this PR, won't close this because the unit tests may be useful.
However this project is not actively maintained by anyone anymore so not clear if any code owner will review & merge this in the near future

